Issue #2, opened 2022-03-05
Add docs to this repo


skyl:
## What

Add the jekyll site served at https://free2z.com/docs to this repo

## Why

More people can contribute to them

## Reference

#1

skyl:
we can tackle this later - easier to iterate faster in the private monorepo for now. Also, most docs will become zPages over time and we might just get rid of the jekyll site.

PR #3, opened 2022-03-05
open source the docs

skyl:
ref #2


skyl:
We may tackle this later when the site is more mature. But, for now easier to iterate in the private monorepo

PR #16, opened 2022-08-18
ZUULI next

skyl:
Fixes, cleanup, maybe restore ... first release?

### DIFF
--- a/ex/zuuli/src/electron/main.js
+++ b/ex/zuuli/src/electron/main.js
@@ -4,7 +4,7 @@ const path = require("path");
 const url = require("url");
 
 const { ipcMain } = require('electron');
-const { forkWarp } = require("./fork");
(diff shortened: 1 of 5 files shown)


ex/zuuli/src/electron/main.js
```diff
@@ -54,6 +54,19 @@ function createWindow() {
         console.log("OPEN", event, arg)
         shell.openExternal(arg)
     })
+    ipcMain.on('send', (event, id, json) => {
+        console.log("SEND", event, id, json)
+        // Zcash
+        warp.setActiveAccount(0, id)
+        const res = warp.sendMultiPayment(json)
```
skyl, line 61: TODO: try to put into fork process? Even in the main process (not in the renderer), this is a lot of work and can cause BBOD in the Mac version.

PR #19, opened 2022-09-01
ex/websites/signupz

skyl:
## What

Basic signup site for now. Someday with private signups using Zcash

## Why

I need a signup site anyways. But, it doesn't seem like the existing products are mindful of privacy. So, I'm going to be stubborn and build my own signup site for emails. Someday this site could be even more private with Zcash signups.


ex/websites/signupz/templates/base.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
```
skyl, line 1: delete, rename index

birdify (commented):
Artwork looks dope, where'd you find that?  The UI for this site looks sweet!

skyl:
> Artwork looks dope, where'd you find that? The UI for this site looks sweet!

yeah, I just googled it. It was on some "freepics" type site. But, should figure out who the artist is and do some combination of a) credit b) pay c) teamup

ex/websites/signupz/README.md
```diff
@@ -0,0 +1,13 @@
+# Signups
+
+Example website that can collect pre-launch signup information.
+
+Currently email-only.
+But, could incorporate Zcash and other payments as condition of signup.
```
skyl, line 6: (or as a private way to sign up for communications privately and anonymously)

ex/websites/signupz/requirements.txt
```diff
@@ -0,0 +1,2 @@
+Django==3.2
```
skyl, line 1: Tested with `Python 3.9.2`

PR #86, opened 2024-09-13
ts/react/free2z gh actions

skyl:
Add build for f2z frontend


.github/workflows/ts-react-free2z.yml
```diff
@@ -0,0 +1,32 @@
...
+        working-directory: ts/react/free2z
+        run: npm install
+
+      # - name: Run tests
+      #   working-directory: ts/react/free2z
+      #   run: npm test
+
+      - name: Build the project
+        working-directory: ts/react/free2z
+        # TODO: we want to set this to true soon!
```
skyl, line 31: this is important really - we would like to fix all warnings or explicitly silence them. Perhaps we can make a bounty.

ts/react/free2z/src/AI2.tsx
```diff
@@ -213,9 +214,9 @@ const reducer = (state: AIState, action: AIAction): AIState => {
 export default function AI() {
     const { conversationId } = useParams();
     const [open, setOpen] = useState(false)
-    const [user, setUser] = useGlobalState('creator')
```
skyl, line 216: some random warnings silenced in this file. But, it still has a couple ;/

```
npx eslint "src/AI2.tsx" --fix


/Users/skyl/Code/github/f2z/tuzi/ts/react/free2z/src/AI2.tsx
  335:8   warning  React Hook useEffect has missing dependencies: 'connect', 'loadPromptResponses', 'setLoginModal', and 'state.currentConversation.id'. Either include them or remove the dependency array                                                                           react-hooks/exhaustive-deps
  364:46  warning  The ref value 'bottomRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'bottomRef.current' to a variable inside the effect, and use that variable in the cleanup function  react-hooks/exhaustive-deps

✖ 2 problems (0 errors, 2 warnings)
```

Some of these are not so easy to fix and can't be fixed automatically because they could take some brain power. These actually seem pretty irrelevant at a glance.

skyl:
https://free2z.com/docs/legal/privacy-policy

PR #89, opened 2024-09-17
feat(zPage): add popout preview window

skyl:


skyl:
Opens new window that live updates and automatically scrolls to the bottom if the editor is scrolled to the bottom

Screenshot 2024-09-16 at 7 10 37 PM

PR #103, opened 2024-10-16
Add warning dialog when open external link

0bkevin:
## Description
Add "are you sure" warning when user opens external link external links

### This pull request includes the following changes:

- **Add external link warning dialog to MLink component**
    - [x] Attractive warning dialog.
    - [x] The dialog is not annoying, allow users to opt-out of future warnings.

## Changes
- Introduced a confirmation dialog that warns users before navigating to external links.
-  Implemented a checkbox in the dialog to allow users to opt-out of future warnings.
- Added a localStorage key to remember user preference for not showing the warning again.
- Ensured internal links bypass the warning dialog.

## Issue
Closes #84

PR #105, opened 2024-10-19
feat(ts/react/free2z): some small changes to the warning modal for clicking link

skyl:
adding to #103


skyl:
@0bkevin - I'm not sure if this is completely idiomatic but it make sense to show the full link in the warning to help people decide ... another possible improvement would be to turn off the warning for certain domains instead of turning it off globally ... 🤔

PR #110, opened 2024-11-04
Fix zpages series pagination request

### LINKED ISSUE #104: Problems with Zpages series
Greetings! I have detected a bug that puzzles me and does not let me organize my Zpages in series. I do not know if it has to do with a maximum number of series that I can have?

But I just created a series of my Europe tour, and when I create a new Zpage and try to add it to that series, the drop-down list doesn't show it to me. The interesting thing is that yes, the series is created, but it doesn't come up for me to add new Zpages. It's really crazy.
Captura de pantalla de 2024-10-18 00-53-33

0bkevin:
## Description
The configuration of the zpage series fetch has been updated. Instead of using useQuery, it now utilizes useInfiniteQuery to asynchronously fetch the subsequent pages. In the autocomplete object, when scrolling to the bottom, the system evaluates if there are more pages available. If additional pages exist, it fetches the next page and displays the results in the options.

## This pull request includes the following changes:
- Changes useQuery for useInfinteQuery
- Uses the onScroll prop from the Autocomplete component to detect when the user scrolls to the bottom of the options list.
- Quit warning errors

## Showcase

https://github.com/user-attachments/assets/ee632660-fbba-4fca-ad53-9fdf0dedaf39

## Issues

Closes #104

### DIFF
--- a/ts/react/free2z/src/components/EditSelectSeries.tsx
+++ b/ts/react/free2z/src/components/EditSelectSeries.tsx
@@ -1,4 +1,4 @@
-import { useQuery, useMutation } from 'react-query';
+import { useMutation, useInfiniteQuery } from 'react-query';
 import { DragDropContext, Droppable, Draggable, DropResult } from 'react-beautiful-dnd';
 import axios from 'axios';
 import React, { useState, useEffect, useRef } from 'react';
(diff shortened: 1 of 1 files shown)


0bkevin:
Another option could be wrapping the fetcher function inside a while loop and fetch all the pages from the beginning.   I take this approach because I think could be more performant in case someone have a lot of zpages series.  

Let me know what you think @skyl

skyl:
going to test now ... https://test.free2z.com/

skyl (approved):

skyl:
we should maybe work on the devcontainer? you use VSCode @0bkevin ?

Issue #92, opened 2024-09-21
Improve zpage rendering - spacing, fonts, style details


skyl:
- [ ] inline code should look better, stand out more.

Compare github to free2z:

`git` and `npm`

<img width="371" alt="Screenshot 2024-09-21 at 11 44 44 AM" src="https://github.com/user-attachments/assets/5637fc1f-fd2b-4449-a6d7-1c124b1270db">

- [ ] better looking italics (?)
- [ ] line height adjustments - less space in paragraph (?), more space between paragraphs.

<img width="862" alt="Screenshot 2024-09-21 at 11 49 00 AM" src="https://github.com/user-attachments/assets/7fa9ebaf-b7e3-4557-a776-5db1fde1772f">

<img width="795" alt="Screenshot 2024-09-21 at 11 50 23 AM" src="https://github.com/user-attachments/assets/83f038f8-dadc-4386-a53a-2047bf967073">

In general there could be a lot of small adjustments that make people _feel_ like free2z is quality. We haven't done anything much with the typography and it's basically just default MUI5.

skyl:
Closing as superseded by the SvelteKit article renderer introduced in #147. Typography and Markdown rendering now use a dedicated prose/code styling system rather than the legacy default-MUI surface. Please file focused current visual defects against the Svelte app; #150 tracks final cutover.

PR #111, opened 2024-11-04
Improve zpage rendering

0bkevin:
## Description
Improve markdown typography rendering, was made using parts of Spotify design system. 

## This pull request includes the following changes:
- Improve inline code rendering
- Improve block code rendering
- Improve italics rendering
-  Subtract line height 
- Add space between paragraphs

## Showcase
Right images: new version
Left images: old version

image

image

image

I think there is still some changes that could be made. For example with the height of this card
image

@skyl do you have any other improvements in mind?

## Issues 
#92

### DIFF
--- a/ts/react/free2z/src/components/CodeCopyButton.tsx
+++ b/ts/react/free2z/src/components/CodeCopyButton.tsx
@@ -29,23 +29,26 @@ export default function CodeCopyButton(props: CodeCopyProps) {
 
     return (
         <IconButton
-            // variant="contained"
             size="small"
             color="primary"
             onClick={handleCopyClick}
             sx={{
                 position: 'absolute',
-                top: '3px',
-                right: '-4px',
+                top: '20px',
+                right: '20px',
                 zIndex: (theme: Theme) => theme.zIndex.mobileStepper - 1,
                 borderColor: copySuccess ? 'success.main' : undefined,
                 color: copySuccess ? 'success.main' : undefined,
             }}
         >
             {copySuccess ?
-                <Check fontSize="small" />
+                <Check fontSize="small" style={{
+                    color: 'green'
+                }} />
                 :
-                <ContentCopy fontSize="small" />
+                <ContentCopy fontSize="small" style={{
+                    color: 'whitesmoke'
+                }}/>
             }
         </IconButton>
     )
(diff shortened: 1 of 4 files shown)


skyl:
The reason I put the copy-paste widget up high was so that on narrow screens it would be less likely to cover the text. Make sure to check down to about 320px for older phones ...

skyl:
I was thinking maybe the header could be grander on big screens ... make it look more flashy like nytimes or something ;9 but also we need to figure out a way to fit more features in gracefully - for example I would love to have a good text-to-speech voice so I could listen to a zpage like a podcast sometimes. Maybe there could be AI features like TL;DR summary from AI, translations potentially (although I think leaving translations to browsers is probably smart...) ... but there could be some more things in the header maybe ... IDK, I like how clean it is now OTOH ...

chat the zpage ;9

IDK, maybe it's pretty good the way it is - I'm not really a designer or a frontend person and I just pulled the whole thing out of my ass so I'm curious what a person with a better eye for design might do. The core theme is up-for-grabs, I'd be up to see some radical changes potentially.

Let's rewrite the whole thing with a hotter UI library and change the name or something ;0

skyl:
on test now ...

https://test.free2z.com/free2z/zpage/flavored-markdown

https://free2z.com/free2z/zpage/flavored-markdown

IDK, I kinda like what's on prod now better ...

0bkevin:
> The reason I put the copy-paste widget up high was so that on narrow screens it would be less likely to cover the text. Make sure to check down to about 320px for older phones ...

I see your point
image

also, I noticed that the code it's difficult to read, since I wrap the text 😭, should let them scroll it horizontally

0bkevin:
> I would love to have a good text-to-speech voice so I could listen to a zpage like a podcast sometimes. Maybe there could be AI features like TL;DR summary from

Loving the idea of text-to-speech, this could be a really great integration to add,

Also, the TL;DR summary, this could be added in a side section, alongside with an input that the user could use to make questions about the post. We could even integrate this to the user profile page, generating a resume of every user based on its blog post, converse coments,

0bkevin:
The AI translations could be useful to avoid the duplication of content, as I been seeing, a lot of users publish their post in Spanish and the same post in English. Or to avoid using AI, worried a little about the token cost, we could allow the user to submit versions of their post. and by versions, I mean lang. Allowing users to create a post and having a section in where they can submit other version of the same post but in another language. 

When another user tries to see that post, it would have an option to change the language.

The only thing that I see that could be a potential problem its what version post the page should show when searching or in the main page.

We could add a global lang setting, and depending on the user selection, the page shows one version of the blog or another.

skyl:
Should we merge? What do you think?

PR #118, opened 2024-11-15
Fix: avoid infinite loading state

### LINKED ISSUE #117: Fix: infinite loading state in storytime edit page
## Description
There is an infinite loading state when navigating to "/storytime/edit/:creator/:slug" and the Storytime does not exist

https://github.com/user-attachments/assets/aa4ed765-9baf-4529-82f9-c6edebf75ce3

0bkevin:
## Description

Avoid infinite loading state on "/storytime/edit/:creator/:slug" when story is not found

## Issues
Closes #117


skyl (approved):
thanks!

PR #273, opened 2026-08-08
fix(zuuli): preserve conflicting legacy wallet on launch

skyl:
Closes #271

## Outcome

ZUULI no longer panics when both legacy `com.2zinc.zuuli` and canonical `cash.free2z.zuuli` app-data trees contain state. It validates and opens canonical state only, preserves the complete legacy tree byte-for-byte, and surfaces a structured `legacyAppData.state = importPending` diagnostic with a persistent notice in ZUULI and Zuuallet.

Automatic migration is narrowed: only a canonical directory proven empty and unchanged may be removed using a journal-first, non-recursive, identity-checked OS operation. Any content refuses automatic migration and becomes preserved/import-pending; links, reparse points, special objects, or identity replacement still fail closed. No populated directory is deleted, moved, quarantined, merged, or implicitly selected as an import source.

Full explicit legacy import with UFVK validation and per-directory salt/native custody handling is tracked in #272.

## Verification

- Rust plugin: 85/85 tests pass (`cargo +1.88.0 test --locked --all-targets`)
- New focused coverage: populated conflict preservation, empty removal, content-arrival race, identity-replacement race, crash recovery, serialized diagnostic
- ZUULI: typecheck, 19 tests, production build pass
- Zuuallet: typecheck and production build pass
- Real macOS `npm run tauri dev`: process initialized and remained live against the actual two-populated-wallet state; current canonical wallet opened and the `Earlier wallet preserved` banner rendered
- Before/after SHA-256: all six legacy files (DB/WAL/SHM/manifest/seed salt/ciphertext) and both canonical seed files were unchanged
- No migration journal or quarantine artifact was created

Claude focused adversarial review is running against the exact PR diff; GitHub CI will provide native Windows coverage.

PR #297, opened 2026-08-08
fix(zuuli): remove fake article upvote

### LINKED ISSUE #249: ZUULI Articles: remove the fake local-only upvote or make it server-backed
## Problem

`/articles/:slug` presents a working upvote control and shows “Thanks for the upvote,” but `Reader.tsx` explicitly implements only local React state because there is no vote endpoint. Anonymous users can vote, refresh resets the result, and no other reader sees it.

This is deceptive product behavior rather than an optimistic UI: success is reported for an action that never persists.

## Acceptance criteria

Choose one honest behavior:

- [ ] Implement an authenticated, idempotent server-backed vote whose state survives reload/device changes, prevents duplicate voting, and rolls back on failure; or
- [ ] Remove/disable the control and success message until the contract exists.
- [ ] Logged-out users get an explicit sign-in path rather than a phantom vote.
- [ ] The displayed score always comes from authoritative state after reconciliation.

skyl:
## Summary

- replace the article reader fake upvote button with a passive server-provided score
- remove local score mutation, pressed state, and false success toast
- cover the non-interactive score contract with a focused rendering test

## Contract audit

The backend route POST /api/zpage/{free2zaddr}/vote/ is authenticated, but each request spends 1 2Z and mutates the score again. It provides no idempotency or reconciled per-user vote state, so ZUULI cannot honestly expose it as a toggle.

## Verification

- npm test — 20 passed
- npm run typecheck
- npm run build
- git diff --check

Closes #249

### DIFF
--- a/wallet/zuuli/src/features/articles/components/ArticleScore.test.tsx
+++ b/wallet/zuuli/src/features/articles/components/ArticleScore.test.tsx
@@ -1,0 +1,15 @@
+import { renderToStaticMarkup } from "react-dom/server";
+import { describe, expect, it } from "vitest";
+import { ArticleScore } from "./ArticleScore";
+
+describe("ArticleScore", () => {
+  it("renders the authoritative score as non-interactive status", () => {
+    const markup = renderToStaticMarkup(<ArticleScore score={128} />);
+
+    expect(markup).toContain('aria-label="Article score: 128"');
+    expect(markup).toContain("Score");
+    expect(markup).toContain("128");
+    expect(markup).not.toContain("<button");
+    expect(markup).not.toContain("aria-pressed");
+  });
+});
--- a/wallet/zuuli/src/features/articles/components/ArticleScore.tsx
+++ b/wallet/zuuli/src/features/articles/components/ArticleScore.tsx
@@ -1,0 +1,19 @@
+import { ArrowUp } from "lucide-react";
+
+interface ArticleScoreProps {
+  score: number;
+}
+
+/** Displays the server-provided article score without implying it is editable. */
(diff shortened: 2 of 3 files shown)


skyl:
## Exact-head adversarial review\n\nReviewed head: `80ea138925cd3568a8bcc80fa0707b3a565d9ced` against base `cf04b5283ff15fb51d6dc9cb75c291afd0d10d8e`.\n\nNo blocking findings. The backend vote POST has no idempotency, request/response vote state, or per-user reconciliation contract, so presenting it as a toggle would remain dishonest. This change correctly removes local score mutation, vote/removal action semantics, and the false success toast while retaining the server-derived ranked score as passive visible text. Real and mock modes use the same passive behavior; signed-out users see no phantom voting affordance.\n\nInformational only: the static-markup test cannot prove screen-reader behavior for the generic div, but visible `Score 128` text remains accessible and false button/pressed semantics are gone; `article.votes ?? 0` affects only manually constructed articles because production mapping and mock fixtures supply numeric scores.\n\n**VERDICT: APPROVE**

skyl:
## Rebased exact-head revalidation\n\nReviewed head: `f538352b728cc99689d1e60fa6c27740b6ec4ee0` on current main `1e67b7bfd5499b91e2673646b9bd5fc656b22771`. The sole patch is range-diff identical to the previously approved head. Merged #273 changed wallet migration/status/AppShell surfaces only; article API/types/mock data and Reader inputs are unchanged. No new findings: the patch still removes local vote mutation, toggle/button semantics, and false success, while displaying GET-derived `f2z_score` passively and consistently in real/mock modes.\n\n**VERDICT: APPROVE**

PR #306, opened 2026-08-08
fix(wallet): restore sync after aborted transitions

### LINKED ISSUE #241: Wallet: restore previous sync after aborted transition errors
## Problem

Create/restore currently stop and join the active wallet sync before network birthday/tree-state work. A later network, database, custody, or manifest error preserves wallet state correctly but can leave the previous wallet sync stopped until a manual restart. Similar early-return paths must be audited in switch/delete.

## Acceptance

- transition guard records whether the previous context was syncing
- every pre-commit failure restarts exactly that unchanged wallet context when appropriate
- no restart occurs after a new context has committed or during shutdown
- cancellation/timeout/panic/drop boundaries are deterministic
- regression tests cover every early-return stage and prove context identity
- full wallet gates and Claude adversarial review

Tracked by #227; observed during #237 review.

skyl:
## Summary

- retain an identity-bound recovery obligation whenever create, restore, switch, or active-delete stops the current wallet sync
- replace the split sync flag/handle lifecycle with one generation-aware `Idle -> Running -> Stopping -> Idle` supervisor
- transfer and abort the old handle and launch an independent join-finalizer before the first cancellable wait, so Drop recovery cannot overlap or overwrite it
- restore the exact prior wallet synchronously on ordinary pre-commit errors and through a transition-lock-serialized Drop path on cancellation, timeout, or unwind
- keep create/restore auto-start registration inside the wallet transition lock and make shutdown a task-creation linearization boundary
- retain cancellation-atomic manifest publication and live DB/read/seed/proposal/broadcast context swaps

## Verification

Exact head: `cc6c2ecc6fe01b267b14d8130640bea5743abae5`, rebased on live `origin/main` `3f7845c5cef48fa9f928f566a29d63896da64539`.

- `CARGO_INCREMENTAL=0 cargo +1.88.0 check --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml`
- `CARGO_INCREMENTAL=0 cargo +1.88.0 test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml` — 93 passed
- scheduler-aware supervisor regressions — 3 passed, exercising real Tokio task handles, abort, task Drop, lifecycle locks, blocked finalization, replacement registration, duplicate rejection, and shutdown
- Drop transition-lock regression plus recovery policy tests — 3 passed
- `CARGO_INCREMENTAL=0 cargo +1.88.0 build --locked --manifest-path wallet/zuuli/src-tauri/Cargo.toml`
- `CARGO_INCREMENTAL=0 cargo +1.88.0 test --locked --manifest-path wallet/zuuli/src-tauri/Cargo.toml` — 12 passed
- `CARGO_INCREMENTAL=0 cargo +1.88.0 build --locked --manifest-path wallet/zuuallet/src-tauri/Cargo.toml`
- `git diff --check`

Repository-wide `cargo fmt --check` remains baseline-red on pre-existing formatting across this crate; no bulk formatting changes are included in this focused repair.

## Safety invariants

- at most one live sync task and one owned handle exist at a time
- a recovery start waits until the aborted predecessor is joined and its matching-generation finalizer reaches `Idle`
- a stale generation cannot clear a replacement task flag or overwrite its handle
- cancellation before handle transfer leaves the original task running; cancellation after transfer cannot orphan finalization
- a previously stopped wallet stays stopped, and recovery requires the same non-empty active wallet ID
- a committed transition can never revive its predecessor
- shutdown and sync task creation share a synchronous gate, making shutdown an irreversible no-start boundary
- Drop recovery reacquires the wallet transition lock before checking identity and starting sync

Closes #241

### DIFF
--- a/wallet/plugins/tauri-plugin-zcash/src/commands.rs
+++ b/wallet/plugins/tauri-plugin-zcash/src/commands.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::sync::atomic::Ordering;
 
 use tauri::{command, AppHandle, Runtime};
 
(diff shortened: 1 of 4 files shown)


skyl:
Adversarial self-review of exact head `3593a9dd658fb818812c078072984761e69680f9`: APPROVE. Audited cancellation/drop ordering, transition-lock reacquisition, exact active-wallet identity, previously-stopped behavior, shutdown races, same-wallet switch, manifest publication boundaries, and atomic DB/read/seed/proposal/broadcast swaps across create/restore/switch/active-delete. No unresolved correctness or security findings. Verification: Rust 1.88 locked check green; 91/91 all-target tests green; exact-head recovery tests 4/4 green. `clippy -D warnings` is independently baseline-red on seven unrelated pre-existing warnings and produced no finding in this implementation.

skyl:
REQUEST_CHANGES — adversarial review of exact head 3593a9dd658fb818812c078072984761e69680f9 against live base 4bee0cdfa777d1be45e70629c8f517a4bc2bb87c.

Blocking cancellation race: StoppedSyncRecovery::stop constructs the recovery guard and then awaits wallet::sync::stop_sync (commands.rs:80-95). stop_sync first writes syncing=false, then separately awaits sync_handle.lock() before taking/aborting the old task (sync.rs:1300-1311). If the transition future is cancelled at that lock boundary, Drop schedules recovery while the old sync task and handle are still live. start_sync consults only the shared boolean and never the existing handle (sync.rs:135-161), so recovery can set the flag true, spawn a second DB-mutating sync, and overwrite the old handle. Alternatively, if the old task already observed false, its unconditional final write of syncing=false (sync.rs:969-975) can occur after recovery spawns the replacement, immediately cancelling that replacement and reproducing #241. This violates the PR guarantees of deterministic cancellation and exactly restored sync context.

The four new tests do not exercise this mechanism: PRECOMMIT_FAILURE_STAGES and ABORT_BOUNDARIES are string lists, and every assertion calls only the pure should_restart predicate. They cannot detect the stop/join race, Drop spawning behavior, handle replacement, lock reacquisition, or any actual early-return stage. Please make stop/take/abort recovery cancellation-safe and add a controlled async regression that cancels at the post-flag/pre-handle boundary and proves one live handle plus continued sync for the exact prior identity.

Evidence: all exact-head CI is green, including gate, Rust/shared plugin, zuuallet Rust, macOS and Windows custody, Android AAB, iOS app, Linux packages, and macOS universal packaging. Those build/unit gates do not cover this scheduler interleaving. VERDICT: REQUEST_CHANGES.

skyl:
Addressed the blocking cancellation review at exact head `cc6c2ecc6fe01b267b14d8130640bea5743abae5` (live base `3f7845c5cef48fa9f928f566a29d63896da64539`).

The split boolean/handle protocol is gone. `SyncTaskSupervisor` now owns a generation-tagged `Idle / Running { handle } / Stopping` lifecycle under one mutex. A stop that reaches `Running` transfers the handle, enters `Stopping`, aborts, and spawns an independent handle join-finalizer without any intervening await. Cancelling the stop caller before transfer changes nothing; cancelling it after transfer cannot orphan the join. Recovery start waits through `Stopping`, and only the matching generation can clear the shared flag and return to `Idle`, so it cannot duplicate a task, overwrite a handle, or accept a stale finalizer write.

I also closed adjacent races found during review: start registration and shutdown now share a synchronous linearization gate, and create/restore keep the wallet transition lock through automatic task registration.

Scheduler-aware evidence uses actual Tokio handles, `abort`, task `Drop` sentinels, held lifecycle/sync locks, a cancelled stop caller, a blocked detached finalizer, concurrent recovery, duplicate starts, and shutdown. It proves one live generation before and after recovery and that the replacement flag remains true. Exact-head results: plugin `check --locked --all-targets` green; plugin `test --locked --all-targets` 93/93; ZUULI backend build + 12/12 tests; Zuuallet backend build green; `git diff --check` green. Full commands are in the updated PR body.

skyl:
APPROVE — independent adversarial re-review of exact head cc6c2ecc6fe01b267b14d8130640bea5743abae5 against live base 3f7845c5cef48fa9f928f566a29d63896da64539.

The prior blocking race is closed. SyncTaskSupervisor linearizes lifecycle state under one mutex: Running transfers its generation-tagged handle to Stopping, aborts it, and registers the independent join-finalizer before the first cancellable wait. A cancelled stop before transfer changes nothing; after transfer, the detached finalizer owns completion. Starts wait through Stopping, and finish_running / finish_stopping mutate the flag and phase only for the matching generation. Because replacement registration occurs only after the old handle joins, task-body flag writes cannot land after the replacement. Duplicate starts return AlreadyRunning without polling a second task. Lock ordering is consistently supervisor-state then syncing, with no reverse acquisition in production paths.

Shutdown and spawn share the synchronous start_gate, so begin_shutdown cannot return before an in-flight registration finishes and no later registration can spawn. Drop recovery reacquires wallet_transition before identity revalidation/start, and create/restore retain that transition through automatic registration; the former stop/start and post-commit context races are therefore serialized.

Local exact-head scheduler evidence is real rather than model-only: supervisor_tests 3/3 passed using Tokio JoinHandles, abort, Drop sentinels, held lifecycle/sync locks, cancelled stop callers, blocked finalization, concurrent recovery, duplicate starts, and shutdown. sync_recovery_tests 3/3 passed, including actual async Drop recovery waiting behind wallet_transition. git diff --check is clean.

The complete exact-head CI matrix is green: required gate; ZUULI backend and shared plugin; Zuuallet Rust; macOS and Windows custody; Android AAB; iOS unsigned app; Linux packages; macOS universal app; release identity and both frontends. No unresolved correctness, deadlock, identity, shutdown, or test-strength findings. VERDICT: APPROVE.

Issue #305, opened 2026-08-08
Epic: world-class E2EE messaging — research, options, and recommended architecture


skyl:
# Epic: world-class E2EE messaging for free2z — research, options, and recommended architecture

This is the research/specification issue for rebuilding p2pe2e into a genuinely
world-class end-to-end encrypted messaging system. **No code ships from this
issue.** Its deliverable is a committed design document plus a set of scoped
child issues. The purpose is to make the architectural decisions *deliberately
and in writing*, before anyone implements.

It supersedes nothing — it builds directly on the prior analysis in #131
(unreliable networking), #132 (DKG/FROST requirements), #133 (the server is
trusted today), and #134 (proposal for a secure data structure). Those four
issues are the best thinking we have on this and much of the recommendation
below is an attempt to honor them rather than replace them.

---

## 1. Where we actually are today

Verified against the tree, so we are arguing from facts rather than memory:

- `ts/react/free2z/src/lib/p2pe2e/` is **642 lines** across
  `webrtcmanager.ts`, `peerconnectionmanager.ts`, `datachannelmanager.ts`, and
  `websocketmanager.ts`. Grepping it for `encrypt|crypto|subtle|fingerprint`
  returns **nothing**. There is no application-layer cryptography anywhere in
  the p2pe2e client.
- The only crypto-ish dependency in `ts/react/free2z/package.json` is
  `@github/webauthn-json` — passkeys for participant authentication, not
  message encryption.
- `py/dj/apps/efm/consumers.py` (210 lines) is a Django Channels
  `SignalingConsumer`: it authenticates a participant by JWT, tracks presence in
  a cache, and relays `offer` / `answer` / `ice_candidate` messages between
  participants (`handle_signaling_message`).

**Therefore:** today's confidentiality guarantee is *entirely* WebRTC's DTLS
transport encryption, keyed by certificate fingerprints that the server itself
relays inside the SDP. The server can hand participant A an offer pointing at a
machine it controls and MITM the whole session, undetectably, exactly as #133
states. The "E2E" in the current prototype is transport encryption to a peer the
server chose for you.

That is not a criticism of the prototype — it was a prototype, and it proved the
UX. But it means we are not "hardening" an E2EE system; **we are building one for
the first time**, and we get a free hand on architecture.

## 2. Requirements

Restating, with the relaxation the Least Authority feedback prompted:

1. **The server must be incapable of reading messages.** Non-negotiable.
[...]

Issue #134, opened 2025-03-02
proposal for secure datastructure


dominic-leastauthority:
The objective here is to implement all the security properties required by FROST/DKG in (such as authentication and confidentiality) in a verifiable data structure. The goal is that all of the properties security can be verified not just when running the protocol, but during the design stage. This means that the networking used to exchange the messages now does not need to be secure or have any special properties for the DKG to be secure.

First message: initiator announces participant list. this is not explicitly part of the FROST protocol but it is necessary, because participants must agree to the participant list. Also, there is nonce, which makes it possible to run multiple DKG sessions with the same participants, and also ensures that every encryption and signature in the session is unique to that session.  

```
session =
{
  participants:[{SigningKey},...], //the public signing keys of the participants
  nonce: <32byte_random>, //a nonce, so that this run is clearly distinct from other runs with the same participants
  time: <date/time> //optional, not strictly required but probably useful in practice
  //perhaps other FROST parameters such as threshold?
}
```

then each participant broadcasts a part1 message. In the frost protocol, this is a commitment to their secret share. but I am also including some extra information to ensure the required security properties.

```
participant_part1_message = 
{
  session: hash(session), //session hash will be the same on every message
  commitment: part1, //created by frost
  encryption: {EncryptionKey}, // a public curve25519 key, to be used by other participants to encrypt part2
  signature: {Signature} //signed by this participants public signing key
}
```
the encryption key MUST be generated fresh for this session, and the associated private key MUST be discarded after the session is complete. As per frost, part1 must be a fresh random number. The hash of the session is included to prevent replaying participant messages from other sessions. The signature is included so that other participants can verify this message is generated by the correct participant.

when a participant has received a valid set of `participant_part1_message`s they may proceed to part2  
```
part1_id = hash(participant_part1_messages), //all the participant_part1_messages, in the same order as the session participant list
participant_part2_message =
{
  part1: part1_id, //part1 hash will be the same on every message
  shares:[ //same order as session participant list,
    {
       //using libsodium crypto_box (message, nonce, receiver_public_key, sender_private_key)
       box(
         FROST.part2(participants[i]),
         part1_id,
         participants[i].encryption.publicKey,
         self.encryption.privateKey
       )
       //each part 2 message is encrypted to the corresponding participant with the encryption keys announced in part1.
      //note that the part1_id is used as the encryption nonce. This could be random but deterministic is better, because it guarantees that encryptions between the same participants are unique.
    }
  ],
[...]

Issue #133, opened 2025-03-02
current implementation has trusted server


dominic-leastauthority:
Although the name contains "end 2 end" the server can man in the middle any participant. The server is responsible for setting up a p2p connection, by relaying webrtc offers, however, the server could give a participant an offer to connect to a mitm, who would then connect to another participant. This would not be apparent to the participants unless they compared the offers, and every participant would need to compare with every other participant, each time they initiated a connection.

But what is really needed here isn't p2p communication but just an encrypted message, that is:

1. receiver can verify it's really from the sender
2. no one else can read it
3. it cannot be a replay from another session, or previous encrypted message

assuming that participants in the group are identified by keys that are just for the ceremony, and that these are verified out of band then 1 can be satisfied by signing the messages, and 2 can be satisfyed by encrypting to a value that can only 
(every participant can see the other ids, and the ids are also shared out of band, for example, in a group chat that is considered not likely to be under attacker control. (It's best to minimise manual checking, and better if many participants check the same thing (which means everyone has to miss something for it to get through) rather than each participant checking different things (so that it anyone can make a mistake and there isn't any backup))

The replay defence (3rd part) is easy by ensuring that the encryption is always a unique value. For example, by including a known unique value (such as the hash of the whole of part1 part of the data) as part of the encryption key

skyl:
Superseded by the current E2EE architecture and identity/key-transparency trackers #305 and #311; this is not a claim that the trusted-server/MITM problem is implemented away. Seed-derived messaging identity, key transparency, bootstrap-safe anti-equivocation, client gossip, and substitution detection remain open in #311 under #305.

Issue #132, opened 2025-03-01
requirements to support DKG/FROST


dominic-leastauthority:
To implement DKG/FROST on top of this, some there are some technical requirements.
* it must be possible to get part1 from every participant to every other participant (i.e. reliable broadcast)
* it must be possible to get part2 from each participant to every other participant but without showing the server or any other participant.
* participants need to have an agreed list of participants before they generate the key. You can then use it to make threshold signatures but for DKG the theshold is 100%.
* once a participant has received a part1 from every other participant they can move to part2, once they have received a part2 from every participant they can move to part3, but they do not need to know that everyone has received all the parts yet.

There needs to be some way to decide the participants, such as an initiator inviting them.
(there are a lot of ways you could go here, does the initiatior just choose everyone? do the participants know who the others are and can they verify that? is it tied to established cryptographic identities? but I'll just assume there is some way to produce a list of participants and all the other participants have the same list)

broadcasting a message to every participant is quite simple, but requires that networking failures are handled, for example participants need to be able to (automatically) rerequest missing messages if they were disconnected

currently, sending part2 confidentially is implemented using direct p2p connections that are encrypted. As I have described in [this issue](https://github.com/free2z/zuu/issues/131) that will be unreliable due to network issues. Instead I suggest using the same broadcast mechanism as part1, but encrypt each message so that only it's intended recipient can receive it. p2p connections requires a lot of external factors to work, but encryption can be designed and verified on paper. (I'll make another issue about encryption recommendations)

skyl:
Superseded by #314; this is not a claim that the DKG/FROST requirements are implemented. Participant agreement, reliable reconnect behavior, confidential round delivery, retry safety, key deletion, and offline-verifiable transcripts remain open as acceptance criteria in #314.

PR #318, opened 2026-08-08
docs(e2ee): Phase 0 architecture, threat model, and decision records

skyl:
Phase 0 deliverable for #305: the committed design documentation for free2z's
end-to-end encrypted messaging system.

**This is documentation only. No application code, no protocol implementation,
no changes outside `docs/e2ee/`.** Nine new markdown files, 1,873 insertions,
zero deletions.

## What's here

| File | Contents |
|---|---|
| `docs/e2ee/ARCHITECTURE.md` | The design document. |
| `docs/e2ee/THREAT-MODEL.md` | Adversaries, defenses, and the honest gaps. |
| `docs/e2ee/decisions/0001`–`0006` | One ADR per owner decision. |
| `docs/e2ee/README.md` | Index. |

### `ARCHITECTURE.md`

Goals and non-goals; the five-layer stack (client integrity → identity + key
transparency → MLS session crypto → delivery → opportunistic P2P); the key
hierarchy and derivation with an explicit domain-separation table; delivery
service semantics; hash-linked causal ordering; the retention model; the
metadata model and its limits; federation and the relay trust model; the
FROST/DKG application design; a disposition table for #131/#132/#133/#134; and
thirteen open questions left explicitly open rather than invented.

Points worth reviewer attention:

- **Key hierarchy.** Identity keys are seed-derived (restorable from the
  mnemonic) via a hardened ZIP-32-style tree rooted at its own BLAKE2b
  personalization, so it cannot collide with the Sapling or Orchard trees.
  Device keys are generated on-device, never leave it, and are **not**
  seed-derivable, so seed compromise does not retroactively yield every device's
  MLS state. A signed `DeviceCredential` binds the two and rides as the MLS
  `Credential`, so every peer validates the binding during ordinary MLS
  processing.
- **"Delivered" is four states, not one** — `accepted`, `queue-delivered`,
  `device-delivered`, `delivered` (D2's "every registered device"). The relay
  implements only `queue-delivered`, per-queue and unconditionally, because
  cross-queue state would mean knowing the device set. A device MUST NOT ACK
  before its durable local write completes; ACK-on-read plus a crash is
  permanent message loss under delete-on-ack.
- **Retention change semantics are specified**: forward-only from the epoch the
  Commit lands, with a separate, explicitly-labeled `PurgeRequest` that is a
  request and not a guarantee. And the honest version of "negotiated": MLS
  cannot enforce unanimity, so what we deliver is **"no silent change," not "no
  unilateral change."**
- **Hash links detect gaps, not tail truncation.** Stated, with the reason.

### `THREAT-MODEL.md`

Ten adversaries — malicious/compromised free2z server, network observer,
compromised relay operator, compromised participant, compromised device, coerced
server serving targeted JavaScript, global passive adversary, future quantum
adversary, malicious directory witness, lost/stolen device — each with its
defense or a plain statement that there is none.

The targeted-JavaScript entry says **"Defense: there is none"** and explains why
it is unfixable at the protocol layer, which is the entire justification for D1.

Eight known limits are stated up front rather than buried: directory lookup
reveals interest in a handle at connection time; delivery-receipt timing is a
published deanonymization vector against sealed sender; ephemeral mode is
best-effort; tail truncation is undetectable; server-side deletion is auditable
but not verifiable; **no deniability** (MLS signs application messages, so we do
not claim it); push notifications are an unresolved leak; Nostr fallback means
accepting retention.

Also stated: hybrid PQ (X-Wing) gives confidentiality against harvest-now-
decrypt-later, **but authentication remains Ed25519 and is not post-quantum.**

## The six decisions designed to

| ADR | Decision |
|---|---|
| 0001 Platform | ZUULI-first; the web client speaks the same protocol via WASM from the same Rust core, with an explicitly weaker trust model stated in the UI. |
| 0002 Multi-device | Design the protocol, queue model and key hierarchy for per-device fan-out from day one; ship single-device in the UI. |
| 0003 History | Per-conversation retention, negotiated and authenticated inside the MLS group; FROST transcripts exempt and always retained. |
| 0004 Metadata | Key-transparency directory for `@handle` discovery, then sealed sender over opaque pairwise unidirectional queues. |
| 0005 Federation | Full multi-relay; the relay is a separate small Rust service, open source and self-hostable from the first release. |
| 0006 Zcash | Client-side cryptography only; seed-derived messaging identity; no blockchain in the critical path; never reuse Sapling/Orchard keys. |

Each ADR states context, decision, consequences, and alternatives rejected.

## Prior art honored

- **#131** — hash-linked messages give transport-independent causal ordering and
  gap detection; wall-clock timestamps are advisory and never order anything.
  #131's suggestion of server-assigned sequence numbers is **consciously
  declined** with the reason stated (it hands the server reordering and
  censorship power).
- **#132** — reliable broadcast plus confidential per-recipient delivery plus an
  agreed participant list, with the honest caveat that we provide eventual
  delivery with retry, **not** Byzantine reliable broadcast: safety holds against
  an adversarial relay, liveness does not.
- **#133** — key transparency for identity, and DTLS fingerprints authenticated
  inside the MLS group, which makes the server-MITM impossible rather than merely
  detectable.
- **#134** — the signed, session-hash-bound payload structure retained
  essentially verbatim, including the follow-up comment's `participant_id` and
  message `type` fields; the mandatory deletion of the per-session encryption key
  after part 2 made enforceable and testable rather than conventional; and the
  participant-list agreement step upgraded to a real echo broadcast per
  BlockstreamResearch/bip-frost-dkg `CertEq` and ZcashFoundation/frost#577.

## Verified against the tree

- `ts/react/free2z/src/lib/p2pe2e/` is 642 lines across four files; grepping for
  `encrypt|crypto|subtle|fingerprint|sodium|nacl` returns nothing. No
  application-layer cryptography exists.
- `py/dj/apps/efm/consumers.py` is a 210-line Django Channels
  `SignalingConsumer` that relays `offer`/`answer`/`ice_candidate` verbatim.

So #133's MITM is real today, and there is no wire format to preserve.

Refs #305

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -1,0 +1,977 @@
+# free2z E2EE — Architecture
+
+**Status:** Phase 0 design. Nothing described here is implemented.
+**Refs:** [#305](https://github.com/free2z/zuu/issues/305) (epic and decisions),
+[#131](https://github.com/free2z/zuu/issues/131),
+[#132](https://github.com/free2z/zuu/issues/132),
+[#133](https://github.com/free2z/zuu/issues/133),
+[#134](https://github.com/free2z/zuu/issues/134).
+**Companion:** [`THREAT-MODEL.md`](./THREAT-MODEL.md),
+[`decisions/`](./decisions/).
+
+This document is written to be read by a third-party security auditor. Where
+something is undecided, it is listed in [§13 Open questions](#13-open-questions)
+rather than answered by invention. Where a property does not hold, it is stated
+as not holding. Constants, labels and wire encodings marked *proposed* are
+placeholders to be fixed in the Phase 1 specification; they are written down so
+the structure can be reviewed, not so they can be implemented as-is.
+
+---
+
+## 1. Where we are today
+
+Verified against the tree at the time of writing, so the starting point is a
+fact rather than a memory:
+
+- `ts/react/free2z/src/lib/p2pe2e/` is **642 lines** across
+  `webrtcmanager.ts` (151), `peerconnectionmanager.ts` (256),
+  `websocketmanager.ts` (160) and `datachannelmanager.ts` (75). Grepping it for
+  `encrypt|crypto|subtle|fingerprint|sodium|nacl` returns **nothing**. There is
+  no application-layer cryptography in the p2pe2e client at all.
+- `py/dj/apps/efm/consumers.py` is a 210-line Django Channels
+  `SignalingConsumer`. It authenticates a participant with an HS256 JWT signed
+  by `settings.SECRET_KEY`, tracks presence in the Django cache, and in
+  `handle_signaling_message` relays `offer` / `answer` / `ice_candidate`
+  messages verbatim from one participant's channel to another's.
+
+The consequence is exactly what [#133](https://github.com/free2z/zuu/issues/133)
+says: today's confidentiality is *only* WebRTC's DTLS transport encryption, and
+the certificate fingerprints that key it travel through the server inside the
+SDP. The server can hand participant A an offer pointing at a machine it
+controls, and MITM the session undetectably. **We are not hardening an E2EE
+system; we are building one for the first time.** There is no wire format to
+preserve, so we are architecturally unconstrained.
+
+## 2. Goals and non-goals
+
+### 2.1 Goals
+
+1. **The server cannot read message content.** Non-negotiable. No server-side
+   key escrow, no server-held plaintext, no server-mediated key agreement that
+   the server could substitute into.
+2. **The server destroys ciphertext on acknowledged delivery.** The relay may
+   buffer ciphertext for an offline recipient and must delete it once delivery
+   is acknowledged. It is a buffer, not an archive. (Relaxed from the original
+   "must be online" requirement, which made the product unusable.)
+3. **Forward secrecy and post-compromise security** at group scale, not just
+   for 1:1.
+4. **The server cannot lie about who you are talking to.** Key changes are
+   publicly auditable; a server that equivocates leaves non-repudiable evidence.
+5. **A credible, written metadata story** — not just AEAD over the body.
+6. **P2P is an optimization, never a security dependency.** It may fail, be
+   blocked, or be attacked without weakening confidentiality or authenticity.
+7. **Support cryptographic "apps"**, with FROST/DKG
+   ([#132](https://github.com/free2z/zuu/issues/132),
+   [#134](https://github.com/free2z/zuu/issues/134)) as the first one.
+8. **Hybrid post-quantum confidentiality from day one**, not retrofitted.
+9. **Federated by construction** — anyone can run a relay, and clients treat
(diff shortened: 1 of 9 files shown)


skyl:

Reviewed on the owner's behalf per repo doctrine. `gate` green, docs-only, no code paths touched. The document is honest where it matters: it declines to claim deniability, names the PQ authentication gap, admits tail truncation is undetectable, and states that server-side deletion is auditable but not verifiable. That posture is worth more than a longer feature list.

Two findings. Neither blocks Phase 0 — both are traps for the implementation phases, so recording them here and on the relevant phase issues.

### F1 — `part1_id` as a deterministic box nonce is safe only if the part-2 plaintext is deterministic (§11.1, blocks #314)

The design (inherited from #134 and correct as stated) uses `nonce = part1_id` for the per-recipient `box()`. Within a session this is sound: each (sender, recipient) pair has a distinct shared key, and same-nonce-different-key is fine.

**The trap is retry.** `crypto_box` is XSalsa20-Poly1305. If a participant retries part 2 after re-randomizing its FROST polynomial, it encrypts *different* plaintext under the *same* key and *same* nonce — keystream reuse, which XORs the two share sets together and is catastrophic for a DKG.

This is safe only because FROST round-2 shares are deterministic given the round-1 polynomial. That precondition is currently implicit. It needs to be an explicit, tested MUST: **a retry re-sends byte-identical ciphertext, or it does not happen.** Re-running `FROST.part2` with fresh randomness under a retained `part1_id` must be structurally impossible in the state machine, not merely avoided by convention.

The doc's claim that determinism is "stronger than random for this purpose" holds under that precondition and is misleading without it.

### F2 — witness cosigning provides zero anti-equivocation at launch (§9.3, affects #311)

§9.3 presents cross-relay witness cosigning as the anti-equivocation mechanism, and §3.9 of the threat model honestly covers a malicious witness and operator/witness collusion. But the **bootstrap window** is not stated: on day one free2z runs the only relay and is therefore the only witness, and a client's configured witness set is a set of one — us. A log operator that is also the entire witness set can equivocate freely and no cosignature detects it.

The property is real but it is *emergent*, and it does not exist until independent operators do. Until then the only genuine protection is **client gossip** (§9.3, second bullet) and manual safety-number verification.

Ask: state the bootstrap caveat explicitly, and make client gossip a Phase 1 requirement rather than a nice-to-have — it is the sole anti-equivocation mechanism that works with one relay. Also worth defining what a client should do when its witness set has only one member: warn, refuse, or proceed silently. Proceeding silently would overclaim.

### Smaller notes, no action required

- §5.2 offers "AES-128-GCM (or ChaCha20-Poly1305...)" while the X-Wing suite discussed elsewhere is the ChaCha20-Poly1305 one. Already covered by open question §13-B (pin the exact codepoint before Phase 2).
- §13-D (history transfer to a new device vs. forward secrecy) is the right thing to have left open. It is the decision most likely to quietly erode the FS claim later — worth an ADR of its own when D2's multi-device UI ships.

Merging via admin after confirming `gate` green, per the review-governance section of `CLAUDE.md`.

PR #345, opened 2026-08-09
ci(zuuallet): narrow the plugins path filter to tauri-plugin-zcash

### LINKED ISSUE #339: ci(zuuallet): narrow the wallet/plugins path filter to tauri-plugin-zcash
## Problem

`.github/workflows/zuuallet.yml` filters on `paths: wallet/plugins/**` (~lines 21 and 29). That directory currently holds only `tauri-plugin-zcash`, which zuuallet genuinely consumes — so today it is correct by accident.

But it means **any** future crate placed under `wallet/plugins/` triggers zuuallet's full Rust gate — a **45-minute job** — on every PR touching that crate, forever, even for a crate zuuallet never builds against.

## Fix

Narrow both occurrences:

```yaml
paths:
  - 'wallet/plugins/tauri-plugin-zcash/**'
```

## Why now

It is a two-line change that costs nothing today and prevents a standing 45-minute tax the moment a second plugin crate exists. Cheaper to fix before the fact than to notice later as "CI got slow."

## Acceptance criteria

- [ ] Both `paths:` occurrences in `zuuallet.yml` narrowed to `tauri-plugin-zcash/**`.
- [ ] A PR touching only `wallet/plugins/tauri-plugin-zcash/**` still triggers the zuuallet Rust job.
- [ ] A PR touching an unrelated path under `wallet/plugins/` does not.

skyl:
## Why

`.github/workflows/zuuallet.yml` filtered on `wallet/plugins/**` in both its
`pull_request` and `push` triggers. That is correct *only by accident*: the
directory currently holds exactly one crate, `tauri-plugin-zcash`, and that is
the one zuuallet actually consumes.

The moment a second crate lands under `wallet/plugins/`, every PR touching it
would fire zuuallet's full Rust gate — a ~45-minute job spanning
`ubuntu-latest` plus the `custody-platforms` macOS/Windows matrix — for a crate
zuuallet never builds against. That is a standing tax that only gets noticed
later as "CI got slow," by which time the cause is non-obvious. Two lines now,
free forever.

## What I verified

- **Exactly two occurrences.** `wallet/plugins/**` appeared only at lines 21
  and 29, both `paths:` filters, both the same "zuuallet consumes this plugin"
  case. The workflow's other four `wallet/plugins` references were already
  specific to `tauri-plugin-zcash` (`--manifest-path`, rust-cache `workspaces`).
- **Zuuallet's real path dependencies.** `wallet/zuuallet/src-tauri/Cargo.toml`
  has exactly one path dep under `wallet/plugins/`:
  `tauri-plugin-zcash = { path = "../../plugins/tauri-plugin-zcash" }`.
  The plugin's own path deps all point into `z/zcash/librustzcash`, which the
  filter already covers via its `z/zcash/librustzcash` entry. So the narrowed
  glob is a complete cover of the actual dependency graph — no transitive input
  loses coverage.
- **`zuuli.yml` is untouched.** The required `gate` check derives from
  `zuuli.yml`'s `changes` job, which uses a separate shell `case` glob list
  (`wallet/zuuli/*|wallet/plugins/*|...`) evaluated against a local `git diff` —
  a different mechanism entirely from GitHub's `on.paths`. This change cannot
  affect it. (`zuuli-packaging.yml` likewise keeps its own `wallet/plugins/**`
  filter; both are ZUULI's to decide and out of scope here.)
- **YAML parses**, and the two filter lists round-trip as intended.

## Acceptance criteria

- [x] Both `paths:` occurrences narrowed to `wallet/plugins/tauri-plugin-zcash/**`.
- [x] A PR touching only `wallet/plugins/tauri-plugin-zcash/**` still matches the
      filter, so the zuuallet Rust job still triggers.
- [x] A PR touching an unrelated path under `wallet/plugins/` no longer matches.

Note: this PR itself changes only `.github/workflows/zuuallet.yml`, which is its
own filter entry — so zuuallet's jobs do run here, exercising the edited file.

Closes #339

### DIFF
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -18,7 +18,7 @@ on:
   pull_request:
     paths:
       - 'wallet/zuuallet/**'
-      - 'wallet/plugins/**'
+      - 'wallet/plugins/tauri-plugin-zcash/**'
       - 'z/zcash/librustzcash'
       - '.gitmodules'
       - '.github/workflows/zuuallet.yml'
@@ -26,7 +26,7 @@ on:
     branches: [main]
     paths:
       - 'wallet/zuuallet/**'
-      - 'wallet/plugins/**'
+      - 'wallet/plugins/tauri-plugin-zcash/**'
       - 'z/zcash/librustzcash'
       - '.gitmodules'
       - '.github/workflows/zuuallet.yml'


skyl:

All checks green, including the 18m21s `rust` job — which is itself the acceptance evidence that the job still fires for a change matching the narrowed filter. Diff is exactly the two lines and nothing else.

Verification I checked rather than took on trust: `wallet/zuuallet/src-tauri/Cargo.toml` has exactly one path dependency under `wallet/plugins/` (`tauri-plugin-zcash`), and that plugin's own path deps all point into `z/zcash/librustzcash`, which the filter already lists separately. So the narrowed glob loses no transitive input.

Good call flagging `zuuli-packaging.yml` (lines 7, 19) and `zuuli.yml`'s `case` list carrying the same latent breadth, and better judgment leaving them out of a two-line PR. Filing the follow-up.

Issue #311, opened 2026-08-08
E2EE Phase 1: seed-derived identity + key transparency directory


skyl:
Phase 1 of #305. **Blocked on** the Phase 0 architecture document landing.

## Goal
Give every user a messaging identity the server cannot lie about. **This is the issue that closes #133** — today the server relays WebRTC SDP and can substitute its own fingerprints, so it can MITM any session undetectably.

## Scope
- **Seed-derived identity** (D6): hardened, ZIP-32-style derivation on a dedicated messaging-only branch of the existing wallet mnemonic. Restore your wallet, restore your identity.
  - **Hard constraint:** never reuse Sapling/Orchard spending or viewing keys. Different curves (Jubjub/Pallas vs X25519/Ed25519), and key separation is non-negotiable. Derive a distinct seed, then generate messaging keys from it.
  - Separate levels for *identity* keys and *device* keys from the first commit (D2), even though only one device ships.
  - Separate key contexts for deniable conversational authentication vs. signed ceremony payloads. These must never be interchangeable.
- **Key transparency directory** mapping `@handle` → identity key: CONIKS/Parakeet-style auditable log, Merkle prefix tree, signed epoch roots.
- **Anti-equivocation via cross-relay witness cosigning** (D5): each relay signs the log root it observed; two conflicting signed roots for one epoch are non-repudiable proof of misbehavior. No blockchain (D6).
- **Client-side verification UX**: safety numbers / QR, and automatic epoch-root cross-checking.

## Acceptance criteria
- [ ] Key derivation implemented in the Rust core, with test vectors committed.
- [ ] Domain separation between all derived key types documented and tested.
- [ ] KT log with signed epochs; clients fetch and verify inclusion proofs.
- [ ] Witness cosigning implemented; a client presented with conflicting roots produces a portable proof of equivocation.
- [ ] A test demonstrates that a server substituting an identity key is **detected**.
- [ ] No Zcash spending or viewing key is reachable from any messaging key path (test asserts this).

Refs #305, closes #133 when merged.

skyl:
## Finding from the Phase 0 review (PR #318): the bootstrap window has no anti-equivocation

`ARCHITECTURE.md` §9.3 presents cross-relay witness cosigning as the anti-equivocation mechanism, and the threat model §3.9 honestly covers malicious witnesses and operator/witness collusion. But the **launch window** is not stated:

[...]

PR #346, opened 2026-08-09
docs(e2ee): correct unlinkability claim, supersede retention model, decouple witness

### LINKED ISSUE #342: docs(e2ee): correct two false security claims and the superseded retention model
## Problem

Two claims in the merged Phase 0 docs (PR #318) are **false as written**, and the docs are public and linked from a forthcoming free2z post. They should be corrected before anyone — an auditor, a contributor, or a user — relies on them.

### 1. Queue addresses are NOT unlinkable at the relay

- `docs/e2ee/ARCHITECTURE.md` §6.2: *"The two addresses are unlinkable to each other at the relay."*
- `docs/e2ee/THREAT-MODEL.md` §3.3: lists *"send/receive addresses for the same queue are distinct and unlinkable at the relay"* under what a relay **cannot** do.

Both are wrong. An `APPEND` to `send_addr` must be delivered to `recv_addr`'s message list, so **the relay necessarily holds the mapping** — that is what makes delivery work. A relay operator dumping that table pairs correspondents directly. SimpleX does not achieve this property either.

**What the split actually buys, and what the docs should say instead:** a sender cannot read, ACK or delete; a compromised sender-side key cannot drain the queue; and an observer of the relay's front door cannot pair addresses from traffic alone. Those are real and worth stating — they are just not unlinkability against the operator.

### 2. Retention is described as an enforceable group property

`ARCHITECTURE.md` §8 and `decisions/0003-history-retention.md` specify a `GroupContext` retention extension listed in `required_capabilities`, with a two-phase negotiated-unanimity change protocol.

This was **superseded by owner decision** (recorded on #305): enforcing another participant's retention on my device is client-trust theater — a modified client simply lies about its capabilities and keeps everything. The revised model is per-user local retention plus a non-binding, authenticated ephemeral *hint*.

## Also worth fixing while in there

3. **§9.3 couples witnessing to relay operation** (*"each relay in the federation observes and signs the KT log root"*). A witness needs no inbound port, no TLS, no domain, and no database — coupling the roles makes the volunteer ask far larger than it needs to be. They should be separate roles that one operator *may* combine.
4. **§13-A (OpenMLS audit status) is answerable.** OpenMLS received an independent security audit published 2026-05-27; the one High-severity finding was fixed in 0.8.1. Record the floor and close the question.

## Acceptance criteria

- [ ] §6.2 and THREAT-MODEL §3.3 corrected to state what the address split does and does not provide.
- [ ] §8 rewritten to the per-user model; ADR 0003 superseded by a new ADR 0007.
- [ ] §9.3 decouples witness from relay.
- [ ] §13-A closed with the audit reference and the `openmls >= 0.8.1` floor.
- [ ] No security property is stated that the design does not actually deliver.

## Notes

Small, self-contained, and worth doing even though the implementation is shelved — these documents are our public statement of what we intend to build, and an overclaim left standing is worse than no document.

Refs #305.

skyl:
**Documentation only. No code changes.** Corrects false and superseded security
claims in the merged Phase 0 E2EE docs before an auditor, a contributor, or a
user relies on them.

## 1. Queue addresses are not unlinkable at the relay (`ARCHITECTURE.md` §6.2, `THREAT-MODEL.md` §3.3)

Both documents claimed the send and receive addresses of a queue are
"unlinkable at the relay." **This is false.** An `APPEND` to `send_addr` has to
be delivered into `recv_addr`'s message list, so the relay necessarily holds the
mapping — that mapping *is* delivery. A relay operator dumping that table pairs
correspondents directly, with no traffic analysis. SimpleX does not achieve this
property either.

Replaced with what the split actually buys:

- a sender cannot read, ACK or delete;
- a compromised sender-side key cannot drain the queue;
- an observer of the relay's front door cannot pair addresses from traffic alone
  — and this is the only one of the three that fails against the operator.

In `THREAT-MODEL.md` this moves out of §3.3's **cannot** column into **Not
defended**, with a new known limit **§4.9 — The relay knows which queue addresses
are paired**. The §5 claims table's "Relay learns no social graph" row is
corrected from **Yes** to **No**, split from a new "no handles or identity keys"
row that *is* true.

ADR 0004 carried the same false sentence and is corrected too.

## 2. Retention: superseded model replaced (`ARCHITECTURE.md` §8, ADR 0003 → ADR 0007)

The negotiated-unanimity model — a `GroupContext` retention extension in
`required_capabilities`, `RetentionChangeProposal`/`RetentionChangeAck`, two-phase
unanimity — was rejected by the owner on #305 as unenforceable client-trust
theater: a modified client simply lies about its capabilities and retains
everything.

§8 is rewritten to the revised model: **retention is a per-user local choice**,
plus a **non-binding, authenticated ephemeral hint** that conforming clients
honor. Preserved from the old text:

- the server still holds nothing after acknowledged delivery in every mode, and
  it is now labeled as the *only* retention property that is a security claim;
- `PurgeRequest` survives unchanged, because it was already correctly labeled a
  request rather than a guarantee;
- ceremony transcripts are retained by default because the participant wants the
  evidence — no exemption machinery needed.

The hint is a courtesy feature and is **explicitly excluded from every security
claims table**, including a note under `THREAT-MODEL.md` §5 saying so.

**The knock-on is stated honestly:** dropping `required_capabilities` removes the
join-time exclusion of non-conforming clients, and nothing replaces it.

Adds `decisions/0007-retention-per-user.md`; marks 0003 **Superseded** with a
link and a "do not implement" banner, preserved otherwise as the record of a
decision taken and reversed. Open question §13-J is deleted (its premise is gone).

## 3. Witness decoupled from relay (`ARCHITECTURE.md` §9.3)

Was: "each relay in the federation observes and signs the KT log root." Now they
are separate roles that one operator MAY combine. A witness needs **no inbound
port, no TLS certificate, no domain, no public IP and no database** — only
outbound HTTPS and a tiny state file. Coupling the roles makes the volunteer ask
far larger than necessary, and the witness *set* is what makes anti-equivocation
real. ADRs 0005/0006 and `THREAT-MODEL.md` §3.9 updated to match.

## 4. §13-A closed — OpenMLS audit and version floor

OpenMLS received an independent assessment by SRLabs, sponsored by the Sovereign
Tech Agency, [published 2026-05-27](https://blog.openmls.tech/posts/2026-05-27-independent-audit/):
8 findings, one **High** ("improper authentication of MACs"), fixed in 0.8.1 and
0.7.3. Records the floor **`openmls >= 0.8.1`** with that rationale and notes that
the one residual Low-severity finding must be re-checked before release.

§13 gains a **13.1 Closed** subsection so letters stay stable identifiers and old
citations (§13-E, §13-F, §13-G, …) keep resolving; the gaps at A and J are
deliberate and explained.

## Verification

All internal Markdown links and heading anchors across `docs/e2ee/` were
programmatically validated against GitHub's slug rules — 0 broken links, 0
dangling anchors.

Closes #342. Refs #305.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -64,8 +64,9 @@ preserve, so we are architecturally unconstrained.
    ([#132](https://github.com/free2z/zuu/issues/132),
    [#134](https://github.com/free2z/zuu/issues/134)) as the first one.
 8. **Hybrid post-quantum confidentiality from day one**, not retrofitted.
-9. **Federated by construction** — anyone can run a relay, and clients treat
-   relays as untrusted, replaceable infrastructure.
+9. **Federated by construction** — anyone can run a relay, and separately anyone
+   can run a witness (§9.3); clients treat both as untrusted, replaceable
+   infrastructure.
 
 ### 2.2 Non-goals
 
@@ -89,7 +90,7 @@ preserve, so we are architecturally unconstrained.
 | Layer | Choice | Standard |
 |---|---|---|
 | 0. Client integrity | ZUULI native (strong) / web WASM (stated-weaker) | [ADR 0001](./decisions/0001-platform-priority.md) |
-| 1. Identity + directory | Seed-derived identity; key-transparency log with cross-relay witness cosigning | CONIKS / SEEMless / Parakeet lineage |
+| 1. Identity + directory | Seed-derived identity; key-transparency log with independent witness cosigning (§9.3) | CONIKS / SEEMless / Parakeet lineage |
 | 2. Session crypto | **MLS**, hybrid PQ (X-Wing) from day one | [RFC 9420](https://datatracker.ietf.org/doc/html/rfc9420), [RFC 9750](https://datatracker.ietf.org/doc/rfc9750/), [draft-ietf-mls-extensions](https://www.ietf.org/archive/id/draft-ietf-mls-extensions-04.html) |
 | 3. Delivery | SMP-style opaque pairwise unidirectional queues, delete-on-ack, federated | [SimpleX SMP](https://simplex.chat/docs/simplex.html) design, clean-room implementation |
 | 3b. Fallback delivery | Nostr relays, optional, retention consciously accepted | [NIP-40](https://github.com/nostr-protocol/nips/blob/master/40.md) is advisory only |
@@ -134,8 +135,24 @@ tightly coupled to Nostr transport, which conflicts with delete-on-ack (§6.7).
 We should study MDK closely and reuse its lessons without adopting its transport
 coupling.
 
-> **Open question (§13-A):** OpenMLS's third-party audit status must be
-> established in writing before release. Do not treat "widely used" as audited.
+**Audit status — established, and it is why the version floor exists.** OpenMLS
+received an independent security assessment by SRLabs, sponsored by the Sovereign
+Tech Agency, published
+[2026-05-27](https://blog.openmls.tech/posts/2026-05-27-independent-audit/).
+Eight findings, one of them **High severity — "improper authentication of
+MACs."** The High finding and all but one of the rest are fixed in **0.8.1** (and
+0.7.3 on the previous line).
+
+**Floor: `openmls >= 0.8.1`.** Any earlier version contains a High-severity
+authentication defect and MUST NOT be shipped. One Low-severity finding was still
+open at publication; **review its status before release** rather than assuming the
+audit closed everything. Note also that the version floor interacts with the
+crypto provider: `openmls_rust_crypto` does not implement X-Wing, so the PQ
+ciphersuite (§5.2) requires the libcrux provider.
+
+This closes §13-A. It does not remove the case for our own review — an audit of
+the library is not an audit of how we use it — but it does mean "widely used" has
+been replaced by a published assessment with a named auditor and a fixed version.
 
 ## 4. Identity and the key hierarchy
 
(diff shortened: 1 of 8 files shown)


skyl:

`gate` green, docs-only. Reviewed on the owner's behalf per repo doctrine.

The §6.2 rewrite is exactly right, and the line that earns the merge is *"Only the third of these survives against a relay operator with database access, and against that adversary it fails completely."* That is the sentence an auditor needed and the original text actively denied.

**Three out-of-brief calls, all correct:**

1. **ADR 0004 carried the same false sentence.** Catching and fixing the identical overclaim where the issue didn't list it is the right instinct — a corrected §6.2 sitting next to an uncorrected ADR would have been worse than neither.
2. **"Relay learns no social graph" flipped to No**, split from a new *"no handles or identity keys"* row that is genuinely true. The two were conflated before, and the honest answer is that the relay holds a pseudonymous graph over rotating opaque addresses — which §4.9 now states precisely, including that rotation is what limits the damage.
3. **§13 letters kept stable** rather than renumbered, with A and J moved to a Closed subsection. Renumbering would have silently broken the live §13-E/F/G/H citations elsewhere in the tree. Validating anchors afterward was the right check.

Excluding ephemeral hints from the claims table *with an explicit note explaining the omission* is a better reading of the rule than a softened "No" row would have been. A reader scanning for security properties should not meet a courtesy feature there, and the note stops the absence reading as an oversight.

**One thing verified, not a finding:** the table asserts **Yes** for "Server cannot MITM the identity", which is only true once independent witnesses exist. `THREAT-MODEL.md` line 76 already conditions it on "an independent witness set exists", so the claim is bounded rather than glossed. The launch-window gap — where free2z is the only relay and only witness, making cosigning detect nothing — is tracked on #311 as a Phase 1 requirement (client gossip is the only mechanism that works with a witness set of one). No change needed here.

Issue #338, opened 2026-08-09
Rust toolchain: single source of truth for the version pin, plus a bump policy


skyl:
## Problem

The Rust toolchain version is **duplicated in 11 places** with no single source of truth, and the pinned version (1.88.0, released June 2025) is now ~14 months old.

Every literal pin found today:

| Location | Value |
|---|---|
| `wallet/rust-toolchain.toml:2` | `channel = "1.88.0"` |
| `wallet/zuuli/src-tauri/Cargo.toml:5` | `rust-version = "1.88"` |
| `wallet/zuuallet/src-tauri/Cargo.toml:5` | `rust-version = "1.88"` |
| `wallet/plugins/tauri-plugin-zcash/Cargo.toml:5` | `rust-version = "1.88"` |
| `.github/workflows/zuuli.yml:152,194` | `toolchain: 1.88.0` (×2) |
| `.github/workflows/zuuallet.yml:142` | `dtolnay/rust-toolchain@1.88.0` |
| `.github/workflows/zuuli-packaging.yml:49` | `ZUULI_RUST_VERSION: "1.88.0"` |
| `.github/workflows/zuuli-packaging.yml:115,209,308` | `toolchain: 1.88.0` (×3) |
| `.github/workflows/zuuli-release.yml:42` | `ZUULI_RUST_VERSION: "1.88.0"` |
| `.github/workflows/zuuli-release.yml:215,362,…` | `toolchain: 1.88.0` |

Note `ZUULI_RUST_VERSION` is declared **twice** (packaging and release) rather than once, and the release workflows verify it with `rustc --version | grep -F "rustc $ZUULI_RUST_VERSION "` — so a partial bump fails loudly in release rather than at PR time, which is the worst place to find out.

There is also **no stated policy**: no cadence, no criteria for bumping, no owner. The pin just ages.

## Why it matters now

- Modern crates increasingly declare MSRVs above 1.88. This has already been hit in dependency evaluation: `openmls 0.9` requires **1.91**, and `openmls 0.8.1` declares no `rust-version` at all, so its true floor is unknown against our pin.
- `AGENTS.md` already documents rustc-version skew between macOS dev machines and Linux CI as a known bug source.
- A bump today means editing 11 places correctly, in lockstep, across 4 workflows. That friction is itself the reason it hasn't happened.

## Proposal

[...]

PR #347, opened 2026-08-09
ci(rust): single source of truth for the toolchain pin, with drift check

skyl:
## The version is UNCHANGED at `1.88.0`

Nothing in this PR moves the Rust toolchain. Every `1.88.0` / `1.88` in the tree
still says `1.88.0` / `1.88`. This is purely the deduplication half of #338 —
one source of truth plus a check that fails on drift. **The bump stays open**;
it is an owner-gated decision that moves every mobile and desktop release target
at once.

## What changed

`wallet/rust-toolchain.toml` is now the only place the version is decided, and
its `channel` grew a header saying so.

**The required `wallet/zuuli` gate now holds no literal at all.** Its two Rust
jobs read the channel out of the file and feed it to the toolchain action:

```yaml
- name: Resolve the pinned Rust toolchain
  id: rust_toolchain
  run: |
    version=$(scripts/check-rust-toolchain.sh --print-channel)
    echo "version=$version" >> "$GITHUB_OUTPUT"

- uses: dtolnay/rust-toolchain@stable
  with:
    toolchain: ${{ steps.rust_toolchain.outputs.version }}
```

The seven `toolchain: 1.88.0` literals in `zuuli-packaging.yml` and
`zuuli-release.yml` became `${{ env.ZUULI_RUST_VERSION }}`, which those workflows
already declare and already verify at runtime.

Counting honestly, rather than claiming a number that sounds better than it is:

| | before | after |
|---|---|---|
| version literals in the required `zuuli.yml` gate | 2 | **0** — it reads the file |
| `toolchain:` literals in packaging + release | 7 | **0** |
| `ZUULI_RUST_VERSION` declarations | 2 | 2 (see below) |
| version-branch action refs (`uses:` cannot take an expression) | 1 | 1 |
| commit-pinned SHAs that silently *are* the version | 7, unlabelled | 7, each labelled `# 1.88.0` |
| **restatements verified against a source of truth** | **0 of them** | **all of them** |

The point is not the count. Before this PR, twelve version literals and seven
invisible SHA pins agreed only by luck and by whoever last edited them. Now
nothing restates the version without CI proving it still matches.

## A finding worth reading: the SHA *is* the pin

`zuuli-packaging.yml` and `zuuli-release.yml` pin
`dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b`. That commit is
the tip of the action's `1.88.0` branch, and on the version branches the action

- **hardcodes** `toolchain: 1.88.0` in `action.yml`, and
- **does not declare a `toolchain` input at all**.

So the commit-pinned SHA selects the compiler entirely on its own, and the
`toolchain:` sitting beside it has been inert (and emitting "Unexpected input(s)"
warnings) all along. It was effectively a *twelfth*, undocumented pin — the one
place a bumper would be most likely to miss, in the two workflows where missing
it costs the most.

That is now recorded where it can be checked:

```yaml
- uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
```

The `# <version>` comment is the only human- and machine-readable record of what
the SHA installs, and the drift check requires it to match the source of truth.
The inert input was kept (pointed at the verified env var) rather than deleted,
because deleting it is a behavioural claim about workflows that cannot be
exercised from a PR; it costs nothing to leave, and the comment now carries the
truth.

## What the check covers

`scripts/check-rust-toolchain.sh` parses `channel` from
`wallet/rust-toolchain.toml` and asserts:

| Pin | Rule |
|---|---|
| `rust-version` in the 3 crate manifests | equals the **two-component MSRV floor** (`1.88`) of the three-component channel (`1.88.0`) — compared as the different forms they are, not forced equal |
| `ZUULI_RUST_VERSION` in packaging + release | equals the channel, and is still declared in both (the release jobs' `rustc --version \| grep -F` safety net depends on it) |
| every `dtolnay/rust-toolchain@<sha>` | carries a `# <version>` comment equal to the channel |
| every `dtolnay/rust-toolchain@<version>` | equals the channel |
| every `toolchain:` input | equals the channel, or is one of two expressions that provably lead back to the file — anything else is rejected rather than trusted |
| MSRV / `cargo +<version>` prose in 2 READMEs, the plugin `CLAUDE.md`, `wallet/zuuli/docs/releasing.md`, and `AGENTS.md` | says exactly what it should |

Plus blindness guards: if the patterns ever stop matching any
`dtolnay/rust-toolchain` step or any `toolchain:` input, the check fails rather
than passing vacuously. The two deliberately-floating `@stable` steps in
`zuuallet.yml` (the reference-wallet build and the upstream canary) are reported,
not failed.

## Demonstration: fails on drift, passes on the tree as merged

Passing, unmodified:

```
note: 2 rust-toolchain step(s) intentionally float on a moving ref (upstream canaries).
Every Rust toolchain pin agrees with wallet/rust-toolchain.toml: channel 1.88.0, MSRV 1.88.
$ echo $?
0
```

Failing, after changing **one** literal
(`ZUULI_RUST_VERSION: "1.88.0"` → `"1.90.0"` in `zuuli-packaging.yml`):

```
drift: .github/workflows/zuuli-packaging.yml:52 declares ZUULI_RUST_VERSION: 1.90.0, expected 1.88.0
drift: .github/workflows/zuuli-packaging.yml does not declare ZUULI_RUST_VERSION: "1.88.0"
2 pin(s) disagree with wallet/rust-toolchain.toml (channel 1.88.0, MSRV 1.88).
$ echo $?
1
```

Failing, after bumping **only** the source of truth to `1.89.0` — this is the
one-line-bump case, and the check turns it into a worklist:

```
drift: .github/workflows/zuuallet.yml:145 uses dtolnay/rust-toolchain@1.88.0, expected @1.89.0
drift: .github/workflows/zuuli-packaging.yml:121 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-packaging.yml:215 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-packaging.yml:314 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-packaging.yml:52 declares ZUULI_RUST_VERSION: 1.88.0, expected 1.89.0
drift: .github/workflows/zuuli-release.yml:221 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-release.yml:368 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-release.yml:584 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-release.yml:681 pins dtolnay/rust-toolchain by commit commented '# 1.88.0', expected '# 1.89.0'
drift: .github/workflows/zuuli-release.yml:45 declares ZUULI_RUST_VERSION: 1.88.0, expected 1.89.0
drift: wallet/zuuli/src-tauri/Cargo.toml declares rust-version = "1.88", expected "1.89" (the MSRV form of channel 1.89.0)
drift: wallet/zuuallet/src-tauri/Cargo.toml declares rust-version = "1.88", expected "1.89" (the MSRV form of channel 1.89.0)
drift: wallet/plugins/tauri-plugin-zcash/Cargo.toml declares rust-version = "1.88", expected "1.89" (the MSRV form of channel 1.89.0)
drift: wallet/zuuallet/README.md no longer says: - [Rust](https://rustup.rs/) 1.89+ (edition 2024)
drift: wallet/plugins/tauri-plugin-zcash/README.md no longer says: cargo +1.89.0 check --locked
...
21 pin(s) disagree with wallet/rust-toolchain.toml (channel 1.89.0, MSRV 1.89).
```

(Both mutations were reverted; the branch is the passing tree.)

### The check is itself checked

`--self-test` stages the files into a scratch tree, confirms the untouched copy
passes, then breaks one pin at a time and confirms each break is caught. A check
nobody has seen fail is not a check. CI runs it before trusting the verdict:

```
self-test: the unmodified tree passes.
self-test: caught drift when a commit-pinned action loses its version comment.
self-test: caught drift when a workflow declares a stale ZUULI_RUST_VERSION.
self-test: caught drift when a version-branch action ref drifts.
self-test: caught drift when the required gate stops reading the source of truth.
self-test: caught drift when a crate manifest MSRV drifts.
self-test: caught drift when documentation still names the old version.
self-test: caught drift when only the source of truth is bumped and nothing follows.
self-test: 7 drift case(s) caught.
```

Each case also asserts its own mutation actually changed the file, so a future
refactor cannot silently turn a case into a no-op.

## Wiring, and what was deliberately left alone

The check runs as a step in `zuuli.yml`'s `changes` job — the only job that runs
on **every** event. `changes` is already in `gate`'s `needs`, so drift fails the
required check without touching `gate`'s `needs:` or its `results` logic, and
without touching the `changes` path filters. All three were left byte-identical
on purpose: a toolchain pin can drift from a workflow the filters intentionally
skip, so the check had to live somewhere unconditional.

Also unchanged: the two floating `@stable` steps in `zuuallet.yml` (pinning the
reference wallet or the canary is a separate call), and the release workflows'
`rustc --version | grep -F "rustc $ZUULI_RUST_VERSION "` verification, which is
kept as the last line of defence.

`ZUULI_RUST_VERSION` is still declared twice. GitHub cannot compute a
workflow-level `env:` from a file, and both workflows' Rust cache keys interpolate
it before any job step runs, so sourcing it once would mean restructuring every
job onto a new `needs:` dependency — real risk to the release path for no gain
now that the two declarations are held to the source of truth by CI. The
rationale is recorded in a comment beside each declaration.

## Policy

`AGENTS.md` gains a "The Rust toolchain pin" section: where the version lives, a
table of every restatement and *why* it cannot read the file, the bump procedure
(change one line, run the check, fix exactly what it names), and — proposed, not
enacted — a cadence of **stable minus one or two releases, reviewed quarterly,
with a named owner**, noting the current pin is ~14 months old. The current-pin
sentence is itself covered by the drift check, so it cannot rot.

## Verification

- `scripts/check-rust-toolchain.sh` and `--self-test` pass on this branch.
- All 8 workflows parse (`yaml.safe_load`); `gate.needs` is unchanged.
- Every rewritten `toolchain:` value parses to the intended expression, including
  the two flow-mapping styles in `zuuli-release.yml`.
- `wallet/zuuli/scripts/verify-ci-cache-policy.mjs` still passes.
- Script is bash 3.2 compatible (verified on `/bin/bash` 3.2.57) and committed `0755`.

Refs #338 — the version bump itself remains open there.

### DIFF
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -139,6 +139,9 @@ jobs:
       - name: Fetch librustzcash submodule
         run: git submodule update --init z/zcash/librustzcash
 
+      # `uses:` cannot take an expression, so this version-branch ref restates
+      # wallet/rust-toolchain.toml; scripts/check-rust-toolchain.sh holds it to
+      # the source of truth.
       - uses: dtolnay/rust-toolchain@1.88.0
 
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -46,6 +46,9 @@ concurrency:
 env:
   CI: true
   ZUULI_NODE_VERSION: "24.18.0"
+  # Restates wallet/rust-toolchain.toml because GitHub cannot compute a
+  # workflow-level `env:` from a file. scripts/check-rust-toolchain.sh fails
+  # the pull request if this ever stops matching the source of truth.
   ZUULI_RUST_VERSION: "1.88.0"
   ZUULI_JAVA_VERSION: "21.0.12"
   ZUULI_XCODE_VERSION: "26.6"
@@ -110,9 +113,14 @@ jobs:
           distribution: temurin
           java-version: "21.0.12"
           cache: gradle
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b
+      # dtolnay/rust-toolchain's version branches hardcode the compiler in
+      # action.yml, so this commit is itself the version pin and the trailing
+      # comment is its only readable record. Bumping the toolchain means moving
+      # the SHA to the new version branch and updating the comment with it;
+      # scripts/check-rust-toolchain.sh fails the pull request otherwise.
+      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
         with:
-          toolchain: 1.88.0
+          toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: ${{ github.event_name == 'pull_request' && 'aarch64-linux-android' || 'aarch64-linux-android,armv7-linux-androideabi,i686-linux-android,x86_64-linux-android' }}
       - name: Install exact Android SDK components
         run: |
@@ -204,9 +212,9 @@ jobs:
           node-version: "24.18.0"
           cache: npm
           cache-dependency-path: wallet/zuuli/package-lock.json
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b
+      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
         with:
-          toolchain: 1.88.0
+          toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: aarch64-apple-ios
       - name: Restore credential-free iOS Rust dependencies
         if: github.event_name != 'schedule' && github.event.inputs.cold_rust_cache != 'true'
@@ -303,9 +311,9 @@ jobs:
           node-version: "24.18.0"
           cache: npm
           cache-dependency-path: wallet/zuuli/package-lock.json
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b
+      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
         with:
-          toolchain: 1.88.0
+          toolchain: ${{ env.ZUULI_RUST_VERSION }}
       - name: Install macOS universal Rust targets
         if: runner.os == 'macOS'
         run: rustup target add aarch64-apple-darwin x86_64-apple-darwin
--- a/.github/workflows/zuuli-release.yml
+++ b/.github/workflows/zuuli-release.yml
@@ -39,6 +39,9 @@ concurrency:
 env:
   CI: true
   ZUULI_NODE_VERSION: "24.18.0"
+  # Restates wallet/rust-toolchain.toml because GitHub cannot compute a
+  # workflow-level `env:` from a file. scripts/check-rust-toolchain.sh fails
+  # the pull request if this ever stops matching the source of truth.
   ZUULI_RUST_VERSION: "1.88.0"
   ZUULI_JAVA_VERSION: "21.0.12"
   ZUULI_XCODE_VERSION: "26.6"
@@ -210,9 +213,14 @@ jobs:
         with:
           distribution: temurin
           java-version: "21.0.12"
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b
+      # dtolnay/rust-toolchain's version branches hardcode the compiler in
+      # action.yml, so this commit is itself the version pin and the trailing
+      # comment is its only readable record. Bumping the toolchain means moving
+      # the SHA to the new version branch and updating the comment with it;
+      # scripts/check-rust-toolchain.sh fails the pull request otherwise.
+      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
         with:
-          toolchain: 1.88.0
+          toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: aarch64-linux-android,armv7-linux-androideabi,i686-linux-android,x86_64-linux-android
       - name: Install exact Android SDK components
         run: |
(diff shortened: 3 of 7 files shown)


skyl:

`gate` green on the merged head (post-#345). Reviewed on the owner's behalf.

**Verified independently rather than taken on trust:**

- **Gate logic byte-identical.** Diffed for any change to `needs:`, `results=` or `*_RESULT` — zero hits. The required check's own semantics are untouched, which was the main risk in a PR that edits four workflows.
- **The drift check actually fails on drift.** Ran it myself: clean tree passes; injecting `rust-version = "1.92"` into `wallet/zuuli/src-tauri/Cargo.toml` produces `drift: … declares rust-version = "1.92", expected "1.88" (the MSRV form of channel 1.88.0)`; restore returns to clean. It names the file and both values, so a future bump is a worklist rather than a hunt.
- **Version unchanged at 1.88.0**, as scoped.

**The finding is the real value here, and I confirmed it against upstream.** At the pinned SHA `2eae45d`, `dtolnay/rust-toolchain`'s `action.yml` declares only `targets`, `target` and `components` — **no `toolchain` input** — and line 51 hardcodes `toolchain: 1.88.0` in the `env:` block. So those commit-pinned refs *were* the version pin, an undocumented twelfth location sitting in the two workflows where missing it costs most, and the `toolchain:` beside them has been inert. Anyone bumping that literal alone would have silently kept building on 1.88.0, caught only by the release workflow's `rustc --version | grep -F` — which is now visibly earning its keep.

**Two restraint calls I agree with:** keeping the inert `toolchain:` input rather than deleting it (deletion is a behavioural claim a PR cannot exercise), and leaving `ZUULI_RUST_VERSION` declared twice, since both workflows interpolate it into cache keys before any step runs and collapsing it would mean restructuring every job onto a new `needs:` — real release risk for no gain now that CI holds both to the source of truth.

The `--self-test` asserting each mutation actually changed the file, so cases cannot decay into no-ops, is the right paranoia for a checker whose failure mode is silent success.

#338 stays open for the version bump itself. Merging via admin after confirming `gate` green.

Issue #351, opened 2026-08-09
deps(rust): shrink the cargo-deny advisory backlog


skyl:
Split out of #340. The `cargo deny` gate landed with the *measured* backlog
written down as individually commented `ignore` entries in `wallet/deny.toml`,
so new advisories fail CI. This issue is the work of shrinking that list.

Baseline on the day the gate landed: **36 findings** — 13 vulnerabilities, 1
unsoundness, 18 unmaintained crates, 1 yanked crate. All transitive; none is a
direct dependency of ours.

## Tier 1 — reachable by a plain `cargo update`, do these first

Semver-compatible bumps. Should be one PR whose diff is three lockfiles.

| Advisory | Crate | Current | Fixed in |
| --- | --- | --- | --- |
| RUSTSEC-2026-0049 | rustls-webpki | 0.103.9 | 0.103.13 (also clears -0098/-0099/-0104 for the 0.103 line) |
| RUSTSEC-2026-0009 | time | 0.3.37 | 0.3.47 |
| RUSTSEC-2026-0204 | crossbeam-epoch | 0.9.18 | 0.9.20 |
| RUSTSEC-2026-0097 | rand | 0.8.5 | 0.8.6 |

Delete the matching entries from `wallet/deny.toml` in the same PR — the gate
then proves the fix is real.

## Tier 2 — needs an intermediate crate to move

- **`aws-lc-sys` 0.37.1 — five vulnerabilities** (RUSTSEC-2026-0044 … -0048),
  two of them requiring 0.39.0. For a `0.x` crate that is a semver-major bump,
  so `cargo update` cannot reach it: `aws-lc-rs` 1.15.4 has to widen its
  requirement first. This is the highest-value item on the list — it sits under
  the rustls/tonic stack the wallet uses to reach lightwalletd. Track
  aws-lc-rs upstream and bump as soon as it allows 0.39.
- **`rustls-webpki` 0.101.7 — RUSTSEC-2026-0098/-0099/-0104.** The 0.101 line has
  **no fixed release at all**; the fix line starts at 0.103.12. Clearing this
  means finding whatever still pulls `rustls 0.21` and getting it onto 0.23, not
  bumping a version. Start with `cargo tree -i rustls@0.21` on each lockfile.
- **`quick-xml` 0.38.4 — RUSTSEC-2026-0194/-0195.** Fixed in 0.41.0, a
  semver-major bump owned by the Tauri bundler chain. Build-time XML only, so
[...]

PR #353, opened 2026-08-09
chore(rust): bump toolchain 1.88.0 -> 1.97.1

### LINKED ISSUE #338: Rust toolchain: single source of truth for the version pin, plus a bump policy

skyl:
Closes #338.

PR #347 built the machinery — one source of truth plus a drift check. This is the other half of the issue: the bump itself.

## The jump

`1.88.0` (June 2025) → **`1.97.1`** (2026-07-14, `8bab26f4f`) — **nine stable releases, ~14 months**, across the librustzcash graph and Tauri. Verified as current stable against `https://static.rust-lang.org/dist/channel-rust-stable.toml`.

One decision (`channel` in `wallet/rust-toolchain.toml`), then mechanical follow-through. Every file below was **named by `scripts/check-rust-toolchain.sh`**, with line numbers — nothing was found by grepping around.

| File | What moved |
|---|---|
| `wallet/rust-toolchain.toml` | `channel` — **the only decision** |
| `wallet/zuuli/src-tauri/Cargo.toml` | `rust-version` `1.88` → `1.97` |
| `wallet/zuuallet/src-tauri/Cargo.toml` | `rust-version` `1.88` → `1.97` |
| `wallet/plugins/tauri-plugin-zcash/Cargo.toml` | `rust-version` `1.88` → `1.97` |
| `.github/workflows/zuuli-packaging.yml` | `ZUULI_RUST_VERSION` + **3 commit-pinned action SHAs** |
| `.github/workflows/zuuli-release.yml` | `ZUULI_RUST_VERSION` + **4 commit-pinned action SHAs** |
| `.github/workflows/zuuallet.yml` | `dtolnay/rust-toolchain@1.88.0` → `@1.97.1` (version-branch ref) |
| `wallet/zuuallet/README.md` | prerequisite line |
| `wallet/plugins/tauri-plugin-zcash/README.md` | two `cargo +<ver>` lines + MSRV |
| `wallet/plugins/tauri-plugin-zcash/CLAUDE.md` | MSRV prose |
| `wallet/zuuli/docs/releasing.md` | release-train pin |
| `AGENTS.md` | cadence section — proposal → adopted policy |
| `wallet/zuuli/scripts/release-identity.mjs` | a **twelfth** restatement the drift check did not know about — see below |

`scripts/check-rust-toolchain.sh` passes, and `--self-test` still catches all 7 drift cases.

## The SHA swap — the part that silently doesn't work if you skip it

`dtolnay/rust-toolchain`'s **version branches hardcode the compiler and declare no `toolchain` input at all**. The commit-pinned SHA *is* the version pin; the `toolchain:` value beside it is inert and ignored. Updating only the `# 1.88.0` comments would have left CI building on 1.88.0 while every comment claimed 1.97.1 — surfacing first in a protected store release, at the `rustc --version | grep -F "rustc $ZUULI_RUST_VERSION "` step.

So all 7 pinned refs moved:

`2eae45db285e407f22119950686d47e1101e071b` → **`032958afbdc797a9164d3bc0b56325c1308924a5`**

Proof the new SHA installs 1.97.1:

```console
$ gh api "repos/dtolnay/rust-toolchain/contents/action.yml?ref=032958afbdc797a9164d3bc0b56325c1308924a5" \
    -q .content | base64 -d | grep -A2 "^      env:"
      env:
        toolchain: 1.97.1
      shell: bash
```

That SHA is the head of the upstream `1.97.1` branch (`refs/heads/1.97.1`). Its `inputs:` block declares only `targets`, `target`, and `components` — confirming the version branch takes no `toolchain` input, exactly as the doctrine in `AGENTS.md` describes. (`@stable`, used by the required gate, *does* declare `toolchain`, which is why the gate can feed it `--print-channel` and hold no literal.)

CI confirms it end to end. The Android packaging job installed, from the commit-pinned action:

```
1.97.1-x86_64-unknown-linux-gnu installed - rustc 1.97.1 (8bab26f4f 2026-07-14)
rustc 1.97.1 (8bab26f4f 2026-07-14)
```

and GitHub emitted exactly the warning the doctrine predicts, proving the adjacent `toolchain:` value is inert:

```
##[warning]Unexpected input(s) 'toolchain', valid inputs are ['targets', 'target', 'components']
```

## MSRV policy consequence — deliberate, not a side effect

The drift check defines each crate's `rust-version` as the two-component MSRV form of the channel, so bumping the channel necessarily raises all three from `1.88` to `1.97`.

That is the intended policy: **we support what we build with.** None of these crates is published to crates.io, so `rust-version` states our build floor rather than a compatibility promise to downstream consumers. `AGENTS.md` now says so explicitly, alongside the cadence, which this PR turns from a proposal into an adopted policy (**stable or stable-minus-one, reviewed quarterly, owner-gated**) — closing the third acceptance criterion of #338.

## Verification

Built and tested locally on `1.97.1-x86_64-apple-darwin` (`rustc 1.97.1 (8bab26f4f 2026-07-14)`), with `--locked` throughout and `z/zcash/librustzcash` initialised:

- `cargo +1.97.1 test --locked --all-targets` — `wallet/plugins/tauri-plugin-zcash`
- `cargo +1.97.1 build --locked` — `wallet/zuuli/src-tauri`
- `cargo +1.97.1 build --locked` — `wallet/zuuallet/src-tauri`

| Crate | Result |
|---|---|
| `wallet/plugins/tauri-plugin-zcash` | `cargo test --locked --all-targets` — **94 passed, 0 failed** (exit 0) |
| `wallet/zuuli/src-tauri` | `cargo build --locked` — clean in 35m40s; `cargo test --locked` — **12 passed, 0 failed** |
| `wallet/zuuallet/src-tauri` | `cargo build --locked` — clean in 34m15s |

**No source changes were needed.** Nine versions' worth of new and promoted lints — `unsafe_op_in_unsafe_fn`, stricter trait resolution, the usual churn — produced nothing that broke us. `tauri-plugin-zcash` was already edition 2024.

**No `Cargo.lock` changed.** `--locked` held across all three graphs, on cold rebuilds from scratch. Consequently the 32 advisory ignores in `wallet/deny.toml` are untouched and `Rust / supply chain` passes unchanged — nothing from #351 was disturbed.

**`cargo fmt` produced zero changes.** This is the surprising one and worth recording. `scripts/check-rust-fmt.sh` exists precisely because rustfmt's output is not stable across compiler versions, so the expectation going in was a reformat pass. Across nine releases, rustfmt 1.97.1 agrees with rustfmt 1.88.0 on every crate under `wallet/` — verified locally (`RUSTUP_TOOLCHAIN=1.97.1 scripts/check-rust-fmt.sh`) and confirmed green by the `Rust / formatting` job. **There is no reformat commit in this PR because there was nothing to reformat.**

The one thing that did break is below.

## A twelfth restatement the drift check did not know about

The first CI run failed **all five packaging jobs** — including `Release identity`, which finishes in 18 seconds and compiles no Rust:

```
ZUULI release identity is inconsistent:
- Rust toolchain: expected "1.88.0", got "1.97.1"
```

`wallet/zuuli/scripts/release-identity.mjs` reads `wallet/rust-toolchain.toml` and asserts the channel equals a hardcoded literal, so the release train fails closed if the source of truth is edited silently. That is a genuinely useful assertion — its whole job is to disagree with the file — but it was the one restatement `scripts/check-rust-toolchain.sh` had never been taught about. The drift check passed locally, and the bump still broke every installable artifact.

Second commit fixes both halves: the literal moves to `1.97.1`, **and the file joins `DOC_PINS`** so the next bump is named by the drift check like everything else rather than discovered 18 seconds into the packaging matrix. `--self-test` still catches all 7 cases.

This is the acceptance criterion of #338 working as intended, one restatement late.

A local macOS build is not the last word — per `AGENTS.md`, feature unification and platform-gating only show up in a clean Linux build. CI is the real gate.

## Device testing

`wallet/rust-toolchain.toml` is in the path filter of `zuuli-packaging.yml`, so the full packaging matrix runs on this PR and produces installable artifacts:

**All four built green on 1.97.1.** Download from the [`ZUULI / packaging smoke` run](https://github.com/free2z/zuu/actions/runs/31294684142) → **Artifacts** (retained until 2026-08-23):

| Artifact | Job | Time | Size |
|---|---|---|---|
| `zuuli-android-unsigned-7c44a24` | Android / unsigned AAB | 12m30s | 17.1 MB |
| `zuuli-ios-unsigned-7c44a24` | iOS / unsigned target app | 10m14s | 12.5 MB |
| `zuuli-Linux-unsigned-7c44a24` | Desktop / Linux packages | 11m49s | 119.8 MB |
| `zuuli-macOS-unsigned-7c44a24` | Desktop / macOS universal app | 19m3s | 55.7 MB |

Each packaging job runs `rustc --version | grep -F "rustc $ZUULI_RUST_VERSION "` **before** building — those four steps are the running proof that the SHA swap took effect rather than being cosmetic, and all four passed.

## Full CI result

Every check green, including the required `gate`:

- **Gate:** `gate`, `changes`, `Rust / shared plugin` (10m8s), `Rust / ZUULI backend` (12m39s), `Rust / formatting`, `Rust / supply chain`, `frontend`
- **Packaging:** `Release identity`, Android, iOS, Desktop Linux, Desktop macOS — the four installable targets
- **Zuuallet:** `rust` (19m5s), `fmt`, `deny`, `frontend`, `custody / macos-latest`, `custody / windows-latest` (19m15s)

Windows and macOS custody adapters — the platform-specific Keychain and Credential Manager code that only compiles on its native host — both build on 1.97.1.

## Scope

Version bump only, plus the drift-check coverage gap the bump exposed. The `changes` path filters and the `gate` job's `needs:`/`results` logic are untouched, as is `wallet/deny.toml`.

Rebased onto `main` after #349 (`cargo fmt`), #352 (`cargo-deny`), and #344 landed; the `publish = false` addition from #352 is preserved alongside the new `rust-version` in all three manifests.

### DIFF
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -208,7 +208,7 @@ jobs:
       # `uses:` cannot take an expression, so this version-branch ref restates
       # wallet/rust-toolchain.toml; scripts/check-rust-toolchain.sh holds it to
       # the source of truth.
-      - uses: dtolnay/rust-toolchain@1.88.0
+      - uses: dtolnay/rust-toolchain@1.97.1
 
       - uses: Swatinem/rust-cache@v2
         with:
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -49,7 +49,7 @@ env:
   # Restates wallet/rust-toolchain.toml because GitHub cannot compute a
   # workflow-level `env:` from a file. scripts/check-rust-toolchain.sh fails
   # the pull request if this ever stops matching the source of truth.
-  ZUULI_RUST_VERSION: "1.88.0"
+  ZUULI_RUST_VERSION: "1.97.1"
   ZUULI_JAVA_VERSION: "21.0.12"
   ZUULI_XCODE_VERSION: "26.6"
   ZUULI_ANDROID_NDK_VERSION: "27.0.12077973"
@@ -118,7 +118,7 @@ jobs:
       # comment is its only readable record. Bumping the toolchain means moving
       # the SHA to the new version branch and updating the comment with it;
       # scripts/check-rust-toolchain.sh fails the pull request otherwise.
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: ${{ github.event_name == 'pull_request' && 'aarch64-linux-android' || 'aarch64-linux-android,armv7-linux-androideabi,i686-linux-android,x86_64-linux-android' }}
@@ -212,7 +212,7 @@ jobs:
           node-version: "24.18.0"
           cache: npm
           cache-dependency-path: wallet/zuuli/package-lock.json
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: aarch64-apple-ios
@@ -311,7 +311,7 @@ jobs:
           node-version: "24.18.0"
           cache: npm
           cache-dependency-path: wallet/zuuli/package-lock.json
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           toolchain: ${{ env.ZUULI_RUST_VERSION }}
       - name: Install macOS universal Rust targets
--- a/.github/workflows/zuuli-release.yml
+++ b/.github/workflows/zuuli-release.yml
@@ -42,7 +42,7 @@ env:
   # Restates wallet/rust-toolchain.toml because GitHub cannot compute a
   # workflow-level `env:` from a file. scripts/check-rust-toolchain.sh fails
   # the pull request if this ever stops matching the source of truth.
-  ZUULI_RUST_VERSION: "1.88.0"
+  ZUULI_RUST_VERSION: "1.97.1"
   ZUULI_JAVA_VERSION: "21.0.12"
   ZUULI_XCODE_VERSION: "26.6"
   ZUULI_ANDROID_NDK_VERSION: "27.0.12077973"
@@ -218,7 +218,7 @@ jobs:
       # comment is its only readable record. Bumping the toolchain means moving
       # the SHA to the new version branch and updating the comment with it;
       # scripts/check-rust-toolchain.sh fails the pull request otherwise.
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: aarch64-linux-android,armv7-linux-androideabi,i686-linux-android,x86_64-linux-android
@@ -365,7 +365,7 @@ jobs:
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
         with:
           node-version: "24.18.0"
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           toolchain: ${{ env.ZUULI_RUST_VERSION }}
           targets: aarch64-apple-ios
@@ -581,7 +581,7 @@ jobs:
           {
             node-version: "24.18.0",
           }
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with: { toolchain: "${{ env.ZUULI_RUST_VERSION }}" }
       - name: Install package dependencies
         run: |
@@ -678,7 +678,7 @@ jobs:
           {
             node-version: "24.18.0",
           }
-      - uses: dtolnay/rust-toolchain@2eae45db285e407f22119950686d47e1101e071b # 1.88.0
+      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
         with:
           {
             toolchain: "${{ env.ZUULI_RUST_VERSION }}",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -188,16 +188,20 @@ A bump that touches the source of truth and nothing else **cannot merge** — th
 gate fails. That is the point: the version is one edit plus a mechanical
 follow-through, not an eleven-place archaeology exercise.
 
-### Cadence — proposed, not yet adopted
+### Cadence
 
-The pin is currently `1.88.0`, released June 2025 and roughly fourteen months
-old at the time of writing. Nothing here changes it; the bump is a separate,
-owner-gated decision because it moves every mobile and desktop release target at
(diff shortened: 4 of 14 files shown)


skyl:

Zero failures, zero pending. Required `gate` green, and the **full packaging matrix passed**: Android AAB, iOS, Desktop-Linux, Desktop-macOS, plus `Release identity`.

**Verified independently rather than taken on trust:**

- **The SHA swap is real.** Fetched `032958afbdc797a9164d3bc0b56325c1308924a5`'s `action.yml` — its `env:` block hardcodes `toolchain: 1.97.1`. This was the one step that fails *silently*: updating the adjacent comment while leaving the SHA would have kept building on 1.88.0 with everything still green. The predicted `Unexpected input(s) 'toolchain'` warning appearing in CI is the confirmation that the input is inert and the SHA is the actual pin.
- **Diff scope is pins and docs only** — no `.rs`, no `Cargo.lock`, no `gen/schemas`. A nine-version jump across ~14 months over the librustzcash graph and Tauri, with zero source changes and `--locked` holding on cold rebuilds, is a genuinely good outcome.

**Three findings worth recording:**

1. **The 12th restatement.** `wallet/zuuli/scripts/release-identity.mjs` asserts the channel against a hardcoded literal and the drift check did not know about it — so the check passed locally and then CI failed all five packaging jobs, including an 18-second `Release identity` job that compiles no Rust. Adding it to `DOC_PINS` in the same PR is the right fix: the next bump gets *named* rather than discovered. That is exactly the class of pin #347 was built to eliminate, found the only way it could be.
2. **`cargo fmt` produced zero changes.** I expected a reformat commit — `check-rust-fmt.sh` exists precisely because rustfmt output is not stable across versions. Across nine releases, rustfmt 1.97.1 agrees with 1.88.0 on every crate. Recording the null result is more useful than silence.
3. **`.github/workflows/zuuli.yml` is absent from this diff entirely**, because #347 made the gate *read* the channel from the file. The single source of truth paid for itself on the very next change.

Good judgment leaving the stale `gen/schemas/` drift reverted and out of scope rather than folding an unrelated fix into a version bump. Filing that separately.

PR #358, opened 2026-08-09
deps(rust): clear tier-1 security advisories

skyl:
Tier 1 of #351: the advisories a targeted `cargo update` can actually reach.
Each lockfile change traces to a specific advisory — no blanket update.

## Cleared

| Advisory | Crate | Old → new | Lockfiles changed |
| --- | --- | --- | --- |
| RUSTSEC-2026-0049 | rustls-webpki | 0.103.9 → 0.103.13 | plugin, zuuallet |
| RUSTSEC-2026-0009 | time | 0.3.37 → 0.3.55 | plugin, zuuallet |
| RUSTSEC-2026-0204 | crossbeam-epoch | 0.9.18 → 0.9.20 | plugin, zuuallet |
| RUSTSEC-2026-0097 | rand | 0.8.5 → 0.8.7 | plugin |

All four are gone from all three lockfiles, verified by re-running cargo-deny
with the ignore list emptied. The four matching `ignore` entries are deleted
from `wallet/deny.toml`: **32 → 28**.

`time` 0.3.55 also moves its own private dependencies (`deranged` 0.3.11 →
0.5.8, `num-conv` 0.1.0 → 0.2.2, `time-core` 0.1.2 → 0.1.9, `time-macros`
0.2.19 → 0.2.32). Both lockfile diffs are exactly 8 version swaps with no
crate added or removed.

## The third lockfile was already clean

`wallet/zuuli/src-tauri/Cargo.lock` needed no change — it already resolves
rustls-webpki 0.103.13, time 0.3.53, crossbeam-epoch 0.9.20 and rand 0.8.7.
Only the plugin and zuuallet lockfiles had drifted.

## Partially cleared — ignores kept, reasons corrected

RUSTSEC-2026-0098 / -0099 / -0104 each hit **two** rustls-webpki copies in the
plugin and zuuallet graphs. The webpki bump cleared the 0.103 copy; the 0.101.7
copy behind `rustls` 0.21 remains, and the 0.101 line has no fixed release at
all. These three stay ignored — that is tier 2 (get the last `rustls` 0.21
consumer onto 0.23), and the reasons in `deny.toml` now say so precisely instead
of describing a 0.103 problem that no longer exists.

## Finding for tier 2: the aws-lc-sys blocker is gone

Out of scope for this PR, but worth recording. `deny.toml` said aws-lc-sys was
stuck because `aws-lc-rs` 1.15.4 pinned 0.37. That is no longer true —
`zuuli/src-tauri` already resolves **aws-lc-rs 1.17.3 → aws-lc-sys 0.43.0** and
reports none of RUSTSEC-2026-0044…-0048. So all five are now reachable in the
other two lockfiles by bumping `aws-lc-rs`, not by waiting on upstream. Same
story for quick-xml: zuuli is already on 0.41.0.

Left for a separate PR because it moves the TLS stack the wallet uses to reach
lightwalletd and deserves its own build evidence. The stale reasons in
`deny.toml` are corrected so the file no longer asserts a blocker that has
lifted.

## Not touched

`spin` 0.9.8 is still yanked in the two updated lockfiles (zuuli has 0.9.9).
`yanked = "warn"` by design and it is not one of the four tier-1 advisories, so
it is left out to keep every lockfile line traceable to an advisory.

## Verification

- `scripts/check-rust-deny.sh` passes all four checks (advisories, bans,
  licenses, sources) on all three crates with the reduced 28-entry ignore list,
  locally on rustc 1.97.1. Licences matter here: the `time` chain pulled new
  crate versions and the allow-list still covers them unchanged.
- Every cleared advisory was confirmed gone by re-running cargo-deny against
  each lockfile with the ignore list emptied — before and after — so the
  deletions are proven, not assumed.
- CI is fully green: **18/18 checks pass**, `upstream-canary` correctly skipped.
  That includes the required `gate`, both supply-chain jobs (`deny` and
  `Rust / supply chain`), and the build/test jobs that compile each changed
  lockfile — `Rust / shared plugin` (`cargo test --locked --all-targets` on the
  plugin), `rust` (zuuallet `cargo build --locked` + `cargo test --locked`) and
  `Rust / ZUULI backend`.
- The whole packaging matrix passes too, which is what proves this change is
  safe for what actually ships: Android unsigned AAB, iOS unsigned target app,
  Desktop Linux packages, Desktop macOS universal app, plus custody on
  macos-latest and windows-latest.

Refs #351 (tier 2 and tier 3 stay open).

### DIFF
--- a/wallet/deny.toml
+++ b/wallet/deny.toml
@@ -45,40 +45,48 @@ yanked = "warn"
 ignore = [
   # ---------------------------------------------------------------------------
   # Vulnerabilities. Each of these is transitive; none is a crate we depend on
-  # directly, and none has a fix reachable from the current lockfile without a
-  # semver-major bump in an intermediate crate. Fixing them is the follow-up
-  # issue, deliberately not this one: bumping them touches the lockfile that
-  # every release build is pinned to, and that belongs in a PR whose diff is
-  # the bump rather than in the PR that installs the check.
+  # directly. The advisories that a plain `cargo update` could reach — the
+  # rustls-webpki 0.103 line, time, crossbeam-epoch and rand 0.8 — were cleared
+  # by the tier-1 sweep in #351 and their entries are gone from this list. What
+  # remains needs an intermediate crate to move first, so `cargo update` cannot
+  # reach it from the committed lockfiles.
   # ---------------------------------------------------------------------------
 
   # aws-lc-sys 0.37.1; reached via rustls 0.23 -> aws-lc-rs -> aws-lc-sys, which
-  # tonic's TLS stack pulls in. Fixed in 0.38.0/0.39.0, which is a semver-major
-  # bump for a 0.x crate: `cargo update` cannot reach it, aws-lc-rs must widen
-  # its requirement first.
-  { id = "RUSTSEC-2026-0044", reason = "aws-lc-sys 0.37.1: X.509 name-constraint bypass; needs aws-lc-sys >= 0.39.0, a semver-major bump held by aws-lc-rs 1.15.4" },
-  { id = "RUSTSEC-2026-0045", reason = "aws-lc-sys 0.37.1: AES-CCM tag-verification timing side channel; needs >= 0.38.0, same semver-major bump" },
-  { id = "RUSTSEC-2026-0046", reason = "aws-lc-sys 0.37.1: PKCS7_verify chain-validation bypass; needs >= 0.38.0, same semver-major bump. We do not call PKCS7_verify" },
-  { id = "RUSTSEC-2026-0047", reason = "aws-lc-sys 0.37.1: PKCS7_verify signature-validation bypass; needs >= 0.38.0, same semver-major bump. We do not call PKCS7_verify" },
-  { id = "RUSTSEC-2026-0048", reason = "aws-lc-sys 0.37.1: CRL distribution-point scope logic error; needs >= 0.39.0, same semver-major bump. We do not consume CRLs" },
+  # tonic's TLS stack pulls in. Only the plugin and zuuallet lockfiles are
+  # affected: zuuli/src-tauri already resolves aws-lc-rs 1.17.3 -> aws-lc-sys
+  # 0.43.0 and reports none of these five.
+  #
+  # That is the useful signal here. aws-lc-sys 0.38/0.39 is a semver-major bump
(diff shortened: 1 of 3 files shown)


skyl:

18/18 green including the full packaging matrix, which matters because this changes what ships.

**Verified independently:** net 4 RUSTSEC ignores removed (32 → 28, 14 removed against 10 reworded); 8 version swaps per lockfile; and every crate name in the lockfile diff appears in matched +/- pairs, so nothing was added or removed — purely version bumps. Diff is 3 files.

**Re-running cargo-deny with the ignore list emptied, before and after, is the right method.** It proves each deletion rather than trusting the issue's list, and it is what surfaced the finding below.

## The finding is worth more than the four advisories

**#351's tier 1 / tier 2 split was partly wrong, and I had repeated it.** `deny.toml` claimed `aws-lc-rs` 1.15.4 pinned `aws-lc-sys` to 0.37 behind a semver-major bump. But `wallet/zuuli/src-tauri` already resolves **aws-lc-rs 1.17.3 → aws-lc-sys 0.43.0** and reports none of RUSTSEC-2026-0044…-0048. The intermediate crate moved some time ago; only the plugin and zuuallet lockfiles were stale.

That reclassifies the **five highest-severity items in the whole backlog** — X.509 name-constraint bypass, two PKCS7 verification bypasses, a CRL scope error, an AES-CCM timing side channel — from "blocked on upstream" to "reachable by bumping aws-lc-rs in two lockfiles." `quick-xml` is the same story: zuuli is already on 0.41.0.

Correcting the stale `reason` strings rather than leaving them to rot is exactly right. A `deny.toml` whose justifications have drifted from reality is worse than no file, because it launders a stale assumption as a reviewed decision — and this one would have kept five real vulnerabilities parked indefinitely behind a blocker that no longer existed.

Leaving the aws-lc-rs bump out of scope was also right: it moves the TLS stack that reaches lightwalletd and deserves its own build evidence.

**Honest about what is not cleared:** RUSTSEC-2026-0098/-0099/-0104 each matched two rustls-webpki copies; the 0.103 copy is fixed, the 0.101.7 copy behind rustls 0.21 has no fixed release at all. Ignores kept with reasons rewritten to describe the *actual* remaining problem rather than the original one. `spin` 0.9.8 stays yanked-but-warned, deliberately.

Stopping the local cold build once CI went green was the correct call — CI runs the identical `--locked` commands on all three crates at 1.97.1 and is the authoritative evidence.

Merging via admin.

Issue #340, opened 2026-08-09
ci(rust): add fmt, clippy and cargo-deny gates


skyl:
## Problem

`wallet/rust-toolchain.toml` installs `clippy` and `rustfmt` as components — and **CI runs neither**. There is also no `cargo audit` and no `cargo deny` anywhere in the pipeline.

Today's Rust CI (`zuuli.yml` jobs `rust_plugin` and `rust_app`) runs only `cargo build --locked` and `cargo test --locked`. So:

- Lint regressions land freely.
- Formatting is by convention only, which means review time is spent on it.
- **A dependency with a known advisory, or an incompatible licence, is discovered by a human or not at all.**

That last one matters most. The repo vendors 31 submodules and depends on a large Zcash dependency graph in source; a supply-chain advisory has no mechanical path to our attention.

## Proposal

Add to the existing Rust jobs, or a new cheap `rust_lint` job:

- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo deny check` (licences + advisories), with a committed `deny.toml`

## The realistic sequencing problem

Turning `-D warnings` on across ~11,400 lines of existing plugin code will surface a backlog, and fixing it is a separate chunk of work. Suggested approach so this doesn't stall:

1. Land `cargo fmt --check` first — mechanical, one formatting commit.
2. Land `cargo deny` next — it gates *dependencies*, not our code, so the backlog is usually small and the security value is immediate.
3. Introduce clippy last, initially advisory (`continue-on-error`), then flip to blocking once the backlog is cleared.

Any brand-new crate should be gated strictly from its first commit — that is the cheapest moment and it never gets cheaper.

## Acceptance criteria

- [ ] `cargo fmt --check` blocking in CI.
- [ ] `cargo deny check` blocking, with `deny.toml` committed.
- [ ] `cargo clippy -- -D warnings` blocking (after the backlog is cleared).
- [ ] Any new Rust crate is gated strictly from its first commit.

skyl:
Closed by #349 (`cargo fmt`) and #352 (`cargo deny`). Clippy was **measured but
deliberately not enabled** here — the numbers, and why that changes the plan:

`cargo clippy --locked --all-targets` on the pinned 1.88.0 toolchain:

| Crate | Warnings |
| --- | --- |
| `wallet/plugins/tauri-plugin-zcash` | **8** |
| `wallet/zuuli/src-tauri` | **0** |
[...]

Issue #355, opened 2026-08-09
wallet: committed gen/schemas drifts from a fresh build


skyl:
## Problem

Building `wallet/zuuallet/src-tauri` locally regenerates files under `gen/schemas/` that differ from what is committed — reportedly a **stale plugin command list**. The drift was surfaced during the Rust 1.97.1 bump (#353) and deliberately reverted there, to keep an unrelated fix out of a version-pin PR.

These files are generated-but-committed, so committed content drifting from what a build produces means:

- Every developer gets a dirty working tree after a routine build, which trains people to ignore `git status` — the opposite of what this repo's clean-checkout doctrine depends on.
- Nobody can tell whether the committed schema is stale or the build is wrong without regenerating and reading the diff.
- CI does not catch it, because CI builds fresh and never compares.

## What to do

1. Regenerate and inspect the diff. Determine whether the committed files are simply stale (plugin commands added/removed since they were last generated) or whether something else is producing them.
2. Commit the regenerated output.
3. **Decide whether CI should assert it.** A cheap `git diff --exit-code -- gen/schemas/` after a build would make this impossible to reintroduce. Note the same class of file exists for ZUULI (`wallet/zuuli/src-tauri/gen/`), so whatever is decided should cover both.

## Acceptance criteria

- [ ] Root cause identified — stale vs. non-deterministic generation. These need different fixes; if it is non-deterministic, committing the output is not a fix at all.
- [ ] Committed `gen/schemas/` matches a fresh build for both `zuuallet` and `zuuli`.
- [ ] A decision recorded on whether CI asserts generated-file freshness, with rationale either way.

## Notes

Discovered during #353 and consciously left out of scope there. Not urgent — it is a hygiene and trust issue rather than a correctness one — but it quietly erodes the "a dirty checkout means something is wrong" signal that the workflow doctrine relies on.

PR #359, opened 2026-08-09
ci(rust): fix clippy warnings and enable -D warnings

### LINKED ISSUE #354: ci(rust): enable clippy -D warnings — the backlog is 8 warnings, all auto-fixable
Split out of #340, which sequenced clippy last on the assumption that turning
`-D warnings` on across ~11,400 lines of plugin code would surface a backlog
worth its own chunk of work. **It was measured, and it does not.**

## The measurement

`cargo clippy --locked --all-targets` on the pinned 1.88.0 toolchain (run from
`wallet/`, so `wallet/rust-toolchain.toml` selects the compiler — clippy's lint
set moves between versions, so the number is only meaningful against the pin):

| Crate | Warnings |
| --- | --- |
| `wallet/plugins/tauri-plugin-zcash` | **8** |
| `wallet/zuuli/src-tauri` | **0** |
| `wallet/zuuallet/src-tauri` | **0** |

Eight, total, across the whole workspace. All eight are in the plugin, and
clippy reports that all eight are machine-fixable
(`cargo clippy --fix --lib -p tauri-plugin-zcash --tests`):

| Lint | Location |
| --- | --- |
| `unnecessary_cast` — casting `u64` to `u64` | `src/app_data_migration.rs:410` |
| `redundant_closure` | `src/commands.rs:1851` |
| `uninlined_format_args` | `src/commands.rs:1852` |
| `uninlined_format_args` | `src/wallet/accounts.rs:88` |
| `uninlined_format_args` | `src/wallet/accounts.rs:101` |
| `items_after_test_module` | `src/wallet/manifest.rs:433` |
| `let_and_return` | `src/wallet/manifest.rs:197` |
| `unused_async` / manual future — "can be simplified using `async fn`" | `src/wallet/sync.rs:1565` |

(Six of the eight appear in the `lib` target; `--all-targets` surfaces two more
from the test target.)

## What this means for the sequencing in #340

The advisory `continue-on-error` phase #340 proposed as a staging step is not
needed. There is no backlog to clear first — the backlog is one small PR.

Recommended:

1. Run `cargo clippy --fix` on the plugin, review the eight mechanical edits,
   land them.
2. Add a `clippy` step to the existing `rust_fmt` job in
   `.github/workflows/zuuli.yml` with `-D warnings`, blocking from the start.
   That job already installs the pinned toolchain, and
   `wallet/rust-toolchain.toml` already lists `clippy` as a component, so the
   only new cost is compile time — clippy needs the dependency graph, so it
   belongs on `rust_deny`'s side of the fence (submodule + real compilation)
   rather than in the ~15-second `rust_fmt` job. Consider a separate
   `rust_clippy` job reusing the `Swatinem/rust-cache` workspaces the existing
   Rust jobs use.
3. Whatever job it lands in **must** be added to *both* the `gate` job's
   `needs:` list and the inline `results=` string — updating only one makes the
   gate decorative for it. See #349 and #352 for the pattern.

Doing it now, while the number is 8, is far cheaper than doing it later. #340's
final acceptance criterion — that any new Rust crate is gated strictly from its
first commit — is already served by `scripts/check-rust-fmt.sh` and
`scripts/check-rust-deny.sh` discovering crates under `wallet/` rather than
listing them; a clippy check should do the same.

## Reproduce

```
cd wallet
cargo clippy --locked --all-targets --manifest-path plugins/tauri-plugin-zcash/Cargo.toml
cargo clippy --locked --all-targets --manifest-path zuuli/src-tauri/Cargo.toml
cargo clippy --locked --all-targets --manifest-path zuuallet/src-tauri/Cargo.toml
```

skyl:
Closes #354.

## Re-measured first — the backlog is 16, not 8

Issue #354 counted **8** warnings on the old 1.88.0 pin. That count is stale:
#353 moved the pin to **1.97.1**, and a nine-version jump brings a different
lint set. The real backlog, `cargo clippy --locked --all-targets` from `wallet/`
on 1.97.1:

| Crate | Warnings |
| --- | --- |
| `wallet/plugins/tauri-plugin-zcash` | **16** (13 lib + 3 more from the test target) |
| `wallet/zuuli/src-tauri` | **0** |
| `wallet/zuuallet/src-tauri` | **0** |

The `uninlined_format_args` hits the issue listed are gone — that lint no longer
fires on those sites in 1.97.1. What replaced them, and more, is
`collapsible_if`: edition 2024 stabilised let-chains, so clippy now folds
`if cond { if let ... }` into `if cond && let ...`. Nine of the sixteen are that.

## Every warning fixed

| Lint | Location | Fix |
| --- | --- | --- |
| `collapsible_if` | `commands.rs:454`, `commands.rs:690`, `commands.rs:1215`, `commands.rs:1284`, `commands.rs:1631` | folded into let-chains |
| `collapsible_if` | `wallet/cache.rs:55` | folded into a let-chain |
| `collapsible_if` | `wallet/manifest.rs:424` | folded into a let-chain |
| `collapsible_if` | `wallet/send.rs:672`, `wallet/send.rs:822` | folded into let-chains |
| `collapsible_if` | `wallet/sync.rs:650`, `wallet/sync.rs:651` | one three-condition let-chain |
| `unnecessary_cast` | `app_data_migration.rs:410` | **not** removed — narrow `#[allow]`, see below |
| `redundant_closure` | `commands.rs:1851` | `.map(\|a\| u64::from(a))` → `.map(u64::from)` |
| `let_and_return` | `wallet/manifest.rs:197` | `WalletEntry { .. }` returned as the tail expression |
| `items_after_test_module` | `wallet/manifest.rs:433` | `chrono_now` / `days_to_ymd` moved above `mod replacement_tests` |
| `manual_async_fn` | `wallet/sync.rs:1565` (test) | `fn -> impl Future` → `async fn` |

Applied with `cargo clippy --fix`, then `scripts/check-rust-fmt.sh --fix`, then
read line by line. Two places needed a human afterwards, both cosmetic:
rustfmt declines to re-indent the two `create_wallet`/`restore_wallet` bodies
whose let-chain condition is itself multi-line, and the `\`-continued log string
in `sync.rs` kept its old continuation indent (leading whitespace after a
`\`-newline is stripped by the lexer, so the message text is byte-identical).
Both were re-indented by hand and `check-rust-fmt.sh` confirms the result is
rustfmt-stable.

## Behaviour

**No behaviour changed.** Every let-chain preserves short-circuit order and
binding scope; `let_and_return`, `redundant_closure` and `items_after_test_module`
are pure rearrangement. `manual_async_fn` touches test-only helper
`counted_task`: both forms produce a lazy `Send + 'static` future built from
owned arguments, and its `impl Future<Output = ()> + Send + 'static` bound is
still satisfied — the call site's `F: Future<Output = ()> + Send + 'static` bound
proves it at compile time.

### The one lint I refused: `unnecessary_cast` in `app_data_migration.rs`

`cargo clippy --fix` deleted the `as u64` on `stat.st_ino`, and that fix is
**wrong** — it is only correct on the host clippy happened to run on. That
struct literal sits in a `#[cfg(unix)]` function that compiles for macOS, iOS,
Linux and four Android ABIs, and `dev_t`/`ino_t` are not one type across them:
on Apple `st_dev` is `i32` and `st_ino` is `u64`, on Linux/Android both are
`u64`. So on macOS the `st_ino` cast looks redundant and on Linux the `st_dev`
cast does — delete either and you get a compile error on the other target.

Reverted, and pinned in place with a narrow `#[allow(clippy::unnecessary_cast)]`
on the statement plus a comment explaining the target skew, next to the
`st_mode` comment already in the file warning about exactly this hazard. Both
fields are non-negative kernel identifiers, so the widening cannot change the
identity comparison's meaning on any target.

### The same trap, caught by CI: `useless_conversion` three lines above

I ran the whole gate under a clean `ubuntu:24.04` container precisely because of
the above, and it — and then CI — found the mirror image of it. The `st_mode`
file-type test two statements earlier uses `u32::from(...)` on three operands.
On macOS those widen `u16` → `u32` and clippy is silent; on Linux all three are
already `u32`, so `clippy::useless_conversion` fires three times and `-D warnings`
turns them into hard errors. The conversions are exactly as load-bearing as the
file's existing comment says: `mode_t` is `u16` on armv7/i686 Android.

Fixed in **f3d2508** by folding the two statements into one boolean and scoping a
single `#[allow(clippy::useless_conversion)]` to it, with a comment saying why
clippy is wrong here. This is the whole argument for running the gate on the
target CI runs on rather than on a developer's Mac.

Those two are the only `#[allow]`s added. Nothing is allowed at crate level.

## The gate

`scripts/check-rust-clippy.sh`, built to the shape of `check-rust-fmt.sh` and
`check-rust-deny.sh`:

- **discovers** crates by finding `Cargo.toml` under `wallet/` — a new crate is
  gated from its first commit, no list to forget to update (#340's final
  acceptance criterion);
- runs cargo from `wallet/` so `rust-toolchain.toml` selects the pin, and
  **refuses to render a verdict** on any other compiler, because clippy's lint
  set moves between releases and a check on whatever-was-installed reports
  version skew as a code defect;
- `--locked --all-targets -- -D warnings`, so tests are linted too;
- `--fix` mode for the mechanical path, which tells you to read the diff and
  re-run rustfmt afterwards.

Unlike `fmt`/`deny` it genuinely compiles — clippy is a rustc driver — so its
jobs carry the librustzcash submodule, the Tauri system libraries, and a cache
key of their own (clippy and `cargo build` write different artifacts into the
same target-directory names; a shared key would have them evicting each other).

### CI wiring

**`.github/workflows/zuuli.yml`** — new `rust_clippy` job, `zuuli == 'true'`
condition matching its siblings, and all three wiring points updated so the gate
is not decorative:

- `changes` path filter gained `scripts/check-rust-clippy.sh`
- `gate.needs` → `[changes, frontend, rust_fmt, rust_deny, rust_clippy, rust_plugin, rust_app]`
- `gate`'s inline `results=` → `... $RUST_DENY_RESULT $RUST_CLIPPY_RESULT $RUST_PLUGIN_RESULT ...`

**`.github/workflows/zuuallet.yml`** — mirrored `clippy` job and path filters,
for the same reason `fmt` and `deny` are mirrored there: the required `gate`
lives in zuuli.yml and its change detector does not select
`wallet/zuuallet/**`, so a Zuuallet-only PR could otherwise land unlinted code
the gate never sees — and then break the next unrelated ZUULI PR on files it
never touched.

`scripts/check-rust-toolchain.sh --self-test` and the real run both pass; the
new jobs take the channel from `--print-channel` and hold no version literal.

## Verification

- `scripts/check-rust-clippy.sh` — clean on all three crates, macOS 1.97.1.
- **Linux, not just macOS.** Clippy's verdict is target-dependent (see the
  `st_ino` note), so the whole gate was also run under a clean `ubuntu:24.04`
  container on the pinned toolchain. Clean there too.
- `scripts/check-rust-fmt.sh` — clean, so the fixes are rustfmt-stable.
- `cargo build` + `cargo test --locked --all-targets` — pass on all three crates.
- **The gate can fail, not just pass.** Proved in CI: see below.

## Proving the gate fails

Not asserted — done, on the real CI, on this branch:

- **d04bd1c** appended a `pub fn clippy_gate_smoke_test()` to
  `wallet/zuuallet/src-tauri/src/lib.rs` containing `format!("{}", "…")` and a
  `let`-then-return — `useless_format` + `let_and_return`. Deliberately in
  **zuuallet**, not the plugin, so it also proves the script's crate discovery
  reaches a crate nobody named and that the zuuallet mirror job earns its place.
- Both lint jobs went red, in about two minutes:

  ```
  Rust / lints   fail   2m1s
  clippy         fail   2m8s
  ```
  ```
  error: useless use of `format!`
    --> src/lib.rs:14:13
     = help: to override `-D warnings` add `#[allow(clippy::useless_format)]`
  1 crate(s) have clippy warnings: wallet/zuuallet/src-tauri
  ```

- And the required `gate` went red **because of it**, which is the part that is
  easy to get wrong:

  ```
  RUST_CLIPPY_RESULT: failure
  ZUULI changed: true; job results: success success success failure success success
  Expected every applicable job to be success, got: failure
  ```

- **f035006** reverts it. `git diff f3d2508 HEAD` is empty, so the branch head is
  byte-identical to the tree that CI already proved green (all 21 checks pass,
  including `Rust / lints`, `clippy` and `gate`).

## Known limitation, deliberately not fixed here

`#[cfg(windows)]` and macOS-only code — the custody backends — is compiled by
`zuuallet.yml`'s `custody-platforms` matrix, which runs `cargo test`, not clippy.
So the two new jobs lint every crate but only the Linux view of it. Extending
`custody-platforms` to lint is a real follow-up; it is not in this PR because it
would surface an unmeasured second backlog on two more targets, and #354 is
about closing the measured one.

### DIFF
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -23,6 +23,7 @@ on:
       - '.gitmodules'
       - 'scripts/check-rust-fmt.sh'
       - 'scripts/check-rust-deny.sh'
+      - 'scripts/check-rust-clippy.sh'
       - 'wallet/deny.toml'
       - '.github/workflows/zuuallet.yml'
   push:
@@ -34,6 +35,7 @@ on:
       - '.gitmodules'
       - 'scripts/check-rust-fmt.sh'
       - 'scripts/check-rust-deny.sh'
+      - 'scripts/check-rust-clippy.sh'
       - 'wallet/deny.toml'
       - '.github/workflows/zuuallet.yml'
   workflow_dispatch:
@@ -109,6 +111,69 @@ jobs:
 
       - name: Check advisories, licences and sources
         run: scripts/check-rust-deny.sh
+
+  # ---------------------------------------------------------------------------
+  # Lints, for the same reason as `fmt` and `deny`: the required `gate` in
+  # zuuli.yml does not select wallet/zuuallet/**, so without this a
+  # Zuuallet-only pull request could land code that fails `clippy -D warnings`
+  # — and then fail the next unrelated ZUULI pull request on files it never
+  # touched. Same script, so the two workflows cannot disagree about what
+  # "lint-clean" means.
+  #
+  # This one compiles (clippy is a rustc driver), so unlike `fmt` it needs the
+  # submodule, the Tauri system libraries and a cache.
+  # ---------------------------------------------------------------------------
+  clippy:
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Fetch librustzcash submodule
+        run: git submodule update --init z/zcash/librustzcash
+
+      - name: Install Tauri system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            libwebkit2gtk-4.1-dev \
+            libgtk-3-dev \
+            libayatana-appindicator3-dev \
+            librsvg2-dev \
+            libsoup-3.0-dev \
+            libjavascriptcoregtk-4.1-dev \
+            libxdo-dev \
+            libdbus-1-dev \
+            libssl-dev \
+            build-essential curl wget file pkg-config
+
+      - name: Resolve the pinned Rust toolchain
+        id: rust_toolchain
+        run: |
+          set -euo pipefail
+          version=$(scripts/check-rust-toolchain.sh --print-channel)
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      # clippy is absent from the minimal profile this action installs.
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: ${{ steps.rust_toolchain.outputs.version }}
+          components: clippy
+
+      # Separate key from the `rust` job: clippy writes different artifacts into
+      # the same target directory names, so a shared cache would have the two
+      # jobs evicting each other's work on every run.
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: |
+            wallet/plugins/tauri-plugin-zcash
+            wallet/zuuli/src-tauri
+            wallet/zuuallet/src-tauri
+          key: zuuallet-clippy
+
+      - name: Lint every Rust crate under wallet/ at -D warnings
+        run: scripts/check-rust-clippy.sh
 
   # ---------------------------------------------------------------------------
   # Frontend: TypeScript typecheck + production bundle.
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -71,7 +71,7 @@ jobs:
           zuuli=false
           while IFS= read -r file; do
             case "$file" in
-              wallet/zuuli/*|wallet/plugins/*|wallet/rust-toolchain.toml|wallet/deny.toml|scripts/check-rust-fmt.sh|scripts/check-rust-deny.sh|z/zcash/librustzcash|.gitmodules|.github/workflows/zuuli.yml|.github/workflows/zuuli-packaging.yml|.github/workflows/zuuli-release.yml|.github/workflows/zuuli-play-testers.yml|.github/workflows/cache-cleanup.yml|.github/actions/zuuli-rust-cache/*)
+              wallet/zuuli/*|wallet/plugins/*|wallet/rust-toolchain.toml|wallet/deny.toml|scripts/check-rust-fmt.sh|scripts/check-rust-deny.sh|scripts/check-rust-clippy.sh|z/zcash/librustzcash|.gitmodules|.github/workflows/zuuli.yml|.github/workflows/zuuli-packaging.yml|.github/workflows/zuuli-release.yml|.github/workflows/zuuli-play-testers.yml|.github/workflows/cache-cleanup.yml|.github/actions/zuuli-rust-cache/*)
                 zuuli=true
                 ;;
             esac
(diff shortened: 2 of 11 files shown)


skyl:

All green. Verified independently: exactly **two** `#[allow]`s, both scoped to a single statement with real reasoning, **zero** crate-level suppressions, and all three gate wiring points touched (`changes` filter, `gate.needs`, and the inline `results=` string).

## The wrong auto-fix is the headline

`cargo clippy --fix` deleting `stat.st_ino as u64` is exactly the failure this repo's `AGENTS.md` warns about, in its most treacherous form. On Apple `st_dev` is `i32` and `st_ino` is `u64`; on Linux and Android both are `u64`. Clippy sees **one target at a time**, so on whichever host it runs, precisely one of the two casts looks redundant *there* and is load-bearing everywhere else. Applying the fix compiles locally and breaks a different target — and this is a `#[cfg(unix)]` function spanning macOS, iOS, Linux and four Android ABIs.

The mirror image is worse, because macOS cannot see it at all: `u32::from(stat.st_mode)` is a genuine widening on macOS (`mode_t` is `u16`) and the reflexive identity on Linux.

**Running the whole gate in a clean `ubuntu:24.04` container before CI did is the right instinct and should stay in the loop for this repo.** Dev is macOS, the gate is Linux, and `AGENTS.md` documents that skew as a real historical bug source. This is a live example rather than a hypothetical.

The comments explaining *why* each cast is load-bearing on targets clippy is not looking at are what make these `#[allow]`s legitimate rather than suppression. A future reader will not delete them.

## Other notes

- **The count was 16, not 8** — #354's figure was measured on the 1.88.0 pin before #353 moved us to 1.97.1. New compilers ship new lints; `uninlined_format_args` disappeared and edition-2024 let-chains made `collapsible_if` fire nine times. Documenting in `AGENTS.md` that a toolchain bump is now the one moment new lints appear, and that they belong in the bump's own PR, is the right systemic response.
- **The gate was proven able to fail** with a real commit (`d04bd1c`) that went red in ~2 min, then reverted — and deliberately planted in *zuuallet* rather than the plugin, which also proves crate discovery and the mirrored job. That is a better test than planting it where it was already expected to be caught.
- The script telling contributors not to blanket-allow at crate level is a nice touch; the guardrail travels with the tool.

**Follow-ups, correctly left out of scope:** `gen/schemas` staleness is already tracked as #355. The observation that `#[cfg(windows)]`/macOS-only custody code is never clippy-linted is a real coverage gap — filing it separately rather than expanding this diff was right, since it would surface an unmeasured backlog on two more targets.

Merging via admin.

PR #356, opened 2026-08-09
fix(zuuli): respect mobile safe areas

skyl:
Refs #331

## What changed
- enables viewport-fit=cover and keyboard-aware viewport resizing together with centralized env(safe-area-inset-*, 0px) variables
- bounds the app with ordered 100vh and 100dvh sizing, locks document scrolling, and keeps route scrolling inside the shell
- protects top chrome, bottom tabs, desktop sidebar, normal pages, AI, articles, login, and crash fallback in portrait and landscape
- moves /login into the persistent shell with one bounded internal scroll owner and overflow-safe centering
- configures Android adjustResize and fixed-light system-bar icons for the always-dark UI
- adds regression coverage for viewport/inset contracts, routing/scroll ownership, auth overflow, and Android edge-to-edge ordering

## Verification
- independent exact-head technical APPROVE
- full exact-head CI and packaging matrix green, including Android AAB and iOS target
- 110 Vitest tests plus 5 safe-area contracts, typecheck, production build, and release verification green
- Pixel 7 Android API 36 emulator: portrait Discover/login, focused email field, gesture navigation, and landscape pass

## Physical follow-up
Keep issue #331 and needs-device open until Skylar validates a signed Android build in portrait and landscape, including the real soft keyboard and available gesture/3-button navigation modes.

Issue #331, opened 2026-08-09
ZUULI Mobile: the login screen breaks the safe area and unpins the top/bottom bars


skyl:
## Problem

Found while testing 0.1.0 (6) on mobile. Opening the login screen breaks the app frame: the top bar scrolls off the top of the screen and the bottom bar is no longer pinned to the bottom. Every other route holds its chrome correctly.

Two separate defects compound here.

### 1. `/login` renders outside the app shell

`src/App.tsx:93` registers `<Route path="/login" element={<AuthFeature />} />` **outside** the `<Route element={<AppShell/>}>` layout route at `src/App.tsx:97`. So `/login` gets no `TopBar`, no `MobileTabBar`, and none of the `h-screen overflow-hidden` framing from `src/components/layout/AppShell.tsx:43`.

The auth screen then builds its own frame incorrectly:

- `src/features/auth/index.tsx:39` — `<div className="grid min-h-screen …">` uses `min-h-screen` (`100vh`), not `h-screen`/`dvh`.
- `src/features/auth/index.tsx:42` — `<main className="… min-h-screen … overflow-y-auto">` nests a *second* `min-h-screen` with its own `overflow-y-auto`. Because the parent grid cell has no bounded height, that `overflow-y-auto` never activates and scrolling falls through to the document/body — the one element that never scrolls anywhere else in the app.

On iOS/Android `100vh` is the *large* viewport (it excludes system chrome), so the card is taller than the visible area and the whole page rubber-bands under the status bar and home indicator.

### 2. Safe-area insets are inert app-wide

- `index.html:5` — `<meta name="viewport" content="width=device-width, initial-scale=1.0">` has **no `viewport-fit=cover`**. Without it the webview resolves `env(safe-area-inset-*)` to `0px`, so the existing bottom-inset work is a no-op: `src/components/layout/Sidebar.tsx:62` (`h-[calc(3.5rem+env(safe-area-inset-bottom))] pb-[env(safe-area-inset-bottom)]`) and `src/features/ai/index.tsx:318`.
- There is **no top inset anywhere**. `src/components/layout/TopBar.tsx:52` is a bare `h-14` with no `pt-[env(safe-area-inset-top)]`. No `env()` appears in `src/index.css` and `tailwind.config.cjs` defines no safe-area utilities (only `.min-tap`, line 117).
- No Tauri safe-area plugin is registered (`src-tauri/tauri.conf.json:41-52` registers only `deep-link`).

[...]

PR #360, opened 2026-08-09
release(zuuli): prepare internal build 9

skyl:
Refs #331

Advances the immutable ZUULI store identity from 0.1.0+8 to 0.1.0+9. The source is exact main after the Android safe-area repair in #356.

Verification:
- npm run release:verify
- git diff --check
- release-only diff across canonical Android and iOS build metadata

After gated merge, the protected release workflow automatically builds, signs, validates, and uploads this exact commit to Play internal and TestFlight.

### DIFF
--- a/wallet/zuuli/release.json
+++ b/wallet/zuuli/release.json
@@ -4,7 +4,7 @@
   "applicationId": "cash.free2z.zuuli",
   "iosUsesNonExemptEncryption": false,
   "version": "0.1.0",
-  "build": 8,
+  "build": 9,
   "minimums": {
     "ios": "18.0",
     "android": 29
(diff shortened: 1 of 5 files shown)


skyl:
Exact-head founder review: APPROVE at 212d79cae440ca04d891cddbcdff4adcbf6bbc25 on base 299721ed97454c854c27e7480f94abb48aa081f4. The diff is limited to the five canonical release identity surfaces; all resolve to 0.1.0+9, the monotonic parent check passes, git diff --check is clean, and the entire exact-head application plus Android/iOS/Linux/macOS packaging matrix is green. Proceeding through the documented admin squash path because the shared owner account cannot formally approve its own PR.

Issue #321, opened 2026-08-09
CI: `gate` never compiles Android/iOS targets — a target-specific compile error reached the protected release (build 7 Android failed)


skyl:
## Summary

The required `gate` check never compiles the Rust workspace for **Android** (or
iOS) targets, so a target-specific compile error passes `gate` green and is only
discovered by the **protected release workflow** — after a store train has
already been kicked off.

This is not hypothetical: it just happened. Build `0.1.0+7`'s
`Android / Play internal` job failed
([run 31270095364](https://github.com/free2z/zuu/actions/runs/31270095364))
with a hard compile error in the shared plugin:

```
error[E0277]: no implementation for `u32 & u16`
  --> wallet/plugins/tauri-plugin-zcash/src/app_data_migration.rs:391:21
error[E0308]: mismatched types: expected `u32`, found `u16`
error[E0277]: can't compare `u32` with `u16`
error: could not compile `tauri-plugin-zcash` (lib) due to 4 previous errors
```

`stat.st_mode` is `u32` on the Android targets while `libc::S_IFMT`/`S_IFDIR`
are `u16`; on iOS/macOS/Linux the widths agree. The PR that introduced it was
`gate`-green because `gate` aggregates `changes`, `frontend`, `rust_plugin`,
`rust_app` (`.github/workflows/zuuli.yml:210-211`) and the Rust jobs build for
the **host** only.

Net effect: iOS uploaded build 7 to TestFlight while Android never produced an
AAB — a half-shipped release identity, discovered at the most expensive possible
moment (inside the protected, credential-bearing lane).

## The gap

`gate` proves "compiles on the CI host." The product ships to iOS, Android,
macOS, Linux. Any `#[cfg]`-conditional code, any libc type-width difference, any
target-conditional dependency can diverge without `gate` noticing. The repo's
`AGENTS.md` already documents this class of trap for macOS-vs-Linux; Android and
iOS are the same trap with more targets.

## Proposed fix

Add a cheap **type-check-only** matrix to the ZUULI CI workflow and make `gate`
depend on it — `cargo check`, not a full build/link, so it stays fast:

- `aarch64-linux-android`, `armv7-linux-androideabi` (Android release ABIs)
- `aarch64-apple-ios` (iOS)

`cargo check --target …` needs the target std (via `rustup target add`) but not
the NDK/Xcode toolchain for pure type-checking, which is what catches this whole
error class. If linking turns out to be required for some crate in the graph,
[...]

PR #386, opened 2026-08-16
docs(agents): record crate constraints that already cost us a release

skyl:
Documentation only — one file, `AGENTS.md`. No code, no workflow changes.

Three constraints we paid for once and should not rediscover. Each is folded into the section that already owns its subject rather than added as a new section.

### 1. `rusqlite` 0.37 is a repo-wide singleton — *What NOT to do*

`libsqlite3-sys` declares `links = "sqlite3"`, and Cargo hard-errors on two versions of a `links` package in one graph. Everything under `wallet/` reaches SQLite through `wallet/plugins/tauri-plugin-zcash` (`rusqlite = { version = "0.37", features = ["bundled", "array"] }`), and `wallet/zuuli/src-tauri` links that plugin together with the rest of the app — so any new crate needing SQLite must resolve to 0.37. Verified: all three lockfiles are on `rusqlite 0.37.0` / `libsqlite3-sys 0.35.0`.

Worked example included: `openmls_sqlite_storage` 0.2.0 (latest release) requires `rusqlite ^0.32` and is unusable here; its `0.3.0-rc` line moved to `^0.37` and would fit. Both requirements verified against the crates.io API. The conflict only appears once the app links both, so a crate that builds standalone can still be off-limits.

### 2. Platform integer widths vs. one-target-at-a-time clippy — *Verifying before you push*

Added as a fourth trap alongside the existing platform-gated-dependency one. `cargo clippy --fix` will delete a conversion that is the reflexive identity on the target it happens to run on and load-bearing on the others. Not hypothetical: build `0.1.0+7` went to TestFlight on iOS while Android never produced an AAB (`error[E0277]: no implementation for u32 & u16`). Points at the live `#[allow(clippy::useless_conversion)]` / `#[allow(clippy::unnecessary_cast)]` sites in `wallet/plugins/tauri-plugin-zcash/src/app_data_migration.rs`, whose comments explain the skew in full, and references #321 for the CI gap. The width details are taken from those comments rather than restated loosely.

Also fixed the stale "Two specific traps" lead-in, which already described three bullets.

### 3. `wallet/deny.toml` justifications rot silently — *Guardrails that keep us honest*

Its `ignore` reasons were true when written and drift as dependencies move; a stale one launders an out-of-date assumption as a reviewed decision. Real example, verified in the lockfiles: five high-severity `aws-lc-sys` advisories parked behind "aws-lc-rs 1.15.4 pins the vulnerable line" while `wallet/zuuli/src-tauri` had already resolved aws-lc-rs 1.17.3 → aws-lc-sys 0.43.0. Guidance is to re-derive rather than re-read — empty `ignore` locally, run `scripts/check-rust-deny.sh advisories`, diff against the file. References #351.

Nothing here duplicates the existing gate, toolchain, or macOS-vs-Linux feature-unification material.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -136,6 +136,19 @@ integrate, and move it forward alongside everything else.
   could never reproduce the build.)
 - ❌ Rely on transitive/implicit behavior (e.g. a Cargo feature enabled only by
   another crate's unification). Declare what we use.
+- ❌ Add a crate that brings a second `rusqlite`. `libsqlite3-sys` declares
+  `links = "sqlite3"`, and Cargo refuses outright to build a graph containing
+  two versions of a `links` package — a hard error, not a warning. Everything
+  under `wallet/` reaches SQLite through `wallet/plugins/tauri-plugin-zcash`'s
+  `rusqlite = { version = "0.37", features = ["bundled", "array"] }`, and
+  `wallet/zuuli/src-tauri` links that plugin together with the rest of the app,
+  so **rusqlite 0.37 is a repo-wide singleton**: any new crate that needs SQLite
+  must resolve to it. All three lockfiles currently agree on `rusqlite 0.37.0` /
+  `libsqlite3-sys 0.35.0`. Worked example: `openmls_sqlite_storage` 0.2.0 — the
+  latest *release* — requires `rusqlite ^0.32` and is therefore unusable here;
+  its `0.3.0-rc` line moved to `^0.37` and would fit. Check the requirement
+  before adopting, because the conflict appears only once the app links both: a
+  crate that builds fine on its own can still be off-limits.
 
 ## Guardrails that keep us honest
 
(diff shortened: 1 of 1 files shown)


skyl:
## Review — approved, merging

Docs-only, `gate` green, +43/-1 in one file. Each addition went into the section that already owns its subject rather than a new one, which is right for a file people skim under pressure.

**Two corrections to my brief, both verified and both mine to own:**

1. **The width skew is not "macOS vs Android."** I wrote that; the code says otherwise. `st_mode` is `c_uint` on *every* Android ABI, `mode_t` is `u16` on the **32-bit Android ABIs (armv7/x86)** and u32 on 64-bit Android and Linux, and **Apple is self-consistent at u16** — so the break was armv7/i686-android only. Writing the code's version instead of the brief's was correct, and I have fixed my own notes.

2. **`openmls_sqlite_storage` is not unusable full stop.** 0.2.0 (latest release) needs `rusqlite ^0.32`, but `0.3.0-rc` moved to `^0.37` and would fit. "Check the requirement before adopting" is the useful rule; "this crate is blocked" would have been wrong and would have sent someone to hand-write a 56-method `StorageProvider` unnecessarily.

Good catch on the `aws-lc-sys` example already being partly corrected in `deny.toml` — describing it in past tense as the worked example, while noting the residual split (zuuli on aws-lc-rs 1.17.3, plugin and zuuallet still on 1.15.4), is more accurate than either extreme.

The **re-derive instead of re-reading** framing is the sentence I most want to survive: empty the `ignore` list locally, run the check, diff what it *reports* against what the file *claims*. That is the method that found the stale blocker, and it generalises past cargo-deny.

Also fixing the pre-existing "Two specific traps" lead-in above three bullets — small, but the kind of drift that makes a reference stop being trusted.

Merging via admin.

PR #427, opened 2026-08-19
fix(zuuli): restore scroll by history entry

### LINKED ISSUE #333: ZUULI: scroll position carries across tab switches (Discover → Articles lands mid-page)
## Problem

Found while testing 0.1.0 (6). Scroll down on Discover (home), then tap Articles: you land part-way down the Articles page instead of at the top, and the top of the feed is not what loads.

### Root cause

`AppShell` is a layout route (`src/App.tsx:97`) — only the `<Outlet/>` child swaps on navigation. The Radix `ScrollArea` viewport at `src/components/layout/AppShell.tsx:53` is therefore the **same persistent DOM node** across every tab switch, so its raw `scrollTop` in pixels carries straight over from the previous route.

Nothing resets it: there is no `<ScrollRestoration>` (unavailable anyway with the `BrowserRouter` at `src/App.tsx:89`) and no `window.scrollTo`/`scrollTop = 0` anywhere in `src` — the only matches for scroll APIs are `scrollIntoView` in `src/features/ai/index.tsx:174` and `src/components/common/Markdown.tsx:105`.

This will read as intermittent: `src/components/layout/AppShell.tsx:40` toggles `isFullBleed` for `/ai`, which unmounts and remounts the container, so Discover → AI → Articles *does* reset while Discover → Articles does not.

### Why it also breaks loading, not just position

The carried-over offset interacts badly with the Articles infinite scroll (`src/features/articles/pages/Feed.tsx:57-69`, `IntersectionObserver` with `rootMargin: "600px 0px"`, sentinel at `Feed.tsx:258`):

1. On mount the list is empty/skeleton (`Feed.tsx:204`), so content is short and the retained `scrollTop` is clamped.
2. Page 1 lands (`src/features/articles/useArticleFeed.ts:50-77`) and content grows underneath a non-zero offset.
3. The sentinel is immediately inside the 600px band, so `loadMore` (`useArticleFeed.ts:79-105`) chain-fires.

Separately, that observer's `root` is `null` (the document viewport) rather than the Radix viewport, so `rootMargin` is measured against the wrong box and the 600px prefetch band is misaligned with the actual scroller. The feed is a plain grid (`Feed.tsx:74-83`) — no virtualization involved.

## Acceptance criteria

- [ ] Navigating to a different route scrolls the shared container to the top before or as the new route paints — no flash of mid-page content.
- [ ] Back/forward navigation *restores* the prior scroll position rather than resetting it (browser-standard behavior, distinct from a fresh navigation).
- [ ] Re-tapping the tab you are already on returns to the top of that route.
- [ ] The Articles feed's `IntersectionObserver` observes the actual scroll container, so `rootMargin` prefetching is measured against the right box.
- [ ] Arriving at Articles fresh loads page 1 at the top without chain-firing `loadMore`.
- [ ] Behavior is consistent whether or not the route change crosses the `isFullBleed` boundary at `AppShell.tsx:40`.

## Notes

The minimal fix is a `useEffect` on `location.pathname` in `AppShell` that resets the Radix viewport's `scrollTop`. The viewport in `src/components/ui/scroll-area.tsx:14` currently exposes no ref or `data` hook of its own, so this needs either a forwarded ref or a `[data-radix-scroll-area-viewport]` query.

Worth deciding as part of this: whether any routes should *retain* their scroll position across tab switches (a common pattern for feeds), which would call for a per-route saved-offset map rather than a blanket reset.

skyl:
Closes #333

## Summary
- restore the actual registered route scroller by React Router history-entry key on POP while resetting PUSH/REPLACE entries before paint
- expose the typed Radix viewport, register full-bleed AI/login owners, and bind Articles IntersectionObserver to the real viewport with a reactive sentinel
- return already-active exact-root desktop/mobile destinations to top on first pointer, touch, or keyboard activation while nested routes navigate normally to their section root
- bound saved offsets to 64 entries and async POP correction to 30 seconds, observing actual direct content boxes with border-box ResizeObserver and using MutationObserver only to track replacements
- cancel pending correction on route/ref/timeout and window-captured pointer/touch/wheel/key input, so late content cannot override genuine user scrolling

## Local verification
- `npm ci`
- `npm run typecheck`
- `npm run build`
- Vitest: 289/289
- boundary Node tests: 7/7
- Playwright exact full suite, zero retries: 43/43
- POP back/forward/app-back soak, 2 workers, zero retries: 5/5
- focused >3s border-box padding + intrinsic min-height growth and outside-viewport pointer/wheel/key cancellation: 1/1
- `npm run test:store-listing` (37/37)
- `npm run store:validate`
- `npm run icons:check`
- `npm run release:verify`
- `node scripts/verify-ci-cache-policy.mjs`
- `npm audit --audit-level=high` (no high/critical findings; existing moderate advisories remain)
- `git diff --check`
- gitleaks exact PR diff: no leaks

## Adversarial review
- clean-head Codex review on exact `011e5dc975dc03d009a66343ed48e74010c9c28f`: no actionable correctness regressions
- prior review findings fixed: nested prefix-active routes now navigate to section roots; feature code imports the documented shared hook layer
- prior review findings fixed: slow data is covered by a hard 30s bound; size-only growth uses explicit border-box ResizeObserver while replacements refresh the direct-box observation set
- verified BrowserRouter initial `{idx: 0}` normalization is exact-shape, idempotent, and unknown state fails closed
- verified native recording freezes on browser/React entry-key mismatch for PUSH, REPLACE, back, and forward
- verified StrictMode replay, route switch, ref detach, timeout, and user input clean observers/listeners
- verified 64-entry eviction, full-bleed handoff, deliberate hash policy, reactive sentinel lifecycle, exact observer root, and page-1-only fresh Articles behavior

### DIFF
--- a/wallet/zuuli/src/components/layout/AppShell.tsx
+++ b/wallet/zuuli/src/components/layout/AppShell.tsx
@@ -4,6 +4,8 @@ import { TopBar } from "./TopBar";
 import { ScrollArea } from "@/components/ui/scroll-area";
 import { TriangleAlert } from "lucide-react";
 import { useWallet } from "@/store/wallet";
+import { useRouteScroll } from "@/hooks/useRouteScroll";
+import { RouteScrollProvider } from "./route-scroll";
 
 function LegacyWalletNotice() {
   const legacy = useWallet((state) => state.status?.legacyAppData);
@@ -29,8 +31,9 @@ function LegacyWalletNotice() {
   );
 }
 
-export function AppShell() {
+function AppShellContent() {
   const location = useLocation();
+  const { registerViewport } = useRouteScroll();
   // AI and login each own a bounded internal scroll region while the header
   // and mobile tabs stay pinned. Keep those routes in a plain, height-bound
   // main instead of nesting their scroll owners inside the generic page
@@ -49,7 +52,11 @@ export function AppShell() {
             <Outlet />
           </main>
         ) : (
-          <ScrollArea className="min-h-0 min-w-0 max-w-full flex-1" data-route-scroll>
+          <ScrollArea
+            className="min-h-0 min-w-0 max-w-full flex-1"
+            data-route-scroll
+            viewportRef={registerViewport}
+          >
             <main className="app-scroll-content mx-auto w-full min-w-0 max-w-6xl px-4 pt-6 md:px-8">
               <Outlet />
             </main>
@@ -58,5 +65,13 @@ export function AppShell() {
       </div>
       <MobileTabBar />
     </div>
+  );
+}
+
+export function AppShell() {
+  return (
+    <RouteScrollProvider>
+      <AppShellContent />
+    </RouteScrollProvider>
   );
 }
--- a/wallet/zuuli/src/components/layout/Sidebar.tsx
+++ b/wallet/zuuli/src/components/layout/Sidebar.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, type MouseEvent } from "react";
 import { NavLink, useLocation } from "react-router-dom";
 import { Wordmark } from "@/components/brand/Logo";
 import {
@@ -19,6 +19,7 @@ import {
   NAVIGATION_GROUP_ORDER,
   type NavigationRoute,
 } from "./navigation";
+import { useRouteScroll } from "@/hooks/useRouteScroll";
 
 const desktopLinkClass = ({ isActive }: { isActive: boolean }) =>
   cn(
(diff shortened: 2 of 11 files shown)


skyl:
Exact-head acceptance for `011e5dc975dc03d009a66343ed48e74010c9c28f`:

- required wallet run [32287480827](https://github.com/free2z/zuu/actions/runs/32287480827): aggregate gate `96181868719` SUCCESS; frontend `96180393985` and all Rust jobs SUCCESS
- packaging run [32287480809](https://github.com/free2z/zuu/actions/runs/32287480809): release identity `96180340607`, iOS `96180340772`, Linux `96180340852`, Android `96180340863`, and macOS universal `96180340872` all SUCCESS
- local exact full Playwright: 43/43, zero retries; focused >3s border-box/intrinsic growth and pointer/wheel/key cancellation green
- third clean-head `codex review --commit 011e5dc...`: no actionable correctness regressions
- worktree clean, remote/local head match, mergeable, zero unresolved review threads

Held unmerged for owner coordination.

PR #447, opened 2026-08-20
fix(zuuli): resolve relative article links against real routes

### LINKED ISSUE #337: ZUULI: relative links in article bodies dead-end in NotFound (and /uploadz file links never reach the media host)
## Summary

PR #334 fixed origin-relative **images** in article bodies (`` now
absolutized against the media host). Relative **links** in the same bodies have the
same root cause but need a different fix, so they were deliberately left out of that
PR — this is the follow-up.

`MarkdownLink` (`wallet/zuuli/src/components/common/Markdown.tsx:53-62, 87-143`)
classifies any href starting with `/` as internal and hands it to the SPA router:

```tsx
if (href.startsWith("#") || href.startsWith("/") || href.startsWith("?"))
  return false;   // → not external → navigate(href)
```

That rule is right for ZUULI's own paths, but article bodies are authored on
**free2z.cash**, where relative hrefs mean free2z routes — which mostly do not exist
in ZUULI.

## Failure

ZUULI's routes are `/login`, `/search/*`, `/creator/:username/*`, `/profile`,
`/kyc/*`, `/wallet/*`, `/ai/*`, `/live/*`, `/articles/*`, `/buy/*`, then
`<Route path="*" element={<NotFound />} />` (`src/App.tsx:93-180`;
articles resolve as `/articles/:slug`, `src/features/articles/index.tsx:12`).

So in a body authored for the web:

| Link in article body | Web meaning | ZUULI today |
|---|---|---|
| `/palmar/some-zpage` | that creator's article | `NotFound` |
| `/uploadz/public/x/report.pdf` | download the upload | `NotFound` (never reaches the media host) |
| `/palmar` | creator profile | `NotFound` (ZUULI uses `/creator/:username`) |

The reader is silently dumped on a Not Found page inside the app, with no indication
the destination exists on free2z. The `/uploadz/…` case is the worst: a linked PDF or
media file can never be reached at all, because routing intercepts it before it
becomes an HTTP request.

Note the sample article that prompted this work
(https://free2z.cash/palmar/7th-meetup-zcash-club-queretaro) uses absolute `https://`
links throughout, so nothing is currently broken *there* — this is latent for bodies
that use relative links, which is the common case for cross-linking on free2z.

## Why this needs a decision, not a prefix

Unlike images, there is no single correct target. Each class wants different handling:

1. **Media/upload paths** (`/uploadz/…`) — absolutize against `MEDIA_BASE` and open
   externally, like `mediaUrl()` does for images.
2. **free2z content paths** (`/{username}/{slug}`) — either map onto the in-app
   equivalent (`/articles/:slug`, `/creator/:username`) so navigation stays native, or
   open on free2z.cash externally. Mapping is nicer but needs a reliable way to tell a
   zpage path from an arbitrary free2z route.
3. **Genuine ZUULI paths** — must keep using the router as they do now.

A reasonable default: treat `/uploadz/…` as media, attempt a mapping for
`/{username}/{slug}`, and fall back to opening unknown relative paths on the free2z
web host rather than dead-ending in `NotFound`.

## Constraints for whoever picks this up

- Do **not** reintroduce the `components` map footgun documented in
  `Markdown.tsx` — never `cond ? X : undefined` (issue #319).
- Keep the comment variant's external-link hardening intact: untrusted comment links
  must still carry `rel="noopener noreferrer"` and open out-of-app.
- In-page `#anchor` links must keep smooth-scrolling (footnotes/heading slugs) — see
  the existing branch at `Markdown.tsx:95`.
- `Markdown.test.tsx` exists (vitest + `renderToStaticMarkup` in `MemoryRouter`, no
  jsdom) — extend it rather than adding a framework.

Related: #319, #334.

Jhonattan2121:
## Summary

Article bodies are authored on free2z.cash, where relative hrefs mean
free2z routes — which mostly don't exist in ZUULI. `MarkdownLink`
classified any href starting with `/` as an in-app path and handed it
straight to the router, dead-ending in `NotFound`.

Classifies relative hrefs before routing:

- `/uploadz/…` — absolutizes against the media host, opens out-of-app
  (same as `mediaUrl()` already does for images).
- Genuine ZUULI routes (`/wallet`, `/articles`, `/creator`, `/kyc`, …) —
  unchanged, still routed in-app.
- `/{username}/{slug}` — maps onto the slug-only `/articles/:slug`
  route.
- `/{username}` — maps onto `/creator/:username`.

## Testing

- `npm run typecheck` — clean
- `npm run test -- src/components/common/Markdown.test.tsx` — 14/14
  pass (4 new cases covering the mapping above)
- Full suite: 422 pass, 1 pre-existing unrelated failure
  (`TopBar.signed-in.test.tsx`, locale number formatting — reproduces
  on `main` before this change too)
- Manually verified in a running dev server (`VITE_MOCK=1`): typed all
  four link shapes into the article editor's live preview, confirmed
  rendered hrefs, clicked through — lands on `/articles/:slug` inside
  the router instead of the app-level `NotFound`, no console errors

## Note for reviewer

`APP_ROUTE_SEGMENTS` is a manual allowlist mirroring `src/App.tsx`'s
top-level routes. A future new top-level route needs to be added here
too, or it'll be misclassified as a creator username.

Fixes #337

### DIFF
--- a/wallet/zuuli/src/components/common/Markdown.test.tsx
+++ b/wallet/zuuli/src/components/common/Markdown.test.tsx
@@ -69,6 +69,38 @@ describe("Markdown links", () => {
     expect(markup).toContain('href="/articles/guide"');
     expect(markup).toContain(">the guide</a>");
   });
+
+  // free2z article bodies are authored on free2z.cash, where relative hrefs
+  // mean free2z routes — which mostly don't exist in ZUULI (#337). These must
+  // resolve to something real instead of dead-ending in `NotFound`.
+
+  it("keeps a genuine ZUULI app route routed as-is", () => {
+    const markup = render("[wallet](/wallet/send)", "article");
+
+    expect(markup).toContain('href="/wallet/send"');
+  });
+
+  it("maps a free2z {username}/{slug} content link onto the in-app article route", () => {
+    const markup = render("[post](/palmar/7th-meetup-zcash-club-queretaro)", "article");
+
+    expect(markup).toContain('href="/articles/7th-meetup-zcash-club-queretaro"');
+  });
+
+  it("maps a free2z {username} link onto the in-app creator route", () => {
+    const markup = render("[palmar](/palmar)", "article");
+
+    expect(markup).toContain('href="/creator/palmar"');
+  });
+
+  it("routes an /uploadz link as out-of-app media instead of the SPA router", () => {
+    const markup = render(
+      "[report](/uploadz/public/x/report.pdf)",
+      "article",
+    );
+
+    expect(markup).toContain('target="_blank"');
+    expect(markup).toContain('rel="noopener noreferrer"');
+  });
 });
 
 /**
--- a/wallet/zuuli/src/components/common/Markdown.tsx
+++ b/wallet/zuuli/src/components/common/Markdown.tsx
@@ -21,6 +21,7 @@ import rehypeSlug from "rehype-slug";
 
 import { openUrl } from "@tauri-apps/plugin-opener";
 import { isTauri } from "@/lib/platform";
+import { mediaUrl } from "@/lib/api/http";
 import { cn } from "@/lib/utils";
 
 import remarkOembed from "@/lib/markdown/remark-oembed";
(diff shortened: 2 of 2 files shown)


Jhonattan2121:
@skyl

skyl (approved):
Reviewed adversarially on the owner's behalf. **Approving** — CI `gate` is green (frontend tsc + vitest + 98 Playwright contracts all pass).

### What I verified beyond the description

- **`APP_ROUTE_SEGMENTS` actually mirrors `src/App.tsx`.** Checked segment-for-segment against the live `<Route path=…>` list: `login`, `search`, `creator`, `profile`, `kyc`, `wallet`, `ai`, `live`, `articles`, `buy` — plus `""` for the `index` route. No omissions, so no genuine in-app route gets misrouted to `/creator/:username`.
- **`/articles/:slug` is real** (`src/features/articles/index.tsx:12`), so the `{username}/{slug}` → `/articles/{slug}` mapping lands on a mounted route rather than the feed fallback.
- **No open-redirect through the media branch.** `mediaUrl()` only prefixes the `MEDIA_BASE` env constant and returns already-absolute inputs untouched; an `/uploadz/…` href can't be steered at an attacker host. The `?? href` fallback is only reachable when `mediaUrl` gets an empty path, which `classifyInternalHref` can't produce for the `uploadz` branch.
- **Untrusted-comment safety is preserved.** The new media branch opens only on an explicit click, so it introduces no silent network beacon — the one-item consent boundary that `variant="comment"` exists to enforce is untouched.
- **The `href === undefined` refactor is behavior-preserving.** The old branch rendered `href="#"` with an onClick that bailed before `preventDefault()`; the new one renders `href="#"` with no handler. Same result.

### Non-blocking follow-ups

1. **The manual allowlist is a real trap** — you flagged it yourself, and it's the right call to flag. A new top-level route added to `App.tsx` without a matching entry here silently becomes a "creator username" and 404s. Worth a cheap guard later: export the route table from `App.tsx` and derive `APP_ROUTE_SEGMENTS` from it, so the two can't drift.
2. **Query/hash are dropped on the mapped paths.** `classifyInternalHref` strips `?`/`#` to classify, then builds `to` from segments only — so `/palmar/my-post#section` routes to `/articles/my-post` and loses the anchor. Minor, and arguably fine given free2z and ZUULI don't share anchor schemes, but it's a silent loss.
3. **Schemeless relative hrefs fall through here too.** `isExternalHref` returns false for `foo/bar` and `../foo`, so they reach `classifyInternalHref` and get mapped as `{username}/{slug}`. Previously they were handed to `navigate()` as relative paths. Both are wrong-ish for genuinely relative markdown; neither is worse than the other. Not worth blocking on.

Good, tightly-scoped fix with tests that pin the actual mapping behavior rather than the implementation. Merging.

PR #448, opened 2026-08-20
fix(zuuli): parse and hint 2Z amounts in the user's own locale

### LINKED ISSUE #324: ZUULI: 2Z amount entry rejects the numbers the app itself displays (non-US locales)
## Problem

**This is a live bug, not future i18n work.** 2Z amounts are *displayed* with locale-aware grouping but *parsed* with a regex that only accepts ASCII commas. Any user whose locale groups numbers differently sees a number the app itself will reject.

`src/lib/format.ts:57-68` `parseTuzis`:

```js
if (!/^(?:0|[1-9]\d*|[1-9]\d{0,2}(?:,\d{3})+)$/.test(raw)) return null;
```

while `src/lib/format.ts:150` `formatTuzis` uses `toLocaleString()` with the runtime locale.

### What breaks where

| Locale | App displays | Parser accepts? |
|---|---|---|
| `de-DE` | `1.000.000` | no — periods rejected |
| `fr-FR` | `1 000 000` (U+202F narrow NBSP) | no |
| `hi-IN` / `en-IN` | `10,00,000` | no — regex demands strict 3-digit groups |
| Arabic-Indic digits | `١٬٠٠٠٬٠٠٠` | no — non-ASCII digits |

### Self-contradicting UI

`src/features/buy/BuyTab.tsx:245` renders the error `Max ${MAX_TUZIS.toLocaleString()} 2Z per purchase.` — locale-grouped — while `:247` instructs "commas may separate thousands" and the parser accepts only commas. So a German user is told the max in a format the app refuses.

Same pattern at: `src/features/buy/SendTab.tsx:181`, `src/features/creator/index.tsx:751`, `src/features/articles/components/TipDialog.tsx:114`, `src/features/articles/components/Comments/CommentForm.tsx:177`, `src/features/live/GoLiveDialog.tsx:195`, `src/features/profile/index.tsx:197`.

Also:
- `src/features/buy/BuyTab.tsx:220` — placeholder `"e.g. 3,000"` hardcodes ASCII grouping.
- `src/lib/format.ts:44-47` `tuziInputMaxLength()` assumes 3-digit groups, so it computes the wrong `maxLength` for `hi-IN`/`en-IN`.

**Severity is limited by luck:** preset buttons clear rather than prefill the input (`src/features/buy/SendTab.tsx:145` `setCustom("")`), so there is no automatic bad round-trip. The damage is user-facing inconsistency and rejected valid input, not silently wrong amounts.

## Must not change

`src/lib/format.ts:105-131` (`formatZec`/`formatZecTrim`/`splitZec`) and `:89-102` (`parseZecToZatoshis`) are **deliberately locale-insensitive** — BigInt, hard `.` decimal, 8dp, ASCII-only. That is correct for ZEC. Making them locale-aware would be a correctness regression.

Please add a comment saying so, or someone will eventually "fix" this into a consensus bug.

## Acceptance criteria

- [ ] Any 2Z amount the app displays is accepted verbatim when typed back in, in every supported locale.
- [ ] Parsing handles non-ASCII grouping separators (including U+202F/U+00A0), non-3-digit grouping (`hi-IN`), and non-ASCII digit systems.
- [ ] Input hint copy and placeholders describe the user's actual locale format rather than hardcoding commas.
- [ ] `tuziInputMaxLength()` is correct for non-3-digit grouping.
- [ ] ZEC formatting/parsing stays ASCII-only and locale-independent, with a comment stating the invariant and why.
- [ ] Round-trip tests cover at least `en-US`, `de-DE`, `fr-FR`, `hi-IN`, and an Arabic-Indic digit locale.

## Notes

Part of the i18n epic and a prerequisite for it — shipping more locales on top of this multiplies the bug. But it stands alone and is worth fixing regardless of whether the broader i18n work proceeds, since it affects users on non-US-format systems **today**, with no language setting involved.

Either make `parseTuzis` locale-aware (via `Intl.NumberFormat().formatToParts()` to discover the locale's separators) or introduce an explicitly ASCII-grouped `formatTuzisForInput` for values destined for input fields. The first is better UX; the second is a smaller change.

Epic: #322

Jhonattan2121:
## Summary

2Z amounts are displayed with locale-aware grouping (`toLocaleString()`)
but parsed with a regex that only accepted ASCII commas in strict
groups of three. Any user whose locale groups numbers differently
(de-DE periods, fr-FR narrow spaces, hi-IN's 3-then-2 grouping,
Arabic-Indic digits, ...) saw the app reject a number it had just
displayed.

- `parseTuzis`, `validateTuzis`, and `tuziInputMaxLength` now derive
  the grouping separator and digit glyphs from `Intl.NumberFormat` for
  the given locale (default: runtime locale, same as `formatTuzis`).
- Grouped input must match the locale's canonical grouping for that
  value exactly, so malformed shapes (`"1,00"`) are still rejected;
  ungrouped digit strings are always accepted.
- Hint copy and placeholders across Buy/Send/Tip/GoLive/Profile/Creator
  now show a locale-formatted example instead of a hardcoded
  `"e.g. 3,000"`.
- `formatZec`/`parseZecToZatoshis` (ZEC, not 2Z) are untouched and now
  carry an explicit comment: they stay ASCII-only by design, since ZEC
  is a consensus amount, not a display number.

## Testing

- `npm run typecheck` — clean
- `npm run test -- src/lib/format.test.ts` — 63/63 pass, including new
  round-trip coverage for `en-US`, `de-DE`, `fr-FR`, `hi-IN`, and an
  Arabic-Indic digit locale
- Full suite: 425 pass, 1 pre-existing unrelated failure (locale
  formatting in `TopBar.signed-in.test.tsx`, reproduces on `main`
  before this change too)
- Manually verified in a running dev server on a pt-BR system locale:
  the custom-amount field placeholder showed `"e.g. 3.000"`, typing
  `3.000` parsed correctly to 3,000 2Z with no error, typing `3.00`
  (wrong grouping) correctly showed the validation error

Fixes #324

### DIFF
--- a/wallet/zuuli/src/features/articles/components/TipDialog.tsx
+++ b/wallet/zuuli/src/features/articles/components/TipDialog.tsx
@@ -18,6 +18,7 @@ import { tuzi } from "@/lib/api/free2z";
 import {
   formatTuzis,
   MAX_TUZIS,
+  tuziInputExample,
   tuziInputMaxLength,
   validateTuzis,
 } from "@/lib/format";
@@ -163,7 +164,7 @@ export function TipDialog({ author }: { author: SimpleCreator }) {
               {amountResult.error === "tooLarge"
                 ? `Max ${MAX_TUZIS.toLocaleString()} 2Z per tip.`
                 : amount.length > 0 && amountResult.error !== null
-                  ? "Enter a positive whole 2Z amount; commas may separate thousands."
+                  ? `Enter a positive whole 2Z amount, e.g. ${tuziInputExample()}.`
                   : null}
             </p>
           </div>
--- a/wallet/zuuli/src/features/creator/index.tsx
+++ b/wallet/zuuli/src/features/creator/index.tsx
@@ -48,6 +48,7 @@ import {
   initials,
   MAX_TUZIS,
   timeAgo,
+  tuziInputExample,
   tuziInputMaxLength,
   validateTuzis,
 } from "@/lib/format";
@@ -976,7 +977,7 @@ function TipButton({
               {amountResult.error === "tooLarge"
                 ? `Max ${MAX_TUZIS.toLocaleString()} 2Z per tip.`
                 : amount.length > 0 && amountResult.error !== null
-                  ? "Enter a positive whole 2Z amount; commas may separate thousands."
+                  ? `Enter a positive whole 2Z amount, e.g. ${tuziInputExample()}.`
                   : null}
             </p>
           </div>
--- a/wallet/zuuli/src/features/live/GoLiveDialog.tsx
+++ b/wallet/zuuli/src/features/live/GoLiveDialog.tsx
@@ -20,6 +20,7 @@ import { useSession } from "@/store/session";
 import {
   formatTuzis,
   MAX_PPV_PRICE_TUZIS,
+  tuziInputExample,
   tuziInputMaxLength,
   validateTuzis,
 } from "@/lib/format";
@@ -194,7 +195,7 @@ export function GoLiveDialog() {
                 {priceResult.error === "tooLarge"
                   ? `Max ${MAX_PPV_PRICE_TUZIS.toLocaleString()} 2Z for PPV.`
                   : !validPrice
-                    ? "Enter a positive whole 2Z amount; commas may separate thousands."
+                    ? `Enter a positive whole 2Z amount, e.g. ${tuziInputExample()}.`
                     : null}
               </p>
             </div>
--- a/wallet/zuuli/src/features/profile/index.tsx
+++ b/wallet/zuuli/src/features/profile/index.tsx
@@ -18,6 +18,7 @@ import { profile } from "@/lib/api/free2z";
 import {
   initials,
   MAX_MEMBER_PRICE_TUZIS,
+  tuziInputExample,
   tuziInputMaxLength,
   validateTuzis,
 } from "@/lib/format";
@@ -196,7 +197,7 @@ function ProfileForm({ user }: { user: AuthUser }) {
               {memberPriceResult.error === "tooLarge"
                 ? `Max ${MAX_MEMBER_PRICE_TUZIS.toLocaleString()} 2Z per membership.`
                 : !memberPriceValid
-                  ? "Enter a non-negative whole 2Z amount; commas may separate thousands."
+                  ? `Enter a non-negative whole 2Z amount, e.g. ${tuziInputExample()}.`
                   : null}
             </p>
           </div>
--- a/wallet/zuuli/src/features/wallet/funding/BuyTab.tsx
+++ b/wallet/zuuli/src/features/wallet/funding/BuyTab.tsx
@@ -38,6 +38,7 @@ import {
   formatUsd,
   formatZecTrim,
   MAX_TUZIS,
+  tuziInputExample,
   tuziInputMaxLength,
   tuzisToUsd,
   validateTuzis,
@@ -246,7 +247,7 @@ export function BuyTab() {
             <Input
               id="custom-tuzis"
               inputMode="numeric"
-              placeholder="e.g. 3,000"
+              placeholder={`e.g. ${tuziInputExample()}`}
               maxLength={tuziInputMaxLength(MAX_TUZIS)}
               value={custom}
               onChange={(e) => setCustom(e.target.value)}
@@ -270,7 +271,7 @@ export function BuyTab() {
             {hasCustomAmount && customAmount.error === "tooLarge"
               ? `Max ${MAX_TUZIS.toLocaleString()} 2Z per purchase.`
               : hasCustomAmount && customAmount.error !== null
-                ? "Enter a positive whole 2Z amount; commas may separate thousands."
+                ? `Enter a positive whole 2Z amount, e.g. ${tuziInputExample()}.`
                 : null}
           </div>
         </div>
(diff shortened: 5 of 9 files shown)


skyl (changes requested):
Reviewed adversarially on the owner's behalf. The core fix is **right** and I want it in — but **CI `gate` is red**, so I can't merge it yet. One concrete thing to fix.

### Blocking: a Playwright contract still asserts the old hint copy

`tests/paid-actions.pw.ts:136` pins the exact string you replaced:

```
tests/paid-actions.pw.ts:115:1 › send keeps malformed restored money text invalid instead of coercing it to a preset
  Locator: getByText('Enter a positive whole 2Z amount; commas may separate thousands.')
  Error: element(s) not found
```

97 passed, 1 failed. The vitest suite you ran doesn't cover this — the Playwright contracts under `tests/` are a separate runner in the same `frontend` job, so `npm run test` alone won't surface it. Worth running the full `frontend` job locally before the next push.

**Careful how you fix it.** The obvious move — paste the new literal in — reintroduces exactly the bug this PR is about, because the new copy interpolates `tuziInputExample()`, which is runtime-locale-dependent. Hardcoding `"…, e.g. 3,000."` would make that test pass in CI's en-US and fail on any dev machine with a different locale. Assert the locale-independent part instead:

```ts
await expect(
  page.getByText(/Enter a positive whole 2Z amount, e\.g\./),
).toBeVisible();
```

### What I verified in the parser itself

- **Round-trip logic is sound.** The `hasGroupSeparator && num.toLocaleString(locale) !== raw` re-check is the right guard — it's what keeps `"1,00"` from silently parsing as `100` now that the strict `(?:,\d{3})+` shape is gone. Good instinct to keep the strictness rather than trade it for leniency.
- **Iterating `stripped` by code point** (`for…of`, not index) is correct for numbering systems with astral-plane digits, and `tuziInputMaxLength`'s `.length` counting UTF-16 code units happens to match how `<input maxLength>` counts. Both right.
- **The length bound still fires before `BigInt`** — `"9".repeat(10_000)` returns `null` at the `MAX_SAFE_INTEGER_DIGITS` check, so no unbounded BigInt construction.
- **Leaving `parseZecToZatoshis`/`formatZec` ASCII-only is the correct call**, and the explicit "don't make this locale-aware too" comment is exactly the right artifact to leave behind. ZEC is a consensus amount; a display parser has no business near it.
- **All call sites are covered.** I grepped every `parseTuzis`/`validateTuzis`/`tuziInputMaxLength` caller — `CommentForm.tsx`, `comment-weight.ts`, and `SendTab.tsx:113` all pick up the runtime-locale default, consistent with `formatTuzis`.

### Non-blocking notes

1. **`SendTab.tsx:113` parses persisted state with a display parser.** `parseTuzis(restoredSend.amount)` reads a value written at line 319 as `custom || String(selectedAmount)`. The preset path is ASCII and safe everywhere; the `custom` path persists the user's raw locale-formatted text. If the system locale changes between sessions, the restored draft no longer parses. The failure is graceful — it falls through to `setCustom(restoredSend.amount)` and shows the validation error — so this isn't a blocker, but persisting a display string is the underlying smell. Consider persisting the parsed number.
2. **`TopBar.signed-in.test.tsx:46` is the same bug class, still latent.** It asserts `Balance 4,210 2Z` with a hardcoded ASCII comma, which is why it fails for you locally and passes in CI's en-US. Your description is right that it's pre-existing and independent of this change — but it's #324 wearing a different hat. Worth a follow-up that pins the locale the way you did in `format.test.ts`.
3. **`AMOUNTS` has `1_000_000` twice** — once as `1_000_000` and once as `10_00_000`. Harmless, and I get that the second spelling documents the lakh grouping, but it's not adding a case.
4. **fr-FR users can no longer type an ASCII comma.** With `group` = U+202F, `"3,000"` now returns `null` where it used to parse. That's a deliberate consequence of mirroring display, and typing a narrow no-break space isn't realistic either — ungrouped `3000` still works, which is what those users will do. Flagging it as a known behavior narrowing, not a defect.

Push the Playwright fix and I'll re-review and merge as soon as `gate` is green.

skyl (approved):
Unblocked and **approving** — `gate` is green.

To keep you moving rather than round-tripping on a one-line test fix, I pushed it to this branch myself (`878355b..451dcfd`, then a merge of `main` to satisfy the up-to-date branch requirement after #447 landed). The change is exactly what I suggested above — `tests/paid-actions.pw.ts` now asserts the locale-independent prefix:

```ts
// The example in the hint is rendered in the browser's own locale, so match
// only the locale-independent prefix.
await expect(
  page.getByText(/Enter a positive whole 2Z amount, e\.g\./),
).toBeVisible();
```

That was the only stale assertion in the repo — a sweep for the old copy across `*.ts`/`*.tsx`/`*.md` found no other Playwright or vitest file pinning it.

Verified before pushing, and again in CI on the final SHA:

- `npx playwright test tests/paid-actions.pw.ts` — 15 passed, including the previously-failing `paid-actions.pw.ts:115`
- `npx vitest run` — 426 passed, 5 skipped
- **Locale-independence proved at the real conditions**, not assumed: temporarily set `locale: "de-DE"` in `playwright.config.ts` and re-ran the test — passes. Under de-DE the hint renders `3.000`, so the hardcoded-literal version of this fix would have failed there. Config reverted.
- CI on `365bd49`: **430 vitest passed / 5 skipped, 98 Playwright passed** (was 97 + 1 failed), `gate` green.

The four non-blocking notes in my earlier review still stand as follow-ups — particularly the `SendTab.tsx:113` persisted-display-string smell and the still-latent `TopBar.signed-in.test.tsx:46` hardcoded `4,210`, which is #324 in a different costume. Neither belongs in this PR.

Nice fix — deriving the separator and digit glyphs from `Intl.NumberFormat` instead of widening the regex is the right shape, and keeping the canonical-grouping re-check so `"1,00"` still fails is the detail that makes it safe. Merging.

PR #453, opened 2026-08-21
test(zuuli): pin test runner locale

### LINKED ISSUE #449: test(zuuli): both test suites inherit the machine locale — green in CI, red on any non-en-US machine
## Problem

Neither test runner pins a locale, so both suites inherit the **machine's** locale. Assertions that hardcode ASCII-grouped numbers pass on CI (`ubuntu-latest`, en-US) and fail on any contributor's machine set to a locale that groups differently.

This is not hypothetical — it cost real review cycles on #448. The contributor (on a pt-BR system) reported a failing test in `main` that CI showed as green, and it had to be triaged as pre-existing-and-unrelated before their actual fix could be evaluated.

## The two assertions

```
src/components/layout/TopBar.signed-in.test.tsx:46
  expect(markup).toContain('aria-label="Buy 2Zs. Balance 4,210 2Z"');

tests/paid-actions.pw.ts:362
  await expect(page.getByText(/Balance: 4,210 2Z/)).toBeVisible();
```

Both render through `formatTuzis` → `toLocaleString()` with no explicit locale. Under `de-DE` the UI produces `4.210`, under `fr-FR` `4 210` (U+202F), under `hi-IN` `4,210` only coincidentally. Both assertions fail on the first two.

## Root cause — the configs, not the assertions

- `playwright.config.ts` `use:` sets `baseURL`, `browserName`, `channel`, `colorScheme`, `trace` — **no `locale`**.
- `vite.config.ts` has no `test:` block pinning a locale for vitest either.

So the suite is only deterministic by accident of where it runs.

## Why this matters beyond the two lines

This is the same defect class as #324 (which #448 just fixed in the parser): **display formatting treated as a stable string**. Fixing only these two assertions leaves the door open — the next hardcoded `1,234` lands green in CI and breaks the next non-en-US contributor. The config gap is the actual bug.

## Suggested fix

1. Pin the locale in `playwright.config.ts`:
   ```ts
   use: { locale: "en-US", timezoneId: "UTC", /* … */ }
   ```
   (`timezoneId` closes the identical latent hole for any date/time assertion.)
2. Pin the equivalent for vitest so `toLocaleString()` is deterministic under `jsdom`/`happy-dom` regardless of the host.
3. Then fix the two assertions above — with the locale pinned they become legitimately deterministic and can keep asserting `4,210`.
4. Optional follow-up: add a second Playwright project running a non-en locale over a small smoke subset, so locale regressions are *proven* rather than merely not-tripped-over. #448 established the pattern — `src/lib/format.test.ts` now round-trips across `en-US`/`de-DE`/`fr-FR`/`hi-IN`/Arabic-Indic digits.

## Acceptance

- Both suites produce identical results on an en-US and a de-DE machine.
- A newly-added assertion hardcoding a grouped number cannot pass on one machine and fail on another.

Related: #322 (i18n epic), #325, #326. Found while reviewing #448.

henriquejsza:
## Summary

- pin Vitest worker locale and timezone to en-US/UTC, including `LC_ALL` so it cannot override `LANG` on POSIX hosts
- pin Playwright browser contexts to en-US/UTC
- keep production locale-aware formatting and the existing `4,210` assertions unchanged

## Validation

- reproduced before the fix: the focused TopBar test passed under `en_US` and failed under `de_DE` (`4.210` vs `4,210`)
- focused Vitest and Playwright assertions pass under both `en_US`/`America/New_York` and `de_DE`/`Europe/Berlin`
- full frontend workflow surface under an outer `de_DE`/`Europe/Berlin` environment with Node 24.19.0:
  - 430 Vitest tests passed, 5 skipped
  - 77 Node tests passed
  - 98 Playwright tests passed
  - typecheck and production build passed
  - cache, icon, store, TestFlight, and high/critical audit gates passed
- action-pin and Rust-toolchain policy checks passed

Closes #449

### DIFF
--- a/wallet/zuuli/playwright.config.ts
+++ b/wallet/zuuli/playwright.config.ts
@@ -16,6 +16,8 @@ export default defineConfig({
     browserName: "chromium",
     channel: process.env.CI ? "chrome" : undefined,
     colorScheme: "dark",
+    locale: "en-US",
+    timezoneId: "UTC",
     trace: "retain-on-failure",
   },
   webServer: {
--- a/wallet/zuuli/vite.config.ts
+++ b/wallet/zuuli/vite.config.ts
@@ -1,4 +1,4 @@
-import { defineConfig } from "vite";
+import { defineConfig } from "vitest/config";
 import react from "@vitejs/plugin-react";
 import path from "node:path";
 
@@ -14,6 +14,15 @@ const apiTarget = process.env.VITE_F2Z_PROXY || "https://stage.free2z.cash";
 // reference wallet (1421). Tauri drives this via beforeDevCommand.
 export default defineConfig(async () => ({
   plugins: [react()],
+  test: {
+    // Workers must not inherit a contributor's locale or timezone. LC_ALL is
+    // pinned with LANG because it takes precedence on POSIX systems.
+    env: {
+      LANG: "en_US.UTF-8",
+      LC_ALL: "en_US.UTF-8",
+      TZ: "UTC",
+    },
+  },
   // `mathjax-full` (pulled in by rehype-mathjax for offline SVG math in the
   // Markdown renderer) reads its version via `eval('require')(...)` UNLESS a
   // global `PACKAGE_VERSION` is defined — see mathjax-full/js/components/


skyl (approved):
Reviewed adversarially on the owner's behalf. **Approving** — `gate` is green, and more importantly I verified this one **at the real conditions**, because CI cannot.

Welcome, and thank you — this is a well-aimed first contribution.

### Why CI passing means nothing here, and what I did instead

The `frontend` job runs on `ubuntu-latest` in en-US. `TopBar.signed-in.test.tsx:46` asserts `4,210` and passes there **with or without this change**. A green gate is necessary but proves nothing about the actual fix.

So I reproduced a non-en-US contributor machine and ran a real before/after. Same command both times:

```
LANG=de_DE.UTF-8 LC_ALL=de_DE.UTF-8 TZ=Europe/Berlin \
  npx vitest run src/components/layout/TopBar.signed-in.test.tsx
```

**Baseline (`main`, 7a60687) — fails:**
```
× keeps money controls while rendering unknown ZEC as an em dash
  Expected: aria-label="Buy 2Zs. Balance 4,210 2Z"
  Received: aria-label="Buy 2Zs. Balance 4.210 2Z"
```

**This PR (4ead06b) — passes:**
```
✓ src/components/layout/TopBar.signed-in.test.tsx (1 test) 28ms
```

Full suite under the same de_DE host: `1 failed | 33 passed` before, `34 passed` after. That's the proof the issue was asking for.

### I expected the vitest half to be a no-op, and was wrong

Worth writing down, because the reasoning nearly cost you a bogus round of changes. Node fixes ICU's default locale at **process boot** — assigning `process.env.LC_ALL` at runtime does not move it:

```
LANG=de_DE.UTF-8 node -e '(4210).toLocaleString()            // 4.210
                          process.env.LC_ALL="en_US.UTF-8"
                          (4210).toLocaleString()'           // still 4.210
```

(`TZ` is the exception — it *does* re-read at runtime.) So `test.env` only works if vitest applies it at worker **spawn**. It does: vitest 3.2.7 defaults to `pool: 'forks'`, and the workers are real child processes. A probe inside a worker under this PR confirms it:

```
PROBE isMainThread= true      <- separate process, not a worker_thread
PROBE LANG= en_US.UTF-8  LC_ALL= en_US.UTF-8  TZ= UTC
PROBE resolvedLocale= en-US   PROBE toLocaleString= 4,210
```

Your fix is right, and right for a subtler reason than the diff shows.

### One non-blocking nit — please add a comment, it's load-bearing

The same config forced onto the threads pool (`--pool=threads`):

```
PROBE isMainThread= false
PROBE LANG= en_US.UTF-8  LC_ALL= en_US.UTF-8  TZ= UTC   <- env IS set...
PROBE resolvedLocale= de-DE                              <- ...but ICU already booted
PROBE toLocaleString= 4.210                              -> test fails
```

So this fix silently becomes a **no-op under `pool: 'threads'`**, for locale *and* timezone. It works today purely because of vitest's default. Anyone adding `pool: 'threads'` as a perf tweak un-fixes #449 with zero signal on en-US CI — exactly the failure mode #449 exists to kill.

Your existing comment explains the `LC_ALL`/`LANG` precedence well. Worth one more line next to it:

```ts
// Requires the `forks` pool (vitest's default): env is applied at worker
// spawn, and Node fixes ICU's locale at process boot — under `threads` the
// worker inherits the parent's already-booted locale and this silently no-ops.
```

Not blocking; I'd rather land this now than round-trip. Happy to take it as a follow-up.

### On your Playwright validation

Flagging a subtlety in case it misled you: on macOS, `LANG=de_DE.UTF-8 npx playwright test` does **not** reproduce the bug — Chromium takes its locale from system preferences, not `LANG`. Baseline passes all 15 under that command, so it can't falsify anything. I confirmed the `locale` option is genuinely load-bearing a different way — by setting `locale: "de-DE"` on the **baseline** config, which fails as expected:

```
✘ tests/paid-actions.pw.ts:352 › low-balance top-up diagnosis appears only after authentication
  Locator: getByText(/Balance: 4,210 2Z/)  ->  element(s) not found
  1 failed, 14 passed
```

Since Playwright's `locale` overrides the machine default in that direction, your `locale: "en-US"` pins it in the other. Both halves are real.

### Also checked

- `npm run typecheck` — clean; the `vite` → `vitest/config` import change typechecks fine and doesn't disturb `vite build` (the config object is a superset, and `vitest` is already a devDependency alongside `vite` itself).
- The `node --test scripts/*.node-test.mjs` suites aren't covered by `test.env`, but none are locale-sensitive — all 77 pass under de_DE.
- Good call leaving the `4,210` assertions alone. With the locale pinned they're legitimately deterministic; rewriting them as regexes would have been strictly worse.

Merging. Thanks again — the "reproduced before the fix" line in your PR description is exactly the right instinct, and it's what made this quick to verify.

PR #455, opened 2026-08-22
test(zuuli): guard worker locale

### LINKED ISSUE #454: test(zuuli): the locale pin silently no-ops under vitest's threads pool
## Context

#453 fixed #449 by pinning the runners' locale/timezone. It works — verified with a real failing baseline under a de_DE host. This is a small hardening follow-up, not a defect in that PR.

## The fragility

`wallet/zuuli/vite.config.ts` pins the vitest locale via:

```ts
test: {
  env: { LANG: "en_US.UTF-8", LC_ALL: "en_US.UTF-8", TZ: "UTC" },
}
```

This works **only because vitest defaults to `pool: 'forks'`**. Workers are child processes, so `test.env` is applied at spawn and Node's ICU boots with the pinned locale.

Node fixes ICU's default locale at **process boot**; assigning `process.env.LC_ALL` afterwards does not move it:

```
LANG=de_DE.UTF-8 node -e '(4210).toLocaleString()          // 4.210
                          process.env.LC_ALL="en_US.UTF-8"
                          (4210).toLocaleString()'         // still 4.210
```

So under `pool: 'threads'` the pin silently stops working. Measured, same config, `--pool=threads`:

```
PROBE isMainThread= false
PROBE LANG= en_US.UTF-8  LC_ALL= en_US.UTF-8  TZ= UTC   <- env IS set...
PROBE resolvedLocale= de-DE                              <- ...but ICU already booted
PROBE toLocaleString= 4.210                              -> TopBar test fails
```

Both locale **and** timezone revert. `pool: 'threads'` is a common performance tweak, and switching to it would regress #449 with **no signal on CI**, since CI runs en-US where the assertions pass either way. That's precisely the failure mode #449 was filed to eliminate.

## Suggested fix

Cheapest version — a comment recording the dependency, next to the existing `LC_ALL`/`LANG` note:

```ts
// Requires the `forks` pool (vitest's default): env is applied at worker
// spawn, and Node fixes ICU's locale at process boot — under `threads` the
// worker inherits the parent's already-booted locale and this silently no-ops.
```

Stronger version, if it's cheap: assert the pin inside the suite so drift fails loudly rather than silently — a one-line test that `Intl.NumberFormat().resolvedOptions().locale === "en-US"`. That makes the guarantee self-enforcing under any pool, and fails on the config change rather than on some unrelated assertion months later.

Either is fine. The second is worth slightly more because it holds regardless of what a future vitest version defaults to.

## Acceptance

- Switching to `pool: 'threads'` either keeps the locale pinned, or fails a test that names the reason.

Good first issue — self-contained, and the mechanism is fully written up above. Found while verifying #453.

henriquejsza:
## Summary

- add a dedicated Vitest environment test for the configured `en-US` ICU locale
- make a future switch away from fork workers fail at the locale contract instead of an unrelated UI assertion

## Validation

- under `LANG=de_DE.UTF-8 LC_ALL=de_DE.UTF-8 TZ=Europe/Berlin` with Node 24.19.0:
  - the focused test passes with Vitest's default forks pool
  - the focused test fails as intended with `--pool=threads` (`de-DE` received, `en-US` expected)
  - 431 Vitest tests passed, 5 skipped
  - 77 Node tests passed
  - 98 Playwright tests passed
  - typecheck and production build passed
  - cache, icon, store, TestFlight, dependency-audit, action-pin, and Rust-toolchain policy checks passed

Closes #454


skyl (approved):
Reviewed adversarially on the owner's behalf. **Approving.**

### What you got right, and it's the part that's easy to get wrong

You asserted on `Intl.NumberFormat().resolvedOptions().locale` rather than on `process.env.LANG`. That distinction is the whole guard. Under `pool: 'threads'` the environment variables **are** set correctly — a `process.env` assertion would pass green while the bug is live and the suite silently formats in the host locale. Checking what ICU actually resolved is the only version of this test that has teeth.

Naming the fork-pool dependency in the test title is the other half: when this fails, the failure explains its own cause instead of sending someone hunting through `vite.config.ts`.

And you again proved it fails under `--pool=threads` before claiming it works. That's twice now.

### I pushed one addition

The guard covered the locale but not the timezone, and `vite.config.ts` pins both (`TZ: "UTC"`). The #454 measurement showed both revert together under threads — `resolvedLocale= de-DE` *and* `resolvedTZ= Europe/Berlin` — so a date or time assertion could regress with exactly the same silence, and the guard wouldn't have noticed. Added as a sibling case in your file, matching your naming:

```ts
it("uses fork workers so ICU boots with the configured UTC timezone", () => {
  expect(Intl.DateTimeFormat().resolvedOptions().timeZone).toBe("UTC");
});
```

Verified both directions under `LANG=de_DE.UTF-8 LC_ALL=de_DE.UTF-8 TZ=Europe/Berlin`:

```
# default (forks) pool
✓ src/test-environment.test.ts (2 tests) 14ms

# --pool=threads
× ...configured en-US locale   → expected 'de-DE' to be 'en-US'
× ...configured UTC timezone   → expected 'Europe/Berlin' to be 'UTC'
Test Files  1 failed (1)
```

That reproduces #454's probe exactly. Full gate green: typecheck clean, `build ✓ 9.86s`, vitest 432 passed / 5 skipped, 98 Playwright passed.

Thanks — this closes the loop on #449 properly. The locale pin now protects itself.

Issue #382, opened 2026-08-16
Profile dashboard: port the real mobile-overflow fix — #366 shipped the pre-fix markup


skyl:
PR #366 ported the profile-dashboard fixes from the production repo, but it was cut from a revision that was **itself still broken**. The upstream follow-up fix landed after the port, so `zuu` currently carries the pre-fix markup and reproduces both bugs.

## Bug 1 — horizontal scroll at mobile widths

In `ts/svelte/free2z/src/routes/[username]/dashboard/profile/+page.svelte`, the draft and published row containers are:

```svelte
<div class="flex flex-col items-start justify-between gap-4 sm:flex-row sm:items-center">
```

`align-items: flex-start` on a **column** flex container sizes children to fit-content on the **cross (horizontal)** axis. So the inner `<div class="min-w-0 flex-1 space-y-1">` measured **479px inside a 254px parent** — `min-w-0` and `flex-1` are powerless there, because they govern the vertical main axis. The `<h3 class="truncate">` (which implies `white-space: nowrap`) then sets that fit-content width to the entire title length.

Measured on the deployed production build before the fix: **801px wide at a 386px viewport — 415px of overflow.**

**Fix:** drop `items-start`, keeping `sm:items-center`. That single change measured 801px → 386px.

## Bug 2 — UI copy is ellipsized

The stat tiles carry `truncate` on both the value and the label, so they render `64,6…` and `2Z BALAN…`. At 386px the tile's text box is only 75px, because the icon chip + gap + padding consume 88px of the 161px tile, while "Published pages" needs 120px and the balance needs 123px.

**Fix (measured to give zero clipping and zero overflow at 320/360/390):**
- remove `truncate` from every stat value and label — UI copy never truncates;
- stack the tile on mobile: `flex flex-col items-start gap-2 p-4 sm:flex-row sm:items-center sm:gap-3`;
- value `text-xl sm:text-2xl`; labels wrap;
- `col-span-2 sm:col-span-1` on **both** the balance and fans tiles — with only one spanning, four tiles occupy five grid cells and leave a visible hole beside Fans.

A 3-up variant was tried and rejected: at three columns the label box is 75px while the single word "PUBLISHED" needs ~77px.

## Bug 3 — zpage titles
Replace `truncate` on the row `<h3>` with `line-clamp-2 break-words`. Rule: **UI copy never truncates; user-authored content may wrap and clamp.**

## Testing note
[...]

PR #456, opened 2026-08-22
fix(zuuli): UI copy never truncates, and make the rule enforceable

skyl:
## The rule

**UI copy never truncates.** Layouts are designed so the words fit. Only *opaque
identifiers* may truncate, and they truncate in the **middle, tail-weighted**,
because the trailing characters are the checksum a human verifies. User-authored
body content may wrap and `line-clamp`.

It was violated in ~20 places for one reason: **the rule was written nowhere an
agent reads it.** It lived in the body of issue #382 and a doc comment at
`src/lib/format.ts:216-223`. This PR writes it into `wallet/zuuli/CLAUDE.md`'s
Design rules, makes it enforceable, and then fixes the sites.

## The test suite was hiding the defects

`tests/viewport.pw.ts:222-224` treated `.truncate` / `text-overflow: ellipsis` /
`line-clamp` as a **legitimate exemption** from the overflow audit. Every
violation below was whitelisted by the thing meant to catch it.

The inverse assertion is now in place, in two layers:

| Layer | Catches |
| --- | --- |
| `scripts/ui-copy-truncation.node-test.mjs` | Ellipsis utilities anywhere in `src/`, and any `line-clamp-*` whose own JSX element lacks `data-user-content`. Covers components **no route renders** — `SeedReveal` among them. |
| `tests/viewport.pw.ts` | Ellipsis/clamp that arrives through **computed style** rather than a utility class, at 320/360/390/414px, signed in and signed out. |

**How UI copy is told apart from user content:** `data-user-content`, an
annotation on the clamped element itself. Not an allowlist — nothing is
grandfathered, and a *new* `truncate` fails on its first run without anyone
remembering to update a file. The annotation is also what now grants the
overflow exemption, so an unannotated clip can no longer be whitelisted twice.

## The fixture gap #382 documents

> *"the upstream Playwright suite passed while production overflowed by 415px,
> because its fixture titles were 21 characters."*

`src/lib/api/mock-data.ts` and `src/lib/wallet/mock.ts` now carry 60-90 character
titles, long display names, real-sentence memos, and **two fixtures with an
unbroken token** that has nowhere to wrap. Strings other suites assert verbatim
(`Zooko`, `Free2Z`, `Why Shielded-by-Default Matters`, `Remote Media Consent
Audit`) are unchanged — their subtitles and bios carry the length instead.

## Failing before

Both new assertions were run against **unfixed** code first.

`node --test scripts/ui-copy-truncation.node-test.mjs` — **38 + 11 violations, 2/2 failing**:

```
not ok 1 - no source file ships a CSS ellipsis clip
  UI copy never truncates, and identifiers truncate in the middle via
  truncateAddress()/truncateMiddle() — never with a CSS clip on top. Remove these:
    src/features/ai/ModelPicker.tsx:56 uses "truncate"
    src/features/ai/ModelPicker.tsx:92 uses "truncate"
    src/features/ai/PersonalityManager.tsx:214 uses "truncate"
    src/features/ai/PersonalityPicker.tsx:55 uses "truncate"
    src/features/ai/PersonalityPicker.tsx:136 uses "truncate"
    src/features/ai/index.tsx:372 uses "truncate"
    src/features/articles/components/ArticleCard.tsx:109 uses "truncate"
    src/features/auth/SeedReveal.tsx:66 uses "truncate"
    src/features/creator/index.tsx:261 uses "truncate"
    src/features/home/ArticlesGrid.tsx:49 uses "truncate"
    src/features/home/CreatorsRow.tsx:37 uses "truncate"
    src/features/home/LiveRail.tsx:62 uses "truncate"
    src/features/home/parts.tsx:37 uses "truncate"
    src/features/home/parts.tsx:42 uses "truncate"
    src/features/live/Room.tsx:249 uses "truncate"
    src/features/live/Room.tsx:250 uses "truncate"
    src/features/live/Room.tsx:937 uses "truncate"
    src/features/live/Room.tsx:944 uses "truncate"
    src/features/live/StreamCard.tsx:108 uses "truncate"
    src/features/live/StreamCard.tsx:109 uses "truncate"
    src/features/profile/LinkedAccounts.tsx:174 uses "truncate"
    src/features/profile/LinkedAccounts.tsx:296 uses "truncate"
    src/features/profile/index.tsx:237 uses "truncate"
    src/features/profile/index.tsx:247 uses "truncate"
    src/features/search/index.tsx:286 uses "truncate"
    src/features/search/index.tsx:308 uses "truncate"
    src/features/search/index.tsx:373 uses "truncate"
    src/features/search/index.tsx:377 uses "truncate"
    src/features/search/index.tsx:388 uses "truncate"
    src/features/wallet/components.tsx:71 uses "truncate"
    src/features/wallet/components.tsx:176 uses "truncate"
    src/features/wallet/funding/ActivityTab.tsx:160 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:498 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:499 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:518 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:523 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:633 uses "truncate"
    src/features/wallet/funding/SendTab.tsx:637 uses "truncate"

not ok 2 - every line-clamp is annotated as user-authored content
  line-clamp is reserved for user-authored body content, and the markup must say
  so. Add data-user-content to the clamped element — or, if it is UI copy, let it wrap:
    src/features/ai/PersonalityManager.tsx:226 clamps "line-clamp-2" without data-user-content
    src/features/ai/PersonalityPicker.tsx:145 clamps "line-clamp-1" without data-user-content
    src/features/articles/components/ArticleCard.tsx:87 clamps "line-clamp-2" without data-user-content
    src/features/creator/index.tsx:1070 clamps "line-clamp-2" without data-user-content
    src/features/creator/index.tsx:1074 clamps "line-clamp-2" without data-user-content
    src/features/home/ArticlesGrid.tsx:39 clamps "line-clamp-2" without data-user-content
    src/features/home/ArticlesGrid.tsx:43 clamps "line-clamp-2" without data-user-content
    src/features/home/CreatorsRow.tsx:47 clamps "line-clamp-2" without data-user-content
    src/features/home/LiveRail.tsx:58 clamps "line-clamp-2" without data-user-content
    src/features/live/StreamCard.tsx:94 clamps "line-clamp-2" without data-user-content
    src/features/search/index.tsx:383 clamps "line-clamp-2" without data-user-content

1..2
# pass 0
# fail 2
```

`npx playwright test tests/viewport.pw.ts` — **8 of 9 failing.** Excerpt from
`/` at 320px signed out (32 failures on that route alone), with the lengthened
fixtures also surfacing genuine overflow the old exemption had been swallowing:

```
  ✘  1 signed out routes fit at 320px (1.5s)
  ✘  2 signed out routes fit at 360px (1.2s)
  ✘  3 signed out routes fit at 390px (861ms)
  ✘  4 signed out routes fit at 414px (871ms)
  ✘  5 signed in routes fit at 320px (1.3s)
  ✘  6 signed in routes fit at 360px (922ms)
  ✘  7 signed in routes fit at 390px (949ms)
  ✘  8 signed in routes fit at 414px (940ms)

  1) signed out routes fit at 320px ─────────────────────────────────────────
    Error: / has clipped descendants
    + Array [
    +   "h2 “Live now” ellipsizes UI copy (text-overflow: ellipsis, line-clamp: none) — UI copy must never truncate",
    +   "p “Creators broadcasting this moment” ellipsizes UI copy (text-overflow: ellipsis, line-clamp: none) — UI copy must never truncate",
    +   "p “Creators broadcasting this moment” scrollWidth 228 exceeds clientWidth 201",
    +   "h3 “Shielded & Chill — building a Zcash light client live, from ” ellipsizes UI copy (text-overflow: clip, line-clamp: 2) — UI copy must never truncate",
    +   "span “@nine” ellipsizes UI copy (text-overflow: ellipsis, line-clamp: none) — UI copy must never truncate",
    +   "h2 “Fresh articles” ellipsizes UI copy (text-overflow: ellipsis, line-clamp: none) — UI copy must never truncate",
    +   "p “Long-form writing from the community” scrollWidth 250 exceeds clientWidth 201",
    +   "span “Nine Anonymous-Broadcaster” scrollWidth 174 exceeds clientWidth 155",
    +   "div “Zooko” ellipsizes UI copy (text-overflow: ellipsis, line-clamp: none) — UI copy must never truncate",
    +   "a “View Maya Andonovska-Rasmussen ⛏️'s profile” scrollWidth 191 exceeds clientWidth 134",
    +   ... 22 more
    + ]
```

## Passing after

```
$ node --test scripts/ui-copy-truncation.node-test.mjs
ok 1 - no source file ships a CSS ellipsis clip
ok 2 - every line-clamp is annotated as user-authored content
# tests 2
# pass 2
# fail 0

$ npx playwright test tests/viewport.pw.ts
  ✓  1 signed out routes fit at 320px (14.4s)
  ✓  2 signed out routes fit at 360px (13.9s)
  ✓  3 signed out routes fit at 390px (13.9s)
  ✓  4 signed out routes fit at 414px (13.9s)
  ✓  5 signed in routes fit at 320px (14.5s)
  ✓  6 signed in routes fit at 360px (14.1s)
  ✓  7 signed in routes fit at 390px (14.3s)
  ✓  8 signed in routes fit at 414px (14.2s)
  ✓  9 signed-out card checkout returns to Wallet funding after login at 320px (2.3s)
  9 passed
```

Full gate, from `wallet/zuuli/`:

```
$ npm run typecheck                  # clean
$ npm run build                      # ✓ built in 9.98s
$ npm run test
   Test Files  34 passed (34)
        Tests  430 passed | 5 skipped (435)
   # tests 79   # pass 79   # fail 0     (node contract tests)
   98 passed (3.3m)                      (playwright)
$ npm run test:store-listing         # fail 0
$ npm run store:validate             # clean
```

### Highest consequence

| Site | Was | Now | Why |
| --- | --- | --- | --- |
| `auth/SeedReveal.tsx:66` | `truncate font-mono` on **each BIP-39 word** | `min-w-0 break-words` | It did not clip today — the margin was about half a character. A wider mono face, one more pixel of padding, or a non-English wordlist turns this into a silently mis-transcribed recovery phrase and an unrecoverable wallet. `wallet/Onboarding.tsx` already had no clip here; nothing to change. |
| `profile/LinkedAccounts.tsx:174` | `truncate` **wrapping** `truncateAddress(...)` | `break-all` | The exact double truncation `format.ts:220-222` forbids. It clipped the already-shortened string and clobbered the checksum tail with a second ellipsis, regressing #365. |
| `home/parts.tsx:37,42` | `truncate` on section `<h2>` + `<p>` | wrap | Every section header on Discover — the first screen anyone sees. Pure UI copy. |
| `wallet/components.tsx:71` | `truncate` on the sync **error** string | wrap | The actionable part of the error was the part being cut, behind a `title` no touch user can reach. |
| `home/CreatorsRow.tsx:37` | `truncate` on a **flex container** | `w-full min-w-0` + name in a `min-w-0 break-words` span | `text-overflow: ellipsis` does nothing on an anonymous flex item: it hard-clipped with no ellipsis. Broken on its own terms. |
| `live/Room.tsx:937,944` | meeting ID / room name **end**-truncated in CSS | `truncateAddress()` + `break-all` + `title` | Identifiers, but end-truncation drops the tail. Routed through the helper: middle, tail-weighted. |
| `search/index.tsx:377` | `max-w-[40%] shrink-0 truncate` on `article.category` | `shrink-0 whitespace-nowrap` | Category is UI copy from a small controlled vocabulary. It keeps its words; the title (user-authored, `line-clamp-2`) absorbs the width via `min-w-0 flex-1`. |
| `wallet/funding/ActivityTab.tsx:160` | one `truncate` slot holding `@handle` **or** UI copy | `min-w-0 break-words` | The `"Added 2Zs"` / `meta.label` branch is UI copy. |

### Same rule, lower blast radius

- **Display names and usernames** (`search/index.tsx:286,308`, `profile/index.tsx:237,247`, `live/StreamCard.tsx:108,109`, `live/Room.tsx:249,250`, `home/ArticlesGrid.tsx:49`, `home/LiveRail.tsx:62`, `creator/index.tsx:261`, `ArticleCard.tsx:109`, `SendTab.tsx:498,499,518,523,633,637`, `ai/*` display names): arguably user-authored, but they are **the primary label of their row** — clip them and the row loses its subject. Made the layout accommodate them (`min-w-0` + `break-words`) rather than clipping. `ArticleCard.tsx:109` and `LiveRail.tsx:62` were not on the brief; same rule, same fix.
- **AI picker triggers** (`ModelPicker.tsx:56`, `PersonalityPicker.tsx:55`): the fixed `h-10` plus the Button base's `whitespace-nowrap` made wrapping impossible, so removing the clip alone would have produced a *worse* silent overflow. `h-10 → min-h-10`, plus `whitespace-normal py-1.5 text-left`: the trigger grows instead of eating the model name. `min-tap` is preserved.
- **`wallet/components.tsx:176` vs `wallet/History.tsx:139`** — two tx-row components, two rules for the same `tx.memo`: one clipped, one did not. Reconciled toward History (no clip, wraps, `title` retained), because that is the variant the doctrine already sanctions.
- **`ai/index.tsx:372`** — "Anonymous, multi-provider — metered in 2Zs" is pure UI copy, and only visible at `sm:` and up where it has room. Clip removed.
- **`search/index.tsx:373`** — article title, not on either list. Treated as user-authored, matching `creator/index.tsx:1070`: `truncate → line-clamp-2` + `data-user-content`, so the card height stays bounded without an end-clipped title.
- **`components/layout/Sidebar.tsx:216`** — added the `max-w-full` guard its sibling at `:136` already had. At `grid-cols-5` each cell is 64px at 320px, so a longer localized "More" label clipped.

### Annotated, not changed

`line-clamp` on genuinely user-authored body content keeps its clamp and gains
`data-user-content`: `home/ArticlesGrid.tsx:39,43`, `creator/index.tsx:1070,1074`,
`ArticleCard.tsx:87`, `search/index.tsx:383`, `LiveRail.tsx:58`,
`StreamCard.tsx:94`, `CreatorsRow.tsx:47`, `PersonalityPicker.tsx:145`,
`PersonalityManager.tsx:226`.

### Left alone

Correct middle truncation, untouched: `wallet/components.tsx:128,181`,
`wallet/History.tsx:145`, `wallet/Send.tsx:308,354`, `auth/ZcashLoginFlow.tsx:269`,
`profile/LinkedAccounts.tsx:197,333`.

## Scope

No colours, gradients, glows, `shadow-glow`, fonts, type sizes, `src/index.css`,
`tailwind.config.cjs`, or brand assets were touched — including in
`home/parts.tsx`, `home/CreatorsRow.tsx`, `live/Room.tsx` and `creator/index.tsx`,
which the concurrent restyle also edits. The `CLAUDE.md` change **adds one
bullet** to Design rules without reflowing its neighbours. No landscape or
dialog-overflow coverage was added; that remains a separate follow-up.


skyl:
Reviewed adversarially on the owner's behalf. **Approving** — all 13 checks green, and the enforcement half holds up under scrutiny.

### I verified the "before" independently

A test that catches nothing is worse than no test, so I checked the pre-fix state against `main` myself rather than taking the PR's numbers:

```
truncate:          39
text-ellipsis:      1
line-clamp-*:      11
data-user-content:  0
```

That matches the reported 38 + 11 (my 40 counts two hits the scanner correctly blanks — they're inside the `format.ts:220` doc comment that warns *about* truncate, which is exactly what `stripComments` exists for). The assertions fail on real pre-existing defects. Not a tautology.

### The exemption inversion is the right fix

This is the part that mattered. `viewport.pw.ts` previously treated any `.truncate` / `text-overflow: ellipsis` / `line-clamp` as a **legitimate exemption** from the overflow audit — so the suite was actively whitelisting every violation it should have been catching. Now the exemption requires `data-user-content`, and an unannotated clip is reported rather than excused. The audit can no longer be silenced by adding the very class that constitutes the defect.

Asserting on **computed style rather than on whether the element overflows today** is the detail that makes this real. `SeedReveal` does not clip at current metrics — I measured ~72px of cell against a 67px worst-case English BIP-39 word — so an overflow-based check would sail past it. A clip one character away from firing on a recovery phrase is already the defect, and this catches it.

### Scanner robustness — checked the obvious bypasses

- `(?<![\w-])truncate(?![\w-])` correctly does **not** match `truncateAddress` / `truncateMiddle`. The lookahead kills it on the capital `A`/`M`.
- Comments are blanked with **offsets preserved**, so reported line numbers stay accurate and prose about the rule can't trip it.
- `enclosingOpeningTag` walks at brace depth zero, so a `>` inside `cn("x", n > 1 && "y")` doesn't terminate the tag early. That's the failure mode a naive version would have.
- **No allowlist.** Nothing is grandfathered, so a new violation fails on its first run rather than waiting for someone to remember to update a list.
- Good instinct pairing the static scanner with the runtime one: the scanner reaches components **no route renders**, which is the only reason `SeedReveal` is covered at all.

### The one soft spot, named honestly

`data-user-content` is a claim the tooling cannot semantically verify. Someone can annotate genuine UI copy and silence both checks. That does **not** make this theatre — the annotation is a deliberate, conspicuous line in a diff, and the alternative (inferring authorship from markup) isn't achievable. But it is the seam, and it's worth knowing that this rule is now enforced *up to a reviewer noticing a suspicious annotation*. Please push back on any future PR that adds `data-user-content` to something that isn't a bio, an excerpt, or a message body.

### Judgement calls I agree with

- **AI picker triggers** — spotting that removing the clip alone would be *worse* is the best catch in this PR. The Button base carries `whitespace-nowrap` with a fixed `h-10`, so unclipping without `min-h-10 whitespace-normal` would have traded a visible ellipsis for silent overflow. That's the kind of thing a mechanical find-and-replace gets wrong.
- **`components.tsx:176` vs `History.tsx:139`** — two tx-row components with two different memo rules was a genuine inconsistency; reconciling toward the unclipped variant is right.
- **`search/index.tsx:377`** — keeping `whitespace-nowrap shrink-0` on the category badge and letting the clamped title absorb width via `min-w-0 flex-1` is the correct shape. A long category now trips the horizontal-overflow audit loudly instead of being silently eaten.
- **Fixtures** — lengthening subtitles and bios while leaving the strings other suites assert verbatim (`Zooko`, `Why Shielded-by-Default Matters`) is careful work. That's the #382 trap closed without collateral breakage.

### Note for the concurrent restyle

`SeedReveal`'s margin was ~5px at 320px. A wider monospace face — which the foundation PR is introducing — eats that. Since the clip is now gone the word will wrap rather than silently clip, which is the safe failure, and the new assertion will catch a regression. Worth re-checking that screen once both have landed.

Merging.

PR #457, opened 2026-08-22
design(zuuli): restrained violet, bundled Plex, numerals as the signature

skyl:
## The direction

ZUULI's UI had drifted into the generic "AI slop" look — gradient headline text, stacked
violet/magenta radial glows, a sparkle badge, a glow shadow behind *every* primary button,
and four separate hash→gradient generators painting procedural card art across a full 360°
hue wheel at 70–75% saturation. This reforms the token layer, the type, and the motion
around **restrained violet**:

```
GROUND   true neutral graphite      0 0%  7%   (was 240 8% 8% — blue-tinted grey)
SURFACE  lifted one step, separated by a 1px hairline — never by glow
TEXT     two steps: 0 0% 96% primary, 0 0% 64% secondary
ACCENT   264 56% 66%  (was 271 91% 65%) — focus ring, primary action, links.
         Never decoration.
ZEC 42 90% 56% (#f4b728) and LIVE 350 89% 60% (#f43f5e) keep their meanings.
```

**The signature: numerals are the hero.** A wallet's characteristic material is amounts and
addresses, so money is now set the way a precision instrument sets a readout — IBM Plex Mono,
fixed-width figures, −0.03em tracking, whole part at display size with the fraction stepped
down. The house rule is one line: **digits are always `text-foreground`; the unit beside them
carries the currency colour.** That is spent hardest in `funding/BalanceHero.tsx` and
`wallet/Overview.tsx`; everything else stays quiet.

**Before / after, side by side:** https://claude.ai/code/artifact/cf0ebed7-1c5f-4567-bf55-ef7ce2249e0d
— five surfaces captured from both trees at the same viewports, fixtures, clock and locale.

## Why IBM Plex

`tailwind.config.cjs` declared `Inter`, but **Inter was never loaded** — no `@font-face`, no
`<link>`, no dependency. Every screen rendered in system UI, which is why the UI had to lean
on gradients for personality.

Plex Sans is engineered and institutional rather than neutral-default, and it is genuinely
characterful at display sizes. Plex Mono has excellent tabular figures and a `zero` feature
for the slashed zero — which is the whole signature, plus a real functional win on addresses
where `0` and `O` are transcribed by hand.

The fonts **must be bundled**: the Tauri CSP is `default-src 'self'` with no `font-src`
(`src-tauri/tauri.conf.json`), so a Google Fonts link would be blocked and fall back silently
to system UI — exactly the bug being fixed. `src/fonts.css` writes the `@font-face` rules by
hand against the latin woff2 files rather than importing `@fontsource`'s own stylesheets,
because those pull in cyrillic / greek / vietnamese / latin-ext faces that Vite would copy
into the app bundle even though `unicode-range` means they are never fetched.

One deviation from the brief: **`@fontsource-variable/ibm-plex-mono` does not exist** (404 on
npm — upstream Plex only ships a variable Sans). Mono uses the static
`@fontsource/ibm-plex-mono` at 400/500/600, latin. Total shipped type is **91 KB** of woff2:

```
dist/assets/ibm-plex-sans-latin-wght-normal-IvpUvPa2.woff2   45.71 kB
dist/assets/ibm-plex-mono-latin-400-normal-DMJ8VG8y.woff2    14.71 kB
dist/assets/ibm-plex-mono-latin-500-normal-DSY6xOcd.woff2    14.89 kB
dist/assets/ibm-plex-mono-latin-600-normal-BgSNZQsw.woff2    15.62 kB
```

## Removed

| Gone | Was |
|---|---|
| `.zuuli-gradient-text` | violet→pink→gold `bg-clip-text`, 3 call sites |
| `.zuuli-hero-glow` | stacked violet+magenta radials, 5 call sites |
| `.skeleton-shimmer` | sweeping white gradient on every loading state → `animate-skeleton-pulse` |
| `shadow-glow` token | **18 call sites**, including `ui/button.tsx` so every primary button glowed |
| 3 × decorative `blur-3xl` orbs | BrandPanel ×2 (stacked *on top of* hero-glow), BalanceHero ×1 |
| 4 × hash→gradient generators | → one `src/lib/cover.ts` on six curated near-neutral tones, chroma capped at 12% |
| Tailwind gradient utilities | `AiCta`, `BalanceHero` |
| The sparkle badge | `<Sparkles>` on the AI call-to-action |
| `accent` / `accent-foreground` | defined, zero uses — removed from both the CSS and the Tailwind theme |

Black scrims over media are **kept** — those are legibility, not decoration.

**One item from the brief I could not do.** The brief asked to de-gradient the in-app
`<Logo>`, but `scripts/brand-assets.mjs --check` — which the brief also put out of bounds —
asserts that `Logo.tsx` carries the same `from-primary` / `to-fuchsia-600` endpoints as
`assets/brand/logo.svg`, so the in-app mark can never drift from the installed icon. CI fails
outright without them. The gradient stays on the 36px mark and the glow behind it is gone;
flattening it needs a brand call plus a change to that guard, which is yours to make.

## Semantic tokens

`zec` / `tuzi` / `live` promoted from hex literals to `hsl(var(--…))`; new
`success` / `warning` / `info` absorb the ad-hoc raw palette (**emerald 40×, amber 49×,
sky 6×**, plus stray violet/orange/blue/red). `grep` for any
`(bg|text|border|ring|…)-(slate|…|rose)-\d{2,3}` in `src/` now returns **nothing**. Badges move to a hairline border plus a 10% tint, which raises the
worst chip contrast from 2.88:1 to 4.3–5.0:1. A new **`<Callout>`** replaces the warning note
that had been hand-rolled eight times.

Violet also stopped being ornament: static icon tiles and avatar placeholders that were
`bg-primary/15 text-primary` moved to the neutral surface. Selected/active states keep it —
those *are* interaction. The AI **provider marks** went neutral too: they had been a
seven-hue categorical palette off the raw ramp, and my first pass mapping them onto the
status tokens was itself wrong — a green mark beside a model name reads as "OK". The icon and
the written label already identify the provider; the one entry that keeps the accent is the
option we host ourselves.

## Type scale

12px is now the floor. All **38 sub-12px arbitrary sizes** (`text-[9px]`/`[10px]`/`[11px]`)
and the stray `[13px]`/`[15px]` are gone, and all **18 uppercase micro-labels** now go
through one `eyebrow` class instead of drifting between three sizes and three tracking
values. Each step in the scale gained considered leading and optical tracking. Two new
utilities carry the signature: `.numeral` (money) and `.mono-id` (addresses, txids, DIDs —
slashed zero on).

## Motion

There was **no `prefers-reduced-motion` block anywhere**; there is now, and it suppresses
animation, transitions and scroll behaviour. `slide-up` — which fires on 41 route roots —
drops from 10px/350ms to 4px/240ms, and the Discover stagger halves from 0–320ms to 0–160ms.
`CLAUDE.md`'s "Entrances via `animate-slide-up`" line is updated to match.

## Verification

Everything below was run on this branch, rebased onto `origin/main` (283328c — after #456,
whose `ui-copy-truncation` audit this branch also passes; the four conflicts with it were
resolved in favour of its `break-words`/`break-all` and this branch's tokens).

```
$ npm run typecheck
> tsc --noEmit
(clean)

$ npm run build
✓ built in 10.25s

$ npm run test
 Test Files  35 passed (35)
      Tests  470 passed | 5 skipped (475)
# node --test contract suites: all ok, including ui-copy-truncation
  98 passed (2.9m)          # playwright
```

Plus the rest of the `frontend` gate: `icons:check`, `test:store-listing`, `store:validate`,
`test:asc-testflight`, `verify-ci-cache-policy`, and `npm audit --audit-level=high` (clean —
the two new font packages add no advisories; the 5 pre-existing moderates are unchanged).

`tests/viewport.pw.ts` (20 routes × 320/360/390/414) and `tests/touch-targets.pw.ts` are
green. The bigger type did cost real vertical space once: `auth-density.pw.ts` failed at
320×568 with the login card 7px into vertical scroll. Fixed in the layout, not by shrinking
type back — the auth card's vertical padding tightened (`p-5 pb-4` → `px-5 pb-3 pt-4`) and the
wordmark's sub-line lost its extra leading.

Note for whoever runs this locally: `playwright.config.ts` hardcodes port 1432, so two
worktrees cannot run the suite at the same time.

## Left for follow-up

- **`RtkMeeting`** (`@cloudflare/realtimekit-react-ui`, `live/Stage.tsx`) renders its own
  themed video UI and does not follow these tokens.
- **Mermaid** hardcodes `theme: "dark"` in `common/mermaid.worker.ts`.
- **`common/markdown.css`** keeps the VSCode Dark+ syntax hues deliberately — a reading
  convention developers already know, and near-neutral enough not to fight graphite. Its code
  face now points at the bundled mono.
- **QR codes** keep their literal `#ffffff` / `#1a1206` / `#f4b728`. Contrast-critical;
  untouched.
- **Out of scope, untouched:** brand assets, `store/*`, `index.html`'s `theme-color`, the
  Android launcher background, and the installed app icon's violet→fuchsia gradient. No light
  theme was added — `:root` remains a dark fallback, not a second design.
- The **AI nav icon is still lucide `Sparkles`**. It is conventional iconography for "AI"
  rather than a badge, so it stayed; say the word and it goes.
- `src/main.tsx`'s crash fallback stays Tailwind-free; its inline hex was updated by hand.


skyl:
Reviewed adversarially on the owner's behalf. Every check green including the required `gate`. Merging via admin (I can't approve a PR opened under my own account; `gate` was confirmed green first, per the governance note in `AGENTS.md`).

### Verified rather than taken on trust

Removals, counted on the branch:

```
shadow-glow           0     zuuli-gradient-text   0
zuuli-hero-glow       0     blur-3xl              0
skeleton-shimmer      0     text-[9|10|11px]      0
```

Gradients that remain are exactly the ones that should: the black/background scrims over media (legibility, not decoration), the consolidated `cover.ts` treatment with chroma capped at 12%, and `Logo.tsx` — see below.

**The font is real, which is the whole point.** The bug being fixed here was a typeface that *looked* configured and never loaded, so "I added a font" needed proof rather than assertion. The build output settles it:

```
dist/assets/ibm-plex-sans-latin-wght-normal-*.woff2   45.71 kB
dist/assets/ibm-plex-mono-latin-400-normal-*.woff2    14.71 kB
dist/assets/ibm-plex-mono-latin-500-normal-*.woff2    14.89 kB
dist/assets/ibm-plex-mono-latin-600-normal-*.woff2    15.62 kB
                                              total   90.93 kB
```

Same-origin, so the `default-src 'self'` CSP is satisfied. Hand-writing the `@font-face` rules instead of importing `@fontsource`'s stylesheets is the right call and the reasoning is correct — those stylesheets pull cyrillic/greek/vietnamese/latin-ext faces that Vite would copy into the bundle even though `unicode-range` means they'd never be fetched.

### The `<Logo>` compromise is my fault, not yours

You were told to de-gradient the in-app mark *and* not to touch `scripts/brand-assets.mjs`. Those instructions contradict each other: `brand-assets.mjs:69-73` reads `Logo.tsx` and throws if it loses `from-primary` / `to-fuchsia-600`, precisely so the in-app mark can't drift from the installed icon. I verified that guard exists and does what you said it does.

Surfacing the conflict and stopping at the glow was the right resolution — silently breaking CI or quietly editing a fenced file would both have been worse. The tile stays gradient until there's a brand call, which is consistent with the owner's "in-app only" decision anyway.

### Judgement calls I want to endorse specifically

- **`auth-density.pw.ts` went red at 320×568 and you fixed the layout instead of shrinking the type back.** That was the explicit instruction and it's the one most likely to be quietly ignored under time pressure, because reverting a font size is a one-character fix and reflowing a login card isn't. Thank you for taking the harder path.
- **Reverting the provider marks to neutral.** Mapping them onto `success`/`info` and then noticing a green mark beside a model name reads as "this one is OK" is exactly the kind of second-order error a token sweep produces. Correct call.
- **`@fontsource-variable/ibm-plex-mono` not existing** — upstream Plex only ships a variable Sans. Falling back to static 400/500/600 rather than substituting a different mono keeps the pairing intact.
- **The `Sparkles` AI icon stays.** Agreed. Lucide's sparkle as a nav glyph is conventional iconography, not a slop badge; the badge was the thing that had to go.

### Follow-ups, none blocking

1. **Nothing asserts the font actually applies at runtime.** This is the same shape as the bug we just fixed — `tailwind.config` named Inter for months and CI never noticed it wasn't loading. Build output proves the bytes ship; it doesn't prove `body` resolves to Plex. A ~10-line check (`document.fonts.check()` plus a computed-style assertion) would close it permanently. Filing this as an issue.
2. The 1px grid textures at `live/Room.tsx:178` and `live/StreamCard.tsx:42` survived, still duplicated verbatim. At `opacity-.12` over video they read as texture rather than decoration, so I'm not treating it as a defect — but it's the last copy-pasted visual idiom left.
3. `playwright.config.ts` hardcoding port 1432 blocking concurrent worktree runs is a real friction point for parallel agents. Worth an issue of its own; it'll bite again.

Good work. The type fix is the part that actually changes how this reads — the gradients were the symptom, and a UI with no typographic voice is what made them feel necessary.

PR #471, opened 2026-08-23
fix(zuuli): validate branded social-link hosts

### LINKED ISSUE #378: security(profile): validate platform-labelled social links against their real hosts
## Summary

Creator profile fields display trusted platform branding/icons while accepting arbitrary HTTP(S) destinations. A creator can present a link as GitHub, X/Twitter, YouTube, or another supported identity platform while sending the reader to an unrelated phishing host; plain HTTP is also accepted.

The trusted icon and platform label materially increase the credibility of the attacker-selected destination.

## Evidence

- `wallet/zuuli/src/features/profile/bio.ts:71-72` validates only an `http`/`https` prefix.
- Platform URL builders at `bio.ts:82-118` preserve already-absolute URLs rather than enforcing the platform host.
- Platform labels are selected independently of the destination host at `bio.ts:133-173`.
- Raw values are stored around `bio.ts:312-324`.
- Creator pages render the platform icon with that raw destination at the creator index around lines 230-247.

## Required remediation

- Require HTTPS.
- For branded platform fields, parse the URL and require the exact expected registrable host (plus an explicit reviewed subdomain allowlist), rejecting userinfo, alternate ports, confusing Unicode/punycode, suffix matches, and redirector hosts.
- Canonicalize supported handles into URLs rather than preserving arbitrary absolute input.
- Treat arbitrary websites and federated/Mastodon links as untrusted generic links and display the actual hostname prominently.
- Revalidate existing stored links before rendering platform branding.

## Acceptance criteria

- A GitHub-labelled link can resolve only to the reviewed GitHub host set; equivalent tests exist for every branded platform.
- Tests reject `http:`, userinfo tricks, `github.com.attacker.tld`, Unicode lookalikes, encoded host confusion, and non-default ports.
- Generic links display the effective destination hostname before navigation.
- Existing invalid branded links lose the trusted icon/label until corrected.

skyl:
Closes #378

## Summary
- require HTTPS and exact reviewed host allowlists before resolving supported platform links
- keep trusted branding only for structurally unambiguous profile namespaces: YouTube /@handle, LinkedIn /in/handle, and Reddit /u or /user
- reject absolute input for ambiguous root-path platforms; canonicalize their syntactically valid plain handles as generic links that prominently show the effective x.com, github.com, instagram.com, facebook.com, t.me, or njump.me hostname
- reject ambiguous authorities before URL parsing, including userinfo, Unicode, encoded-host, suffix, redirector-subdomain, and non-default-port inputs
- fail closed when an existing stored platform URL is invalid
- render websites and Mastodon/federated destinations as generic links with the browser-canonical effective host visibly shown

No backend or storage migration is required: stored values are revalidated by the pure bio parser each time a creator profile renders.

## Verification
- npx vitest run src/lib/utils/bio.test.ts src/features/creator/SocialLinks.test.tsx (123 passed)
- npm test (full Vitest and 79 Node contract tests passed; 98 Playwright tests passed)
- npm run typecheck
- npm run build


skyl (commented):
Material trust-boundary bypass: `brandedUrl()` validates the authority, but then returns the attacker-selected pathname/query unchanged. For example, a stored value like `facebook: https://facebook.com/l.php?u=https%3A%2F%2Fattacker.example` passes this parser and renders with the Facebook icon/label; Facebook’s `l.php` is a live outbound redirect flow (the first response currently redirects into its warning flow for that external `u`). Thus the branded link can still lead off the reviewed host despite the exact-host checks, contrary to “can resolve only to the reviewed host set” and the requirement to canonicalize handles instead of preserving arbitrary absolute input. Please reconstruct only reviewed per-platform profile URL shapes from their identifier (or reject non-profile paths/query/fragment), and add redirect-path cases—at minimum Facebook `l.php`, plus equivalent redirect surfaces for the other branded hosts—to the fail-closed table.

skyl:
Addressed the redirect-surface review in 9ba7188. Branded absolute inputs now must match a reviewed per-platform profile path and handle grammar, contain no query, fragment, or encoded path, and are reconstructed onto the canonical host rather than preserved. Added on-host redirect/non-profile regressions for all nine branded platforms plus canonical reconstruction coverage. Focused tests passed (107), as did typecheck and production build. Please re-review this correction.

skyl (commented):
The reconstruction closes the listed redirect endpoints, but a material denylist bypass remains for branded platforms whose profile shape is an ambiguous root segment. `RESERVED_PROFILE_SEGMENTS` is not platform-complete: `twitter: https://x.com/about` passes `profileHandle()`/`validPlatformHandle()` and renders branded as `https://x.com/about`; that URL currently returns `301 Location: https://about.twitter.com/about`, an unreviewed subdomain outside the exact X host allowlist. This is the same outbound-resolution failure via an unlisted reserved route. Please make the reserved identifiers platform-specific and cover the platform route namespace rather than relying on the current short generic set (or use a mechanism that cannot confuse root service routes with profiles), with `x.com/about` as a regression. The new query/fragment/encoded-path rejection and the listed redirect-surface tests otherwise look sound; focused tests pass 107/107.

skyl:
Second review correction pushed in a36716c. The incomplete shared reserved-segment list is removed. Absolute branded values are now rejected for every ambiguous root-path platform (X, GitHub, Instagram, Facebook, Telegram, and Nostr); only plain handles are canonicalized there. Absolute input remains supported only for syntactically explicit profile namespaces: YouTube /@handle, LinkedIn /in/handle, and Reddit /u or /user. Added x.com/about and corresponding platform-route namespace regressions for all nine branded platforms. Focused tests passed (116), as did typecheck and production build. Please re-review.

skyl (commented):
The absolute `x.com/about` form is now rejected, but the prior finding is not fully resolved because the parallel plain-handle path recreates the same URL. This revision removed `RESERVED_PROFILE_SEGMENTS`, so `twitter: about` passes `validPlatformHandle()` and `handlePath()` and renders branded as `https://x.com/about`; that still 301s to the unreviewed `about.twitter.com` host. The same ambiguity applies to other root-path platform service names even though their absolute forms are now blocked. Please apply platform-specific reserved-identifier validation to plain handles before reconstruction (with `twitter: about` as the regression), while keeping the new `absoluteProfile: false` boundary. Focused tests pass 116/116, but they currently exercise only normal plain handles and reserved absolute URLs.

skyl:
Third review correction pushed in 63a01cd. Plain handles are now checked against platform-specific reserved root identifiers before canonical reconstruction, while absolute URLs remain disabled for ambiguous root-path platforms. The exact twitter: about case now fails closed, with a cross-platform matrix covering reserved X, GitHub, Instagram, Facebook, Telegram, and Nostr names. Focused tests passed (128), as did typecheck and production build. Please re-review.

skyl (commented):
The exact `twitter: about` case is fixed, but equivalent live off-host routes remain because the manually curated reserved lists are still incomplete. Two concrete values accepted by `validPlatformHandle()` today:

- `twitter: help` → branded `https://x.com/help` → currently `302 Location: https://help.twitter.com` (not in the reviewed exact host set). `twitter: rules` similarly redirects to `help.twitter.com/...`.
- `telegram: contact` → branded `https://t.me/contact` → currently `302 Location: //telegram.org/` (also outside the reviewed set).

These are the same fail-open class as `about`, so the trust boundary is not yet clean. Please add the missing live service identifiers and regressions, and consider sourcing/documenting a platform-specific route registry or otherwise making completeness mechanically auditable—the current generic-looking hand lists will regress whenever a route is omitted. The prior `about` value and case variants are now rejected, and the other parser hardening still looks sound. Focused tests pass 128/128.

skyl:
Mechanically fail-closed trust redesign pushed at 8adccc7 after rebasing onto d922d3b. Ambiguous root-path platforms now accept only syntactically valid plain handles, render them generic/unbranded, and visibly show the canonical hostname; their absolute inputs remain rejected. Only YouTube /@, LinkedIn /in/, and Reddit /u or /user retain branded trust. Route-like about, help, rules, and contact regressions verify generic host disclosure and absence of platform icons/labels. Verification passed: 123 focused tests, full npm test including 79 Node contracts and 98 Playwright tests, typecheck, and build. Please re-review.

PR #406, opened 2026-08-19
fix(zuuli): recover native checkout returns

skyl:
## Summary

- send the explicit `zuuli_mobile` return mode only from authenticated packaged iOS/Android checkout launches
- register the exact `cash.free2z.zuuli://checkout/return` path in Tauri and Android while keeping iOS at one deterministic scheme registration
- strictly parse and deduplicate cold-start and warm-start return URLs, navigate to canonical `/wallet/fund`, and retry a bounded authenticated claim after transient delivery failure
- refresh session/balance best-effort without discarding a truthful cancel, processing, or webhook-backed credit result
- treat a correlated `processing` status token as non-authoritative and show success only after the status endpoint confirms a webhook-authored ledger credit
- preserve the merged compact-auth, touch-target, loading-state, TopBar/KYC, and legacy `/buy` → `/wallet/fund` compatibility contracts

Progresses #388.

## Release truth

Native checkout return depends on an unshipped backend capability tracked
internally. This PR makes **no claim that live card checkout is release-ready**,
and no live charge or completed production checkout flow was attempted.

When native return is unavailable, the client retries exactly once in existing
web mode so checkout can still return to the HTTPS page. A 4xx, or any 5xx
other than the explicitly handled unavailable response, never downgrades the
mode. This fallback is covered by the client tests.

The private-use scheme is registered but **not uniquely owned**. The
client requires a bounded, authenticated claim before showing authoritative
credit; a verified Universal Links / App Links rollout remains necessary to
remove custom-scheme delivery-denial risk. Tracked in #461.

## Client contract

- Packaged iOS/Android launches request native return explicitly; browser flows
  continue to use the existing web return.
- The app accepts only the exact registered scheme, host, and path, and strictly
  parses and deduplicates cold-start and warm-start return URLs.
- Correlated processing state is non-authoritative; success appears only after
  the status API confirms the credit.
- Request deadlines, listener re-arming after transient IPC failure, and
  truthful balance messaging are covered by the implementation and tests.

## Exact verification at `f9bfb51` (post-rebase, measured — not carried over)

- `npm test`: **35 Vitest files / 470 passed + 5 skipped**, **77/77 Node contract tests**, **98 Playwright tests passed** (3.2 min)
- `npm run typecheck`, `npm run build`, `npm run release:verify`: passed (`0.1.0+10`, `zuuli-v0.1.0+10`)
[...]

PR #473, opened 2026-08-23
fix(android): restore deep-link marker in the generated manifest

skyl:
## Unblocks the in-flight ZUULI `0.1.0+11` release

The **Android / Play internal** job built the AAB successfully and then died on
`scripts/mobile-release.sh`'s source-integrity assertion:

```
             </intent-filter>
+            <!-- DEEP LINK PLUGIN. AUTO-GENERATED. DO NOT REMOVE. -->
         </activity>
##[error]Process completed with exit code 1
```

The assertion was right. This PR makes the tree match it; nothing about the
assertion, `release.json`, the version identity (`0.1.0+11`), or any credential
/ provenance control is touched.

## What the plugin actually generates, and how I determined it

Read verbatim from the pinned dependency sources, and then confirmed by a real
`tauri android build --ci --aab` on Linux CI (see the evidence section below):

- `tauri-plugin-deep-link-2.4.9/build.rs` renders **one `<intent-filter>` per
  `plugins.deep-link.mobile` entry**, `join`ed with `\n`, and hands the whole
  string to `tauri_plugin::mobile::update_android_manifest("DEEP LINK PLUGIN", "activity", ...)`.
- `tauri-utils-2.9.3/src/build.rs::insert_into_xml` **toggles** on the marker
  line: content *between* the two `<!-- DEEP LINK PLUGIN. AUTO-GENERATED. DO NOT REMOVE. -->`
  markers is discarded and re-emitted, content *outside* them is user content and
  is preserved. It then writes `marker + contents + marker` immediately before
  `</activity>`.

So the plugin's output is **one marker pair wrapping both filters** — not one
marker per filter.

## Root cause

cc56fc1 (#406) did the *config* half correctly: `tauri.conf.json` already
declares both mobile deep links, so **it is already the source of truth and is
unchanged by this PR.** What #406 also did was hand-write an `<intent-filter>`
into the generated manifest **after the closing marker**, where
`insert_into_xml` reads it as user content.

The consequence is worse than the one-line diff suggests. Because the
hand-written filter lives outside the managed block, a real build **keeps it and
adds the plugin's own copy inside the block**. Regenerating the committed file
verbatim yields a *duplicated* checkout filter:

| occurrences | committed (before) | what a build regenerates from it | after this PR |
|---|---|---|---|
| `android:scheme="cash.free2z.zuuli"` | 2 | **3** | 2 |
| `android:host="oauth"` / `android:path="/callback"` | 1 / 1 | 1 / 1 | 1 / 1 |
| `android:host="checkout"` / `android:path="/return"` | 1 / 1 | **2 / 2** | 1 / 1 |
| `android.intent.category.BROWSABLE` | 2 | **3** | 2 |

Those regenerated counts also violate the pinned Rust test
`generated_android_manifest_registers_exact_mobile_redirect` in
`src-tauri/src/oauth.rs`. The committed file was simply never the plugin's
output.

## The fix

Rewrite the manifest to be exactly what the plugin emits — both filters inside a
single marker pair, in the plugin's own formatting (`<intent-filter >`, the
ChromeOS ARC++ comment, the empty path-pattern/prefix/suffix lines that
`scripts/normalize-generated-android-manifest.mjs` blanks). No config change was
needed, which is the durable outcome: generated output now derives entirely from
`tauri.conf.json`.

Verified as a **fixed point**: applying the plugin's exact algorithm to the new
file reproduces the new file byte-for-byte, so repeat builds converge instead of
re-drifting.

## Evidence: before vs. after

**Before** — the protected release job on `main`, after the AAB had already built:

```
             </intent-filter>
+            <!-- DEEP LINK PLUGIN. AUTO-GENERATED. DO NOT REMOVE. -->
         </activity>
##[error]Process completed with exit code 1
```

**After** — this branch's `Android / unsigned AAB` job ran a real
`tauri android build --ci --aab --target aarch64 -- --locked` on `ubuntu-24.04`
and then the newly added assertion, with no diff output and a zero exit:

```
    Finished 1 AAB at:
        /home/runner/work/zuu/zuu/wallet/zuuli/src-tauri/gen/android/app/build/outputs/bundle/universalRelease/app-universal-release.aab

##[group]Run node scripts/normalize-generated-android-manifest.mjs
node scripts/normalize-generated-android-manifest.mjs
git diff --exit-code
...
##[endgroup]
##[group]Run mkdir release-artifacts        <- next step; nothing printed, exit 0
```

Job steps, all green:

```
  Build unsigned release bundle: success
  Verify generated Android project remains clean: success
```

`Rust / ZUULI backend` also passes on this branch, and that job runs
`cargo test`, which executes
`generated_android_manifest_registers_exact_mobile_redirect` against this exact
manifest.

## Deep links both survive, unaltered

Neither registration's scheme, host, or path changes, and neither is widened —
`cash.free2z.zuuli://oauth/callback` (load-bearing for mobile OAuth; `wallet/zuuli/CLAUDE.md`
forbids widening it) and `cash.free2z.zuuli://checkout/return` (#406's reason for
existing) are both present exactly once, each with `DEFAULT` + `BROWSABLE`,
`appLink: false`. The exact-count table above is the evidence; the pinned Rust
test asserts the same invariants and this file satisfies them, whereas the
regenerated old file would not.

## Closing the coverage gap

`scripts/mobile-release.sh` runs `build → normalize → git diff --exit-code`, but
that pair existed **only** inside the protected release script. The PR-level
Android packaging job built the AAB and never checked the generated tree, so the
drift #406 introduced stayed invisible from that PR until a live store upload.

This adds the same two lines to `.github/workflows/zuuli-packaging.yml`'s Android
job, right after the build — mirroring what the iOS job already does for the
Xcode project and what the desktop job already does with a bare
`git diff --exit-code`. Build outputs under `gen/android` are gitignored, so the
assertion sees only tracked generated sources.

`.github/workflows/zuuli-release.yml` is deliberately untouched (owned by a
concurrent fix).

### DIFF
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -164,6 +164,13 @@ jobs:
             # Play so the protected release can restore a complete cache.
             ./node_modules/.bin/tauri android build --ci --aab -- --locked
           fi
+      - name: Verify generated Android project remains clean
+        # scripts/mobile-release.sh runs exactly this pair after its build, so a
+        # stale gen/android tree only surfaced during a protected store upload.
+        # Repeating it here fails the pull request that introduces the drift.
+        run: |
+          node scripts/normalize-generated-android-manifest.mjs
+          git diff --exit-code
       - name: Collect package
         run: |
           mkdir release-artifacts
--- a/wallet/zuuli/src-tauri/gen/android/app/src/main/AndroidManifest.xml
+++ b/wallet/zuuli/src-tauri/gen/android/app/src/main/AndroidManifest.xml
@@ -44,16 +44,20 @@
 
 
             </intent-filter>
-            <!-- DEEP LINK PLUGIN. AUTO-GENERATED. DO NOT REMOVE. -->
-            <intent-filter>
+            <intent-filter >
                 <action android:name="android.intent.action.VIEW" />
+                <!-- ChromeOS ARC++ uses a different action for deep links -->
                 <action android:name="org.chromium.arc.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data android:scheme="cash.free2z.zuuli" />
                 <data android:host="checkout" />
                 <data android:path="/return" />
+
+
+
             </intent-filter>
+            <!-- DEEP LINK PLUGIN. AUTO-GENERATED. DO NOT REMOVE. -->
         </activity>
 
         <provider


skyl:
Reviewed adversarially on the owner's behalf. Merging via admin (own-account PR; `gate` and `Android / unsigned AAB` confirmed green first).

### You found a worse bug than the one you were sent to fix

I briefed this as "a missing marker". It wasn't. Reading `insert_into_xml` and establishing that it **toggles on the marker line** — discarding what's between the markers and preserving what's outside as user content — is what turned a cosmetic-looking diff into the real defect:

#406 hand-wrote its `<intent-filter>` **after** the closing marker, so the plugin treats it as user content, keeps it, *and* emits its own copy inside the block. A real build produces `android:host="checkout"` **twice** and `BROWSABLE` **three times**. We were one successful release away from shipping a manifest with a duplicated deep-link registration.

Proving that by porting both functions verbatim and applying them to the committed file — rather than inferring it — is the right method for a question where a guess re-breaks a live release.

### Verifying it as a fixed point is the part that makes this durable

Applying the algorithm to your output reproduces it byte-for-byte. That means repeat builds converge instead of re-drifting, which is the actual acceptance condition for a generated file under a `git diff --exit-code` gate. A fix that merely passed once would have re-broken on the next build.

### Independently confirmed

- **2 host entries, 2 markers** in the proposed manifest — one pair wrapping both filters, no duplication. Checked directly on your branch.
- `Rust / ZUULI backend` passes, and that job runs `generated_android_manifest_registers_exact_mobile_redirect` (`src-tauri/src/oauth.rs`) against this exact manifest. That's executed proof the OAuth registration is intact, not an assertion that it looks intact — which matters, because `CLAUDE.md` forbids widening the mobile scheme/host/path and OAuth is load-bearing.
- Diff is 2 files. No `release.json`, no version file, `zuuli-release.yml` untouched as instructed.

### The smoke coverage is the most valuable line in this PR

`build → normalize → git diff --exit-code` existed only inside the protected release script, which is precisely why #406 sailed through PR CI and detonated during a real store upload. Mirroring it into the Android job in `zuuli-packaging.yml` means the next PR that hand-edits a generated file fails on **its own PR**, where it's cheap, instead of during a release. Seven lines, and it closes the structural gap rather than the instance.

### On not weakening the assertion

You made the tree match the check rather than relaxing the check. That was the instruction, but it's worth naming: `git diff --exit-code` on the generated tree caught a genuine defect that would have shipped, and it's exactly the source-bound guarantee it exists to provide. It earned its keep today.

Also: the stray `0` and `currentPage` files you flagged are gone from the primary checkout — working tree is clean. Good call reporting rather than deleting another agent's files.

**Unblocks the Android half of the in-flight 0.1.0+11 release.**

PR #475, opened 2026-08-23
fix(wallet): keep generated Tauri schemas current

### LINKED ISSUE #355: wallet: committed gen/schemas drifts from a fresh build

skyl:
## Summary

- regenerate Zuuallet’s stale Linux Tauri schema and remove obsolete `send_transaction` permissions from every generated layer
- force app-schema and plugin-permission generation on every CI attempt even when Cargo artifacts are restored
- enforce parity among the plugin build registry, runtime handlers, command TOMLs, generated permission schema/reference, and every tracked Zuuallet target schema
- fail on tracked or untracked generated drift; preserve ZUULI’s intentional ignored app-schema policy

## Root cause

Tauri generates target-specific app schemas and does not prune removed autogenerated command TOMLs. The macOS/desktop outputs were refreshed in #441 while `linux-schema.json` retained an older command set; separately, obsolete `send_transaction.toml` kept removed permissions alive in the generated reference, permission schema, ACL manifest, and app schemas. Restored Cargo targets could also skip both build scripts and their filesystem side effects.

The corrected 34-command output is deterministic across Linux/macOS/desktop. Build scripts now watch per-attempt CI nonces. Parity checks require the exact 68 namespaced allow/deny IDs plus `zcash:default` in each tracked target schema, and byte identity across Linux, macOS, and the desktop alias prevents a stale non-Linux target from surviving Linux CI.

## Verification

- full pinned Rust 1.97.1 builds for both apps on macOS and native ARM64 Ubuntu 24.04
- warm-cache controls reproduced Zuuallet’s skipped-generation false negative and ZUULI’s missing-schema false positive; changing the nonce forced regeneration and corrected both
- real stale `send_transaction` output rejected before clean regeneration
- parity self-tests reject extra, missing, and duplicate commands
- synthetic untracked permission TOML rejected by parity and untracked-output guards
- stale macOS schema rejected for a missing/extra permission ID
- content-only macOS divergence rejected by the cross-target byte-identity invariant
- exact 34-command / 68-permission parity across build registry, runtime handlers, TOMLs, generated permission outputs, and all tracked target schemas
- `actionlint .github/workflows/zuuallet.yml .github/workflows/zuuli.yml`
- Rust formatting, generated JSON parsing, pinned-toolchain, and action-pin checks

Closes #355


skyl (commented):
Material freshness gap: both new assertions are conditional on cargo build actually rerunning the Tauri app build script, but each job restores a Rust target cache immediately before the build. Cargo can reuse a fresh build-script fingerprint without recreating side effects outside target/. In Zuuallet, the tracked Linux schema already exists from checkout, so a skipped generator makes git diff pass even if the committed file is stale. In ZUULI, the ignored schema is absent after checkout, so the same warm-cache skip makes test -s fail even though the source is healthy. Please make generation deterministic before asserting it, for example by removing the expected target schema and forcing the app build script to rerun (or using a dedicated generation command/targeted clean), then assert regeneration plus tracked/untracked policy. A negative control should demonstrate stale tracked content is repaired/detected even with a warm target cache, and the ignored-output check should stay green under that same cache condition. The schema diff itself is legitimate and minimal: Linux is byte-identical to macOS/desktop and the eight new allow/deny pairs match the plugin command/default permission sources.

skyl (commented):
Material freshness gap: the committed/generated permission inputs are already stale, and the new post-build diff cannot detect this class of command removal. `wallet/plugins/tauri-plugin-zcash/build.rs` and the actual `invoke_handler!` register 34 commands, but `permissions/autogenerated/commands/` contains 35; the extra `send_transaction.toml` has no registered Tauri command. Consequently all three tracked Zuuallet schemas—including the newly regenerated Linux file—still advertise `zcash:allow-send-transaction` and `zcash:deny-send-transaction`. A clean build remains byte-stable because the Tauri generator rewrites current commands but does not delete obsolete autogenerated permission files, then consumes that stale file again; `git diff --exit-code` therefore blesses it.

Please remove/regenerate the obsolete permission artifacts and every affected app schema, and make the assertion detect removals by comparing the plugin `COMMANDS`/registered handler set to autogenerated allow+deny identifiers (or generate from an empty autogenerated directory and assert the complete resulting tree). The current Linux additions do match the eight new commands, and actionlint/pins/toolchain/JSON checks pass, but the stated current-schema invariant is not yet true.

skyl:
Corollary for the ZUULI side of the policy: `test -s` + untracked + ignored proves that a file appeared in the clean job, but it does not prove semantic freshness. With the stale `send_transaction.toml` generator input above, a fresh ignored `linux-schema.json` is generated containing an obsolete permission and this assertion passes. The command-set parity check (or clean generation input) needs to guard both the committed Zuuallet output and ZUULI’s ignored output.

skyl (commented):
One material app-schema gap remains. The new parity script never reads `wallet/zuuallet/src-tauri/gen/schemas/*-schema.json`, and the workflow explicitly regenerates only the current Linux schema. Therefore a stale change already committed in `macOS-schema.json` or `desktop-schema.json` is invisible: the Linux build does not rewrite it, and `git diff --exit-code` compares the post-build tree to that stale commit, so it is clean. I confirmed the bypass with a disposable committed-state negative control by replacing one macOS `zcash:allow-*` const with an obsolete ID; `check-zcash-permissions.mjs`, JSON validation, the tracked schema diff, and a clean status all passed. This can reintroduce exactly the cross-target committed drift the issue is meant to prevent even though all three files happen to be byte-identical now. Please have the checker assert the expected 68 namespaced IDs in every tracked app schema (and preferably assert the currently claimed byte identity), or regenerate/verify each committed target on its native job. The nonce invalidation, removal of `send_transaction` across generated layers, command/handler/TOML/plugin-schema/reference parity, and permission tracked/untracked guards otherwise look sound.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** via admin (own-account PR; `gate` confirmed green first).

### Permission surface: strictly narrows. Verified independently.

This is the part that could have blocked, since the diff touches `send_transaction.toml`, `schema.json` and `acl-manifests.json` on a wallet that moves money. Checked rather than assumed:

```
send_transaction in build.rs COMMANDS:        0
commits ever adding it to build.rs:           0
granted by any capability / default.toml:     none
only live use: send.rs:1720  client.send_transaction(RawTransaction{...})  <- lightwalletd gRPC
```

So `send_transaction` was never a registered IPC command. The orphan TOML has sat unreferenced since #144 in February, and the two ACL identifiers it produced (`allow-`/`deny-send-transaction`) granted nothing because nothing referenced them. Deleting them narrows the compiled ACL by two dead entries and changes nothing reachable. `jq -S 'del(.zcash)'` being byte-identical before/after is the right way to prove no other plugin moved — comparing the compiled manifest rather than the schema text is what makes that conclusive.

The +92 lines in `linux-schema.json` deserve their own note: those 16 identifiers were **already** in `acl-manifests.json` and `macOS-schema.json`. `linux-schema.json` was simply stale, and `*-schema.json` is an authoring schema that grants nothing. Not new access.

### The guard is real, and it's on the required path

Two things I care about after today, both satisfied:

**It can fail.** Breaking it four ways — restoring the pre-fix tree, restoring the stale linux schema, re-serializing a schema to identical content with different bytes, removing a live command from `lib.rs` only — produces four distinct errors. The load-bearing assertion is `assertSameSet(commands, generatedCommands)`, which targets exactly the failure mode Tauri creates: it writes permission TOMLs and never prunes them.

**It actually blocks.** Branch protection requires one context, `gate`; the parity step sits in `changes`, which has a bare `on: pull_request:` with no path filter, so it runs on every PR. A guard that runs but can't block is the same as no guard.

**And #477 doesn't apply here** — every modified job is `ubuntu-latest`, and the diff uses `test`, `git diff --exit-code`, `git check-ignore -q` and node exit status throughout. No bare `[[ ]]` anywhere. Good instinct, given how much of this repo's release path is macOS.

### `build.rs` — correct diagnosis, real cost

The three `cargo:rerun-if-env-changed=TAURI_*_NONCE` lines change *when* regeneration runs, never *what* it emits — no timestamps or path/env-dependent content. The false negative they fix is subtle and correctly identified: with a restored Cargo target dir the build script is skipped, so its filesystem side effects never run and `git diff --exit-code` compares committed files against nothing. Worth knowing the cost: forcing a plugin build-script rerun invalidates the plugin crate and its dependents, so both apps recompile even on a warm cache. Right trade.

### Three followups — filing, not blocking

1. **The "fresh build reproduces committed schemas" assertion is advisory.** `zuuallet.yml` has no `gate` job and isn't a required context, so `Verify committed Tauri schemas are current` can go red without blocking. That step is the *only* thing that would catch drift from a **Tauri version bump** — the parity script compares identifier sets and cross-target byte identity, not schema shape. Given this repo tracks upstream HEAD deliberately, that's the likeliest future drift source.
2. **No documented way to regenerate `linux-schema.json` from macOS.** Dev is macOS; a macOS build refreshes macOS/desktop but not linux. The new byte-identity invariant then fails with no `--fix` path.
3. **`appSchemaNames` is hardcoded** (`check-zcash-permissions.mjs:157`) — a future `windows-schema.json` would be silently unchecked by the required job.

Nits, not worth a round trip: `zuuallet.yml` now pins two different `setup-node` revisions (`8207627…` at :238, `49933ea…` at :187); `permissionIds.sort()` mutates in place at :142 and the mutated array is reused at :172; the identifier regex at :123 won't match under CRLF.

### Honest limit

I can confirm the committed bytes are self-consistent and that the guard detects the real historical drift. I cannot confirm from here that a **fresh** build emits exactly those bytes — that needs the pinned Rust 1.97.1 build on both hosts. The PR claims it; CI's own regeneration step is what will keep it honest.

Good, well-evidenced work. Merging.

Issue #477, opened 2026-08-23
security(release): bare [[ ]] assertions never fail the macOS release jobs (bash 3.2 + set -e)


skyl:
## Problem

The macOS release jobs guard credential handling with bare `[[ ]]` assertions. **On the macOS runners those assertions cannot fail the job.**

GitHub's macOS runners execute `run:` steps with `/bin/bash`, which on macOS is **bash 3.2.57**. In that version `set -e` does not abort on a failing standalone `[[ ]]`:

```
$ /bin/bash --version | head -1
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ /bin/bash -c 'set -euo pipefail; [[ 1 -eq 2 ]]; echo "REACHED"'
REACHED          <- assertion failed, script continued
exit=0

$ /bin/bash -c 'set -euo pipefail; test 1 -eq 2; echo "REACHED"'
exit=1           <- aborts correctly
```

Independently reproduced on this machine. `test` and `false` abort as expected; bare `[[ ]]` falls through.

## How it surfaced

The `0.1.0+11` release failed in **iOS / credential-free unsigned archive**. The archive built correctly, but the step's guard —

```bash
archives=(src-tauri/gen/apple/build/arm64/*.xcarchive)
[[ ${#archives[@]} -eq 1 && -d "${archives[0]}" ]]
ditto -c -k ... "${archives[0]}" ...
```

— was pointed at a directory that does not exist. The glob did not expand, the `[[ ]]` evaluated false, and **execution continued anyway** into `ditto`, which failed with a confusing "Cannot get the real path" error instead of the guard's clear failure. That path bug is fixed in #476 (which uses `test`), but the inert-guard problem is general.

## Scope — this is the part that matters

Every bare `[[ ]]` in the macOS release jobs added by #443 (`security(release): isolate Apple signing credentials`) is decorative, including:

- `ios-sign`: provisioning-profile **UUID**, **name**, **team**, and **application-identifier** checks, and `profile_authorizes_identity`
- `ios-upload`: `expected_members` comparison and the non-symlink checks

These are credential-boundary controls. They were written to fail closed and they do not fail at all — a mismatched provisioning profile or an unexpected archive member would pass straight through. Roughly 40 assertions across the digest-sealed credential jobs.

[...]

Issue #474, opened 2026-08-23
release(zuuli): STATUS.md is 21 commits stale and promotion does not fail closed on it


skyl:
## Problem

`wallet/zuuli/docs/releasing.md:442` makes re-deriving `STATUS.md` **step 1 of promotion**, and says explicitly: *"do not start promotion with a stale commit/date."*

`wallet/zuuli/STATUS.md` currently reads:

> Last re-derived from `origin/main` at `1f752c743473f41c3b745160d3bb662cf5467e52` on 2026-08-19.

That is **21+ commits behind** `main`. The `0.1.0+11` release was promoted anyway, on the owner's explicit instruction, so this is a knowingly-accepted gap rather than an oversight — but it should not stay open.

## Why it matters

`STATUS.md` opens by defining itself as a release-readiness record where **production-observed** means the non-mock path was actually exercised against `https://free2z.cash` or read back from the named store. Its value is precisely that it does not drift: a stale row silently converts "was true three weeks ago" into "is true now", which is the one failure mode the document exists to prevent.

Everything merged since 08-19 is currently undescribed by it, including a complete visual restyle (#457), the truncation-enforcement work (#456), and native checkout returns (#406).

## Why this was not fixed mechanically

Re-deriving requires production and native evidence that cannot be manufactured from the repo — the whole point of the document. Bumping the commit and date without re-running the evidence would make it assert something no one verified, which is strictly worse than being visibly stale.

## What is needed

- Re-derive from current `main`, updating the evidence and disposition of every non-ready row per `releasing.md:442-447`.
- Pay particular attention to rows touching card checkout, since #406 landed and the client contract it targets depends on backend work tracked separately.
- Consider whether promotion should *fail closed* on a stale `STATUS.md` rather than relying on a human following step 1 — the release workflow already pins an immutable source SHA, so comparing it to the recorded re-derivation SHA is mechanically possible. That would turn a documented convention into an enforced one, consistent with how `git diff --exit-code`, the credential boundary, and the store contract are already handled here.

The last point is the durable fix. The current arrangement is a written rule with nothing enforcing it, and it has now been skipped once.

Related: #387 (store listing media), #396 (replace mock-mode readiness claims with verified production evidence).

skyl:
Claimed for implementation. I am adding a public-only, source-bound promotion guard and tests without changing STATUS.md or advancing its evidence date/SHA. The implementation PR will use Refs #474; the issue remains open until production/native evidence is genuinely re-derived.

[...]

PR #479, opened 2026-08-23
security(release): pin tag identity at publication

### LINKED ISSUE #376: security(release): revalidate release-tag identity immediately before publication
## Summary

The release pipeline validates the release tag/source identity during preparation, then hours of platform builds can pass before publication. The final index/publish step checks that the tag exists but does not prove it still resolves to the commit whose artifacts were built.

Because the repository currently has no tag ruleset, a tag can be deleted and recreated or otherwise moved during that window. Artifacts built from commit A could then be attached to a release whose tag resolves to commit B.

## Evidence

- `.github/workflows/zuuli-release.yml:155` performs the early tag/source check.
- The later release-index path around `zuuli-release.yml:794-826` trusts a `tag_exists` condition.
- The publishing script uses `gh release create --verify-tag`, which verifies existence, not equality with the expected source SHA.
- Live repository rulesets do not protect release tags.

## Required remediation

- Record the expected release commit as an immutable workflow value at preparation.
- Immediately before creating or updating the GitHub Release, fetch the tag from the canonical origin, peel annotated tags, and compare its commit byte-for-byte with the expected SHA.
- Repeat the comparison after release creation before publishing the release as non-draft.
- Bind artifact attestations and release-index metadata to that same expected commit.
- Add a tag ruleset that prevents update/deletion of release tag patterns except through a narrowly controlled process.

## Acceptance criteria

- Publication fails if the tag is missing, moved, retargeted, points through an unexpected object type, or resolves to any commit other than the prepared SHA.
- A test retargets a tag between prepare and publish and confirms no release becomes public.
- Release metadata records both tag object and peeled commit identity.
- Release tag update/deletion is repository-policy protected.

skyl:
## Summary

- refetch and verify the canonical annotated release tag immediately before and after GitHub Release creation, and before every asset mutation
- record both the immutable tag object and peeled prepared commit in release-index metadata
- reject artifact provenance whose source commit differs from the prepared workflow SHA
- add a retargeting regression that proves a mid-publication tag move leaves the release draft and blocks uploads

## Verification

- `npm run release:verify`
- `node --test scripts/release-tag-identity.node-test.mjs`
- `actionlint ../../.github/workflows/zuuli-release.yml`
- GitHub Actions pin checker
- full Vitest and Node contract phases
- Playwright: 100/100 on isolated port
- production build

The live release-tag ruleset will be installed after merge so repository policy and the verified workflow land together.

Refs #376

---

_Reviewer note: changed `Closes #376` to `Refs #376`. Two of #376's acceptance criteria — adding the tag ruleset, and repository-policy protection of tag update/deletion — are explicitly deferred to a manual post-merge step, and nothing in the repo verifies the ruleset exists. Auto-closing on merge would mark the issue done on unmet criteria. #376 stays open until the ruleset and its verification land._


skyl:

The mechanism is real, not theatre. I checked the two things most likely to be theatre by executing them.

### No inert assertions — and this one mattered

Given #477 (bare `[[ ]]` doesn't abort under `set -e` on macOS runners, bash 3.2.57), a 71-line assertion-heavy security script was exactly where theatre would hide. It isn't:

`verify-release-tag.sh` — **0 bare `[[ ]]`**, all 10 assertions in `[[ … ]] || { exit N; }` form, which enforces in every bash version. `publish-github-release.sh` likewise: `if`/`elif` conditions, `cmp -s || { exit 75; }`, and the four gate invocations are simple commands. The only caller, `release-index`, is `ubuntu-24.04`, so the trap wouldn't have applied anyway — but the scripts are written correctly regardless of runner, which is the right habit.

### The regression test is real — mutation-verified

Removing all `reverify_tag_identity` call sites makes test 3 fail; unmutated it passes. It isn't a mock that passes regardless — the fake `gh` performs an actual `git tag -f -a` plus force-push against a real bare-repo fixture during `release create`, and the script observes the move through a real `ls-remote`/`fetch`. That's a test with teeth.

Two things to know about it:
- **On macOS it never reaches the loop.** `publish-github-release.sh:50` uses GNU `tar --sort=name`, which BSD tar rejects, so a local `npm test` dies before the pre-upload gate and exercises a shorter path than CI. Not a defect (`release-index` is Linux-only), but don't trust a Mac run of this suite.
- **Only the *pair* of gates is pinned.** Deleting the post-create gate at `:43` alone is caught by neither the test nor the `release-identity.mjs` contract, which only requires the string to appear at least once. Cheap to tighten by counting occurrences.

### Threat model: closed in-run, open after — and that's why I changed the issue link

Traced the actual ordering. Two in-run windows remain: `L21 → L32` (detect-after; the job aborts leaving a draft with no assets) and `L64 → L65` (irreducible TOCTOU, milliseconds). Both acceptable, both fail closed.

The real gap is **after the job ends**. Nothing re-verifies, and the draft→public transition is entirely out-of-band — there's no code path that publishes a non-draft release, and `release-identity.mjs` now actively forbids the publisher from containing `gh release edit --draft=false`. So #376's criterion *"repeat the comparison before publishing the release as non-draft"* is satisfied only vacuously. Someone who moves the tag after CI goes green wins, and the script cannot detect it by construction.

**I changed `Closes #376` to `Refs #376`.** Two of that issue's acceptance criteria — adding the tag ruleset, and repository-policy protection of tag update/deletion — are deferred to a manual post-merge step, and `git grep -rn ruleset` returns **zero hits** anywhere in `.github/`, `scripts/`, or `docs/`. Nothing asserts the ruleset exists; nothing goes red if it never lands. Auto-closing would have marked #376 done on unmet criteria, which is precisely the #474 pattern this repo keeps generating. Filed **#484** to install it *and verify it from CI*, with the in-tree precedent to follow (`z/zakura-core/zakura/docs/release-tag-protection.md`, which keeps `Immutable release tags` as a separate ruleset so the release app can't bypass it).

To be clear about what this PR does achieve: it converts "tag identity is never checked at publication" into "checked at every mutation, and publication aborts on a move." That's a genuine improvement and it fails closed in all three in-run cases. It just isn't the whole of #376.

### Verified clean

- `package.json` +1/-1 is the `test` script gaining the new node test. **No version change**, `release.json` untouched at `0.1.0+12`.
- `release-identity.mjs` +19 only ever appends to `failures` — no identity computation, no `--require-newer-than`/`--require-main` change. `0.1.0+13` will bump exactly as 12 did. All six contract strings match byte-exactly and the negative regex is `false`.
- Tag format compatible: `zuuli-v0.1.0+12` satisfies the script's regex, and `release.schema.json` pins bare `X.Y.Z` so no prerelease suffix can be produced.

### Followups I'm not blocking on

`zuuli-release.yml:1435` asserts only that *one* provenance manifest matches, not one per artifact directory — tight today because all four platform artifacts carry one, but a future artifact without a manifest would be tarred and uploaded unchecked. The `--draft=false` negative regex is a one-line tripwire evaded by a backslash continuation or `gh api -f draft=false` — fine as a tripwire, not a control. And there's no `shellcheck` step anywhere, which is worth adding now that this much security logic lives in shell.

skyl (commented):
Two material gaps remain:

1. The `download-artifact` glob also matches this job’s own `zuuli-release-index-${identity}-${source_sha}` artifact from a prior workflow attempt. Actions artifacts with duplicate names can exist across reruns; before the current release-index upload, the prior index is therefore eligible for download. That artifact recursively contains the old `release-downloads`, and `publish-github-release.sh` packages every top-level directory, so the rerun can add a new `zuuli-release-index-…tar.gz` release asset instead of being idempotent (and repeat recursively). The source-binding loop does not stop this because it only requires `count > 0` anywhere; nested old provenance files satisfy it. Please download an exact allowlist of platform artifacts (or explicitly exclude the release-index artifact), require exactly one correctly bound provenance manifest for every directory that will be packaged, and add a prior-attempt/rerun regression.

2. Issue #376 requires release-tag update/deletion policy, but the live `GET /repos/free2z/zuu/rulesets` result is still empty and this PR uses `Closes #376` while saying policy will be installed after merge. Please either install and record the matching enforced ruleset before merge, or keep the issue open (`Refs #376`) until that acceptance criterion is actually satisfied. This matters for the later manual draft-to-public transition, which this workflow intentionally does not perform.

The tag verifier itself correctly re-reads canonical origin, requires an annotated tag, compares the fetched tag object plus peeled commit, detects the tested create-time retarget before upload, and writes deterministic metadata. Focused Node tests (3/3), `npm run release:verify`, and actionlint all pass at the reviewed content head.

Issue #481, opened 2026-08-23
test(zuuli): mobile audits only visit an empty creator, so the catalog and pagination UI are never checked


skyl:
## Problem

The two mobile audits both enumerate `/creator/skyl`:

- `wallet/zuuli/tests/viewport.pw.ts:8` — per-element overflow audit at 320/360/390/414, signed in and out
- `wallet/zuuli/tests/touch-targets.pw.ts:7` — 44px + accessible-name + overlap audit at 320/360

`skyl` is **not in `mockCreators`**, so `zpages` resolves to `0` and the route renders the EmptyState. That means the creator profile's actual content — the page grid, the catalog count row, and (after #472) the "Load more pages" control — **never exists in either audit**, at any width.

So the most content-dense screen in the app is nominally covered by both mobile suites while contributing almost nothing to either.

## Why it matters now

#472 adds pagination UI to exactly this surface. Its Load-more button happens to be safe by construction (`Button`'s base carries `min-tap`, every size is `h-11`/`h-12`), but that's luck rather than coverage — nothing in the required audits would have caught it if it weren't.

Partial mitigation today: `tests/creator-responsive.pw.ts` does hit `/creator/zooko` at 320 and measures `[data-creator-profile]`, so gross overflow is covered indirectly. But its `actionTargets` scan is scoped to `[data-creator-actions]`, so **tap-target sizing on the catalog controls is not checked anywhere**.

## Suggested fix

Add a populated creator to both ROUTES arrays — `/creator/zooko` is the natural choice, since `genArticles(63)` gives it 11 pages plus 2 featured, i.e. more than one page, which also exercises the paginated state rather than a single-page shortcut.

While there: consider whether `/creator/skyl` should stay. An EmptyState route is worth auditing too, but it shouldn't be the *only* creator route either suite sees.

## Acceptance

- A tap-target or overflow regression in the creator page grid, count row, or pagination control fails `tests/touch-targets.pw.ts` / `tests/viewport.pw.ts`.
- At least one audited creator route renders more than one page of catalog.

Found while reviewing #472. Related: #382 (fixture-shaped text hiding real overflow) — same class of "the suite runs but can't see the thing".

skyl:
Correcting the framing of this issue — the coverage gap is real, but my description of the cause was imprecise.

[...]

PR #482, opened 2026-08-23
fix(zuuli): paginate global search results

### LINKED ISSUE #252: ZUULI Search: preserve backend pagination and make every result reachable
## Problem

`/search` promises to find every creator and page, but both real search methods request only 24 results, discard backend `count`/`next`, and return flat arrays. The tab badges then report the loaded array length as if it were the total, and there is no pagination or load-more path.

## Acceptance criteria

- [ ] Creator and page search preserve authoritative `count` and pagination independently.
- [ ] Result badges show total matches, not only the first page length.
- [ ] Every page of results is reachable with deduped incremental loading.
- [ ] Query, selected tab, and loaded/scroll state remain stable while loading more and across back/forward.
- [ ] Partial failure in one corpus does not hide successful results from the other.

skyl:
Closes #252

## Summary
- preserve validated, independent backend pagination and authoritative counts for creator and page search
- incrementally load every result with ordered deduplication, retryable partial failures, and bounded per-query restoration
- keep the selected tab in the URL and restore loaded rows plus scroll position across back/forward navigation
- add API contract, state-machine, and Playwright coverage for terminal empty, nonterminal empty, overlapping pages, partial failures, and complete traversal

## Verification
- `npm run typecheck`
- `npx vitest run src/features/search/pagination.test.ts src/lib/api/search-pagination-contract.test.ts` (12 passed)
- `ZUULI_PW_PORT=1462 npx playwright test tests/search-pagination.pw.ts` (3 passed)
- `npm run build`
- `ZUULI_PW_PORT=1462 npm test` (616 Vitest assertions / 5 skipped, 79 Node tests, 103 Playwright tests)

### DIFF
--- a/wallet/zuuli/src/features/search/index.tsx
+++ b/wallet/zuuli/src/features/search/index.tsx
@@ -5,6 +5,7 @@ import {
   BookOpen,
   Clock,
   FileText,
+  Loader2,
   Radio,
   Search as SearchIcon,
   Users,
@@ -16,6 +17,7 @@ import {
   AvatarImage,
 } from "@/components/ui/avatar";
 import { Badge } from "@/components/ui/badge";
+import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Skeleton } from "@/components/ui/skeleton";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
@@ -23,15 +25,10 @@ import { EmptyState } from "@/components/common/EmptyState";
 import { PageHeader } from "@/components/common/PageHeader";
 import { SectionLoadError } from "@/components/common/SectionLoadError";
 import { RemoteMedia } from "@/components/common/RemoteMedia";
-import { useAsync } from "@/hooks/useAsync";
-import { discover } from "@/lib/api/free2z";
 import { formatTuzis, initials, timeAgo } from "@/lib/format";
-import {
-  currentResourceData,
-  type KeyedRemoteData,
-} from "@/lib/remote-data";
 import type { Article, SimpleCreator } from "@/lib/api/types";
 import { SEARCH_INPUT_LABEL, withSearchQuery } from "@/lib/search-route";
+import { useCreatorSearch, usePageSearch } from "./pagination";
 
 const DEBOUNCE_MS = 300;
 
@@ -50,32 +47,8 @@ export default function SearchFeature() {
   const debounced = useDebounced(query, DEBOUNCE_MS);
   const inputRef = useRef<HTMLInputElement>(null);
   const searchKey = debounced.trim();
-  const {
-    data: creatorData,
-    loading: creatorsLoading,
-    error: creatorsError,
-    reload: reloadCreators,
-  } = useAsync<KeyedRemoteData<string, SimpleCreator[]>>(
-    async () => ({
-      key: searchKey,
-      value: searchKey ? await discover.searchCreators(searchKey) : [],
-    }),
-    [searchKey],
-  );
-  const {
-    data: pageData,
-    loading: pagesLoading,
-    error: pagesError,
-    reload: reloadPages,
-  } = useAsync<KeyedRemoteData<string, Article[]>>(
-    async () => ({
-      key: searchKey,
-      value: searchKey ? await discover.searchPages(searchKey) : [],
-    }),
-    [searchKey],
-  );
-  const creators = currentResourceData(creatorData, searchKey);
-  const pages = currentResourceData(pageData, searchKey);
+  const creators = useCreatorSearch(searchKey);
+  const pages = usePageSearch(searchKey);
 
   // Focus the box on mount for a keyboard-first feel.
   useEffect(() => {
@@ -85,16 +58,25 @@ export default function SearchFeature() {
   // The route, not the debounce timer, owns visible empty-vs-results state.
   // Clearing `?q=` must clear the screen in the same render.
   const hasQuery = query.trim().length > 0;
-  const creatorCount = creators?.length ?? 0;
-  const pageCount = pages?.length ?? 0;
+  const creatorCount = creators.count ?? 0;
+  const pageCount = pages.count ?? 0;
 
-  const defaultTab = useMemo(
-    () =>
-      !creatorsLoading && !pagesLoading && creatorCount === 0 && pageCount > 0
-        ? "pages"
-        : "creators",
-    [creatorCount, creatorsLoading, pageCount, pagesLoading],
-  );
+  const selectedTab = useMemo(() => {
+    const requested = params.get("tab");
+    if (requested === "creators" || requested === "pages") return requested;
+    return creators.initialized &&
+      pages.initialized &&
+      creatorCount === 0 &&
+      pageCount > 0
+      ? "pages"
+      : "creators";
+  }, [
+    creatorCount,
+    creators.initialized,
+    pageCount,
+    pages.initialized,
+    params,
+  ]);
 
   return (
     <div className="animate-slide-up">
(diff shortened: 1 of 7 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Requesting changes** — one substantive issue, plus a test that likely only passes in dev mode.

First, credit: this **does not** repeat either defect that blocked #472, and I checked both specifically.

- `count` and `next` come from the search endpoint's own envelope (`parseSearchEnvelope`, `free2z.ts` ≈L470). No second metadata field, nothing optional mapping to `0`, and a malformed envelope throws rather than silently rendering an EmptyState. "Empty with zero requests" is unreachable.
- **`mock-data.ts` is not in the diff at all.** The long fixture strings, the 61-article corpus, and the verbatim-asserted titles are all intact. No invariant was made unfalsifiable by shortening a fixture.

It also genuinely walks a cursor rather than raising a cap, and the badges now read backend `count` instead of array length. #252's substance is addressed.

### The problem: fail-closed invariants over conditions the API can legitimately produce

`mergeSearchPage` (`src/features/search/pagination.ts` ≈L61-92) turns three arithmetic disagreements into thrown errors that discard the page it just built:

```
≈L64  response.count !== current.count            -> "count changed during the read"
≈L84  next === null && items.length !== count     -> "ended before every result was loaded"
≈L90  next !== null && items.length >= count      -> "continued past its authoritative count"
```

Against the real API these aren't paranoid guards — they're **reachable states**:

1. **`count` is recomputed per request.** One zpage published or unpublished between two Load-more clicks and the traversal hard-errors mid-flight. Nothing is wrong; the corpus simply changed.
2. **The search ordering has no unique tiebreaker**, so it is not a total order. Offset pagination over a non-total order duplicates *and* skips rows — and a skip makes distinct rows `< count` at the terminal page, tripping `≈L84`. The user loses the final page **and** gets an error banner, caused by our own arithmetic rather than any backend failure. Rank ties are common in the text-search path, so this is not a corner case.

   Worth knowing: this codebase already learned this lesson elsewhere and fixed it with a pk tiebreaker on a different listing endpoint, with a comment describing exactly this symptom. The search endpoint didn't get the same treatment. **That's a backend fix and I'll track it separately** — but it means the client cannot assume a stable total order today.

3. **It's unrecoverable, permanently, for that query.** `retry` is literally `loadMore` (≈L188-196): it refetches the same page and re-compares against a `current.count` frozen at the pre-drift value, so it throws identically forever. `cache.remember()` only runs on success, so the module-level LRU keeps handing back the poisoned snapshot for the rest of the session. There is no path to a clean refetch.

**Requested change:** treat `count` as advisory display data, not a fail-closed authority. Terminate on `next === null`, full stop. Keep hard throws for genuinely malformed payloads and for non-progress (`≈L87` is a real liveness guard). Log the count mismatches instead of throwing. And make `retry` actually recover — re-baseline `count` and re-run the failed page.

### The fixtures don't hide it the way #472's did — but they can't express it either

`mockSearchResultPage` (`free2z.ts` ≈L1907) is constructed so `count` is always `allItems.length` and never drifts, and the `overlap` scenario re-serves a row while shifting the window so nothing is ever *skipped* — dedup exactly compensates and the terminal tie-out always holds. So the mock can express duplicates but **not** a skip or a count change, which are precisely the two conditions the new throws guard.

That's the milder cousin of what blocked #472: you didn't edit a fixture to hide a hazard, but no local run or CI run can observe the failure this design introduces. Please extend the mock to express both.

### A test that probably only passes in dev — please verify before merging

`search-pagination.pw.ts` test 2 (`empty-twice` → one Retry → success) appears calibrated to a **React StrictMode artifact**. `playwright.config.ts` runs `npm run dev`, and `src/main.tsx:64` wraps the app in `StrictMode`, so effects double-invoke; the second invocation resets `loading` to `false` before calling `load`, defeating the `if (stateRef.current.loading) return` guard (`pagination.ts` ≈L142). Mount therefore issues two page-1 requests, the first discarded by `requestIdRef`. The `empty-twice` counter is tuned to exactly that: mount burns 2 failures, Retry is call 3 and succeeds. Against a production build mount burns 1, Retry is call 2 and still fails — the test inverts.

I'm flagging this as **needs confirming, not established** — it's derived from reading the config and the scenario state machine, not from an executed run. Please verify directly; if it holds, the fix is to make the scenario counter independent of mount behaviour rather than to accept a dev-only calibration.

Corollary: the `fail-once` scenario looks like **dead, unreachable code** — the discarded first StrictMode request consumes it, so no error ever surfaces. It ships inside `src/lib/api/free2z.ts`, not a test file. Worth removing.

### Verified clean

- **No truncation regression.** The diff adds no `truncate`/`text-ellipsis`. The two `line-clamp-2` in `PageResultRow` (`search/index.tsx:374,390`) are pre-existing and compliant; the new `data-search-page-result` goes on the parent, so their `data-user-content` annotations are untouched. This matters — it's build-enforced now.
- Semantic tokens only; no raw `purple-*`/`violet-*`/hex.
- **Mobile fit is genuinely covered, better than I expected.** `/search?q=zcash` is in the ROUTES array of *both* `viewport.pw.ts:7` and `touch-targets.pw.ts:7`, and that query yields ~57+ pages in mock, so the Load-more button and the "24 of 61" counter really do render inside both audits at 320px. `Button`'s base carries `min-tap` at `h-11`. Contrast with #481, where the creator equivalent is invisible to both suites.
- Loading/empty/error separation for page 2+ is good: later-page failure keeps loaded rows, swaps the banner, hides Load more. Both corpora fail independently.

### Smaller items

- **Absolute `next` URLs are handled but untested.** DRF emits absolute URLs in production; every case in `search-pagination-contract.test.ts` feeds a relative one. It happens to work (`new URL(value, "http://localhost")`), but it's the untested-path shape that bit the sibling PR. Also the endpoint check compares `pathname` only, so an absolute `next` pointing at a *different host* would be accepted — worth tightening.
- The `it.each` cursor block asserts the same alternation for all five cases, so a case can pass by throwing for the wrong reason. Tighten to per-case messages.
- **New failure surface for the wallet send flow**: `discover.searchCreators` is now a strict parser and is called from `wallet/funding/SendTab.tsx:161`. Recipient suggestions previously tolerated a missing `results` key or rows without `username`; they now throw. It degrades to `searchFailed` (caught at :171) so nothing crashes, but suggestions now vanish where they used to render.
- Completed queries never refresh: the module-level caches hold 20 queries for the app's lifetime, the mount effect skips fetching when `initialized`, and `retry` no-ops at `next === null`. Re-searching a term in the same session returns arbitrarily stale results.

Fix the fail-closed count handling and confirm the StrictMode question, and I'll merge.

skyl:
Addressed the blocking review at current integrated head. Count is now advisory: cursor termination is authoritative, valid deduplicated rows survive live count drift/tied-row skips, and mismatches are diagnostic warnings rather than unrecoverable failures. Retry clears the old count baseline, the StrictMode-calibrated mock was replaced by a deterministic nonterminal failure cleared explicitly before Retry, and dedicated mock/E2E coverage now exercises both count drift and a skipped tied row.

Also restored the permissive legacy one-page recipient suggestion call (so strict global-search parsing does not regress Send), removed the dead `fail-once` scenario, made cursor-negative assertions exact per case, and covered production DRF absolute cursors. After merging current main, creator/search conflict resolution is jointly green.

Verification: 618 Vitest pass + 5 skipped; 87 Node/policy pass; typecheck; production build; focused creator+search Vitest 20/20; combined creator/search Playwright 8/8 after deriving corpus size from the backend count rather than freezing the fixture at 61. CI has restarted.

skyl:
Addressed the adversarial pagination finding in 321030b: a fully duplicated nonterminal offset page now warns and retains the advanced cursor, allowing the following unique page to remain reachable. Added unit and browser regressions for duplicate page 2 followed by new rows on page 3. Focused Vitest 15/15 and search Playwright 4/4 are green.

PR #485, opened 2026-08-23
security(oauth): bind every callback to its session

### LINKED ISSUE #380: security(oauth): bind desktop/web social association to the initiating Knox session
## Summary

Mobile social OAuth binds callback completion to the initiating session, but the desktop and web paths do not enforce the same invariant. Completion operates with whichever Knox token is current when the callback finishes.

If account A starts linking, then the user logs out or switches to account B before completion, the social identity can be applied in the wrong authentication context (or exhibit unsafe anonymous behavior depending on backend handling).

## Evidence

- Desktop callback transport at `wallet/zuuli/src/services/social-oauth.ts:68-84` does not bind the flow to a Knox session digest.
- The web transport around `social-oauth.ts:230-310` likewise lacks session binding.
- Session assertion is effectively a no-op outside native mobile at `social-oauth.ts:369-391`.
- Completion uses the current token in `wallet/zuuli/src/services/free2z.ts:756-773`.
- The mobile path demonstrates the stronger intended design with session hashing/binding at `social-oauth.ts:120-154` and `:190-207`.

## Required remediation

- Bind every OAuth transaction to a non-reversible digest of the exact initiating Knox session and account.
- On callback, require the same authenticated session/account before exchanging or associating anything.
- Cancel outstanding social-link transactions on logout, token rotation, account switch, or session revocation.
- Prefer an explicit backend association transaction that atomically validates OAuth state, PKCE where applicable, provider identity, initiating user/session, and one-time consumption.
- Never infer the target account solely from whatever bearer happens to be current at completion.

## Acceptance criteria

- Start-as-A/finish-as-B, start-authenticated/finish-anonymous, logout/relogin, concurrent-flow, replay, and expired-flow tests all fail closed.
- Successful association is one-shot and can affect only the initiating account.
- Desktop, web, Android, and iOS enforce the same session-binding invariant.

skyl:
## Summary

- bind web, desktop, Android, and iOS OAuth captures to a one-way digest of the exact initiating Knox session
- abort in-flight provider flows immediately when logout, token rotation, or account switching changes that session
- revalidate immediately before backend exchange and pin association to the verified initiating token rather than mutable global state
- clear renderer auth before awaiting server logout while revoking with the captured token

## Verification

- focused OAuth/token Vitest: 26/26
- auth session contract: 4/4
- full suite: 609 Vitest passing / 5 skipped, 80 Node contract tests, 100 Playwright tests
- typecheck
- production build

Refs #380

---

_Reviewer note: changed `Closes #380` to `Refs #380`. The session binding is renderer-side only and is never sent to the backend, so #380's requirement for an explicit backend association transaction with atomic validation and one-time consumption is not addressed. #380 stays open for that._

### DIFF
--- a/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
+++ b/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
@@ -53,11 +53,13 @@ test("every social OAuth transport pins completion to its initiating session", (
 
   assert.match(transport, /sessionBinding: string/);
   assert.match(transport, /transport: OAuthCallbackTransport/);
-  assert.match(transport, /onTokenChange\(\(token\)/);
-  assert.match(transport, /export async function assertOAuthSession/);
+  assert.match(transport, /onTokenChange\(\(\) => controller\.abort\(\)\)/);
+  assert.match(transport, /export async function withOAuthSession/);
   assert.doesNotMatch(transport, /assertMobileOAuthSession/);
-  assert.match(completion, /const initiatingToken = await assertOAuthSession\(capture\)/);
-  assert.match(completion, /authToken: initiatingToken \?\? undefined/);
+  assert.match(completion, /return withOAuthSession<SocialAuthResult>\(capture/);
+  assert.match(completion, /authToken: lease\.initiatingToken \?\? undefined/);
+  assert.match(completion, /auth\.me\(lease\.initiatingToken \?\? undefined, lease\.signal\)/);
+  assert.match(completion, /signal: lease\.signal/);
 
   const logout = between(api, "  async logout():", "  /**\n   * Login with Zcash:");
   assert.match(
--- a/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
+++ b/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
@@ -39,6 +39,33 @@ test("uncommitted profile probes use a private explicit token header", () => {
   assert.match(requestBody, /headers\["Authorization"\] = `Token \$\{token\}`/);
 });
 
+test("every social OAuth transport pins completion to its initiating session", () => {
+  const api = readFileSync(new URL("../src/lib/api/free2z.ts", import.meta.url), "utf8");
+  const transport = readFileSync(
+    new URL("../src/lib/oauth/transport.ts", import.meta.url),
+    "utf8",
+  );
+  const completion = between(
+    api,
+    "  async completeSocialOAuth(",
+    "};\n\n// ─── Profile",
+  );
+
+  assert.match(transport, /sessionBinding: string/);
+  assert.match(transport, /transport: OAuthCallbackTransport/);
+  assert.match(transport, /onTokenChange\(\(token\)/);
+  assert.match(transport, /export async function assertOAuthSession/);
+  assert.doesNotMatch(transport, /assertMobileOAuthSession/);
+  assert.match(completion, /const initiatingToken = await assertOAuthSession\(capture\)/);
+  assert.match(completion, /authToken: initiatingToken \?\? undefined/);
+
+  const logout = between(api, "  async logout():", "  /**\n   * Login with Zcash:");
+  assert.match(
+    logout,
+    /const token = getToken\(\);[\s\S]*?setToken\(null\);[\s\S]*?authToken: token \?\? undefined/,
+  );
+});
+
 test("recovery phrases cannot enter browser persistence, URLs, logs, or toasts", () => {
   const flow = readFileSync(
     new URL("../src/features/auth/useZcashChallengeFlow.ts", import.meta.url),
--- a/wallet/zuuli/src-tauri/src/oauth.rs
+++ b/wallet/zuuli/src-tauri/src/oauth.rs
@@ -18,7 +18,10 @@
 use std::fs::{self, File, OpenOptions};
 use std::io::{BufReader, BufWriter, Write};
 use std::path::{Path, PathBuf};
-use std::sync::Mutex;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex,
+};
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use serde::{Deserialize, Serialize};
@@ -32,13 +35,26 @@ const MOBILE_HOST: &str = "oauth";
 const MOBILE_PATH: &str = "/callback";
 const OAUTH_TTL: Duration = Duration::from_secs(10 * 60);
 const LOOPBACK_TIMEOUT: Duration = Duration::from_secs(5 * 60);
+const LOOPBACK_CANCEL_POLL: Duration = Duration::from_millis(50);
 const MAX_CALLBACK_URL_BYTES: usize = 8 * 1024;
 const MAX_CODE_BYTES: usize = 4 * 1024;
 const MAX_STATE_BYTES: usize = 512;
 const PENDING_VERSION: u8 = 1;
 
+enum LoopbackFlow {
+    Ready {
+        server: Server,
+        nonce: String,
+    },
+    Waiting {
+        server: Arc<Server>,
+        nonce: String,
+        cancelled: Arc<AtomicBool>,
+    },
+}
+
 #[derive(Default)]
-pub struct OauthLoopbackState(Mutex<Option<(Server, String)>>);
+pub struct OauthLoopbackState(Arc<Mutex<Option<LoopbackFlow>>>);
 
 /// Serializes every read/transition/write of the crash-safe mobile record.
 #[derive(Default)]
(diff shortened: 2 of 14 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Requesting changes** — one real defect with a one-line fix, plus a test-coverage problem that's the reason it shipped unnoticed.

The core security property is genuinely solid, and I want to say that first because most of this PR is right.

### Cross-session and cross-intent redemption are blocked — verified by execution

All six scenarios run against unmodified PR-head code, 6/6 correct:

| Scenario | Result |
|---|---|
| associate started as A, redeemed with token B | rejects `/session changed/` |
| associate started as A, redeemed anonymously | rejects |
| associate started as A, redeemed as A | resolves, token pinned |
| login capture forged to `associate: true` | rejects |
| associate capture forged to `associate: false` | rejects |
| login capture completed while signed in | rejects |

Binding is `associate:${sha256hex(knoxToken)}` — full 64 hex chars, compared with `!==` on full-length strings, no `startsWith`/prefix/truncation anywhere. And it **fails closed**: `sessionBinding` is required on `OAuthCapture` (`transport.ts:26`), a missing/expired session makes `buildSessionBinding` **throw** rather than degrade (`protocol.ts:169-176`), and there's no `if (!capture.sessionBinding) skip` escape hatch. I looked for that specifically — backwards-compat fallbacks are how this class of fix usually becomes a no-op.

Every invariant in `CLAUDE.md` holds: desktop's ephemeral loopback + random path nonce, the exact mobile scheme/host/path, PKCE S256, mobile pending state bound to mode + Knox digest, and every native claim/resume/cancel still scoped to its random completion state.

### Blocking: the session watcher never fires

`transport.ts:405-436`:

```ts
try {
  const transport = await abortable(oauthCallbackTransport(), controller.signal);
  if (transport !== "web") {
    return transport === "mobile" ? runMobileDeepLink(...) : runDesktopLoopback(...);
  }
  return runWebPopup(...);
} finally {
  stopWatching();
}
```

In an async function, `try { return p } finally { … }` runs the `finally` **immediately** when the return is evaluated — it does not await `p` first. I confirmed the semantics directly:

```
bare `return p` in try/finally:   finally ran at  0 ms   <- before inner resolves
`return await p` in try/finally:  finally ran at 50 ms   <- after inner resolves
```

So `stopWatching()` unsubscribes the `onTokenChange` listener about one microtask after the provider window opens. `controller.abort()` can never fire during the actual OAuth round trip — the exact window this PR exists to cover. The watcher effectively only guards `oauthCallbackTransport()` and, on web, the `socialStart` fetch.

Executed against unmodified PR head, driving `captureOAuthCode` through the web popup and calling `setToken("account-b")` mid-flight:

```
OUTCOME: STILL PENDING — watcher never fired
```

Changing only the two bare returns to `return await`:

```
OUTCOME: rejected: Your account session changed while the provider was open. Start again.
```

**Fix:** `return await` at `transport.ts:408` and `:426`. Worth noting `recoverMobileOAuth` (`:471-484`) already gets this right with `const capture = await waitForMobileCallback(...)` — only `captureOAuthCode` is affected.

Severity: not account takeover, because the exchange-time gate in §1 still fails closed independently. But it means #380's *"cancel outstanding social-link transactions on logout, token rotation, account switch"* isn't actually delivered, and on desktop the loopback listener keeps running until its 10-minute timeout after a logout.

### The reason this shipped: the contract assertions can't fail

This is the part I'd most like fixed. A text-preserving behavioural kill — `assertOAuthSession` returning the current token unconditionally **and** the watcher's `controller.abort()` disabled, with every asserted string left intact — produces:

```
node --test scripts/auth-session-boundary.node-test.mjs  ->  # pass 4  # fail 0
npx vitest run                                            ->  607 passed | 5 skipped
```

**The entire suite passes with the protection removed.** Patterns like `/onTokenChange\(\(token\)/` match a listener with an empty body; `/sessionBinding: string/` matches any parameter annotation. They're string-presence checks, and one of them would break on a prettier reflow while none of them prove behaviour. (`assert.doesNotMatch(transport, /assertMobileOAuthSession/)` is a genuinely useful anti-regression tripwire — keep that one.)

The only behavioural coverage here is the pure function `assertSessionBinding`, which is real (neutering the comparison produces 2 failures). Nothing covers the **wiring**: that `captureOAuthCode` aborts, that `completeSocialOAuth` calls the gate, that the associate request carries the pinned token. That gap is exactly why the `return await` bug wasn't caught.

**Please add one vitest** that drives `captureOAuthCode` through the web popup with a stubbed `window.open`, calls `setToken` mid-flight, and asserts rejection. It must fail on current HEAD and pass with the fix — that's the standard here.

### Also worth changing before merge

`auth.me()` at `free2z.ts:864` is **unpinned** after a successful associate, re-reading global token state two lines after a comment saying mutable global state is never read a second time. Should be `auth.me(initiatingToken ?? undefined)`.

### I've changed `Closes #380` to `Refs #380`

The binding is entirely renderer-side — it's never sent to the backend. #380's fourth requirement, *"an explicit backend association transaction that atomically validates … initiating user/session, and one-time consumption"*, is not addressed. That's a legitimate scope reduction, not a defect: a client now refuses to redeem a mis-bound callback, but a modified client still can. It just shouldn't auto-close the issue on an unmet criterion — same reasoning as #479/#376.

### Follow-ups, not blocking

- **`oauth_loopback_cancel` is unscoped** (`oauth.rs:557-565` — it just `.take()`s the singleton), and this PR routes more abort paths into it. A stale flow's cancel landing after a newer `oauth_loopback_start` kills the newer flow. Fails closed (DoS, not misbinding) and the trigger is pre-existing, but it's the exact pattern `CLAUDE.md` warns about and is now more reachable. Also, aborting desktop isn't a true cancel — `oauth_loopback_wait` has already taken the server, so the socket stays bound until timeout.
- **`http.ts:41`** — listeners are invoked synchronously and unguarded; one throwing listener aborts the fan-out and propagates out of `setToken`, breaking `logout()`/login. Wrap in try/catch.
- **Doc drift** (pre-existing): `CLAUDE.md:178` links `docs/release/ZUULI-MOBILE-OAUTH.md` relative to `wallet/zuuli/`, but the file is at the repo root.

`http.ts` and `free2z.ts` blast radius is clean — purely additive listener plumbing, `request()`/`buildUrl`/`API_BASE`/`mediaUrl` untouched, no cross-host header attachment, and the `logout()` reordering is strictly better.

Fix the two `return await`s, add the wiring test, and I'll merge.

skyl:
Independent adversarial review at exact head `404ef65ce6f13e7df9386a35b84f32598c2c31c6`: I reproduced/confirmed the existing blocking finding. The bare `return run…()` expressions in `captureOAuthCode` do execute `finally` immediately, so the watcher is removed before the callback settles; `return await` plus a behavioral popup/token-rotation test is required.

I found two additional material lifecycle gaps that should stay in the blocking set:

1. **The session lease does not cover exchange/result publication.** Even after the `return await` fix, the watcher ends when capture resolves. [`assertOAuthSession`](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src/lib/oauth/transport.ts#L504-L525) snapshots `getToken()` once, then awaits SHA-256; there is no watcher/recheck across that digest or the backend request. A token change after that snapshot still permits the request. For association the POST is correctly pinned to A, but a logout can still race a mutation to A and the subsequent unpinned [`auth.me()`](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src/lib/api/free2z.ts#L838-L871) can read B/null after mutating A. For anonymous login, a B session arriving after the snapshot does not stop the anonymous exchange; [`SocialButtons`](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src/components/common/SocialButtons.tsx#L130-L154) has only a component-generation fence, so it can publish the returned provider token over B. Please keep a token-generation/session watcher through exchange and final publication (with checks after awaits), pass its abort signal into `request`, and pin `auth.me(initiatingToken)`.

2. **Native cancellation/cleanup is not effective in two paths.** Desktop `oauth_loopback_wait` [takes the server out of shared state](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src-tauri/src/oauth.rs#L479-L501) before blocking, while [`oauth_loopback_cancel`](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src-tauri/src/oauth.rs#L556-L563) can only take that now-empty slot. Thus a logout makes JS reject but leaves the native socket/thread alive for the 10-minute timeout, and retries can accumulate listeners. Separately, the recovered-mobile armed path [returns `null` on token-change abort](https://github.com/free2z/zuu/blob/404ef65ce6f13e7df9386a35b84f32598c2c31c6/wallet/zuuli/src/lib/oauth/transport.ts#L462-L484) without calling `oauth_mobile_cancel`, leaving the persisted pending record armed until callback/restart/TTL. Please make post-wait desktop cancel terminate the receiver and cancel the expected mobile state in the recovery abort/error path, with negative tests proving native state/listeners are gone.

The existing focused tests, full frontend tests/gate, typecheck, and build are green, but none exercise these timing/cleanup paths. Recommendation: **request changes / do not merge** at this head.

skyl:
Blocking review fixes are pushed and current with `main` at exact head `72e477f`:

- `captureOAuthCode` now deliberately `return await`s every transport and has a behavioral popup/token-rotation regression.
- A process-local generation lease detects A→B→A, spans binding, native claim, exchange, pinned profile read, and post-await result checks; its `AbortSignal` reaches both exchange requests and `auth.me`. Both UI publication sites synchronously fence the returned generation.
- Post-associate `auth.me` is pinned to the initiating token. Logout clears renderer-private user/balance state before best-effort revocation, while revocation remains pinned to the captured token.
- Token listener fanout isolates throwing subscribers.
- Desktop loopback state is flow-scoped. Cancelling after `oauth_loopback_wait` begins sets the matching flow's atomic cancellation flag and actively unblocks its native server; stale cancellation cannot affect a newer listener. Rust negative tests prove prompt termination and exact-flow scoping.
- Recovered armed mobile flows now cancel the exact persisted state on abort/error, including token changes before listener installation; a behavioral regression also proves a late-installed deep-link listener is immediately removed.

Local evidence on the fix commit plus focused post-update verification:

- 613 Vitest tests passed, 5 skipped; focused post-update OAuth/session suite 29 passed
- 83 Node contract tests passed
- `npm run typecheck` and `npm run build` passed
- Playwright: 99/100 passed during a run disrupted by Vite HMR from an in-progress source edit; the sole interrupted test passed on an immediate stable-tree rerun
- 14 Rust OAuth tests passed, including both new loopback cancellation controls
- `cargo clippy --lib -- -D warnings` and rustfmt passed

The PR intentionally remains `Refs #380`: renderer/native binding is now robust, but #380's backend-owned atomic association transaction and one-time-consumption acceptance criterion remains unimplemented. Requesting exact-head adversarial re-review; please do not merge before it and the new CI gate are green.

Issue #484, opened 2026-08-23
security(release): the release-tag ruleset is installed but nothing in CI verifies it


skyl:
## Problem

#479 adds real tag-identity verification at every publication mutation, and it fails closed within a run. But the **durable** guarantee it depends on — a repository ruleset preventing release tags from being updated or deleted — is deferred to a manual post-merge step, and **nothing in the repository verifies that step ever happened**.

`git grep -rn ruleset` across `.github/`, `wallet/zuuli/scripts/`, and `wallet/zuuli/docs/` returns **zero hits**. No workflow asserts the ruleset exists; no check goes red if it is never installed.

This is why #376 is deliberately left open rather than auto-closed by #479 (its `Closes` was changed to `Refs`). Two of its acceptance criteria are unmet:

- *"Add a tag ruleset that prevents update/deletion of release tag patterns except through a narrowly controlled process."*
- *"Release tag update/deletion is repository-policy protected."*

## The gap the ruleset is supposed to cover

#479 verifies tag identity before and after release creation and before every asset upload, so an in-run tag move aborts the job. Two windows remain, and only the second matters:

1. **Within the run** — the `L21 → L32` and `L64 → L65` gaps are detect-after and irreducible TOCTOU respectively. The job aborts, leaving a draft with no assets. Acceptable.
2. **After the run ends** — nothing ever re-verifies. The draft→public transition is entirely out-of-band (`grep -rn draft` finds no code path that publishes a non-draft release, and `release-identity.mjs` now actively *forbids* the publisher from containing `gh release edit --draft=false`). So an attacker who moves the tag after CI is green wins, and nothing detects it.

The script cannot close window 2 by construction. Only repository policy can. That makes the ruleset the load-bearing control, and it is currently unenforced by construction.

## Suggested work

1. **Install the ruleset.** There is in-tree precedent to follow: `z/zakura-core/zakura/docs/release-tag-protection.md` describes a two-ruleset setup, keeping `Immutable release tags` separate so the release app cannot bypass it.
2. **Assert it exists from CI** — e.g. a `gh api /repos/free2z/zuu/rulesets` check in `gate` or in the release workflow's `prepare` job, so a missing or weakened ruleset fails rather than passing silently. Without this, step 1 can be silently undone later and nothing notices.
[...]

PR #486, opened 2026-08-23
fix(release): keep release-index reruns idempotent

### LINKED ISSUE #376: security(release): revalidate release-tag identity immediately before publication

skyl:
Closes #376

Fix-forward for the independently reported rerun gap after #479 merged.

- download only the four exact platform artifact families, excluding prior release-index artifacts
- fail closed on any unexpected top-level package
- require exactly one top-level, source-bound provenance manifest per packaged platform
- add regressions for recursive prior indexes and duplicate/mismatched provenance
- preserve the workflow contract in release verification

The live repository tag ruleset is also installed and API-verified as ruleset `21224853`: active on `refs/tags/zuuli-v*`, no bypass actors, and update/deletion/non-fast-forward protection.

Verification so far:
- release tag/index Node tests: 6/6
- `npm run release:verify`
- `actionlint .github/workflows/zuuli-release.yml`
- full `npm test` running on isolated Playwright port 1478

### DIFF
--- a/.github/workflows/zuuli-release.yml
+++ b/.github/workflows/zuuli-release.yml
@@ -1420,19 +1420,14 @@ jobs:
           persist-credentials: false
       - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
         with:
-          pattern: zuuli-*-${{ needs.prepare.outputs.identity }}-${{ needs.prepare.outputs.source_sha }}
+          pattern: zuuli-{android,ios,linux,macos}-${{ needs.prepare.outputs.identity }}-${{ needs.prepare.outputs.source_sha }}
           path: release-downloads
+          merge-multiple: false
       - name: Verify release-index source binding
         env:
+          RELEASE_IDENTITY: ${{ needs.prepare.outputs.identity }}
           EXPECTED_SOURCE_SHA: ${{ needs.prepare.outputs.source_sha }}
-        run: |
-          set -euo pipefail
-          count=0
-          while IFS= read -r -d '' manifest; do
-            jq -e --arg sha "$EXPECTED_SOURCE_SHA" '.source.commit == $sha' "$manifest" >/dev/null
-            count=$((count + 1))
-          done < <(find release-downloads -type f -name provenance.json -print0)
-          test "$count" -gt 0
+        run: wallet/zuuli/scripts/verify-release-index.sh release-downloads "$RELEASE_IDENTITY" "$EXPECTED_SOURCE_SHA"
       - name: Publish idempotent draft release
         if: needs.prepare.outputs.dry_run == 'false' && needs.prepare.outputs.tag_exists == 'true'
         env:
--- a/wallet/zuuli/scripts/release-identity.mjs
+++ b/wallet/zuuli/scripts/release-identity.mjs
@@ -694,6 +694,8 @@ for (const publishContract of [
 for (const releaseTagContract of [
   'RELEASE_SOURCE_SHA: ${{ needs.prepare.outputs.source_sha }}',
   'Verify release-index source binding',
+  'pattern: zuuli-{android,ios,linux,macos}-${{ needs.prepare.outputs.identity }}-${{ needs.prepare.outputs.source_sha }}',
+  'scripts/verify-release-index.sh release-downloads "$RELEASE_IDENTITY" "$EXPECTED_SOURCE_SHA"',
   'scripts/publish-github-release.sh "$RELEASE_TAG" "$RELEASE_IDENTITY" "$RELEASE_SOURCE_SHA" release-downloads',
 ]) {
   if (!releaseWorkflow.includes(releaseTagContract))
--- a/wallet/zuuli/scripts/release-tag-identity.node-test.mjs
+++ b/wallet/zuuli/scripts/release-tag-identity.node-test.mjs
@@ -9,7 +9,9 @@ import test from "node:test";
 const scripts = dirname(fileURLToPath(import.meta.url));
 const verify = resolve(scripts, "verify-release-tag.sh");
 const publish = resolve(scripts, "publish-github-release.sh");
+const verifyIndex = resolve(scripts, "verify-release-index.sh");
 const tag = "zuuli-v9.8.7+654";
+const identity = "9.8.7+654";
 
 function git(cwd, ...args) {
   return execFileSync("git", args, { cwd, encoding: "utf8" }).trim();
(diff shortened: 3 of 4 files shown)


skyl:
Full local verification is green at `cd86c0a`:

- 609 Vitest assertions (604 passed, 5 skipped)
- 85 Node contract tests
- 100 Playwright tests on isolated port 1478
- `npm run release:verify`
- `actionlint .github/workflows/zuuli-release.yml`
- `npm run typecheck`
- production build

The release-index regression tests specifically cover exact platform acceptance, recursive prior-index rejection, and per-platform provenance cardinality/source binding.

skyl:

I asked three blocking questions of this PR and all three came back clean, verified by execution rather than reading.

### 1. Does "idempotent" quietly mean "skip verification on rerun"? — **No.**

This was the central risk: on a security-gated publication path, the cheap way to make a rerun succeed is to detect "already done" and return early, bypassing the gates #479 just added. There is **no** state check, no early exit, no "already published → return 0" anywhere in the new path:

- `Verify release-index source binding` (`:1426-1430`) has **no `if:`** — unconditional on every attempt.
- `Publish idempotent draft release` keeps #479's `if:` byte-for-byte.
- `publish-github-release.sh` is untouched, so `verify-release-tag.sh` and all four `reverify_tag_identity` calls still fire every attempt.

The word "idempotent" refers to the pre-existing step name and the checksum-compare-and-keep behaviour, not to new skip logic.

### 2. The 9 removed workflow lines — **strictly upgraded, nothing dropped.**

They came from #479, and both enforcing constructs survive: the per-manifest `jq -e` at `verify-release-index.sh:46`, and `test "$count" -gt 0` → `[[ … ]] || { exit 65; }` at `:56`. Three new controls are added on top (top-level name whitelist, top-level provenance requirement, exactly-one-provenance). Net stronger.

**And the bug being fixed is real and nasty.** The old `zuuli-*-<identity>-<sha>` pattern re-downloaded the previous attempt's `zuuli-release-index-…` artifact on a rerun; the inline `find` walked into it and *passed*, because every nested provenance SHA matches. `publish-github-release.sh` would then have attached a recursive `zuuli-release-index-….tar.gz` to the public release. Reproducing that with a fixture — old logic passes, new logic exits 65 before any release mutation — is exactly the right way to establish it.

The brace-pattern verification is the detail I'd have most likely hand-waved: confirming against the pinned `download-artifact` that minimatch brace-expands `zuuli-{android,ios,linux,macos}-…`, then enumerating every `upload-artifact` name in the workflow to prove no artifact family is silently dropped. That's the difference between "the pattern looks right" and "the pattern is right."

### 3. Inert assertions (#477) — **clean, and tested under real bash 3.2.**

Zero bare `[[ ]]` statements across all 56 lines; every assertion is `if`, `|| { exit N; }`, `case … exit`, or a simple command under `set -e`. The `jq -e` at `:46` sits in a `while` body with the loop in the current shell (`< <(…)`), so it does abort — confirmed empirically. The invoking job is `ubuntu-24.04` anyway, but the idioms are 3.2-safe regardless, which is the right habit.

### The real weakness: the tests barely test the new script

Mutation-testing each assertion in a scratch copy:

```
delete the jq -e source-SHA comparison (:46)  ->  6/6 tests STILL PASS
delete artifact_count -gt 0 guard      (:56)  ->  6/6 STILL PASS
delete top-level provenance requirement(:39)  ->  6/6 STILL PASS
delete identity regex                  (:14)  ->  6/6 STILL PASS
delete expected-commit regex           (:18)  ->  6/6 STILL PASS
remove the name whitelist              (:32)  ->  test 5 fails ✓
-eq 1 -> -ge 1                         (:49)  ->  test 6 fails ✓
```

**The step is called "Verify release-index source binding" and the source binding itself is untested.** Deleting it is invisible to the suite — and per the next point, invisible to `release:verify` too.

The root cause is precise: the mismatch case in `release-tag-identity.node-test.mjs:130-145` plants the wrong SHA in a *nested second* provenance, so what it actually observes is the duplicate-count rejection, not the SHA comparison — and it only asserts `notEqual(status, 0)` with no stderr match, so it cannot tell the two apart. The missing case is a single top-level provenance carrying the wrong SHA.

Compounding it: `release-identity.mjs` asserts on the *content* of `publish-github-release.sh` (`:702-709`) but **never reads `verify-release-index.sh`**. An empty `verify-release-index.sh` passes `npm run release:verify`.

I'm merging rather than blocking because the script itself is correct and fails closed on every case constructed against it — but the coverage gap is filed as a followup and I'd like it closed before the next release relies on this.

### Fail-closed matrix — one pre-existing gap

Ten scenarios all reject correctly (prior index artifact, wrong identity, wrong SHA, provenance mismatch, corrupt JSON, partially-written dir, empty root, symlinked dir, bad args). The one gap: **only 1 of 4 platforms present → accepted**, because the check is `artifact_count > 0`. Pre-existing and partly intentional (the job's `if:` tolerates skipped platform jobs), but "something exists" is being treated as sufficient. Worth passing the set of non-skipped platforms in explicitly.

### Noted, not blocking

- **Reruns still aren't fully idempotent.** `:1439-1444` uploads the index with no `overwrite:`, which defaults to `false`, so a rerun where a prior attempt uploaded the index fails at the final step. This PR fixes the *corrupting* half (it now fails before touching the release); the red-X half remains.
- `verify-release-index.sh:46` exits 1 silently on SHA mismatch — no stderr. Hard to triage during a release incident.
- A symlinked platform dir fails with the misleading "must contain exactly one provenance.json".
- **There is no `shellcheck` anywhere in CI.** Given #477, the inert-assertion class has no automated guard at all — only review. That's worth adding now that this much security logic lives in shell.

### Confirmed side claims

The tag ruleset is real: `Immutable ZUULI release tags`, `active`, `refs/tags/zuuli-v*`, rules `update`/`deletion`/`non_fast_forward`, **`bypass_actors: []`**. I've updated #484 accordingly — it's now narrowed to "installed but unverified by CI" rather than "not installed."

PR #487, opened 2026-08-23
fix(zuuli): sanitize public app fixtures

### LINKED ISSUE #483: privacy(zuuli): real personal email ships in the public app fixture and renders in mock mode
## Problem

`wallet/zuuli/src/lib/api/mock-data.ts:33` contains a real personal email address in the `mockUser` fixture:

```ts
export const mockUser: AuthUser = {
  id: 1,
  username: "skyl",
  email: "<the owner's real personal gmail address>",
  ...
```

This repository is **public**, and that fixture renders in the running UI on every `VITE_MOCK=1` session — so it appears in demos, screen recordings, screenshots pasted into issues, and any contributor's local run.

## Why this is inconsistent rather than merely untidy

The project already treats this exact data class as forbidden — in the *store capture* path:

`wallet/zuuli/scripts/store-screenshot-contract.mjs:265`
```js
if (/[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}/i.test(text))
  fail("PRIVATE_CAPTURE_DATA", "capture fixture contains an email-like identity");
```

and `:28` bans `mock|fixture|debug|localhost|playwright|seed phrase|private key|secret key` plus any email pattern from disclosed capture text. The store fixtures under `store/fixtures/` are correspondingly fictional (`identityClassification: "fictional-editorial"`).

So the standard exists and is enforced for screenshots, but the app's own fixture — the one contributors actually look at all day — doesn't meet it.

**Store screenshots are not affected.** `scripts/store-screenshot-capture.mjs` serves `store/fixtures/` through Playwright `route.fulfill` and never reads `mock-data.ts`. This is about the dev/demo surface only.

## Severity: low, but worth fixing

The address is already present in git commit metadata, so this is not a novel disclosure. What's new is that it is *rendered as UI* — which is how it ends up in a screenshot or a recording without anyone noticing.

## Suggested fix

- Replace with a clearly fictional address consistent with the store fixtures' `fictional-editorial` convention (e.g. an `@example.com` address).
- Consider whether the `mockUser` display name should follow — it currently reads as a real person's name.
- Optionally extend the existing email-pattern check to cover `src/lib/api/mock-data.ts` and `src/lib/wallet/mock.ts`, so the app fixtures are held to the same bar as capture fixtures. That is a small addition to an existing node test and would make the rule uniform instead of path-specific.

## Acceptance

- No real personal email or identity appears in any committed fixture.
- The app fixtures are covered by the same email/identity assertion that already guards capture fixtures.

skyl:
Closes #483

- replace the real dev/demo identity with an explicitly fictional creator and reserved example-domain email
- update the signed-in and empty-creator audit fixtures consistently
- add a fixture privacy contract that rejects non-example email domains across both app fixture sources

Verification:
- privacy Node contracts: 2/2
- full Vitest: 609 assertions (604 passed, 5 skipped)
- affected touch-target/viewport Playwright: 13/13
- TypeScript typecheck

### DIFF
--- a/wallet/zuuli/package.json
+++ b/wallet/zuuli/package.json
@@ -8,7 +8,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "test": "vitest run && node --test scripts/safe-area-contract.node-test.mjs scripts/auth-session-boundary.node-test.mjs scripts/mermaid-security.node-test.mjs scripts/send-review-boundary.node-test.mjs scripts/ui-copy-truncation.node-test.mjs scripts/apple-credential-boundary.node-test.mjs scripts/release-tag-identity.node-test.mjs && node scripts/apple-credential-boundary.mjs && playwright test",
+    "test": "vitest run && node --test scripts/safe-area-contract.node-test.mjs scripts/auth-session-boundary.node-test.mjs scripts/mermaid-security.node-test.mjs scripts/send-review-boundary.node-test.mjs scripts/ui-copy-truncation.node-test.mjs scripts/fixture-privacy.node-test.mjs scripts/apple-credential-boundary.node-test.mjs scripts/release-tag-identity.node-test.mjs && node scripts/apple-credential-boundary.mjs && playwright test",
     "test:mermaid-security": "node scripts/check-mermaid-security.mjs && node --test scripts/mermaid-security.node-test.mjs && playwright test tests/mermaid-security.pw.ts",
     "test:send-review": "node --test scripts/send-review-boundary.node-test.mjs && playwright test tests/send-review.pw.ts",
     "test:apple-credential-boundary": "node --test scripts/apple-credential-boundary.node-test.mjs && node scripts/apple-credential-boundary.mjs",
--- a/wallet/zuuli/scripts/fixture-privacy.node-test.mjs
+++ b/wallet/zuuli/scripts/fixture-privacy.node-test.mjs
@@ -1,0 +1,34 @@
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+import { dirname, resolve } from "node:path";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+const root = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const fixturePaths = ["src/lib/api/mock-data.ts", "src/lib/wallet/mock.ts"];
+const emailPattern = /[a-z0-9._%+-]+@([a-z0-9.-]+\.[a-z]{2,})/gi;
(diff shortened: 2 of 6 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving** — will merge once `gate` reports green.

This resolves #483 properly, and the guard is the part that makes it stick.

### The privacy fix is real

Replacing the identity is the easy half; `scripts/fixture-privacy.node-test.mjs` is what stops it coming back. Test 2 has genuine teeth — it scans both fixture sources for **any** email pattern and rejects every domain outside `example.com/net/org`. Reintroducing a real address fails the build, wired into `npm run test` so it runs in the required gate. That's the right shape: not "we fixed the string" but "the string can't come back."

Using RFC 2606 reserved domains rather than an invented one is also correct — those can never resolve to a real mailbox.

### One note on test 1

The fictional-identity contract asserts four exact source strings (`username: "demo-creator"`, etc.). It can fail, so it isn't vacuous — but it's brittle in a specific way: a formatter switching to single quotes, or a property reorder, breaks it without any privacy regression. It also pins *one particular* identity rather than asserting a property, so it won't catch a different real name introduced elsewhere in the file.

Test 2 is the load-bearing one. Worth considering whether test 1 earns its brittleness, or whether the same protection is better expressed as "no email outside reserved domains" (already covered) plus a check for the specific former values never reappearing. Not blocking.

Also worth knowing it currently covers only `mock-data.ts` and `mock.ts`, and only emails — a real personal *name* elsewhere in a fixture would pass. `store/fixtures/` is separately covered by the store contract's `PRIVATE_CAPTURE_DATA` check, so the gap is small.

### #481 is not fixed by this — please don't close it

`tests/viewport.pw.ts:8` and `tests/touch-targets.pw.ts:8` change `/creator/skyl` → `/creator/demo-creator`. That's the correct rename to follow the identity change, but it **renames the problem rather than fixing it**:

```
demo-creator in mockCreators: 0
```

`demo-creator` is `mockUser`, not a member of `mockCreators` (`nine`, `mining_maya`, `zooko`, `f2z`), and no fixture article is authored by it. So the route still resolves to zero pages and renders the EmptyState — meaning the creator page grid, the catalog count row, and (once #472 lands) the pagination control remain **invisible to both mobile audits at every width**.

#481 asks for a *populated* creator in those ROUTES arrays — `/creator/zooko`, which has 11 pages plus 2 featured. That's a one-line addition on top of this change and can be done here or separately; either way #481 stays open until a populated creator is audited.

### Verified clean

`TopBar.signed-in.test.tsx` updates its mocked user to match, and the `4,210` balance assertion is untouched — that one is locale-pinned by #453/#455, so it stays deterministic.

PR #489, opened 2026-08-23
fix(release): make Apple assertions fail closed

### LINKED ISSUE #477: security(release): bare [[ ]] assertions never fail the macOS release jobs (bash 3.2 + set -e)

skyl:
Closes #477

- make every standalone Apple release `[[ ]]` assertion explicitly abort with a diagnostic
- cover source digests, artifact inventories, symlink checks, provisioning-profile identity, signing identity, IPA/app cardinality, and signed DMG layout
- reject any future bare Bash assertion across iOS/macOS build, credential, or finalize jobs
- prove a deliberately false guarded assertion exits before credential work continues
- reseal the three reviewed credential-job execution digests

Verification so far:
- Apple credential boundary tests: 62/62
- Apple boundary verifier
- release identity verifier
- actionlint
- full wallet suite running on isolated Playwright port 1480

### DIFF
--- a/.github/workflows/zuuli-release.yml
+++ b/.github/workflows/zuuli-release.yml
@@ -465,14 +465,14 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           set -euo pipefail
-          [[ "$EXPECTED_ARCHIVE_SHA256" =~ ^[0-9a-f]{64}$ ]]
-          [[ "$EXPECTED_PAYLOAD_SHA256" =~ ^[0-9a-f]{64}$ ]]
+          [[ "$EXPECTED_ARCHIVE_SHA256" =~ ^[0-9a-f]{64}$ ]] || { echo "invalid expected iOS archive digest" >&2; exit 1; }
+          [[ "$EXPECTED_PAYLOAD_SHA256" =~ ^[0-9a-f]{64}$ ]] || { echo "invalid expected iOS payload digest" >&2; exit 1; }
           expected_members=$(printf '%s\n' \
             ./CHECKSUMS.sha256 ./ExportOptions.plist ./ZUULI.xcarchive.zip ./source-record.json | LC_ALL=C sort)
           actual_members=$(cd unsigned-ios && find . -mindepth 1 -print | LC_ALL=C sort)
-          [[ "$actual_members" == "$expected_members" ]]
+          [[ "$actual_members" == "$expected_members" ]] || { echo "unexpected unsigned iOS artifact inventory" >&2; exit 1; }
           for member in CHECKSUMS.sha256 ExportOptions.plist ZUULI.xcarchive.zip source-record.json; do
-            [[ -f "unsigned-ios/$member" && ! -L "unsigned-ios/$member" ]]
+            [[ -f "unsigned-ios/$member" && ! -L "unsigned-ios/$member" ]] || { echo "unsafe unsigned iOS member: $member" >&2; exit 1; }
           done
           test "$(shasum -a 256 unsigned-ios/CHECKSUMS.sha256 | awk '{print $1}')" = "$EXPECTED_PAYLOAD_SHA256"
           (cd unsigned-ios && shasum -a 256 -c CHECKSUMS.sha256)
@@ -486,7 +486,7 @@ jobs:
           mkdir archive-root
           ditto -x -k unsigned-ios/ZUULI.xcarchive.zip archive-root
           archives=(archive-root/*.xcarchive)
-          [[ ${#archives[@]} -eq 1 && -d "${archives[0]}" ]]
+          [[ ${#archives[@]} -eq 1 && -d "${archives[0]}" ]] || { echo "expected exactly one unpacked iOS archive" >&2; exit 1; }
           echo "ZUULI_UNSIGNED_ARCHIVE=${archives[0]}" >> "$GITHUB_ENV"
       - name: Materialize, export, and destroy ephemeral iOS signing credential
         id: sign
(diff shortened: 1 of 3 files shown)


skyl:
Full pre-update local suite passed: 609 Vitest assertions (604 passed, 5 skipped), 84 Node/policy checks, and 100 Playwright tests on isolated port 1480. After merging current main (#486) into the PR branch, the combined exact head also passes all 68 Apple/release focused Node contracts, release verification, and actionlint. The main merge is disjoint from #477; CI is validating the combined head.

skyl:

This is the fix for #477, and it holds up under the only test that matters here — running the idioms on the actual shell.

### Complete: 0 bare `[[ ]]` left in any macOS job

Base `main` had **exactly 29** bare `[[ ]]` statement lines in `zuuli-release.yml`, all in macOS jobs. PR head has **0**. (My "~40" in #477 counted every `[[` occurrence; 29 was the real statement-position population.) `ios-sign`'s provisioning-profile UUID / name / team / application-identifier checks and `profile_authorizes_identity`, and `ios-upload`'s `expected_members` and non-symlink checks, are all now guarded.

Swept the other workflows too: `zuuli-packaging.yml`'s macOS job already used the enforcing form; the one remaining bare `[[ ]]` anywhere is `zuuli-linux-image.yml:128`, which is `ubuntu-24.04` and therefore aborts correctly. Cosmetic only.

### Enforcement verified on `/bin/bash` 3.2.57, not by inspection

```
bare [[ ]]                              exit=0  reached=yes   <- the bug
bare [[ ]] with trailing ;              exit=0  reached=yes
bare [[ ]] inside for loop              exit=0  reached=yes
[[ ]] || { echo; exit 1; }              exit=1  reached=no    <- the fix
[[ =~ ]] || { echo; exit 1; }           exit=1  reached=no
[[ a && b ]] || { echo; exit 1; }       exit=1  reached=no
[[ $(cmd) -eq 1 ]] || { echo; exit 1; } exit=1  reached=no
```

All 29 conversions use the `|| { echo "…" >&2; exit 1; }` form, including the awkward ones — `[[ $(security find-identity … || true) -eq 1 ]]` and the in-loop member checks. **No `|| echo`-only, no `|| true`, no `if`-without-`exit`.** The `[[ ]] || cmd` lines left untouched are deliberate cleanup logic feeding a `cleanup_status` flag that a later `if … then … exit 1` converts — and `if`-guarded `[[ ]]` does enforce under 3.2.

### Digests correct, and genuinely enforced

All three `CREDENTIAL_JOB_SHA256` values recompute to match the head, `ROOT_AUTHORITY_SHA256` correctly unchanged. Worth noting the seal isn't itself a bare `[[ ]]` — it's verified by Node in two places: `apple-credential-boundary.mjs` under the `frontend` job (in `gate`'s `needs`), and again via `release-identity.mjs:563` on the macOS runner *before* any credential job starts. No irony available.

### The new guard can actually fail — mutation-proven

Removing the guard block → `not ok 2`. Rewriting a test idiom back to a bare `[[ 1 -eq 2 ]]` → `not ok 3`. End-to-end: inserting a bare `[[ -d … ]]` into `macos-finalize` makes `node scripts/apple-credential-boundary.mjs` exit 1 with the right message. That's the difference between a guard and a comment.

### No behaviour smuggled in

All 29 removed lines paired programmatically with their 29 added lines: **every added line is byte-identical to the removed line plus the `|| { … exit 1; }` suffix.** Zero non-conforming pairs. The only other change is the three digest constants.

### Filed as #494, not blocking

The guard's regex requires the line to end right after `]]`, so three equally-inert forms evade it: `[[ … ]];`, `[[ … ]] ;`, and line-continued `[[ … && \⏎ … ]]`. For `ios-sign`/`ios-upload`/`macos-sign` that's moot — the digest seal catches any edit (confirmed: a trailing-`;` mutation trips `credential execution program changed`). But **`ios-build`, `macos-build`, `ios-verify`, `ios-finalize`, `macos-finalize` are macOS and not sealed**, and a bare `[[ … ]];` lands there today with zero failures.

### Honest limit on #477's acceptance

#477 asked for this to be "demonstrated by deliberately breaking one and observing a red run." That is **not** satisfied here and can't be by PR CI — `zuuli-release.yml` only fires on a `release.json` push to `main` or `workflow_dispatch`, so these assertion paths never execute on a PR. The bash-3.2 proof and the static guard are strong; the paths themselves stay unexercised, and `ios-sign`/`ios-verify`/`ios-upload`/`ios-finalize`/`macos-*` have still never run. Leaving #477 open for that.

PR #491, opened 2026-08-23
test(release): lock index source binding

### LINKED ISSUE #488: test(release): verify-release-index.sh's source-binding check is untested and can be deleted silently
## Problem

#486 added `wallet/zuuli/scripts/verify-release-index.sh`, invoked by the release workflow step named **"Verify release-index source binding"**. The script is correct and fails closed on every case I constructed against it.

**But the source binding — the check the step is named after — is not covered by any test, and deleting it is invisible to the entire gate.**

Mutation-testing each assertion against the PR's own suite (`scripts/release-tag-identity.node-test.mjs`), in a scratch copy:

```
delete the jq -e source-SHA comparison (:46)   ->  6/6 tests STILL PASS
delete artifact_count -gt 0 guard      (:56)   ->  6/6 STILL PASS
delete top-level provenance requirement(:39)   ->  6/6 STILL PASS
delete identity regex                  (:14)   ->  6/6 STILL PASS
delete expected-commit regex           (:18)   ->  6/6 STILL PASS
delete artifact_root -d check          (:13)   ->  6/6 STILL PASS
delete per-entry -d check              (:38)   ->  6/6 STILL PASS
remove the name whitelist              (:32)   ->  test 5 fails ✓
-eq 1 -> -ge 1                         (:49)   ->  test 6 fails ✓
```

**2 of 8 checks are actually covered.**

## Root cause

`release-tag-identity.node-test.mjs:130-145` — the "mismatch" case plants the wrong SHA in a **nested second** provenance file. So what it actually observes is the *duplicate-count* rejection at `:49`, not the SHA comparison at `:46`. And it asserts only `notEqual(status, 0)` with no stderr match, so it cannot distinguish the two.

The missing case is simply: **one platform dir, one top-level `provenance.json`, wrong `source.commit`** → must exit non-zero.

## Compounding: the contract check doesn't read the script either

`scripts/release-identity.mjs` asserts on the *content* of `publish-github-release.sh` (`:702-709`) but **never reads `verify-release-index.sh`**. It only pins that the script is *invoked* (`:697-698`), not that it checks anything.

So an **empty** `verify-release-index.sh` passes both `npm run release:verify` and the full test suite.

## Why this matters

This script is what stands between a rerun and attaching a recursive `zuuli-release-index-….tar.gz` to a public release — the exact corruption #486 was written to prevent. A guard that can be deleted without a single red test is one refactor away from being gone.

Same family as #474, #477, and #484: a real control with nothing asserting it stays real.

## Suggested work

1. **Add the missing negative test** — single top-level provenance with a wrong `source.commit`, asserting non-zero exit.
2. **Assert stderr** in the existing mismatch case at `:143` so it can't be satisfied by the duplicate-count path.
3. **Add a `verify-release-index.sh` content contract** to `release-identity.mjs`, mirroring what already exists for `publish-github-release.sh`, so the script can't be hollowed out.
4. While there, give `:46` a stderr message — a bare `exit 1` with no output is hard to triage during a release incident.

## Acceptance

- Deleting any assertion in `verify-release-index.sh` fails at least one test.
- An empty `verify-release-index.sh` fails `npm run release:verify`.

Found while reviewing #486 (merged).

skyl:
Closes #488

- add distinct negative cases for missing roots, malformed identity/SHA, non-directory artifacts, missing top-level provenance, wrong source commit, duplicate provenance, and empty indexes
- assert the precise failure reason so one guard cannot accidentally satisfy another test
- give source-binding failures an incident-ready diagnostic
- make release verification read the verifier itself and pin every fail-closed contract, so an empty or hollowed script fails the gate

Verification:
- release tag/index tests: 10/10
- `npm run release:verify`
- diff check

### DIFF
--- a/wallet/zuuli/scripts/release-identity.mjs
+++ b/wallet/zuuli/scripts/release-identity.mjs
@@ -202,6 +202,7 @@ const ascTestFlight = read("scripts/asc-testflight.mjs");
 const gateWorkflow = read("../../.github/workflows/zuuli.yml");
 const releaseWorkflow = read("../../.github/workflows/zuuli-release.yml");
 const githubReleasePublisher = read("scripts/publish-github-release.sh");
+const releaseIndexVerifier = read("scripts/verify-release-index.sh");
 const testFlightRecoveryWorkflow = read(
   "../../.github/workflows/zuuli-testflight-recovery.yml",
 );
@@ -708,6 +709,23 @@ for (const releasePublishContract of [
 ]) {
   if (!githubReleasePublisher.includes(releasePublishContract))
     failures.push(`GitHub release publication identity contract is missing: ${releasePublishContract}`);
+}
+for (const releaseIndexContract of [
+  '[[ -d "$artifact_root" ]]',
+  '[[ "$identity" =~ ^(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\+(0|[1-9][0-9]*)$ ]]',
+  '[[ "$expected_commit" =~ ^[0-9a-f]{40}$ ]]',
+  '"zuuli-android-$identity-$expected_commit"',
+  '"zuuli-ios-$identity-$expected_commit"',
+  '"zuuli-linux-$identity-$expected_commit"',
+  '"zuuli-macos-$identity-$expected_commit"',
+  '[[ -d "$directory" ]]',
+  '[[ -f "$directory/provenance.json" ]]',
+  "'.source.commit == $sha'",
+  '[[ "$provenance_count" -eq 1 ]]',
+  '[[ "$artifact_count" -gt 0 ]]',
+]) {
+  if (!releaseIndexVerifier.includes(releaseIndexContract))
+    failures.push(`release-index verifier contract is missing: ${releaseIndexContract}`);
 }
 if (/gh release edit[^\n]*(?:--draft(?:=|\s+)false|--draft=false)/.test(githubReleasePublisher))
   failures.push("GitHub release publication must not make a draft public without a final tag identity gate");
--- a/wallet/zuuli/scripts/release-tag-identity.node-test.mjs
+++ b/wallet/zuuli/scripts/release-tag-identity.node-test.mjs
@@ -13,6 +13,10 @@ const verifyIndex = resolve(scripts, "verify-release-index.sh");
 const tag = "zuuli-v9.8.7+654";
 const identity = "9.8.7+654";
 
+function runIndex(root, releaseIdentity, sha) {
+  return spawnSync(verifyIndex, [root, releaseIdentity, sha], { encoding: "utf8" });
+}
+
 function git(cwd, ...args) {
   return execFileSync("git", args, { cwd, encoding: "utf8" }).trim();
 }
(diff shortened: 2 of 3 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving** — `gate` is green. Holding the merge briefly under a release freeze while build 13 (#492) ships; I'll merge straight after.

This is a test-quality PR, so the only review worth anything is empirical. Full mutation matrix, base (`ee43f09`) vs head (`7d0fcaf`), run under `/bin/bash` 3.2.57:

| Mutation to `verify-release-index.sh` | BEFORE | AFTER |
|---|---|---|
| **delete `jq -e` source-SHA comparison** | **passed (undetected)** | **FAILS** — "rejects a single provenance bound to another commit" |
| delete `artifact_count -gt 0` | passed | FAILS |
| delete top-level provenance requirement | passed | FAILS |
| delete identity regex | passed | FAILS |
| delete expected-commit regex | passed | FAILS |
| delete `-d "$artifact_root"` | passed | FAILS |
| delete per-entry `-d` | passed | FAILS |
| remove name whitelist | FAILS ✓ | FAILS |
| `-eq 1` → `-ge 1` | FAILS ✓ | FAILS |
| weaken predicate to `.source.commit != null` | passed | FAILS |
| drop `-e` from jq | passed | FAILS |
| hollow the script to `exit 0` | 2 tests | 6 tests + `release:verify` |

**2 of 8 covered → 8 of 8.** The critical row is red, and only that test goes red for it, so the signal is precise rather than incidental.

### The negative case is constructed correctly

`release-tag-identity.node-test.mjs:180-190` builds exactly what #488 asked for: one platform dir via plain `mkdir` (no nesting), one **top-level** `provenance.json` with a wrong SHA, asserting non-zero **and** `assert.match(stderr, /not bound to expected commit/)`. Splitting the old conflated test so each negative pins a distinguishing stderr string is the detail that makes this durable — no guard can now satisfy another guard's test.

### Content contract has teeth, with an honest caveat

`release-identity.mjs:205` now reads the script and `:713-729` pins 12 substrings, so an `exit 0` stub fails `release:verify`. #488's second criterion is met.

Worth recording what it is and isn't: substring presence means a hollow script that keeps all 12 strings **in comments** passes `release:verify` — it's the *tests* that kill that mutant. And conversely a behaviour-identical reformat (`[[ -d … ]]` → `[ -d … ]`) fails the contract with green tests. So neither layer is sufficient alone; together they are. That's the same trade-off already accepted for the `publish-github-release.sh` contract, so it's consistent rather than a new wart.

### Correction to my own issue

#488 said an empty script "passes both `npm run release:verify` and the full test suite." That was slightly overstated — an `exit 0` stub already failed 2 of the 6 base tests. The real hole was `release:verify`, and that is now closed. Noted so the record is accurate.

### Still uncovered — minor, worth a followup rather than a block

- The **`$# -ne 3` usage guard** (`:4-7`) is invisible to both tests and contract. It's a usage guard, not a security control, and the sole caller passes exactly three args — but it's the last uncovered assertion.
- **`jq -e`'s `-e` flag isn't pinned by the contract.** Pinning `'.source.commit == $sha'` doesn't pin the flag that makes it fail-closed. Dropping `-e` is caught by tests, so not a hole — but adding `'jq -e --arg sha'` would close it at both layers.

### Behaviour change is deliberate and inert

The bare `jq -e … >/dev/null` becoming `|| { echo "…not bound to expected commit…" >&2; exit 65; }` changes a silent `exit 1` into a diagnostic `exit 65`. Accept/reject verdicts are unchanged on identical fixtures, and no consumer reads the code — the workflow step only checks non-zero. This directly addresses the "bare exit 1 is hard to triage in a release incident" note from #488.

Small quibble: a **malformed** `provenance.json` (jq parse error) now reports "not bound to expected commit", which is slightly misleading. jq's own parse error still reaches stderr so triage isn't lost, and it still fails closed. Not worth a round trip.

`shellcheck -s bash` clean, no bare `[[ ]]` anywhere — #477 hygiene holds.

PR #495, opened 2026-08-23
ci(wallet): require fresh Zuuallet schemas in the protected gate

### LINKED ISSUE #480: ci(wallet): the Tauri schema freshness check is advisory — zuuallet.yml is not a required context
## Problem

#475 added a genuine parity guard for generated Tauri permissions, and it is correctly wired into the required `gate` path — the `changes` job has a bare `on: pull_request:` with no path filter, so `scripts/check-zcash-permissions.mjs` runs on every PR and can block a merge. That part is solid.

But the **other** assertion it added is advisory only.

`.github/workflows/zuuallet.yml:288` — `Verify committed Tauri schemas are current` — is the step that does a real regeneration and compares it against the committed files. `zuuallet.yml` has **no `gate` job and is not in `required_status_checks.contexts`** (which is exactly `["gate"]`). So that step can go red and a PR still merges.

## Why that specific gap matters

The two checks catch different things:

- `check-zcash-permissions.mjs` (required) compares **identifier sets** and **cross-target byte identity**. It catches an orphaned or missing command.
- The `zuuallet.yml` step (advisory) is the only thing that catches a change in **schema shape** — i.e. drift introduced by a **Tauri version bump** rather than by our own commands.

This repo deliberately tracks upstream HEAD and contributes fixes rather than pinning (see `AGENTS.md` / root `CLAUDE.md`). A Tauri bump changing generated schema shape is therefore not a hypothetical — it is the most likely future source of exactly the drift #355 was filed about. And it is the one case the required check cannot see.

## Suggested fix

Either:

- add `zuuallet.yml`'s job to `required_status_checks.contexts`, or
- wire an equivalent regeneration+diff assertion into the `gate` aggregation in `zuuli.yml`, so one required context covers both apps.

The second is more consistent with how `gate` already works here.

## Related smaller item from the same review

`scripts/check-zcash-permissions.mjs:157` hardcodes `appSchemaNames`. A future `windows-schema.json` would be silently unchecked by the **required** job — only the advisory step above would notice, which is the same gap from the other direction. Derive it instead:

```
git ls-files -- wallet/zuuallet/src-tauri/gen/schemas/*-schema.json
```

## Acceptance

- A drift that only a fresh regeneration can detect fails a **required** check.
- Adding a new target schema file does not silently escape the required job.

Found while reviewing #475 (closes #355). See also #474 and #477 for the neighbouring "guarantee that isn't actually enforced" pattern.

skyl:
Closes #480

## What changed
- add a cache-busted Zuuallet schema regeneration job to the required `gate` aggregation
- select it for backend, plugin, toolchain, submodule, build-image, and gate-policy changes, failing closed when the comparison base is unavailable
- reject tracked diffs and untracked generated permission/schema files after the fresh build
- discover every tracked `*-schema.json` dynamically instead of hardcoding target names
- extend the Linux image policy self-test to prove the new build stays attached to `gate` and its fail-closed contract

## Verification
- `node scripts/check-zcash-permissions.mjs --self-test`
- `node scripts/check-zuuli-linux-image.mjs --self-test` (27 negative cases + libxdo runtime cases)
- `node scripts/check-zuuli-linux-image.mjs`
- `git diff --check`

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -31,6 +31,7 @@ jobs:
     timeout-minutes: 5
     outputs:
       zuuli: ${{ steps.filter.outputs.zuuli }}
+      zuuallet_schema: ${{ steps.filter.outputs.zuuallet_schema }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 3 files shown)


skyl (commented):
Blocking finding on exact head `aa083eba95b1a7a7580134dadcd89948ebb8a22e`: the fresh-build job itself is correctly awaited by `gate`, but the mechanical protections against detaching or silently skipping it are not in the required path.

Branch protection has exactly one required context, `gate`. The new detachment negative control runs only through `.github/workflows/zuuli-linux-image.yml` (`Validate image policy`), while `.github/workflows/zuuli.yml`'s always-run `changes` job never invokes `check-zuuli-linux-image.mjs`. A later PR can therefore remove `zuuallet_schema` from `gate.needs` plus its result assertion; the PR-defined `gate` succeeds, and only the non-required image-policy context fails. That repeats the advisory-vs-required trust gap #480 is intended to close.

There is a second false-green in the same boundary: `check-zuuli-linux-image.mjs:382-424` validates the old `zuuli=true` selector, the new job/gate wiring, and two fail-closed fallbacks, but it never parses or validates the new `zuuallet_schema=true` selector's build-input patterns. Dropping (for example) `wallet/zuuallet/src-tauri/*` is not caught; a future Tauri/Cargo change then yields `zuuallet_schema=false`, the freshness job skips, and `gate` explicitly accepts that skip. The current self-tests have no selector-removal negative control.

Please run the image-policy self-test and validator from the always-run `changes` job (which `gate` already requires), and extend it to parse the schema selector and assert representative paths for every current input class, with a negative control that removes one. Include the policy checker itself in the guarded selector/policy surface. Then accidental detachment or selector erosion fails the protected context rather than only an advisory workflow.

The rest of this review is clean: current selector fallbacks fail closed, cache-busting reaches both build scripts, tracked/untracked drift rejection is complete, and dynamic `git ls-files` target-schema discovery plus exact-byte parity behaves as intended. Local checks passed: actionlint, permission self-test/current check, image-policy 27-case self-test/current check, toolchain/pin checks, and `git diff --check`.

skyl:
Addressed the independent blocking review at new head `9c2053a`:

- the image/schema policy self-test and current-tree verdict now run inside always-required `changes`, so branch protection sees failures through `gate`
- the schema selector now includes its own policy checker
- the validator parses the schema selector and requires representative Zuuallet source, plugin, toolchain, parity-policy, image-policy, submodule, image, workflow, and cache-action paths
- four new negative controls prove schema-job gate detachment, policy-check detachment, source/plugin selector removal, and policy selector removal fail the required contract

Local verification is green: actionlint; dynamic schema parity self-test/current; image/schema policy self-test (31 negative cases + libxdo runtime cases)/current; toolchain; action pins; diff check. Merged current `main` into the branch and CI has restarted.

PR #498, opened 2026-08-23
fix(release): catch multiline and semicolon Apple assertions

### LINKED ISSUE #494: security(release): the bare-[[ ]] guard misses trailing-semicolon and line-continued forms in five unsealed macOS jobs
## Problem

#489 converted all 29 bare `[[ ]]` assertions in the macOS release jobs to the enforcing `[[ … ]] || { echo …; exit 1; }` form, and added a guard in `wallet/zuuli/scripts/apple-credential-boundary.mjs:238-243` to stop them coming back:

```js
/^[\t ]*\[\[[^\n]*\]\](?:[\t ]*#[^\n]*)?[\t ]*$/m
```

The guard is genuinely live — removing it fails `not ok 2`, and inserting a bare `[[ -d … ]]` into `macos-finalize` makes `node scripts/apple-credential-boundary.mjs` exit 1. But the regex requires the line to **end** right after `]]`, so three forms that are equally inert under bash 3.2 slip past it.

## Measured

Mutations inserted into `macos-finalize` / `macos-build` (both macOS, both **not** digest-sealed):

| Form | bash 3.2.57 | Guard |
|---|---|---|
| `[[ … ]]` | inert | caught ✅ |
| `[[ … ]];` | inert | **missed** ❌ |
| `[[ … ]] ;` | inert | **missed** ❌ |
| `[[ … && \`<newline>`… ]]` (line continuation) | inert | **missed** ❌ |
| `[[ … ]] \|\| echo "bad" >&2` | continues, no exit | missed |
| `[ … ]` bare | aborts correctly | missed — correct, no issue |

Reminder of why "inert" matters (from #477):

```
$ /bin/bash -c 'set -euo pipefail; [[ 1 -eq 2 ]]; echo REACHED'   -> REACHED, exit 0
$ /bin/bash -c 'set -euo pipefail; test 1 -eq 2'                  -> exit 1
```

## Scope — which jobs are actually exposed

For `ios-sign`, `ios-upload`, and `macos-sign` this is **covered anyway**: they are digest-sealed by `CREDENTIAL_JOB_SHA256`, and any edit — including a trailing semicolon — trips `credential execution program changed`. Confirmed.

The exposure is the five macOS jobs that are **not** digest-sealed:

- `ios-build`
- `macos-build`
- `ios-verify`
- `ios-finalize`
- `macos-finalize`

A bare `[[ … ]];` added to any of those lands today with zero failures.

## Suggested fix

1. Allow an optional trailing `;` in the pattern.
2. Better: reject any statement-position `[[` whose logical line does not continue into `||`, `&&`, or `then`, which also covers the line-continuation form.
3. Consider flagging `[[ … ]] || <command that doesn't exit>` — `|| echo "bad" >&2` is a plausible bad conversion that logs and continues, and is arguably worse than a bare `[[ ]]` because it looks handled.

## Acceptance

- Each of the three inert forms above, inserted into any of the five non-sealed macOS jobs, fails `node scripts/apple-credential-boundary.mjs`.
- The existing 62 tests still pass.

## Also worth knowing (not this issue's scope)

#477's first acceptance criterion — *"demonstrated by deliberately breaking one and observing a red run"* — is still unmet, and can't be met by PR CI: `zuuli-release.yml` only fires on a `release.json` push to `main` or `workflow_dispatch`, so these assertion paths never execute on a PR. The static guard plus the bash-3.2 proof are strong, but the paths themselves remain unexercised — and `ios-sign`, `ios-verify`, `ios-upload`, `ios-finalize`, `macos-build`, `macos-sign`, `macos-finalize` have still never run at all.

Cosmetic: `zuuli-linux-image.yml:128` has a bare `[[ ]]`, but that job is `ubuntu-24.04` (bash 5) so it aborts correctly. Convert for consistency only.

Found while reviewing #489.

skyl:
Closes #494

## What changed
- normalize explicit Bash line continuations and scan complete double-bracket expressions across physical lines
- reject unguarded assertions at every reviewed statement boundary, including line/semicolon, single background ampersand, then/do/else, group/subshell/case boundaries, AND/OR-list positions, and Bash time/time -p
- preserve deliberate conditional/control-flow forms: if, while, negation, and assertions continued by AND/OR guards
- mutation-test thirteen inert forms across all five unsealed macOS-runner jobs, plus safe false-positive controls

## Verification
- Apple credential boundary: 132/132 Node tests
- 30 independent exact single-cause rejection mutations
- 40 safe-control acceptance mutations
- current release workflow boundary and release verifiers
- actionlint
- git diff --check

### DIFF
--- a/wallet/zuuli/scripts/apple-credential-boundary.mjs
+++ b/wallet/zuuli/scripts/apple-credential-boundary.mjs
@@ -101,7 +101,7 @@ function containsBareBashDoubleBracket(source) {
   // etc. do not begin at one of these boundaries and remain allowed.
   const shell = source.replace(/\\\r?\n[\t ]*/g, " ");
   const starts =
-    /(?:^|[;\n]|(?<!&)&(?!&)|&&|\|\||\bthen\b|\bdo\b|\belse\b|\btime\b|[({)])[\t ]*\[\[/gm;
+    /(?:^|[;\n]|(?<!&)&(?!&)|&&|\|\||\bthen\b|\bdo\b|\belse\b|\btime\b(?:[\t ]+-p)?|[({)])[\t ]*\[\[/gm;
   for (const match of shell.matchAll(starts)) {
     const open = match.index + match[0].lastIndexOf("[[");
     let quote = null;
--- a/wallet/zuuli/scripts/apple-credential-boundary.mjs
+++ b/wallet/zuuli/scripts/apple-credential-boundary.mjs
@@ -101,7 +101,7 @@ function containsBareBashDoubleBracket(source) {
   // etc. do not begin at one of these boundaries and remain allowed.
   const shell = source.replace(/\\\r?\n[\t ]*/g, " ");
   const starts =
-    /(?:^|[;\n]|&&|\|\||\bthen\b|\bdo\b|\belse\b|[({)])[\t ]*\[\[/gm;
+    /(?:^|[;\n]|(?<!&)&(?!&)|&&|\|\||\bthen\b|\bdo\b|\belse\b|\btime\b|[({)])[\t ]*\[\[/gm;
   for (const match of shell.matchAll(starts)) {
     const open = match.index + match[0].lastIndexOf("[[");
     let quote = null;
--- a/wallet/zuuli/scripts/apple-credential-boundary.mjs
+++ b/wallet/zuuli/scripts/apple-credential-boundary.mjs
@@ -91,6 +91,16 @@ function sourceForNode(workflow, node) {
   return node?.range ? workflow.slice(node.range[0], node.range[2]) : "";
 }
 
+function containsBareBashDoubleBracket(source) {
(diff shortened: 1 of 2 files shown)


skyl:
Addressed the review blocker in 19e892a. The guard now scans complete double-bracket expressions at shell command boundaries, including semicolon/newline, then/do, group, and AND/OR-list positions. The regression matrix is 117/117: ten inert forms across all five unsealed jobs plus safe if/while/negated/&&/|| controls. Current verifier and actionlint are green.

skyl:
Addressed the final adversarial `time -p` bypass at exact head `2fee4ce`: statement-boundary detection now recognizes Bash POSIX timing syntax, with the mutation exercised independently across all five unsealed macOS jobs. Local verification: Apple credential boundary suite 132/132, committed release verifier, and actionlint all pass. Exact-head CI and re-review are running.

PR #499, opened 2026-08-23
fix(zuuli): separate AI context switches

### LINKED ISSUE #266: ZUULI AI: make model/personality switching match the conversation context users see
## Problem

After messages exist, changing model or personality silently creates a new backend conversation because the `(model, personality)` key changed, but the old transcript remains rendered as one continuous thread. The new model receives none of the prior context that the screen implies it can see.

## Acceptance criteria

Choose and communicate one coherent model:

- [ ] Switching explicitly starts/confirms a new chat and clears/archives the old transcript; or
- [ ] The visible history is actually migrated/replayed to the new backend conversation.
- [ ] No screen presents messages as context the backend did not send.
- [ ] Cancel preserves the original conversation and selection.
- [ ] Conversation/model/personality identity survives reload if history is presented as durable.

skyl:
Closes #266

## What changed
- require explicit confirmation before model or personality changes discard a visible transcript
- keep model, personality, backend conversation, and transcript unchanged on cancel
- clear transcript, backend conversation identity, and session cost before a confirmed context switch
- reset the visible chat when the active personality is edited or deleted
- keep personality mutation dialogs non-dismissible while save/delete is in flight
- read current identity/transcript state for async mutation completion and generation-fence active sends so stale responses cannot repopulate a reset chat
- switch immediately when the chat is empty

## Verification
- Vitest: 619 passed, 5 skipped
- Node contracts: 90/90
- typecheck
- production build
- focused AI Playwright: 3/3
- independent save and delete race checks
- git diff --check

### DIFF
--- a/wallet/zuuli/src/features/ai/PersonalityManager.tsx
+++ b/wallet/zuuli/src/features/ai/PersonalityManager.tsx
@@ -53,12 +53,14 @@ export function PersonalityManager({
 }: PersonalityManagerProps) {
   const [view, setView] = useState<View>({ kind: "list" });
   const [deletingId, setDeletingId] = useState<string | null>(null);
+  const [saving, setSaving] = useState(false);
   // Armed by a first click on the trash icon; a second click on "Confirm"
   // actually deletes. Avoids a blocking `window.confirm()`, which freezes
   // the page (and is untheme-able) — an inline step keeps it in-app.
   const [confirmingId, setConfirmingId] = useState<string | null>(null);
 
   function close(next: boolean) {
+    if (!next && (saving || deletingId !== null)) return;
     if (!next) {
       setView({ kind: "list" });
       setConfirmingId(null);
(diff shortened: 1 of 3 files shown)


skyl:
Adversarial review found a real async personality CRUD race: closing the manager during an awaited save lets the stale completion switch personality after a new message is sent, recreating the dishonest-context bug. Holding merge; I am adding current-state/generation fencing and a save-close-send regression before rerunning review.

skyl:
Addressed the stale async personality CRUD review blocker at exact head `dba221f`: save/delete now keep the manager non-dismissible, identity decisions read current refs, and send completion is abort/generation-fenced so an obsolete response cannot repopulate a reset chat. Added a Playwright regression for attempted dismissal during an in-flight save followed by a correctly personality-bound message. Local verification: typecheck; focused PW 3/3; Vitest 619 pass / 5 skip; Node 90/90; production build. Exact-head CI and re-review are running.

Issue #248, opened 2026-08-07
ZUULI Articles: replace the unbounded tag wall with intersecting facets


skyl:
## Problem

The `/articles` tag browser is both visually unbounded and logically unsafe. `Feed.tsx` permanently unions every tag seen across loaded and previously filtered pages, renders the full set as wrapping chips, and keeps choices that have no overlap with the current selection.

In the live UI, selecting `AI` grew the historical union to roughly 146 tags; selecting the still-visible `travel` chip then produced `AI + travel` and zero results. Filters also disappear after visiting an article because they are not encoded in the URL.

This is a follow-up to #179: the mechanics landed, but the resulting interaction is not usable at production corpus size.

## Desired behavior

- Use a compact, searchable tag facet/popover rather than an always-expanded wall.
- Keep selected tags visible and easy to remove.
- After each selection, offer only tags that co-occur with the current result set; hide or disable zero-count choices.
- Prefer backend-provided facet counts over counts inferred from the currently loaded page.
- Persist search, sort, selected tags, and return position in the URL/history.

## Acceptance criteria

- [ ] Tag browsing is collapsed/compact by default on desktop and mobile.
- [ ] Every enabled next tag leaves at least one matching article.
- [ ] Selected tags remain visible even when the chooser is collapsed.
- [ ] Removing a tag restores the applicable choices and counts.
- [ ] Facets represent the full filtered corpus, not only already-loaded cards.
- [ ] Back/forward and shareable URLs preserve filter state.
- [ ] Keyboard, screen-reader, and touch interaction are covered.

skyl:
Re-confirmed while testing 0.1.0 (6) — still reproduces exactly as described, and it's the single most page-dominating problem in Articles right now (~14 rows of 5 tags consuming the whole screen before you see any articles).

**More usefully: this may not actually be `needs-backend`. The facet endpoint already exists.**

`GET /api/tagging/autocomplete` accepts `query`, `username`, `type` (`zpage|comment`), and **`selected_tags`**, and returns `[{ name, count }]` — a prefix-searchable, co-occurrence-narrowed tag list with counts, which is essentially the whole ask of this issue.

The classic React web app already consumes it:

- `ts/react/free2z/src/components/TagFilterMultiSelect.tsx:41-53` (params incl. `selected_tags`), `:84-88` renders `{option.name} ({option.count})`
- `ts/react/free2z/src/components/ChipFilterSelect.tsx:44-58`, counts at `:101-107`
[...]

PR #500, opened 2026-08-23
feat(zuuli): round-trip article tags

### LINKED ISSUE #250: ZUULI Articles: make tags round-trip through authoring, reading, and filtered navigation
## Problem

The article model and feed support `tags`, but `/articles/new` cannot set them, `articles.publish()` omits them, and `/articles/:slug` does not display them. The only tag interaction is the feed-level tag cloud.

Authors therefore cannot classify new content and readers cannot understand or pivot from an article's topics.

## Acceptance criteria

- [ ] The composer supports validated tag entry/autocomplete using existing tags.
- [ ] Tags are normalized, deduplicated, bounded, and included in the publish contract.
- [ ] Article detail renders its tags.
- [ ] Clicking a tag opens a shareable, filtered Articles route.
- [ ] Existing articles with tags and newly published articles round-trip consistently in mock and real modes.

skyl:
Closes #250

## What changed
- add accessible, debounced existing-tag autocomplete and validated open-vocabulary tag entry to the article composer
- normalize, deduplicate, bound, and persist up to eight tags through both real and mock publishing
- retrieve the canonical real article after publish so tags and creator data genuinely round-trip
- render reader tag links into URL-backed, shareable AND filters while preserving unrelated query parameters
- sanitize malformed legacy/backend/share tag values without breaking otherwise valid articles

## Verification
- typecheck
- production build
- Vitest: 626 passed, 5 skipped
- focused article-tag unit/API contracts: 7/7 after rebase onto current main
- focused article-tag Playwright: 2/2 at mobile width after rebase
- touch-target audit: 4/4
- UI-copy policy: 2/2
- formatting and git diff checks

### DIFF
--- a/wallet/zuuli/src/features/articles/components/ArticleTagInput.tsx
+++ b/wallet/zuuli/src/features/articles/components/ArticleTagInput.tsx
@@ -1,0 +1,221 @@
+import { useEffect, useId, useMemo, useState } from "react";
+import { Loader2, X } from "lucide-react";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { articles } from "@/lib/api/free2z";
+import type { ArticleTagSuggestion } from "@/lib/api/types";
+import {
+  MAX_ARTICLE_TAG_LENGTH,
+  MAX_ARTICLE_TAGS,
+  validateArticleTag,
+} from "@/lib/article-tags";
+
+interface ArticleTagInputProps {
+  value: string[];
+  onChange: (tags: string[]) => void;
+  disabled?: boolean;
+}
+
+export function ArticleTagInput({
+  value,
+  onChange,
+  disabled = false,
+}: ArticleTagInputProps) {
+  const listId = useId();
+  const helpId = useId();
+  const errorId = useId();
+  const [draft, setDraft] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [focused, setFocused] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [suggestions, setSuggestions] = useState<ArticleTagSuggestion[]>([]);
+  const valueKey = value.join("\u0000");
+  const full = value.length >= MAX_ARTICLE_TAGS;
+
+  useEffect(() => {
+    if (!focused || disabled || full) {
+      setLoading(false);
+      setSuggestions([]);
+      return;
+    }
+
+    let active = true;
+    const timer = window.setTimeout(() => {
+      setLoading(true);
+      void articles
+        .suggestTags(draft, value)
+        .then((next) => {
+          if (active) setSuggestions(next);
+        })
+        .catch(() => {
+          // The API boundary is already fail-soft, but retain that property if
+          // its implementation changes: authors can always enter a valid tag.
+          if (active) setSuggestions([]);
+        })
+        .finally(() => {
+          if (active) setLoading(false);
+        });
+    }, 150);
+
+    return () => {
+      active = false;
+      window.clearTimeout(timer);
+    };
+    // `valueKey` deliberately captures the selected set without retriggering
+    // for a new array containing the same tags.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [disabled, draft, focused, full, valueKey]);
+
+  const describedBy = useMemo(
+    () => (error ? `${helpId} ${errorId}` : helpId),
+    [error, errorId, helpId],
+  );
+
+  function addTag(raw: string) {
+    if (full) {
+      setError(`Articles can have at most ${MAX_ARTICLE_TAGS} tags.`);
+      return;
+    }
+    const result = validateArticleTag(raw);
+    if (!result.tag || result.error) {
+      setError(result.error ?? "Enter a valid tag.");
+      return;
+    }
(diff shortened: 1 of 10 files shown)


skyl:
Addressed the independent real-mode publish blocker at exact head 3abe8bd: once POST returns a committed article id, a transient canonical-detail GET failure now carries that id across a dedicated post-commit boundary; the composer reports Published and navigates to the reader instead of re-enabling Publish and inviting a duplicate POST. Also fail-closes non-array backend tag fields. Added contract regressions for both paths. Rebased integration was published as a fast-forward update (no force-push); typecheck and focused tag contracts 9/9 pass on current main. Exact-head CI and independent re-review are running.

skyl:
Reviewed adversarially on the owner's behalf. **Merge-with-followups** — the round trip is real and mostly well tested. One read-path defect and one theatrical test.

### The round trip genuinely works

Author → normalize → POST `tags` → re-fetch canonical detail → Reader chips → `/articles?tags=…` → feed → filtered result. Encoding is symmetric (`articleTagHref` emits `%20`, `URLSearchParams` emits `+`; both decode identically), and unrelated query params are preserved. The API filters case-insensitively, so client-side lowercasing doesn't break filtering.

**The unit tests are strong** — mutation-tested, 9 of 11 mutants go red: dropping `toLocaleLowerCase`, dropping dedup, dropping `encodeURIComponent`, dropping the `>MAX` throw, gutting `sanitizeArticleTags`, dropping `tags` from the POST body, echoing the create response instead of re-fetching, dropping sanitize from `mapArticle`, dropping suggestion normalization. That's a real suite.

`article-tags.pw.ts:9` is genuine end-to-end proof — breaking filtering makes it red.

### Test 2 is theater

`article-tags.pw.ts:70` ("a direct multi-tag Articles URL remains filtered across reload") **stays green with filtering completely disabled**. It reads its own expected count off the page (`:79 const expected = await resultCount.textContent()`) and then only asserts that some card exists. The `"tagged privacy + zcash"` label is derived from the URL param, not from filtering — so with filtering off it cheerfully asserts "76 articles tagged privacy + zcash".

It tests URL parsing and reload persistence, which is worth testing. It just isn't what its name claims. Hard-assert the count, or rename it.

### The real defect: the read path silently drops legitimate tags

`mapArticle` (`free2z.ts:447`) runs **stored backend data** through `sanitizeArticleTags`, which applies the *authoring* rules. The backend vocabulary is open and allows up to 100 characters. Running the real functions:

| stored tag | ZUULI shows |
|---|---|
| 33–100 chars (legal server-side) | **dropped** — client-invented 32-char cap (`article-tags.ts:29`) |
| `"🏳️‍🌈 pride"`, `"👨‍👩‍👧 family"` | **dropped** — `\p{Cf}` at `article-tags.ts:4` matches the ZWJ |
| Arabic/Hebrew tags carrying RLM/LRM | **dropped**, same rule |
| `"machine learning, deep learning"` | **dropped** (comma) |

Silently, from the Reader chips, from `Feed`'s `knownTags`, and from `suggestTags`.

And here's the part that makes it invisible: **the unit test's emoji case is `"🛡️ shielded"` — VS16 only, no ZWJ — so it passes and creates the appearance of emoji support the ZWJ cases don't have.** That's the recurring shape in this repo: a fixture chosen such that the invariant can't be violated.

Fix is small — split a lenient **read-side** sanitize (keep anything non-empty; reject only what actually breaks a URL) from the strict **write-side** validate. Filing separately.

### Two more, filing separately

- **Publish can report failure after succeeding.** `free2z.ts:1467` re-fetches with `anonymous: true` and throws on any failure; `Author.tsx` catches it into "Couldn't publish. Please try again." A transient blip on that second call shows the author an error for an article that *was* created — inviting a duplicate. Nothing covers create-ok/retrieve-fail.
- **Write-side case folding fragments the platform vocabulary.** Publishing `ai` alongside an existing `AI` creates a second tag row and splits autocomplete counts. Filtering still works (case-insensitive); only the vocabulary degrades.

### Uncovered bound

Removing `.slice(0, MAX_ARTICLE_TAGS)` at `article-tags.ts:77` stays green — the **URL-side** bound is untested, so a crafted `?tags=` with 500 entries is unbounded. (Changing `MAX_ARTICLE_TAGS` 8 → 99 also stays green, but that's cosmetic: tests build fixtures from the constant, so the mechanism is tested and the value isn't.)

### Constraints — all clean, and I checked the one I expected to break

No `truncate`/`text-ellipsis`/`line-clamp` introduced; `ui-copy-truncation` and `fixture-privacy` pass; `mock-data.ts` untouched; semantic tokens only.

I specifically probed the 320px tag-wall worry (#248's shape) with a mutated fixture of **8 tags × 32 chars** on `/articles` and on a reader page: `scrollWidth === clientWidth === 320`, zero overflowing elements. The layout genuinely holds.

Worth knowing the audits don't prove that themselves — they hit `/articles/new` with an **empty** composer (no chips, no listbox ever rendered) and the reader fixture has 3 tags of ≤8 chars. That's the #382 shape again; my probe is the evidence they lack.

Minor: the combobox has no `ArrowDown`/`aria-activedescendant` navigation — options are reachable only by Tab. They're real buttons so it works, but it's off-pattern for `role="combobox"`.

PR #502, opened 2026-08-23
fix(release): fail closed on stale ZUULI status

skyl:
Refs #474

> [!WARNING]
> Do not merge while protected build 13 run 32622069009 is active. This PR deliberately makes stale immutable release sources ineligible for later real recovery.

## What this does

- fails real automatic promotion and real manual recovery in the source-pinning prepare job when `STATUS.md` is stale
- parses the marker from the immutable release source and requires its audited SHA on the release parent first-parent history
- permits unrelated trunk merges, but rejects intervening ZUULI-impacting changes
- restricts the final release commit to `STATUS.md` and files owned by `release:bump`
- keeps dry runs available for packaging validation
- adds temp-Git behavioral coverage and a static workflow ordering contract
- documents the non-self-referential audited-ancestor semantics

This does **not** modify `STATUS.md`, advance its SHA/date, or claim production/native evidence. Issue #474 must remain open until that evidence is genuinely re-derived.

## Verification

- `npm run test:status-freshness` — 21/21 pass
- `npm run release:verify` — pass, identity `0.1.0+13`
- `npm run typecheck` — pass
- `npm test` — pass (Vitest, 111 Node tests, 111 Playwright tests)
- `node scripts/apple-credential-boundary.mjs` — pass
- `actionlint .github/workflows/zuuli-release.yml` — pass
- live check against current stale source — fails closed as intended


skyl:
Reviewed adversarially on the owner's behalf. **The implementation is right and I'd approve it on the merits — but I'm holding the merge for the owner**, because landing it stops the release train until a human re-derives STATUS.md with production evidence. That's a scheduling call, not a code call. See "Operational consequence" below.

### It does what #474 asked, proven by execution

Verified against real repo history from a scratch worktree:

```
STALE  (build-13 source db1d880, STATUS.md recording 1f752c7)
  node scripts/status-freshness.mjs --source-sha=db1d880…   -> EXIT 1
  "STATUS.md was not re-derived after release-impacting changes:" + ~250 paths

FRESH  (STATUS.md recording main tip, then a release.json bump)
  -> EXIT 0

FRESH, single combined commit (STATUS.md + bump together)
  -> EXIT 0

node --test scripts/status-freshness.node-test.mjs -> 21/21
```

### It genuinely blocks, and `--admin` does not bypass it

The check sits in `prepare` ("Pin immutable source"), after `should_release` is decided and **before** `$GITHUB_OUTPUT` is written. `prepare` has no `environment:`, and every credential-bearing job is `needs: prepare` gated on `should_release` — so a non-zero exit skips all of them. Nothing is built, signed, or uploaded.

The important structural point: it lives in the **release workflow**, not branch protection. Since `gh pr merge --admin` is routine here and branch protection only requires `gate`, a check placed there would have been bypassed constantly. This one can't be.

Dry runs are deliberately exempt, so packaging validation still works while STATUS.md is being re-derived. Good call.

### The staleness definition is sound

Ancestry + surface-diff rather than SHA equality: the recorded SHA must be on the first-parent history of the release parent, the diff between them must touch nothing release-impacting except STATUS.md, and the release commit itself may only touch that surface via STATUS.md plus the nine `release:bump`-owned files. The surface list is a faithful copy of the `case` globs in `zuuli.yml`.

And the docs are honest about the limit: this proves STATUS.md was *touched* at a point where nothing ZUULI-impacting has landed since, not that evidence was re-run. "Never advance either field merely to satisfy the guard" is exactly the right sentence to have written down.

### Operational consequence — the reason I'm not merging tonight

As of `2e7b06f` the recorded audit SHA is **50 first-parent commits and 239 release-impacting files** behind `main`. So once this lands, cutting build 14 requires, in order:

1. Re-run production/native evidence per `releasing.md:441-447` — the part that can't be manufactured from the repo.
2. Land STATUS.md recording `origin/main`'s tip **as the next release-impacting commit**.
3. `release:bump` + push.

That's the intended effect of #474 and I'm not arguing against it. But it's a real workflow change and it halts overnight batching, so it should be a deliberate choice about *when*, made by you.

Two things that bear on the timing:

- **There is a ~20-minute merge race.** 68 of the last 73 first-parent commits on `main` were release-impacting; median gap between merges ~20 min (p25 7, p75 41). A re-derived STATUS.md goes stale fast. The escape valve works — putting STATUS.md and the bump in the **same commit** passes, verified — but `releasing.md` still describes them as two separate PRs (steps 1 and 3), which is the arrangement most likely to lose the race. Worth documenting the single-commit path.
- **After this merges, a manual recovery re-run against build 13's source (`db1d880`) with `dry_run=false` fails closed permanently.** If there's any chance build 13 needs a store transaction replayed, do that before merging.

### Two real weaknesses, neither blocking

1. **The static contract in `release-identity.mjs` is `|| true`-defeatable.** I verified: appending `|| true` to the invocation line leaves the contract passing clean while the check no longer fails the job. It's ordering-aware, which is better than plain presence, but it does not prove the check can fail. Defense-in-depth, not a control — worth not describing it as a guarantee. (Reindenting or commenting *does* trip it, since the match includes exact leading whitespace.)
2. **Nothing in the gate parses the real `STATUS.md`.** `parseStatusMarker` is only exercised against synthetic fixtures, and the regex requires the SHA to begin a new line immediately after ``…`origin/main` at ``. Joining lines 10–11 makes it throw — fail-closed, but discovered at promotion time, which is the worst possible moment. Cheap fix: assert `parseStatusMarker(readFileSync("wallet/zuuli/STATUS.md"))` in `npm test`, and add a "this line is machine-parsed, do not reflow" note to STATUS.md itself.

### Also noted

- **The recorded date is validated but never compared to anything** — a fresh SHA with `2020-01-01` passes. #474 said "stale commit/date"; only the commit half is mechanical. Fine, but worth knowing.
- **`zuuli-store-publish.yml` writes to the stores and does not consult STATUS.md freshness.** #474 was about build promotion so this isn't a gap in scope, but it is an uncovered store-writing path.
- `releasing.md` is properly updated (22 lines under step 1, accurate to the implementation), and STATUS.md correctly isn't touched — the PR doesn't advance the SHA or claim evidence, and says #474 stays open until real evidence is re-derived. That restraint is the right instinct.
- #477 hygiene clean: the only added shell is a compound `if [[ … ]]; then … fi`, verified to abort under `set -euo pipefail` on bash 3.2.57 as well as CI's bash 5.

**@skyl — your call on when to land this.** I'll keep batching builds without it in the meantime; say the word and I'll merge.

skyl:
Addressed the release-file exemption blocker at exact head `a8b4e01`.

The release bump and freshness guard now share one deterministic pure transformer. For every release-owned file changed in the source commit, the guard reconstructs the exact expected content from the parent tree and rejects any extra dependency, lockfile, Tauri, Xcode, Gradle, Cargo, or release-metadata change.

Added semantic tamper regressions for all nine release-owned files, plus real unrelated-merge acceptance and release-impacting second-parent rejection. Local evidence:
- `npm run test:status-freshness` — 31/31 pass
- `npm run release:verify` — `0.1.0+13`
- `actionlint .github/workflows/zuuli-release.yml` — pass
- Prettier, Node syntax checks, and `git diff --check` — pass

Requesting a fresh adversarial review and exact-head CI before merge.

PR #503, opened 2026-08-23
fix(converse): prevent async embed scroll jumps

### LINKED ISSUE #137: Scroll on converse on load pushes up under search
Often, seemingly when an embed loads asynchronously, it pushed the feed up under the search (instead of down).

Image

This is annoying and looks bad ;D

Maybe there is a very simple fix ...

skyl:
Closes #137

## What changed
- exclude the window-scrolled Converse virtual feed from native browser scroll anchoring
- keep React Virtuoso responsible for measuring asynchronously resized cards without a second browser-level `window.scrollY` adjustment
- add a focused regression that renders the real Virtuoso boundary while an embed-sized card expands asynchronously

## Why
When an embed above Chrome’s selected scroll anchor grows after loading, native scroll anchoring advances the document scroll position. Because Converse uses `useWindowScroll` underneath a fixed search/filter panel, that compensation moves the feed beneath the panel. `overflow-anchor: none` at the feed boundary prevents the competing browser adjustment while preserving Virtuoso measurement and user scrolling.

## Verification
- `CI=true npm test -- --watchAll=false --runInBand src/ConverseList.test.tsx`
- `CI=false npm run build`
- `npx prettier --check src/ConverseList.test.tsx src/ConverseList.css`
- `git diff --check`
- headless Chromium reproduction: delayed 500px content growth changed `scrollY` 300 → 800 before the boundary; it remained 300 → 300 with the boundary

### DIFF
--- a/ts/react/free2z/src/ConverseList.css
+++ b/ts/react/free2z/src/ConverseList.css
@@ -1,0 +1,4 @@
+.converse-feed {
+  /* Virtuoso remeasures async embeds; native anchoring would move the window too. */
+  overflow-anchor: none;
+}
--- a/ts/react/free2z/src/ConverseList.test.tsx
+++ b/ts/react/free2z/src/ConverseList.test.tsx
@@ -1,3 +1,4 @@
+import { readFileSync } from "fs";
 import { act } from "react";
 import { createRoot, Root } from "react-dom/client";
 import { VirtuosoMockContext } from "react-virtuoso";
@@ -74,6 +75,14 @@ jest.mock("./components/CommentCardInfinite", () => {
       </article>
     );
   };
+});
+
+test("disables native scroll anchoring at the Converse feed boundary", () => {
+  const css = readFileSync(`${__dirname}/ConverseList.css`, "utf8");
+
+  expect(css).toMatch(
+    /\.converse-feed\s*\{[^}]*overflow-anchor\s*:\s*none\s*;?[^}]*\}/
+  );
 });
 
 test("keeps async embed resizes from selecting a native scroll anchor in the virtual feed", async () => {
--- a/ts/react/free2z/src/ConverseList.test.tsx
+++ b/ts/react/free2z/src/ConverseList.test.tsx
@@ -1,0 +1,107 @@
+import { act } from "react";
+import { createRoot, Root } from "react-dom/client";
+import { VirtuosoMockContext } from "react-virtuoso";
+
+import ConverseList from "./ConverseList";
+
+(
+  globalThis as typeof globalThis & { IS_REACT_ACT_ENVIRONMENT: boolean }
+).IS_REACT_ACT_ENVIRONMENT = true;
+
+jest.mock("react-query", () => ({
+  useInfiniteQuery: () => ({
+    data: {
+      pages: [
+        {
+          count: 1,
+          next: null,
+          previous: null,
+          results: [
+            {
+              uuid: "async-embed-comment",
+              author: { username: "alice" },
+              parent: null,
+              headline: "An asynchronously sized embed",
+              content: "::embed[https://example.com/video]",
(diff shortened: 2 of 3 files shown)


skyl:
Addressed the adversarial review gap in `256b07f`: the focused suite now reads `ConverseList.css` and requires `overflow-anchor: none` inside the scoped `.converse-feed` rule. Removing or moving the actual CSS fix now fails the regression instead of merely passing on DOM identity.

Verification:
- `CI=true npm test -- --watchAll=false --runInBand src/ConverseList.test.tsx` — 2/2 passed
- `npx prettier --check src/ConverseList.test.tsx src/ConverseList.css` — passed
- `git diff --check` — passed
- `CI=false npm run build` — passed with the repository's existing warnings

skyl:
Reviewed adversarially on the owner's behalf. **The 7-line fix is right and I'd merge it today. Requesting changes on the test only.**

### The fix addresses the cause, not the symptom

`overflow-anchor: none` is the correct mechanism — not a scroll-restoration hack. Two things were compensating for the same growth: Chromium's scroll anchoring (on `documentElement`, because Converse uses `useWindowScroll`) and Virtuoso's own measurement. I dumped the rendered DOM and confirmed the class lands on `div[data-virtuoso-scroller="true"]`, the ancestor of every item — Chromium's `ScrollAnchor` returns `kSkip` for a subtree rooted at `overflow-anchor: none`, so the whole feed leaves anchor candidacy and only Virtuoso compensates. Right element, right cause.

No layout reservation is introduced, so there's no 320px overflow risk and no blank gap when an embed fails or is blocked by the media-consent boundary. No scrolling behaviour is introduced, so `prefers-reduced-motion` isn't implicated. And it doesn't interact with `tests/scroll-restoration.pw.ts` — that's a different app.

### The test cannot fail if the feature is broken

`ConverseList.test.tsx:79` is named "keeps async embed resizes from selecting a native scroll anchor in the virtual feed". Four mutants:

| mutation | result |
|---|---|
| delete `overflow-anchor: none` from the CSS | **PASS** |
| set `overflow-anchor: auto` (the exact bug) | **PASS** |
| delete `import './ConverseList.css'` entirely | **PASS** |
| remove `className="converse-feed"` | FAIL |

It asserts that `.converse-feed` exists, that the mock grew after 10ms, and that node identity is stable. It never touches `scrollY`, `getComputedStyle`, or `overflow-anchor` — and it *can't*, because CRA's jest maps CSS imports to a stub and jsdom has no layout or scroll anchoring.

So it's a className assertion wearing the name of a scroll-anchoring regression test, at 107 lines of mocking for one string. That's the pattern I've been blocking all night, and it's worth holding the line on even when the underlying fix is good.

### And it never runs in CI

`.github/workflows/ts-react-free2z.yml:25-27` — the `Run tests` step is **commented out**. This is the only test file in the entire `ts/react/free2z` app; CI runs `CI=false npm run build` only. The green `build` check here proves nothing about the test, and the green `gate` is the unrelated zuuli workflow.

### What I'd ask

1. **Make the guard falsifiable, or drop it.** A node test that parses `ConverseList.css` and asserts the declaration, *plus* the existing class assertion — two halves, neither silently removable — would be honest and cheap. A real-browser check would be better. Either is fine; a test that passes with the CSS deleted is not.
2. **Rename it if you keep it** — it doesn't test what its name claims.
3. **Uncomment `Run tests`** in `ts-react-free2z.yml`, or file an issue for it. Otherwise this and every future test in that app is decorative by construction. Filing that one myself.

The PR body's headless repro — "300 → 800 before, 300 → 300 after" — is the only real evidence in this change, and it isn't committed anywhere. If the falsifiable test is more trouble than it's worth, say that plainly in the body and keep the repro; I'd take that over a green assertion that means nothing.

PR #507, opened 2026-08-23
fix(zpages): preview first body image in cards

### LINKED ISSUE #51: Featured picture
Visually, it would be helpful when there is no featured photograph, a photograph from the page is automatically selected for the article preview.
Screenshot 2023-04-11 at 8 31 37 PM

skyl:
Closes #51

## Summary
- keep an explicitly selected featured image as the first choice for classic zpage cards
- otherwise extract the first safe Markdown image from the article body before using the existing Tuzi fallback
- ignore image-like text in comments/code, reject non-HTTP(S) schemes, and support inline/reference Markdown destinations
- keep the fallback dimmed while showing featured/body images at full opacity

## Verification
- `CI=true npm test -- --runInBand --watchAll=false` (3 suites, 11 tests)
- `npx eslint src/lib/page-card-image.ts src/lib/page-card-image.test.ts`
- `CI=false npm run build` (passes with pre-existing repository warnings)
- `git diff --check`

### DIFF
--- a/ts/react/free2z/config-overrides.js
+++ b/ts/react/free2z/config-overrides.js
@@ -27,3 +27,20 @@ module.exports = function override(config, env) {
     return config
 }
 
+module.exports.jest = function overrideJest(config) {
+    // The production Markdown stack is ESM-only. Transform it in Jest so card
+    // image tests execute the same parser instead of a mock or parallel parser.
+    const parserTransformPattern =
+        "node_modules/(?!(?:bail|character-entities|decode-named-character-reference|devlop|is-plain-obj|mdast-util-[^/]+|micromark(?:-[^/]*)?|remark-parse|trough|unified|unist-util-[^/]+|vfile(?:-message)?)/)"
+    config.transformIgnorePatterns = config.transformIgnorePatterns.map(
+        (pattern) =>
+            pattern.includes("node_modules") ? parserTransformPattern : pattern
+    )
+    config.moduleNameMapper = {
+        ...config.moduleNameMapper,
+        "^#minpath$": "<rootDir>/node_modules/vfile/lib/minpath.js",
+        "^#minproc$": "<rootDir>/node_modules/vfile/lib/minproc.js",
+        "^#minurl$": "<rootDir>/node_modules/vfile/lib/minurl.js",
+    }
+    return config
+}
--- a/ts/react/free2z/package-lock.json
+++ b/ts/react/free2z/package-lock.json
@@ -81,6 +81,7 @@
         "remark-gfm": "^4.0.1",
         "remark-math": "^6.0.0",
         "remark-oembed": "^1.2.2",
+        "remark-parse": "^11.0.0",
         "slugify": "^1.6.6",
         "swiper": "^9.4.1",
         "three": "^0.168.0",
--- a/ts/react/free2z/package.json
+++ b/ts/react/free2z/package.json
@@ -76,6 +76,7 @@
     "remark-gfm": "^4.0.1",
     "remark-math": "^6.0.0",
     "remark-oembed": "^1.2.2",
+    "remark-parse": "^11.0.0",
     "slugify": "^1.6.6",
     "swiper": "^9.4.1",
     "three": "^0.168.0",
--- a/ts/react/free2z/src/components/PageListRow2.test.tsx
+++ b/ts/react/free2z/src/components/PageListRow2.test.tsx
@@ -1,0 +1,136 @@
+import { render, screen } from "@testing-library/react";
+
+import PageListRow from "./PageListRow2";
+
+jest.mock("@mui/icons-material", () => ({
+  Edit: () => null,
+  MoreVert: () => null,
+}));
+
+jest.mock("@mui/material", () => {
+  const React = require("react");
+  const Wrapper = ({ children }: { children?: React.ReactNode }) => (
+    <div>{children}</div>
+  );
+  const CardMedia = React.forwardRef(
+    (
+      {
+        image,
+        style,
+        title,
+      }: { image: string; style: { opacity: number }; title: string },
+      ref: React.ForwardedRef<HTMLDivElement>
+    ) => (
+      <div
+        ref={ref}
+        data-testid="page-card-image"
+        data-image={image}
+        data-opacity={style.opacity}
+        title={title}
+      />
+    )
+  );
+  CardMedia.displayName = "CardMedia";
+
+  return {
+    Avatar: Wrapper,
+    Box: Wrapper,
+    Card: Wrapper,
+    CardContent: Wrapper,
+    CardHeader: Wrapper,
+    CardMedia,
+    Divider: Wrapper,
+    Grid: Wrapper,
+    IconButton: Wrapper,
+    Link: Wrapper,
+    ListItem: Wrapper,
+    Popover: Wrapper,
+    Stack: Wrapper,
+    Tooltip: Wrapper,
+    Typography: Wrapper,
+    useMediaQuery: () => false,
+  };
+});
+
+jest.mock("moment", () => () => ({ fromNow: () => "now" }));
+jest.mock("../state/global", () => ({
+  useGlobalState: () => [{ username: "viewer", stars: [] }, jest.fn()],
+}));
+jest.mock("../hooks/useTransitionNavigate", () => ({
+  useTransitionNavigate: () => jest.fn(),
+}));
+jest.mock("./CreatorDonate", () => () => null);
+jest.mock("./UpDownPage", () => () => null);
+jest.mock("./TransitionLink", () => () => null);
+jest.mock("./profile/CreatorSupport", () => () => null);
+
+const page = {
+  creator: {
+    username: "alice",
+    is_verified: false,
+    total: 0,
+    full_name: "Alice",
+    description: "",
+    p2paddr: "creator-address",
+    zpages: 1,
+    member_price: "",
+    can_stream: false,
+    avatar_image: null,
+    banner_image: null,
+  },
+  get_url: "/alice/article",
+  vanity: "article",
+  title: "Article",
+  content: "![Body](/uploads/body.jpg)",
+  description: "",
+  tags: [],
+  free2zaddr: "article-id",
+  p2paddr: "address",
+  featured_image: {
+    name: "featured.jpg",
+    title: "Featured",
+    access: "public" as const,
+    thumbnail: "/uploads/featured.jpg",
+  },
+  is_verified: false,
+  is_published: true,
+  is_subscriber_only: false,
+  total: "0",
+  f2z_score: "0",
+  created_at: "2026-08-23T00:00:00Z",
+  updated_at: "2026-08-23T00:00:00Z",
+  comments: [],
+};
+
+test("uses the selected card image and fallback opacity in the classic card", () => {
+  const { rerender } = render(<PageListRow {...page} mine={false} />);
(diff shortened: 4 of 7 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Requesting changes.** The parser is careful work — 270 lines of real CommonMark handling with masking, reference definitions, escapes and scheme rejection — but I don't think the feature fires in production, and the tests can't tell you that.

### The blocking finding: this is very likely a no-op on real data

`PageListRow2` gets `page.content` from the zPage **list** endpoint (`Find.tsx:180`). That endpoint does not return the article body. It returns the page's `description` when one exists, and otherwise a truncated, tag-stripped rendering of the body capped at **365 characters** — markdown image syntax is removed outright by that stripping.

So:

- **Any zpage with a `description`** — the common case, and exactly the cards in #51's screenshot — has `page.content` = the description, containing no markdown at all. `extractFirstArticleImage` returns `null` and the card is unchanged.
- **A description-less page** exposes at most the first ~365 characters of body, so the image has to be in roughly the opening paragraph.

Every fixture in `page-card-image.test.ts` feeds a full markdown body that this endpoint never sends. That's the fixture-shaped-so-the-defect-can't-appear pattern this repo keeps hitting — the tests are green against input the code will not receive.

**What I'd need to merge:** one test using a genuinely list-shaped `content` value (description populated, and separately the truncated-body form), or a screenshot from the running app showing a card that actually changed. If the list payload genuinely can't carry this, the right fix is a server-derived preview field rather than client-side body parsing.

### And the wiring is untested — the whole change can be deleted silently

| Mutation | Result |
|---|---|
| **revert `PageListRow2.tsx:333` to `featured_image?.thumbnail \|\| "/docs/img/tuzi.svg"` — i.e. delete the entire user-visible change** | **all 11 tests still PASS** |
| revert the `source: "body"` opacity handling | PASS |
| allow every scheme except `javascript:` (so `data:`, `blob:`, `file:` pass) | PASS |
| stop rejecting control characters in URLs | PASS |
| neuter `isEscaped` entirely | PASS |
| disable code/comment masking | RED |
| featured image no longer beats body image | RED |
| reference-style definitions unsupported | RED |

80 of the 80 new test lines cover the pure helper; **zero cover the wiring that is the PR.** A test that fails when `cardImage.url` stops being used is the minimum here.

The scheme one matters independently: the body claims "reject non-HTTP(S) schemes," but the only fixture is `javascript:alert(1)`, so `data:`, `blob:`, `file:` and `vbscript:` are unpinned.

### Security: this escalates a beacon from readers to scrollers

`CardMedia image={cardImage.url}` sets a CSS `background-image`, so a creator can put `` in a body and collect the IP and User-Agent of **everyone who scrolls past the card in search results** — no one need open the article. Measured behaviour on the PR head:

```
//evil.example.com/pixel.gif           ACCEPTED (protocol-relative)
https://attacker.example/p.gif?u=1     ACCEPTED
http://127.0.0.1:9000/probe.png        ACCEPTED
https://user:pass@evil.example/a.png   ACCEPTED
data: / blob: / file: / javascript:    rejected ✓
```

`CustomImage.tsx` already renders body images unproxied on the detail page, so this isn't a brand-new capability — but it moves the exposure from "readers of that article" to "anyone who scrolls the results page," and adds protocol-relative and loopback reachability. That deserves an explicit decision (same-origin-only for card previews, or an image proxy), not a silent landing.

### Latent O(n²) in the parser

`closingBracket()` (`page-card-image.ts:143`) rescans to end-of-string for every `![`. Measured on `"![".repeat(n)`:

```
 25 KB  0.54 s
 50 KB  2.0 s
100 KB  8.5 s
200 KB  34.2 s
```

Synchronous, inside `useMemo`, during render, per card. **Not reachable today** only because the list payload caps content at 365 characters — but the body field allows up to 100,000, so this becomes a hard main-thread freeze the moment anyone reuses this helper against a full body. Bound the input length inside `extractFirstArticleImage`.

### Smaller

- **No `onError` fallback.** A `source: "body"` image renders at full opacity with no recovery, so a dead, hotlink-blocked, or mixed-content `http://` URL now yields a blank box where the old code always showed the dimmed placeholder. That's a regression in the failure case.
- Indented four-space code blocks are **not** masked (fenced ones are, though only `~~~` is tested).
- Raw `` HTML isn't parsed — safe, but it does mean the PR doesn't serve part of #51's use case.
- **Layout is clean:** `height: 0; paddingTop: 56.25%` is pre-existing and unchanged, so the aspect ratio is reserved — no #503-class async-size-change bug, and no 320px overflow risk.
- `Closes #51` is a fair reference, but note #51 also asks for the **detail** view; this touches only the listing row.

### Minimum to merge

1. A test at the real payload shape, or app evidence that a card changes.
2. A test that fails if `PageListRow2` stops using `cardImage.url`.
3. Extend the unsafe-scheme fixture beyond `javascript:` to `data:`, `blob:`, `file:`.
4. Decide the arbitrary-remote-host question — at minimum reject protocol-relative `//host` and loopback.
5. Bound the parser input length.

skyl:
Corrected the Markdown extraction blockers in 14b464e4940d96cdadfc28767be68e94cd85954a.

- Replaced the hand-written scanner with the app stack's `remark-parse`/CommonMark AST, with iterative document-order traversal and first-definition reference resolution. This aligns indented/fenced code, unterminated fences/comments, raw HTML blocks, malformed angle destinations, and character-reference decoding with ReactMarkdown.
- Preserved featured > first safe rendered body image > Tuzi fallback precedence and rejects malformed/control/non-HTTP(S) destinations.
- Added a 100k-character unmatched-image regression to guard against quadratic rescanning.
- Added a `PageListRow2` contract test proving the real component consumes featured/body/fallback results and applies full vs fallback opacity.
- Added the parser as an explicit dependency plus the narrow Jest ESM transform needed to execute the production parser rather than a mock.

Evidence:
- Clean-cache full classic Jest: 4 suites, 26 tests passed.
- Focused parser suite after final timing adjustment: 22/22 passed.
- Focused ESLint: clean.
- Prettier check: clean.
- `git diff --check`: clean.
- Production `CI=false npm run build`: completed successfully (pre-existing source-map/ESLint warnings only).
- Local, remote branch, and PR head all resolve to `14b464e4940d96cdadfc28767be68e94cd85954a`.

Ready for independent re-review; not merged.

skyl (commented):
BLOCKED — fresh independent re-review of exact head `14b464e4940d96cdadfc28767be68e94cd85954a`.

The correction resolves several prior findings well:

- Replacing the hand scanner with `remark-parse` aligns inline/reference images, definitions, escapes, indented/fenced code, HTML blocks/comments, malformed constructs, and character references with ReactMarkdown's CommonMark foundation.
- Both AST passes are iterative and linear. Direct timings for `"![".repeat(n)` were 39 ms at 25k characters, 47 ms at 50k, and 88 ms at 100k; the former O(n²) render freeze is gone.
- The URL tests now pin `data:`, `ftp:`, `javascript:`, control/NUL replacement, malformed `http://`, and malformed angle destinations.
- `PageListRow2.test.tsx` is a meaningful wiring contract: reverting `cardImage.url` or the body/fallback opacity distinction now fails. This fixes the prior “delete the feature and all tests stay green” problem.
- Full clean-cache classic Jest passed locally: 4 suites / 26 tests. Focused ESLint had no errors (one pre-existing `setUser` warning). CI classic `build` is green. The branch also derives conflict-free against current `origin/main`.

Three blockers remain.

### 1. The feature still does not receive article Markdown in the production list flow

The production path is:

`Find.tsx:180` GET `/api/zpage/` → `res.data.results` → `PageListRow2 {...page}` → `getPageCardImage(..., page.content)`.

But the public `zPageListSerializer.get_content()` does this:

```py
if instance.description:
    return instance.description
if hasattr(instance, 'truncated_content'):
    html = markdown2.markdown(instance.truncated_content)
    text = strip_tags(html)
    return f"{text[:365]}..."
```

So populated descriptions replace the body, while the fallback path converts the truncated body to HTML and strips every Markdown image before the client sees it. The new component test still constructs an impossible list payload by assigning a full Markdown body directly to `page.content`; it proves component wiring, not API integration.

This is not theoretical. I queried the current public API and found zpage `07f4d5d1-0a09-4b55-b59c-4ebe277ca66c` (“Zodl summit 2026…”):

- `featured_image` is `null`.
- The detail endpoint returns 6,553 characters and multiple body images; this helper correctly finds `/uploadz/public/Zerodartz/zodl-yes1.webp`.
- The list endpoint returns `content === description` (`"Zodl summit 2026 recap…"`) with no Markdown.
- Running this exact PR helper on that live list object returns `{ url: "/docs/img/tuzi.svg", source: "fallback" }`.

That is precisely the user-visible case #51 asks to change, and the card remains unchanged. The fix needs a list-shaped server field carrying a derived safe preview image (or another real fetch/data contract), plus a test through the actual list response mapping. A direct component fixture containing full body Markdown cannot establish that.

### 2. Arbitrary remote-host promotion remains undecided and unbounded

The revised `safeImageUrl()` still accepts all of these verbatim:

```text
//evil.example/pixel.gif
http://127.0.0.1:9000/probe.png
http://[::1]/probe.png
http://169.254.169.254/latest/meta-data/
https://user:pass@evil.example/a.png
```

Once blocker 1 is fixed, `CardMedia` turns those creator-controlled body URLs into CSS background requests for every results-page scroller. That escalates remote image disclosure from article readers to people who never opened the article, and protocol-relative, loopback/link-local, and credential-bearing destinations are still admitted. The parser's non-HTTP rejection is useful but does not resolve the privacy/SSRF-adjacent browser-fetch policy. Use a same-origin uploaded-media contract or a controlled image proxy; at minimum, make the accepted remote-host policy explicit and test it.

### 3. A failed promoted image regresses to a blank card, and detail scope is untouched

`CardMedia image={cardImage.url}` is a CSS `background-image`; there is no image load event or layered Tuzi background. A dead, blocked, mixed-content, or hotlink-denied body URL therefore leaves the media box blank at full opacity instead of recovering to the prior dimmed Tuzi fallback. The component test mocks `CardMedia` as a data-only `<div>`, so it cannot observe loading failure.

Also, #51's title explicitly says “listing/detail.” This PR changes only `PageListRow2`; `PageHeader` still applies a background only when `featured_image` exists, and `PageRenderer` still renders only a divider when it does not. The same live zpage above has body images but no detail header image. Either implement the detail half or change `Closes #51` to `Refs #51` and leave the issue open.

Exact-head verification:

- `CI=true npm test -- --watchAll=false --runInBand` — 4/4 suites, 26/26 tests pass
- focused parser/component suites — 23/23 pass
- focused ESLint — 0 errors, one unrelated existing warning
- `git diff --check` — pass
- classic `build` CI — pass
- required wallet frontend CI was still running when reviewed; all completed Rust checks were green

No parser-complexity or component-wiring blocker remains. The production payload contract, remote fetch policy, failed-image recovery, and issue-closing scope remain blocking.

skyl:
Closing this client-only implementation after the public-backend ownership audit documented in #51. The list serializer replaces/strips body Markdown before `PageListRow2` receives it, so this PR remains a production no-op for the target case; it also cannot safely define the persisted preview, same-origin/proxy trust boundary, backfill/invalidation, or query-count contract because the public repo lacks the zPage/GenericFile models, migrations, zpage viewset/queryset, and backend endpoint tests. The exact unblock and a live public API reproduction are recorded at https://github.com/free2z/zuu/issues/51#issuecomment-5384906489. No private repository or data was inspected. Keeping #51 open as `blocked` + `needs-backend`; this PR must not merge.

PR #508, opened 2026-08-23
fix(zuuli): preserve stored article tags

### LINKED ISSUE #504: fix(zuuli): the read path silently drops legitimate backend tags by applying authoring rules
## Problem

`mapArticle` (`wallet/zuuli/src/lib/api/free2z.ts:447`) runs **stored backend data** through `sanitizeArticleTags` — a function written to enforce the *authoring* rules. The backend tag vocabulary is open and allows names far longer and more varied than the client's write-side constraints, so legitimate stored tags are silently discarded on read.

Verified by running the real functions:

| stored tag | ZUULI shows |
|---|---|
| 33–100 characters (legal server-side) | **dropped** — client-invented 32-char cap, `article-tags.ts:29` |
| `"🏳️‍🌈 pride"`, `"👨‍👩‍👧 family"` | **dropped** — `\p{Cf}` at `article-tags.ts:4` matches U+200D ZWJ |
| Arabic/Hebrew tags carrying RLM/LRM | **dropped**, same rule |
| `"machine learning, deep learning"` | **dropped** (comma) |
| `"ART"` | shown as `#art` |

The loss is silent and affects the Reader chips, `Feed`'s `knownTags`, and `suggestTags`. A creator who tagged an article on the web sees those tags vanish in ZUULI with no indication.

## Why it wasn't caught

The unit test's emoji fixture is `"🛡️ shielded"` — **VS16 only, no ZWJ** — so it passes and creates the appearance of emoji support that the ZWJ cases don't have.

That's the recurring shape in this repo: the fixture is chosen such that the invariant cannot be violated. Same family as #497 and the search-pagination mock in #482.

## Suggested fix

Split the two concerns, which are genuinely different:

- **Write-side `validateArticleTags`** — strict. Enforces what *this client* is willing to create: length cap, character set, count limit, dedup, casing. Rejecting input here is correct and user-visible.
- **Read-side `sanitizeArticleTags`** — lenient. Keeps anything non-empty; rejects only what would actually break a URL or the layout. Display is not the place to relitigate authoring policy.

Add fixtures that would actually fail today: a ZWJ emoji sequence, an RTL tag with a directional mark, a 100-char tag, and a tag containing a comma.

## Related, worth folding in or filing alongside

**The URL-side bound is untested.** Removing `.slice(0, MAX_ARTICLE_TAGS)` at `article-tags.ts:77` leaves every test green, so a crafted `?tags=` with 500 entries is unbounded on the read path.

## Acceptance

- A tag legal on the server round-trips into ZUULI's UI unchanged, including ZWJ emoji, RTL marks, commas, and lengths up to the server maximum.
- A test fails if read-side sanitisation starts dropping legal stored tags.
- The `?tags=` bound is covered.

Found while reviewing #500 (merged). Related: #250, #248.

skyl:
Closes #504

## What changed

- keep strict normalization and character/count limits on the article authoring path
- preserve server-owned tags on read, including original case, ZWJ emoji, RTL marks, commas, and the backend's 100-character maximum
- continue failing soft for non-string, empty, over-limit, and actual control-character response values
- display comma-bearing tags without offering a misleading filter link, since the backend filter uses an unescaped comma delimiter
- lock the share-URL filter to eight tags with a regression that fails if the bound is removed

## Verification

- `npx vitest run` — 646 passed, 5 skipped
- `npm run typecheck` — pass
- `npm run build` — pass
- focused tag/API contract tests — 12/12 pass
- `git diff --check` — pass

### DIFF
--- a/wallet/zuuli/src/features/articles/pages/Feed.tsx
+++ b/wallet/zuuli/src/features/articles/pages/Feed.tsx
@@ -8,6 +8,7 @@ import { PageHeader } from "@/components/common/PageHeader";
 import { useRouteScroll } from "@/hooks/useRouteScroll";
 import { cn } from "@/lib/utils";
 import {
+  isArticleTagFilterable,
   MAX_ARTICLE_TAGS,
   parseArticleTagsParam,
   sanitizeArticleTags,
@@ -187,6 +188,17 @@ export function Feed() {
         <div className="mb-6 flex flex-wrap items-center gap-2" aria-label="Filter by tag">
           {knownTags.map((tag) => {
             const on = selectedTags.includes(tag);
+            if (!isArticleTagFilterable(tag)) {
+              return (
+                <span
+                  key={tag}
+                  className="min-tap inline-flex max-w-full items-center rounded-full border border-border px-3 py-1 text-xs font-medium text-muted-foreground"
+                  title="This stored tag cannot be represented by the server's comma-delimited filter."
+                >
+                  <span className="min-w-0 break-words">{tag}</span>
+                </span>
+              );
+            }
             return (
               <button
                 key={tag}
@@ -194,7 +206,7 @@ export function Feed() {
                 aria-pressed={on}
                 onClick={() => toggleTag(tag)}
                 className={cn(
-                  "min-tap rounded-full border px-3 py-1 text-xs font-medium transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring",
+                  "min-tap max-w-full break-words rounded-full border px-3 py-1 text-xs font-medium transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring",
                   on
                     ? "border-primary bg-primary/15 text-primary"
                     : "border-border bg-transparent text-muted-foreground hover:bg-secondary hover:text-foreground",
--- a/wallet/zuuli/src/features/articles/pages/Reader.tsx
+++ b/wallet/zuuli/src/features/articles/pages/Reader.tsx
@@ -28,6 +28,32 @@ import { ArticleScore } from "../components/ArticleScore";
 import { coverTone } from "@/lib/cover";
 import { formatPublished } from "../lib";
 
+const articleTagClassName =
+  "min-tap inline-flex max-w-full items-center rounded-full border border-primary/30 bg-primary/10 px-3 py-1 text-xs font-medium text-primary";
+
+function StoredArticleTag({ tag }: { tag: string }) {
+  const href = articleTagHref(tag);
+  const label = <span className="min-w-0 break-words">#{tag}</span>;
+  if (!href) {
+    return (
+      <span
+        className={articleTagClassName}
+        title="This stored tag cannot be represented by the server's comma-delimited filter."
+      >
+        {label}
+      </span>
+    );
+  }
+  return (
+    <Link
+      to={href}
+      className={`${articleTagClassName} transition-colors hover:bg-primary/15 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring`}
+    >
+      {label}
+    </Link>
+  );
+}
+
 export function Reader() {
   const { slug } = useParams<{ slug: string }>();
   const key = slug ?? "";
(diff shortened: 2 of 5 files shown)


skyl:
Addressed the malformed-Unicode review blocker at exact head `c7e8c5b891094bc15f78e18f8c194b8ae54a092f`. The read sanitizer now rejects unpaired UTF-16 surrogates while preserving valid surrogate pairs/emoji, and the regression proves both sanitization and `articleTagHref` fail soft for `\ud800`. Focused tag/API tests (12/12), typecheck, production build, and diff check pass. Requesting fresh exact-head review and CI.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** once `gate` is green — this is the best-verified PR I've reviewed in this repo.

### The trap wasn't repeated

#504 went unnoticed because the emoji fixture was `"🛡️ shielded"` — VS16 only, no ZWJ — so it passed and manufactured the appearance of emoji support. The obvious failure mode for a fix was to add another soft fixture. Instead the new read-side fixtures are the hard cases: a real ZWJ sequence (`"🏳️‍🌈 pride"` = U+1F3F3 VS16 **U+200D** U+1F308), an Arabic tag carrying a real U+200F RLM, a tag of exactly `MAX_STORED_ARTICLE_TAG_LENGTH` (100) characters, a comma-bearing tag, plus negatives at 101 chars and a NUL. The VS16-only fixture correctly survives in the *authoring* test, where it belongs.

### All seven questions, answered empirically

| | | |
|---|---|---|
| 1 | 100-char stored tag survives | **yes** (101 dropped, 33 survives) |
| 2 | ZWJ survives | **yes** — both the pride flag and family sequences |
| 3 | RTL with RLM/LRM survives | **yes** |
| 4 | comma tag survives | **yes** — displayed, and `articleTagHref` returns `null` so it renders as an unlinked chip rather than a broken filter. Nice touch. |
| 5 | fixtures are the hard cases | **yes** |
| 6 | real read/write split | **yes** — two patterns (`AUTHORING_TAG_CONTROL_PATTERN` = `/[\p{Cc}\p{Cf}]/u` vs `STORED_TAG_CONTROL_PATTERN` = `/\p{Cc}/u`) and two limits (32 vs 100). Write side still rejects 33 chars, commas, ZWJ, empties and >8; still dedups and lowercases. |
| 7 | `?tags=` bound covered | **yes** — and mutation-confirmed red when the `.slice()` is removed |

`MAX_STORED_ARTICLE_TAG_LENGTH = 100` matches the server's actual tag-name limit rather than being invented.

### Mutation table — 15 of 17 killed

Including the one that matters most: **M1, reverting the read side to authoring rules — the #504 bug itself — goes red across 6 tests.** Also red: removing the URL-side slice, changing the stored cap 100→32, removing it entirely, re-adding `\p{Cf}` (which kills ZWJ and RTL), removing read-side control rejection, letting `articleTagHref` accept commas, removing read-side dedup, re-lowercasing on read, and all five write-side guards.

The two survivors are honest ones: the `Feed.tsx:191` and `Reader.tsx:37` JSX branches, uncovered because `features/articles/` has **no component tests at all**. The library-level contract *is* pinned (M5 kills it); only the consumption isn't.

### Follow-ups — filing, not blocking

1. **Bidi override characters now render in chips.** `\p{Cf}` was the old blanket guard; `\p{Cc}` lets the override and isolate controls through — U+202D LRO, U+202E RLO, U+2066–U+2069 LRI/RLI/FSI/PDI. Verified: a tag prefixed with U+202E survives sanitisation and renders into a chip. RLM/LRM are legitimate and must survive, but overrides and isolates reorder *neighbouring* text, which is the "breaks the layout" class #504 said to keep rejecting. Suggested: `\p{Cf}` minus the range U+202A–U+202E and U+2066–U+2069.
2. **`suggestTags` can now surface unpublishable tags.** `free2z.ts:1519` runs the authoring autocomplete through the now-lenient read-side sanitize, so a 100-char / ZWJ / comma tag can be suggested and then refused by `validateArticleTag` on click. It also shows case-duplicates, since `selectedTags` is still lowercased. The new contract test asserts that as expected behaviour; I'd rather the authoring path used a stricter projection.
3. **Case-sensitive selected-chip matching in `Feed.tsx`** — `?tags=art` against a stored `ART` leaves the chip unhighlighted, and clicking yields `?tags=art,ART`. Cosmetic, since server matching is case-insensitive.
4. A small render test asserting a comma-bearing tag renders as a `<span>` and not a `<Link>` would kill the two surviving mutants.

Constraints clean: no `truncate`/`text-ellipsis`/`line-clamp`, no raw palette or hex, `ui-copy-truncation` passes.

PR #509, opened 2026-08-23
test(zuuli): audit populated creator catalog on mobile

### LINKED ISSUE #481: test(zuuli): mobile audits only visit an empty creator, so the catalog and pagination UI are never checked

skyl:
Closes #481

Supersedes #493, whose current head only adds a comment and does not change either mobile audit route matrix or assert populated catalog/pagination behavior.

## Summary

- preserve the empty `/creator/demo-creator` coverage
- add deterministic `/creator/zooko` coverage to both mobile viewport and touch-target audit matrices
- require the catalog grid, exactly 12 initial cards, the exact `12 of 13` count, and a visible `Load more pages` control before each generic audit proceeds
- share the route and assertions so the two suites cannot silently drift

## Verification

- `ZUULI_PW_PORT=1491 npx playwright test tests/viewport.pw.ts tests/touch-targets.pw.ts --workers=2` — 13 passed
- `npx vitest run` — 643 passed, 5 skipped
- `npm run typecheck`
- `npm run build`
- `git diff --check`

### DIFF
--- a/wallet/zuuli/tests/helpers/creator-catalog-audit.ts
+++ b/wallet/zuuli/tests/helpers/creator-catalog-audit.ts
@@ -1,0 +1,14 @@
+import { expect, type Page } from "@playwright/test";
+
+export const POPULATED_CREATOR_ROUTE = "/creator/zooko";
+
+export async function expectPopulatedCreatorCatalog(page: Page) {
+  const catalog = page.locator("[data-creator-pages]");
+
+  await expect(catalog.locator(".creator-pages-grid")).toBeVisible();
+  await expect(catalog.locator("[data-creator-page-card]")).toHaveCount(12);
+  await expect(catalog.getByText("12 of 13", { exact: true })).toBeVisible();
+  await expect(
+    catalog.getByRole("button", { name: "Load more pages" }),
+  ).toBeVisible();
+}
--- a/wallet/zuuli/tests/touch-targets.pw.ts
+++ b/wallet/zuuli/tests/touch-targets.pw.ts
@@ -1,4 +1,8 @@
 import { expect, test, type Page } from "@playwright/test";
+import {
+  expectPopulatedCreatorCatalog,
+  POPULATED_CREATOR_ROUTE,
+} from "./helpers/creator-catalog-audit";
 
 const TOUCH_WIDTHS = [320, 360] as const;
 const ROUTES = [
@@ -6,6 +10,7 @@ const ROUTES = [
   "/search",
   "/search?q=zcash",
   "/creator/demo-creator",
+  POPULATED_CREATOR_ROUTE,
   "/profile",
   "/kyc",
   "/wallet",
@@ -233,6 +238,10 @@ for (const signedIn of [false, true]) {
         await page.goto(route);
         await page.locator("[data-app-frame]").waitFor();
         await page.waitForTimeout(500);
+
+        if (route === POPULATED_CREATOR_ROUTE) {
+          await expectPopulatedCreatorCatalog(page);
+        }
 
         const audit = await auditTouchTargets(page);
         expect(audit.targets, `${route} must expose audited controls`).toBeGreaterThan(0);
--- a/wallet/zuuli/tests/viewport.pw.ts
+++ b/wallet/zuuli/tests/viewport.pw.ts
@@ -1,4 +1,8 @@
 import { expect, test, type Page } from "@playwright/test";
+import {
+  expectPopulatedCreatorCatalog,
+  POPULATED_CREATOR_ROUTE,
+} from "./helpers/creator-catalog-audit";
 
 const WIDTHS = [320, 360, 390, 414] as const;
 const ROUTES = [
@@ -6,6 +10,7 @@ const ROUTES = [
   "/search",
   "/search?q=zcash",
   "/creator/demo-creator",
+  POPULATED_CREATOR_ROUTE,
   "/profile",
   "/kyc",
   "/wallet",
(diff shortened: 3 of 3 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving** — and this is a better fix than the one I proposed in #481.

I suggested simply adding a populated creator to the ROUTES arrays. That would have worked, but it would have been silently fragile: if the fixture ever stopped producing a populated catalog, both audits would go on passing while auditing an empty page — the exact failure #481 is about, just relocated.

`expectPopulatedCreatorCatalog` closes that:

```ts
await expect(catalog.locator(".creator-pages-grid")).toBeVisible();
await expect(catalog.locator("[data-creator-page-card]")).toHaveCount(12);
await expect(catalog.getByText("12 of 13", { exact: true })).toBeVisible();
await expect(catalog.getByRole("button", { name: "Load more pages" })).toBeVisible();
```

Asserting the surface is actually rendered **before** running the audit means the coverage can't quietly evaporate. That's the difference between adding a route and guaranteeing the route still means something — and it's the right instinct given how much of tonight has been guards that stopped guarding without anyone noticing.

Sharing it as a helper across both suites is also right: `viewport.pw.ts` and `touch-targets.pw.ts` now cannot drift apart on what "populated" means.

### One fragility worth knowing

`toHaveCount(12)` and `"12 of 13"` are derived from fixture arithmetic — the 13 comes from `genArticles(63)` with `i % 6 === 0` giving 11 zooko articles plus 2 featured. Change `mockCreators`'s length or the featured authorship and these break.

That's acceptable, and arguably correct: it breaks **loudly**, at the assertion that names the number, rather than by silently auditing nothing. Worth a one-line comment above the helper noting where 13 comes from, so whoever hits it knows to recount rather than to soften the assertion.

### What this now covers that nothing did before

The creator page grid, the catalog count row, and the "Load more pages" control from #472 are now audited at 320/360 for touch targets and at 320/360/390/414 for overflow, signed in and out. Previously all three were invisible to both suites at every width, because the only creator route rendered an EmptyState.

Supersedes #493, which claimed to fix this issue with a one-line comment and no behavioural change.

skyl (commented):
CLEAN — independently reviewed exact head `f945551b88210682634fc768f6f1ae2dc6c8bc2e` adversarially.

The new route and guard close the actual #481 gap:

- `/creator/zooko` stays alongside the empty creator route in both mobile audit matrices, so the real populated catalog is audited signed in/out at 320/360 for touch targets and 320/360/390/414 for horizontal layout.
- The state is deterministic: 63 generated articles round-robin over six creators gives Zooko 11 generated pages, and the two featured Zooko pages make 13. `creatorPages()` is the real mock API path, returns the production-shaped first 12 plus `next = 2`, and the helper waits for 12 actual cards, `12 of 13`, and the visible Load-more button before either generic audit runs.
- The wait is race-safe: the initial grid visibility can match the skeleton, but `toHaveCount(12)`, the exact count text, and the button all auto-wait for catalog hydration. A request failure, empty/single-page fixture drift, or missing pagination control therefore fails loudly instead of silently auditing an EmptyState.
- Mutation sensitivity is adequate for the issue: removing/misrendering the grid/cards/count/pagination state fails the helper; shrinking or de-naming the card links/Button fails the real DOM touch scan; clipping or widening the count/grid/cards/Button outside the route boundary, or adding unintended horizontal scroll, fails the per-descendant viewport scan. These are the actual rendered elements, not synthetic stand-ins. Clicking page 2 is outside these layout-audit acceptance criteria and is already exercised by the dedicated creator pagination suite.

Local exact-head checks:

- `npm run typecheck` — pass
- `ZUULI_PW_PORT=15909 npx playwright test tests/touch-targets.pw.ts tests/viewport.pw.ts --workers=2` — 13/13 pass (1.7m)
- `git diff --check HEAD^ HEAD` — pass
- clean merge derivation against current `origin/main` (`b5e89ee7f72417da9945a063ff73d25907ab4a7d`)

Exact-head CI inspection: required frontend `gate` is green, along with frontend/Rust/iOS/Linux checks; Android and macOS packaging smoke jobs were still running when reviewed. No code or test blocker found.

Issue #505, opened 2026-08-23
ci(free2z): the ts/react/free2z test step is commented out, so that app's tests never run


skyl:
## Problem

`.github/workflows/ts-react-free2z.yml:25-27` — the `Run tests` step is **commented out**. CI for `ts/react/free2z` runs `CI=false npm run build` and nothing else.

So every test in that application is decorative by construction: it can be written, reviewed, merged, and silently rot, and no PR will ever go red because of it.

## How it surfaced

#503 (`fix(converse): prevent async embed scroll jumps`) added `ConverseList.test.tsx` — currently the **only** test file in that app. Mutation testing showed it passes even with the CSS import deleted entirely, which is its own problem (raised on the PR). But the deeper issue is that it wouldn't have mattered either way: the step that would run it doesn't execute.

The green `build` check on such a PR proves the app compiles. The green `gate` check comes from the unrelated `wallet/zuuli` workflow.

## Suggested fix

Uncomment the step, or replace it with whatever the app's test runner should be. If the tests are currently red and that's why it was disabled, say so in a comment with a tracking issue rather than leaving a silently-dead step — a commented-out test invocation reads as an oversight and invites people to add tests that will never run.

While there, consider whether the job should be a required context, or whether it should feed an aggregate gate the way `wallet/zuuli`'s jobs feed `gate`. Related: #480, which is the same "the check exists but isn't required" shape for the zuuallet schema freshness step.

## Acceptance

- A deliberately failing test in `ts/react/free2z` turns a PR red.
- No workflow contains a commented-out test invocation without an explanatory comment and a tracking issue.

Found while reviewing #503.

skyl:
Already fixed on `main` by #501 (`2e7b06fa83e5a893fe714dc953c79d2e2c83d160`). The `Run tests` step in `.github/workflows/ts-react-free2z.yml` is active and executes `npm test -- --watchAll=false`; `CustomImage.test.tsx` is present on trunk. Closing as completed by the earlier merged PR rather than opening a duplicate change.

PR #513, opened 2026-08-23
test(free2z): harden iOS 15 regex guard

skyl:
## Summary
- model `mdast-util-find-and-replace` rejected-match retries faithfully and pin email cases containing plus signs, punctuation, and rejected prefixes
- exercise the real bundle scanner API and CLI across multiple JavaScript files, ignored non-JavaScript files, clean bundles, and missing/empty directories
- run the guard from the standard test lifecycle and assert that production builds retain the post-build bundle scan

## Verification
- `npm run test:ios15-regex-compat`
- `CI=true npm test -- --watchAll=false` (2 suites, 3 tests)
- `CI=false npm run build` (164 JavaScript bundles scanned)
- `npx prettier --check scripts/ios15-regex-compat.test.js package.json`
- mutation probes reject removing `+` support, disabling bundle aggregation, or removing the build scanner hook

Refs #82

### DIFF
--- a/ts/react/free2z/package.json
+++ b/ts/react/free2z/package.json
@@ -91,9 +91,10 @@
   "proxy": "https://test.free2z.com",
   "scripts": {
     "start": "react-app-rewired start --host 0.0.0.0",
-    "prebuild": "node scripts/ios15-regex-compat.test.js",
+    "prebuild": "npm run test:ios15-regex-compat",
     "build": "react-app-rewired build && node scripts/check-ios15-bundle.js",
-    "test": "react-app-rewired test",
+    "test": "npm run test:ios15-regex-compat && react-app-rewired test",
+    "test:ios15-regex-compat": "node scripts/ios15-regex-compat.test.js",
     "eject": "react-scripts eject",
     "prettier": "prettier -c './**/*.{tsx,js,jsx,ts,json}'",
     "prettier:format": "prettier --write -c './**/*.{tsx, js, jsx, ts, json}'"
--- a/ts/react/free2z/scripts/ios15-regex-compat.test.js
+++ b/ts/react/free2z/scripts/ios15-regex-compat.test.js
@@ -1,6 +1,14 @@
 const assert = require("node:assert/strict");
+const childProcess = require("node:child_process");
+const fs = require("node:fs");
+const os = require("node:os");
+const path = require("node:path");
 const loader = require("./ios15-regex-compat-loader");
-const { findLookbehindRegexLiterals } = require("./check-ios15-bundle");
+const {
+  checkBundle,
+  findLookbehindRegexLiterals,
+  javascriptFiles,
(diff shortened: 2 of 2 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving** — merging once `gate` is green. This closes the follow-up from #511 properly, and I verified it rather than taking the description's word.

### The fixtures now actually exercise the retry path

I ran the rewritten pattern from the loader against both harnesses — the old `matchAll`-based `guardedMatches` and the new `exec`+retry one:

```
rewritten literal: /([-.\w+]+)@([-\w]+(?:\.[-\w]+)+)/gu

é.foo@example.com        matchAll=[]   exec+retry=["foo@example.com"]     DIVERGE
é-person@example.com     matchAll=[]   exec+retry=["person@example.com"]  DIVERGE
é+person@example.com     matchAll=[]   exec+retry=["person@example.com"]  DIVERGE
éperson@example.com      matchAll=[]   exec+retry=[]                      same
```

**Three of the four new fixtures fail against the old harness.** That's what makes this a real fix: the previous test's fixture list happened to avoid every input where its own model diverged from shipped behaviour, so it agreed with reality by luck. Now it agrees by construction, and if the model drifts again the fixtures will say so.

The retry itself is modelled correctly — `pattern.lastIndex = match.index + 1` on rejection, matching what `mdast-util-find-and-replace` does — and the comment explains *why*, which is the part someone would otherwise "simplify" away later.

### The scanner is now tested end-to-end

`runScanner` spawns the real `check-ios15-bundle.js` as a subprocess against temp directories covering the failing, clean, missing and empty cases. That's a genuine improvement over asserting on the exported helper — the previous version could have passed with the CLI wiring broken.

### And the guard now actually runs under `npm test`

```diff
-    "test": "react-app-rewired test",
+    "test": "npm run test:ios15-regex-compat && react-app-rewired test",
+    "test:ios15-regex-compat": "node scripts/ios15-regex-compat.test.js",
```

Previously it executed only via `prebuild`, which is why #511's body reported "2 suites, 3 tests passed" — those were the pre-existing `src` tests, not this. Now it runs in both paths, and extracting the named script means `prebuild` and `test` can't drift apart.

### Still worth knowing (unchanged by this PR)

The `ts-react-free2z` job is **not** a required context — branch protection requires exactly `["gate"]`, which comes from `zuuli.yml`. So a red iOS-15 scanner still won't block a merge. That's a pre-existing repo condition rather than anything this PR introduced, and it's the same shape as #480 and #505. Worth closing eventually, or this guard remains advisory no matter how good it gets.

I could not run the full suite standalone — the app's `node_modules` isn't installed here, and `check-ios15-bundle.js` needs `acorn` — so the divergence probe above is direct evidence for the semantics fix, not for the whole file. CI's `build` job runs it via `prebuild`; confirm that's green before merging.

PR #514, opened 2026-08-23
fix(android): keep ChromeOS devices installable

### LINKED ISSUE #302: ZUULI Android: Chromebooks get "your device is not compatible with this version" on Play
## Problem

Installing ZUULI on a Chromebook from Google Play fails with:

> **Your device is not compatible with this version.**

That message is Play's **device-catalog filter**, not a runtime crash — Play has
decided the device is excluded from the published release before download
begins. ChromeOS runs Android apps under ARC, and it is one of the most common
targets to accidentally filter out.

## Likely causes, most probable first

**1. Implied camera feature requirements (strongest suspect).**
`gen/android/app/src/main/AndroidManifest.xml` declares
`<uses-permission android:name="android.permission.CAMERA" />`. Play implicitly
converts that into **required** `android.hardware.camera` *and*
`android.hardware.camera.autofocus` feature requirements unless each is
explicitly declared `required="false"`. The manifest currently declares only:

```xml
<uses-feature android:name="android.hardware.camera.any" android:required="false" />
<uses-feature android:name="android.hardware.microphone" android:required="false" />
```

`camera.any` does **not** cancel the implied `android.hardware.camera` /
`android.hardware.camera.autofocus` requirements — those are separate feature
names. Most Chromebooks have no rear autofocus camera, so they get filtered out.
The `microphone` declaration is correct and is the pattern the camera entries
should follow.

**2. Implicit touchscreen requirement.**
Android implies `android.hardware.touchscreen` with `required="true"` when it is
not declared. Non-touch Chromebooks (a large share of the fleet) are excluded by
this unless we explicitly declare it optional. ChromeOS specifically wants
`android.hardware.touchscreen` marked `required="false"`.

**3. Missing x86_64 native code in the published bundle.**
Many Chromebooks are x86_64. `zuuli-release.yml` builds all four ABIs
(`aarch64, armv7, i686, x86_64`), but `zuuli-packaging.yml:153` builds an
**aarch64-only** AAB on the PR path. If the artifact that reached the Play track
came from an aarch64-only build, every x86_64 Chromebook is incompatible by
definition. Needs verification against what is actually live, not against what
the workflow *can* produce.

**4. minSdk floor.**
`app/build.gradle.kts` sets `minSdk = 29` (Android 10). Older Chromebooks still
run ARC++ on Android 9 (API 28) and would be excluded. This one may be a
deliberate trade-off rather than a bug — but it should be a decision, not an
accident.

Causes 1, 2 and 4 are all visible in the Play Console under **Release → Reach
and devices → Device catalog**, which lists excluded devices *with the reason*.
That is the fastest way to confirm which of these is actually biting.

## Suggested fix

Add the explicit opt-outs to `AndroidManifest.xml` alongside the existing ones:

```xml
<uses-feature android:name="android.hardware.camera" android:required="false" />
<uses-feature android:name="android.hardware.camera.autofocus" android:required="false" />
<uses-feature android:name="android.hardware.touchscreen" android:required="false" />
<uses-feature android:name="android.hardware.faketouch" android:required="false" />
```

These are correct on their own merits, not just for ChromeOS: ZUULI already
treats camera and mic as request-at-point-of-use (see the comment already in the
manifest for livestream viewers), so no hardware should be a hard install
requirement. Tablets and camera-less devices benefit identically.

Then confirm the published bundle actually carries `x86_64` native libs.

## Investigation steps

1. Play Console → Release → **Device catalog**, filter to excluded devices,
   search for the reporter's Chromebook model — record the stated exclusion reason.
2. Play Console → **App bundle explorer** → the live version → **Native
   platforms**: confirm `arm64-v8a, armeabi-v7a, x86, x86_64` are all listed.
3. Confirm which workflow produced the live artifact (universal release build vs
   the aarch64-only packaging path).

## Acceptance criteria

- [ ] Root cause identified from the Play Console device catalog and stated in the PR.
- [ ] `AndroidManifest.xml` declares camera, camera.autofocus, touchscreen and
      faketouch as `required="false"`.
- [ ] The published AAB is verified to contain all four ABIs including `x86_64`.
- [ ] Play Console device catalog shows ChromeOS devices as **supported** for the
      new release.
- [ ] ZUULI installs and launches on the reporting Chromebook, verified end to end.
- [ ] If `minSdk = 29` is what excludes the device, that is called out explicitly
      and kept-or-lowered as a deliberate decision rather than left implicit.

## Needed from the reporter

Chromebook **model** and **ChromeOS version** (`chrome://version`), plus whether
the device has a touchscreen. That pins the cause immediately instead of us
guessing between the four above.

skyl:
Refs #302

## What

- explicitly declare camera, autofocus, camera.any, microphone, touchscreen, and faketouch hardware optional in the generated Android manifest
- retain camera permission for livestream hosts while preventing permission-implied Play device filters
- add a regression contract to the standard ZUULI test suite that requires each optional feature exactly once

## Why

Google Play can infer required camera/autofocus hardware from CAMERA permission, and ChromeOS also needs touchscreen declared optional for non-touch devices. These declarations are correct independently of the reporter hardware because ZUULI requests capture access only at point of use.

## Verification

- node --test scripts/android-device-catalog.node-test.mjs
- npm run typecheck
- npm run build
- git diff --check

This intentionally references rather than closes #302. Play Console catalog, published AAB ABI inventory, and a physical Chromebook install still require post-merge release/device verification.

### DIFF
--- a/wallet/zuuli/package.json
+++ b/wallet/zuuli/package.json
@@ -8,7 +8,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "test": "vitest run && node --test scripts/safe-area-contract.node-test.mjs scripts/auth-session-boundary.node-test.mjs scripts/mermaid-security.node-test.mjs scripts/send-review-boundary.node-test.mjs scripts/ui-copy-truncation.node-test.mjs scripts/fixture-privacy.node-test.mjs scripts/apple-credential-boundary.node-test.mjs scripts/release-tag-identity.node-test.mjs scripts/status-freshness.node-test.mjs && node scripts/apple-credential-boundary.mjs && playwright test",
+    "test": "vitest run && node --test scripts/safe-area-contract.node-test.mjs scripts/android-device-catalog.node-test.mjs scripts/auth-session-boundary.node-test.mjs scripts/mermaid-security.node-test.mjs scripts/send-review-boundary.node-test.mjs scripts/ui-copy-truncation.node-test.mjs scripts/fixture-privacy.node-test.mjs scripts/apple-credential-boundary.node-test.mjs scripts/release-tag-identity.node-test.mjs scripts/status-freshness.node-test.mjs && node scripts/apple-credential-boundary.mjs && playwright test",
     "test:mermaid-security": "node scripts/check-mermaid-security.mjs && node --test scripts/mermaid-security.node-test.mjs && playwright test tests/mermaid-security.pw.ts",
     "test:send-review": "node --test scripts/send-review-boundary.node-test.mjs && playwright test tests/send-review.pw.ts",
     "test:apple-credential-boundary": "node --test scripts/apple-credential-boundary.node-test.mjs && node scripts/apple-credential-boundary.mjs",
--- a/wallet/zuuli/scripts/android-device-catalog.node-test.mjs
+++ b/wallet/zuuli/scripts/android-device-catalog.node-test.mjs
@@ -1,0 +1,51 @@
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import test from "node:test";
+
+const manifest = readFileSync(
+  new URL(
+    "../src-tauri/gen/android/app/src/main/AndroidManifest.xml",
+    import.meta.url,
+  ),
+  "utf8",
+);
+
+function featureDeclarations(name) {
+  const escaped = name.replaceAll(".", "\\.");
+  return manifest.match(
+    new RegExp(
+      `<uses-feature\\s+android:name=["']${escaped}["'][^>]*\\/>`,
+      "g",
+    ),
+  ) ?? [];
+}
+
+test("capture and touch hardware do not filter devices in Google Play", () => {
+  const optionalFeatures = [
+    "android.hardware.camera",
+    "android.hardware.camera.autofocus",
+    "android.hardware.camera.any",
+    "android.hardware.microphone",
+    "android.hardware.touchscreen",
+    "android.hardware.faketouch",
+  ];
+
+  for (const name of optionalFeatures) {
+    const declarations = featureDeclarations(name);
+    assert.equal(declarations.length, 1, `${name} must be declared exactly once`);
+    assert.match(
+      declarations[0],
+      /android:required=["']false["']/,
+      `${name} must remain optional`,
+    );
+  }
+});
+
+test("camera permission remains available for hosts without implying hardware", () => {
+  assert.equal(
+    manifest.match(
+      /<uses-permission\s+android:name=["']android\.permission\.CAMERA["']\s*\/>/g,
+    )?.length,
+    1,
+  );
+});
--- a/wallet/zuuli/src-tauri/gen/android/app/src/main/AndroidManifest.xml
+++ b/wallet/zuuli/src-tauri/gen/android/app/src/main/AndroidManifest.xml
@@ -5,10 +5,16 @@
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
 
-    <!-- Livestream viewers can run without capture hardware. Hosts request
-         camera and microphone access at the point of use. -->
+    <!-- Livestream viewers can run without capture or touch hardware. Hosts
+         request camera and microphone access at the point of use. Explicitly
+         override Android's permission-implied camera and touchscreen filters
+         so Google Play keeps camera-less and non-touch ChromeOS devices. -->
+    <uses-feature android:name="android.hardware.camera" android:required="false" />
+    <uses-feature android:name="android.hardware.camera.autofocus" android:required="false" />
     <uses-feature android:name="android.hardware.camera.any" android:required="false" />
     <uses-feature android:name="android.hardware.microphone" android:required="false" />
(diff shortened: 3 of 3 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving in substance — merging once `Android / unsigned AAB` and `gate` are both green.** That specific job matters more than usual here; see below.

### The change is well-targeted

Six `uses-feature` declarations, all `required="false"`, overriding Android's permission-implied filters. The `RECORD_AUDIO` permission implicitly requires `android.hardware.microphone`, and camera permissions imply camera hardware, so Play was filtering out camera-less and non-touch ChromeOS devices — which is exactly #302's symptom. Adding `touchscreen` and `faketouch` as optional is the standard remedy.

The comment explains *why* the declarations exist, which matters for a file people are tempted to tidy.

### The guard can actually fail, and reads the real file

`android-device-catalog.node-test.mjs` reads the **committed manifest** rather than a fixture, and asserts each feature appears exactly once with `required="false"`. Delete a declaration and `declarations.length` is 0; flip `required` and the regex misses. Both fail. That's a real check — and reading the real artifact is what stops it becoming another fixture-shaped assertion.

Wiring it into `npm test` (the `package.json` +1/-1) puts it on the required `gate`.

### Why I want the Android job green first

This hand-edits `src-tauri/gen/android/app/src/main/AndroidManifest.xml`, a **generated** file. Hand-editing it is precisely what broke the `0.1.0+11` release: #406 added an `<intent-filter>` outside the deep-link marker block, and a real build regenerated a duplicate, failing the release's `git diff --exit-code`.

The reasoning here says this should be safe — the `uses-feature` elements sit before `<application>`, well outside the marker block inside `<activity>`, and `insert_into_xml` only rewrites content *between* its markers, preserving everything outside. I verified the invariants hold on your head:

```
DEEP LINK PLUGIN markers: 2
android:host="oauth": 1   android:host="checkout": 1
```

But reasoning is what we had last time. Since #473, the PR-level Android job runs `tauri android build` followed by the normalizer and `git diff --exit-code`, so it will *empirically* confirm the plugin doesn't disturb these lines. That check is the whole reason it exists — worth waiting the few minutes for it rather than merging on the manifest reading alone.

### One thing worth confirming separately

Play's device catalogue is the actual acceptance criterion for #302, and no CI check can read it. Once a build carrying this reaches internal test, it's worth confirming in Play Console that ChromeOS devices are no longer excluded — otherwise this closes on a plausible mechanism rather than an observed outcome. If the PR says `Closes #302`, I'd rather it said `Refs` until that's seen.

skyl:
Exact-head adversarial review: **CLEAN** at `df510352d2ed552cef140620bdb5a889a9d6b6af`.

### Device-catalog semantics and final artifact

The declarations match Android's current Play filtering rules:

- `CAMERA` permission implies both `android.hardware.camera` and `android.hardware.camera.autofocus`; explicitly declaring each with `required="false"` overrides those inferred requirements.
- `camera.any` optional keeps any available camera usable without making one an install prerequisite.
- Android apps default to touch/faketouch requirements; explicitly optional `touchscreen` admits non-touch Chromebooks, and optional `faketouch` deliberately removes the remaining basic-touch filter.
- `RECORD_AUDIO` remains paired with optional microphone hardware.

I downloaded this exact run's credential-free Android AAB and decoded `base/manifest/AndroidManifest.xml` from its protobuf form. After the real Gradle/library manifest merge, all six declarations—camera, autofocus, camera.any, microphone, touchscreen, and faketouch—are present exactly once and each has compiled boolean `required=false`. There is no hidden dependency declaration reintroducing a required hardware filter.

### Manifest durability and regression guard

The committed XML is well-formed and places all `<uses-feature>` elements as direct `<manifest>` children before `<application>`. The test reads that committed manifest rather than a fixture, requires each feature exactly once and optional, and separately retains exactly one CAMERA permission. Deletion, `required="true"`, duplicate-declaration, and permission-removal mutation probes all go red.

The new node test is part of the standard `npm test` command. `wallet/zuuli/**` selects the full ZUULI suite, and required `gate` waits for `frontend`, so the contract is not advisory. More importantly for this generated file, exact-head `Android / unsigned AAB` ran the real Tauri Android build and then passed `normalize-generated-android-manifest.mjs` plus `git diff --exit-code`; regeneration preserved the declarations byte-for-byte.

### Scope and remaining #302 acceptance

This is a safe public-code partial and correctly says `Refs #302`, not `Closes`. The PR-level AAB intentionally carries only the PR packaging ABI, so it does not prove the published release has all four ABIs. It also cannot prove the Play Console exclusion reason/catalog result, the reporter Chromebook's install/launch, or whether API 29 is the model's remaining filter. Those acceptance items properly remain open on #302 for the next release/device pass.

Local checks passed: manifest XML validation, node contract, mutation probes, typecheck, production frontend build, and diff check. Exact-head required `gate`, Android AAB/reproducibility, frontend, Rust, iOS, Linux, and macOS packaging checks are all green. Scope is limited to the generated Android manifest, its focused guard, and test wiring; no runtime, backend, credential, or private-repository changes.

skyl:
`Android / unsigned AAB` is green, which is the check that matters here — it runs a real `tauri android build` followed by the normalizer and `git diff --exit-code`, so the plugin demonstrably leaves these `uses-feature` lines alone. That's the empirical confirmation I wanted rather than reasoning from where the elements sit in the file, given #406 taught us that lesson the expensive way. Merging.

PR #515, opened 2026-08-23
ci(zuuli): reject soft-failing required jobs

### LINKED ISSUE #512: security(ci): job-level continue-on-error silently neuters any gate-feeding job
## Problem

Adding `continue-on-error: true` to any **job** in `.github/workflows/zuuli.yml` silently neuters that job's contribution to the required `gate`, while every policy check and self-test stays green.

The `gate` job decides pass/fail by inspecting `needs.<job>.result`. GitHub sets that to `success` for a job that failed but carried `continue-on-error: true`. So the compile can fail, the gate passes, and nothing anywhere goes red.

## Measured

Found while reviewing #510, which adds a policy script (`scripts/check-zuuli-android-gate.mjs`) that pins its job into the gate. That script correctly rejects eleven escape routes — retargeting the check, dropping `--locked`, removing the job from `gate.needs`, removing its result from the inspected `results=` string, hardcoding the result to `success`, `if: false` at job *or* step level, `continue-on-error` on the **step**, dropping the live policy invocation, and removing the policy file from the path filter. All red.

The twelfth escapes:

```yaml
  rust_android_32:
    continue-on-error: true     # job level, not step level
```

→ policy check green, `gate` green, failing compile ignored.

## Why it matters beyond #510

This is not specific to the Android job. **Every job feeding `gate` is exposed the same way**, including the Rust builds, the frontend suite that carries the truncation/fixture-privacy/locale contracts, and the release-identity check. One line in a workflow disables any of them without turning anything red.

Same family as the guards already fixed in #477, #484 and #494: a real control with a bypass nothing asserts against.

## Suggested fix

Add a workflow-policy assertion — the natural home is the existing gate-policy script, or `scripts/check-github-actions-pins.mjs` which already parses these workflows — that rejects `continue-on-error` on any job listed in `gate.needs`. Step-level `continue-on-error` is sometimes legitimate (an upload that may be absent), so scoping the ban to job level is probably right; if any job genuinely needs it, allow it via an explicit named allowlist so the exception is visible.

Worth also asserting that every job in `gate.needs` appears in the inspected `results=` string, so a job can't be added to `needs` without being evaluated.

## Acceptance

- Adding `continue-on-error: true` to a gate-feeding job fails CI.
- The check is mutation-proven: introduce it, watch a policy check go red.

Found while reviewing #510.

skyl:
## Summary

- extend the always-on GitHub Actions policy scan to reject job-level `continue-on-error` on every direct `gate.needs` dependency and on `gate` itself
- fail closed on alternate YAML spellings that could hide the property (quoted/escaped keys, explicit/decorated keys, aliases/merges, inline job maps, and noncanonical direct-job indentation) while preserving legitimate step-level and non-gate advisory use
- require every direct gate dependency to expose `needs.<job>.result` and consume its environment variable in a gate run step
- re-run the repository policy inside the final gate after checkout, so job-level `continue-on-error` on the initial `changes` policy job cannot mask the earlier failure

## Verification

- `node scripts/check-github-actions-pins.mjs --self-test` (40 policy cases)
- real-workflow mutation proof: adding `continue-on-error: true` to `rust_app` fails with `job-level continue-on-error is forbidden on required-gate job rust_app`

Closes #512

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -39,9 +39,10 @@ jobs:
           fetch-depth: 0
 
       # This runs before change detection on every event. A workflow cannot
-      # replace its own judge with a mutable external action and then skip the
-      # full application suite as an allegedly unrelated change.
-      - name: Verify every external action has an immutable revision
+      # replace its own judge with a mutable external action, or mark a job the
+      # required gate awaits as best-effort, then skip the full application
+      # suite as an allegedly unrelated change.
+      - name: Verify immutable actions and fail-closed required jobs
         run: |
           node scripts/check-github-actions-pins.mjs --self-test
           node scripts/check-github-actions-pins.mjs
@@ -544,6 +545,13 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
+      # Re-run the workflow policy here so even job-level continue-on-error on
+      # `changes` cannot turn that job's deliberately failing policy verdict
+      # into the `success` conclusion this gate consumes.
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Recheck immutable actions and fail-closed required jobs
+        run: node scripts/check-github-actions-pins.mjs
+
       - name: Verify required jobs succeeded or legitimately skipped
         env:
           CHANGES_RESULT: ${{ needs.changes.result }}
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -8,6 +8,10 @@ import { fileURLToPath } from "node:url";
 const FULL_COMMIT_SHA = /^[0-9a-f]{40}$/;
 const EXTERNAL_USES = /^[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+(?:\/[^@\s]+)?@([^@\s]+)$/;
 
+// Required-gate policy deliberately accepts a small, canonical YAML subset.
+// Alternate keys, inline job maps, aliases, and decorators fail closed instead
+// of giving a second spelling to controls this checker must recognize.
+
 function yamlFilesBelow(directory) {
   if (!fs.existsSync(directory)) return [];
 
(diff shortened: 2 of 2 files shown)


skyl:
BLOCKED at exact head `561532e7ecd241d66d71913d8b3454902f50a2bb`.

The canonical job/property parsing is strong: I found no escape through quoted/escaped keys, expressions, aliases/merge keys, inline job maps, duplicate direct properties, or noncanonical direct-job indentation. Current-tree `node --check`, all 40 self-tests, repository scan, `actionlint`, and diff checks pass.

Two fail-open paths remain:

1. **A dependency result only has to be mentioned, not enforced.** On this exact tree I removed `$RUST_DENY_RESULT` from the `results=` aggregate and added:

```sh
echo "Rust deny result (advisory): $RUST_DENY_RESULT"
```

Both `node scripts/check-github-actions-pins.mjs` and `actionlint .github/workflows/zuuli.yml` exited 0. A failing `rust_deny` is then logged but omitted from the comparison loop, so the required gate succeeds. Moving the env/result reference into a step guarded by `if: github.event_name == '__never__'` also passes both checks; `gateResultInputs` aggregates env/run text across all steps without proving the consuming step runs or contributes to the verdict.

2. **The gate-local policy recheck is not itself bound.** Adding `continue-on-error: true` to `Recheck immutable actions and fail-closed required jobs` passes both the repository policy and actionlint. Deleting the checkout/recheck steps entirely also passes both. Therefore a workflow edit can soften, skip, retarget, or remove the second judge that is supposed to defeat job-level `continue-on-error` on the first judge.

Please bind the exact live gate verdict path, not arbitrary textual consumption: each direct `gate.needs` result must participate in the comparison that can fail the gate, dead/skipped/decorative uses must fail mutation tests, and the final checkout + policy invocation must be required, current-source, unconditional, and unable to use step-level `continue-on-error`. Add real-workflow mutations for the log-only result, dead-step env/result, soft-failed recheck, and removed recheck cases.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — `gate` is green and the fix is mutation-proven. Residuals below are filed, not blocking.

### It closes the hole, demonstrably

Same mutation — `continue-on-error: true` added to `rust_app` in the real `zuuli.yml`:

```
PR script:            exit 1  "job-level continue-on-error is forbidden on required-gate job rust_app"
origin/main script:   exit 0
```

That's the before/after that matters. Both pass clean on the unmutated tree.

### The key-based design is the right call

Forbidding the **property** regardless of its value kills the entire family in one stroke rather than playing whack-a-mole with spellings:

```
True / TRUE / yes / on              caught
${{ true }}                          caught
${{ github.event_name == 'push' }}   caught
false (deliberate)                   caught
"continue-on-error": true (quoted)   caught
CRLF, trailing comments              caught
```

Everything the restricted parser can't model — flow mappings, merge keys, anchored keys, explicit `?` keys, tab indentation, non-canonical indent — **fails closed**, which is the correct default for a policy file.

And `gate.needs` completeness came out stronger than #512 asked for: each dependency must be exposed as `X: ${{ needs.<dep>.result }}` *and* referenced in a gate `run` block. Adding a dep to `needs` only, or to `needs` + env but unconsumed, or deleting `$RUST_APP_RESULT` from `results=` — all three caught.

### The self-test is load-bearing

This is the part I most expected to be theatre, given #510. It isn't. Weakening the **real** check seven ways broke the suite six times: dropping `"gate"` from the forbidden list, only banning the literal `"true"`, skipping the consumption check, skipping result-exposure, allowing merge keys, dropping the property-indent rule, allowing inline job maps. Good.

The one survivor: **allowing tabs** passes the self-test. The tab guard is genuinely load-bearing — a tab-indented `continue-on-error` computes indent 1 and would never be seen as a job property — it just has no fixture. Worth one.

### Residuals worth filing

1. **Reusable-workflow callee is a real escape.** Convert a gate dependency to `uses: ./.github/workflows/x.yml` and put the soft-fail in the callee: **exit 0**. `gatePolicyFailures` does run on the callee file, but returns early because it has no `gate` job and isn't `zuuli.yml`. At runtime a soft-failed job in the callee makes the callee run succeed, so the caller's `result` is `success` — exactly #512's mechanism, one file removed. Two files instead of one is more visible in review, but it's the same hole.
2. **The policy is its own only judge.** `changes` and `gate` both enforce it, and both are policed by it. `continue-on-error` on `changes` alone is caught by the gate recheck — that's precisely what the new recheck buys and it works. But `continue-on-error` on **`gate` itself** softens gate's own conclusion, and the only red signal is `changes`, which is **not a required context**. Worth either a second required context or asserting `gate`'s conclusion can't be softened.
3. **Nothing asserts the policy actually runs.** Replacing both `run: node scripts/check-github-actions-pins.mjs` lines with `run: true` and adding job-level `continue-on-error` goes entirely undetected, because the script never executes. Pre-existing, but this PR now leans on the gate recheck as a defence without pinning that defence. #510's script self-pins its own invocation; this one could do the same.
4. **"Consumed" is a mention check, not an assertion check.** Adding a dep, exposing `EXTRA_RESULT`, and writing `echo "ignored: $EXTRA_RESULT"` passes clean while the result is never compared. Same weakness my own suggestion in #512 had, so not a regression — just the obvious next hole.
5. **False-positive surface on other workflows.** The structural rules apply to *every* workflow file, not only `zuuli.yml`, and they reject valid GitHub Actions YAML: `steps:` sequence items at indent 4 (the style prettier/yamlfmt emit) give `job build property must use an undecorated block mapping key`, and 3-space job indentation is rejected. All fail closed, so nothing unsafe — but a formatter run over any workflow would break CI with a message that doesn't explain itself. Recommend scoping the structural strictness to the required workflow.

### On the 548 lines

The zero-dependency constraint is legitimate — this runs as the first step of `changes`, before any `setup-node` or install, and the YAML libraries in this repo live in app-level `package.json` files that aren't installed at that point. Hand-rolling is defensible. A real parser would *resolve* aliases and merges rather than fail closed on them, which would be strictly better, but it isn't reachable there.

No crash or false positive on the current tree: all 13 workflows plus `.github/actions` pass, `actionlint` clean, `node --check` clean. No bare `[[ ]]` anywhere in the diff.

Issue #516, opened 2026-08-23
security(ci): the gate soft-fail ban misses reusable-workflow callees, and the policy is its own only judge


skyl:
## Problem

#515 closed the job-level `continue-on-error` bypass from #512 — mutation-proven, and it correctly forbids the property regardless of value, so `True`/`yes`/`on`/`${{ … }}` all get caught. One escape route remains.

**Convert a gate dependency to a local reusable workflow and put the soft-fail in the callee.**

```yaml
# .github/workflows/zuuli.yml
jobs:
  rust_app:
    uses: ./.github/workflows/x.yml        # gate dependency, no continue-on-error here

# .github/workflows/x.yml
jobs:
  build:
    continue-on-error: true                # never inspected
```

Measured: `node scripts/check-github-actions-pins.mjs` → **exit 0**.

`gatePolicyFailures` does run over the callee file, but returns early because that file has no `gate` job and isn't `zuuli.yml`, so job-level `continue-on-error` there is never inspected. At runtime a soft-failed job inside the callee makes the *callee run* succeed, so the caller job's `result` is `success` — precisely #512's mechanism, one file removed.

Two files instead of one is more visible in review, but the guard is meant to make this mechanical rather than reliant on someone noticing.

## Suggested fix

When a gate dependency's body is `uses: ./.github/workflows/<file>`, recurse the job-level `continue-on-error` ban into that file. And forbid *external* reusable workflows as gate dependencies outright — their contents aren't reviewable from this repo at all.

## Three smaller residuals from the same review

1. **The policy is its own only judge.** `changes` and `gate` both enforce it and both are policed by it. `continue-on-error` on `changes` alone is caught by the new gate recheck — that part works. But `continue-on-error` on **`gate` itself** softens gate's own conclusion, and the only remaining red signal is `changes`, which is **not a required context**. Either add a second required context, or assert that `gate`'s own conclusion cannot be softened.
2. **Nothing asserts the policy actually runs.** Replacing both `run: node scripts/check-github-actions-pins.mjs` lines with `run: true` and then adding a job-level `continue-on-error` goes entirely undetected, because the script never executes. #510's policy script self-pins its own invocation; this one could do the same.
3. **"Consumed" is a mention check, not an assertion check.** Adding a dep to `needs`, exposing `EXTRA_RESULT`, and writing `echo "ignored: $EXTRA_RESULT"` passes clean while the result is never compared. (Same weakness my own suggestion in #512 had — not a regression, just the next hole.)

[...]

PR #517, opened 2026-08-23
fix(ci): bind required gate verdict path

### LINKED ISSUE #512: security(ci): job-level continue-on-error silently neuters any gate-feeding job

skyl:
Fixes forward after #515 by making the required gate enforce structure and runtime results, rather than merely finding textual mentions.

- requires an exact unconditional gate-local checkout and current-source policy recheck
- binds the final enforcing verdict to the policy step outcome and the complete `toJSON(needs)` context
- preserves the independent `changes` / `zuuallet_schema` selector semantics and automatically covers future gate dependencies such as `rust_android_32`
- adds current-workflow mutations for log-only/dead verdicts, deleted or skipped checkout/recheck, soft-failing recheck, and soft-failing required jobs

Verification:
- `node scripts/check-github-actions-pins.mjs --self-test` (46 source-policy, 6 gate-verdict, 8 real-workflow mutation cases)
- `git diff --check`

Closes #512

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -550,47 +550,13 @@ jobs:
       # into the `success` conclusion this gate consumes.
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Recheck immutable actions and fail-closed required jobs
+        id: policy
         run: node scripts/check-github-actions-pins.mjs
 
       - name: Verify required jobs succeeded or legitimately skipped
         env:
-          CHANGES_RESULT: ${{ needs.changes.result }}
-          ZUULI_CHANGED: ${{ needs.changes.outputs.zuuli }}
-          ZUUALLET_SCHEMA_CHANGED: ${{ needs.changes.outputs.zuuallet_schema }}
-          FRONTEND_RESULT: ${{ needs.frontend.result }}
-          RUST_FMT_RESULT: ${{ needs.rust_fmt.result }}
-          RUST_DENY_RESULT: ${{ needs.rust_deny.result }}
-          RUST_CLIPPY_RESULT: ${{ needs.rust_clippy.result }}
-          RUST_PLUGIN_RESULT: ${{ needs.rust_plugin.result }}
-          RUST_APP_RESULT: ${{ needs.rust_app.result }}
-          ZUUALLET_SCHEMA_RESULT: ${{ needs.zuuallet_schema.result }}
+          POLICY_OUTCOME: ${{ steps.policy.outcome }}
+          REQUIRED_JOBS_JSON: ${{ toJSON(needs) }}
         run: |
-          set -euo pipefail
-          [ "$CHANGES_RESULT" = "success" ] || {
-            echo "Change detection did not pass: $CHANGES_RESULT"; exit 1;
-          }
-
-          # Every job in `needs:` above must appear here as well; a job that is
-          # awaited but not inspected makes the gate decorative for that job.
-          results="$FRONTEND_RESULT $RUST_FMT_RESULT $RUST_DENY_RESULT $RUST_CLIPPY_RESULT $RUST_PLUGIN_RESULT $RUST_APP_RESULT"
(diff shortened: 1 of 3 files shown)


skyl:
CLEAN — adversarial review of exact head `4440ef969a19fd293007e93f014bd03c24479467`.

I reproduced the #515 negative controls against the pre-fix base (`298aba011c38bf5c78293e723785fb551068dd24`) and this head. The old policy accepted all nine bypasses; this head rejects all nine: log-only result use, result/env use confined to a dead step, removed checkout, `if: false` checkout, checkout `continue-on-error`, removed policy recheck, `if: false` recheck, recheck `continue-on-error`, and an overwritten shell verdict variable.

The enforcing path is now structurally and dynamically bound:

- the gate has the exact unconditional pinned checkout, exact current-source policy step, and exact enforcing verdict step;
- the verdict binds `POLICY_OUTCOME` to `steps.policy.outcome` and `REQUIRED_JOBS_JSON` to the complete `toJSON(needs)` context, reruns the source policy, then verifies results;
- direct verifier probes passed for a legitimate independent schema selection and rejected schema-result/selector mismatches plus a missing schema selector;
- `check-zuuli-linux-image.mjs` now enforces the same complete-needs/verifier contract and its self-test rejects replacing the verifier with a log-only command.

Hosted run [32628621839](https://github.com/free2z/zuu/actions/runs/32628621839) is fully green. Its gate log proves the current-source policy recheck ran successfully, the verdict received `POLICY_OUTCOME=success` and the full `needs` object (including `zuuallet_schema=true` with a successful schema job), and `--verify-gate-results` reported every required dependency successful. All packaging smoke and Linux image checks are green too.

Local review checks passed: Node syntax check, the 46 source-policy / 6 verdict / 8 current-workflow mutation self-tests, current policy scan, Linux-image policy self-tests/current scan, `actionlint`, and `git diff --check`.

Scope is limited to the workflow and its two CI policy scripts. The PR closes only #512 and does not claim to solve #516's reusable-callee soft-fail/self-judge residuals.

PR #518, opened 2026-08-23
fix(ci): inspect reusable gate workflows

### LINKED ISSUE #516: security(ci): the gate soft-fail ban misses reusable-workflow callees, and the policy is its own only judge

skyl:
## Summary

- scope the gate-only canonical YAML policy to `zuuli.yml` and transitive local reusable workflows that feed its required gate
- recursively forbid job-level `continue-on-error` in those callees and reject external reusable workflows anywhere in the required chain
- pin exact unconditional self-test + live-policy invocations in both `changes` and the required gate
- add negative fixtures for direct/nested callee soft-fail, external callees, replaced/dead/soft-failing policy invocation, gate soft-fail, and tab indentation
- prove ordinary non-required workflows retain valid formatter-style indentation flexibility

## Trust boundary

This mechanically closes every repository-visible one-site bypass described in #516. As with any policy implemented and invoked from the same repository, deleting every invocation and softening the sole required context simultaneously cannot be cryptographically judged by that deleted code. Live branch protection currently requires the GitHub Actions `gate` context; that external configuration and review remain the bootstrap trust root. This PR does not claim otherwise or mutate branch protection.

## Verification

- `node scripts/check-github-actions-pins.mjs --self-test` (54 source-policy, 6 gate-verdict, 13 current-workflow mutations)
- `node scripts/check-zuuli-linux-image.mjs --self-test` (32 negative cases + runtime cases)
- `node scripts/check-zuuli-linux-image.mjs`
- `scripts/check-public-repo-boundary.sh --self-test`
- `scripts/check-public-repo-boundary.sh`
- `git diff --check`

Closes #516

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -551,7 +551,9 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Recheck immutable actions and fail-closed required jobs
         id: policy
-        run: node scripts/check-github-actions-pins.mjs
+        run: |
+          node scripts/check-github-actions-pins.mjs --self-test
+          node scripts/check-github-actions-pins.mjs
 
       - name: Verify required jobs succeeded or legitimately skipped
         env:
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -30,6 +30,9 @@ const REQUIRED_JOB_ENVIRONMENTS = new Map([
     ]),
   ],
 ]);
+const REQUIRED_JOB_DEFAULT_WORKING_DIRECTORIES = new Map([
+  ["frontend", "wallet/zuuli"],
+]);
 const REQUIRED_STEP_ENVIRONMENTS = new Map([
   [
     "changes\0Detect release-impacting ZUULI changes",
@@ -73,6 +76,12 @@ const REQUIRED_STEP_ENVIRONMENTS = new Map([
       ["REQUIRED_JOBS_JSON", "${{ toJSON(needs) }}"],
     ]),
   ],
+]);
+const REQUIRED_STEP_WORKING_DIRECTORIES = new Map([
+  ["rust_clippy\0Verify pinned Linux build image", "/"],
+  ["rust_plugin\0Verify pinned Linux build image", "/"],
+  ["rust_app\0Verify pinned Linux build image", "/"],
+  ["zuuallet_schema\0Verify pinned Linux build image", "/"],
 ]);
 
 // Required-gate policy deliberately accepts a small, canonical YAML subset.
@@ -311,7 +320,14 @@ function mappingEntry(source, kind) {
   return { key: decoded.value, value: match[2].trim() };
 }
 
-function defaultRunShellFailures(relativeFile, lines, property, end, owner) {
+function defaultRunExecutionFailures(
+  relativeFile,
+  lines,
+  property,
+  end,
+  owner,
+  expectedWorkingDirectory,
+) {
   const failures = [];
   if (property.value) {
     failures.push(
@@ -384,6 +400,12 @@ function defaultRunShellFailures(relativeFile, lines, property, end, owner) {
   if (shell && shell.value !== "bash") {
     failures.push(
       `${relativeFile}:${shell.index + 1}: ${owner} defaults.run.shell must be exactly bash`,
+    );
+  }
+  const workingDirectory = runProperties.get("working-directory");
+  if ((workingDirectory?.value ?? undefined) !== expectedWorkingDirectory) {
+    failures.push(
+      `${relativeFile}:${(workingDirectory?.index ?? property.index) + 1}: ${owner} defaults.run.working-directory differs from its exact reviewed value`,
     );
   }
   return failures;
@@ -565,15 +587,23 @@ function requiredJobExecutionFailures(relativeFile, lines, job) {
   const requiredMainWorkflow =
     relativeFile.split(path.sep).join("/") === REQUIRED_WORKFLOW_PATH;
   const defaults = job.properties.get("defaults");
+  const expectedDefaultWorkingDirectory = requiredMainWorkflow
+    ? REQUIRED_JOB_DEFAULT_WORKING_DIRECTORIES.get(job.id)
+    : undefined;
   if (defaults) {
     failures.push(
-      ...defaultRunShellFailures(
+      ...defaultRunExecutionFailures(
         relativeFile,
         lines,
         defaults,
         job.end,
         `required job ${job.id}`,
+        expectedDefaultWorkingDirectory,
       ),
+    );
+  } else if (expectedDefaultWorkingDirectory !== undefined) {
+    failures.push(
+      `${relativeFile}:${job.start + 1}: required job ${job.id} defaults.run.working-directory differs from its exact reviewed value`,
     );
   }
   const jobEnvironment = job.properties.get("env");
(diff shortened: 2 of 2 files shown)


skyl:
BLOCKED at exact head `f5fdf599192763e5446982efe117619a2091dd50` — job-level default shell can make the exact pinned gate controls no-ops.

A valid/actionlint-clean mutation under the required `gate` job is:

```yaml
    defaults:
      run:
        shell: bash -n {0}
```

This leaves all three exact gate steps and both exact command sequences byte-for-byte intact, so the current policy does not reject it. In an isolated copy of this exact head, after adding only that block:

- `node scripts/check-github-actions-pins.mjs --self-test` passed all `54 source-policy, 6 gate-verdict, and 13 current-workflow` cases.
- `node scripts/check-github-actions-pins.mjs` passed.
- `actionlint .github/workflows/zuuli.yml` passed.

But Actions now launches each gate `run:` block with `bash -n`, which parses the scripts without executing them. Consequently neither the gate-local policy/self-test nor `--verify-gate-results` runs, and the always-running required `gate` can conclude success even when a dependency failed. Step-level `shell` decoration is prevented by the exact step-key checks, but the enclosing job's `defaults.run.shell` is currently outside `requiredGateControlFailures` / `gatePolicyFailures`.

Please fail closed on an inherited/default shell override for the required `gate` controls (and mutation-bind this exact case). The same class should be considered for the `changes` policy step; protecting the gate itself is the critical branch-protection blocker.

Baseline exact-head evidence was otherwise green locally: policy self-test/live checker, Android gate self-test/live checker (preserving `rust_android_32`), full actionlint, and diff check.

skyl:
BLOCKED at exact head `12150a1ee2e0e1ab61f15a50b4efd1a731fc861c` — inherited process environment is still an unpoliced way to turn the required scripts into successful no-ops.

The new parser checks `defaults.run.shell` and step `shell`, including transitive local reusable workflows, but `requiredJobShellFailures` does not inspect workflow-, job-, or ordinary-step-level `env`. Only the gate verdict's two exact data variables are parsed. That leaves several valid/actionlint-clean execution controls outside the policy.

The minimal one-line bypass is a workflow or `gate` job environment entry:

```yaml
    env:
      SHELLOPTS: noexec
```

Bash imports `SHELLOPTS` before processing the generated `run:` script. `noexec` parses commands without executing them and exits successfully. Using the same runner-style flags locally:

```text
env SHELLOPTS=noexec bash --noprofile --norc -eo pipefail -c \
  'node scripts/check-github-actions-pins.mjs --self-test; node scripts/check-github-actions-pins.mjs; node scripts/check-github-actions-pins.mjs --verify-gate-results; exit 97'
# exit 0, no output
```

That preserves the exact three gate steps and every pinned command while executing none of them. At workflow scope it also makes the Android policy/type-check command blocks parse-only, so the merged `rust_android_32` control is not preserved against this mutation.

Adjacent variants in the same uninspected surface:

- `BASH_ENV: bypass-gate.sh`, with a tracked file containing `exit 0`, exits before each noninteractive Bash script. In the isolated exact-head mutation already present locally, `actionlint` and `node scripts/check-github-actions-pins.mjs` both return 0; a runner-style Bash invocation traces only `+ exit 0` and returns 0.
- `'BASH_FUNC_node%%': '() { return 0; }'` imports a Bash function that replaces all four exact `node` invocations. Local reproduction reports `node is a function` and the full policy/verdict sequence returns 0.
- `NODE_OPTIONS: "--import=data:text/javascript,process.exit(0)"` at the gate job makes Node exit 0 before resolving or executing even a nonexistent policy script. `actionlint` accepts it and local Node 22 reproduces it; the workflow's pinned Actions use Node 24.
- `PATH` can similarly resolve `node`, `cargo`, or `npm` to a checked-in successful shim after checkout.

The same inheritance exists in every required dependency and every transitive local reusable-workflow callee at workflow/job/step scope. A job- or step-scoped `SHELLOPTS: noexec` can therefore neutralize an ordinary required test while the complete-needs verifier sees `success`.

Please fail closed on execution-shaping environment at all required scopes, including transitive callees, and mutation-bind at least workflow/job/step `SHELLOPTS`, `BASH_ENV`, imported `BASH_FUNC_*`, and the gate's Node startup/command-resolution surface. Quoted keys, inline maps, and merge/alias forms need the same constrained-parser treatment as the existing structural controls.

Other exact-head evidence:

- Prior `bash -n {0}` regression reproduced: Bash returns 0 without executing a trailing `exit 97`; the 58-source/6-verdict/17-current-workflow self-test rejects workflow-, job-, step-, and transitive-callee shell variants.
- Direct and nested local reusable-workflow soft-fail/external-callee fixtures pass; non-required formatter-indentation fixture passes.
- Policy live/self-test, Android live/20-mutation self-test, Linux-image live/32-case self-test, public-boundary live/self-test, `actionlint`, Node syntax, and diff check all pass.
- Every exact-head hosted check is green, including `Rust / Android 32-bit`, the required `gate`, and all four packaging-smoke builds. Green CI does not exercise the accepted hostile-environment mutation.

skyl:
BLOCKED at exact head `50337a7b75c814fb4b802b7da7edabe3d73ef7de` — required-step `working-directory` can redirect an exact required command into an attacker-controlled checked-in tree.

The new execution-policy parser recognizes `defaults.run.working-directory` but permits any value, and `requiredJobExecutionFailures` does not inspect step-level `working-directory` at all. A valid/actionlint-clean mutation on the required `rust_plugin` job is:

```yaml
      - name: Build and test shared Zcash plugin
        working-directory: bypass
        run: cargo test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
```

With a checked-in minimal crate at `bypass/wallet/plugins/tauri-plugin-zcash`, the exact command succeeds while never compiling or testing the real shared plugin. In an isolated archive of this exact head with only that mutation and the minimal crate:

- `node scripts/check-github-actions-pins.mjs --self-test` passed all **65 source-policy, 6 gate-verdict, and 24 current-workflow** cases.
- `node scripts/check-github-actions-pins.mjs` passed.
- `actionlint .github/workflows/zuuli.yml` passed.
- From the runner-equivalent `bypass` directory, the exact pinned `cargo test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml` compiled `bypass v0.0.0`, ran zero tests, and exited 0.

The gate then receives `rust_plugin.result == success`, so its complete-needs verifier cannot distinguish this no-op from the real test. The same surface applies to other required commands and to job/workflow defaults. Please fail closed on required-scope `working-directory`, while retaining the reviewed `working-directory: /` image-verification steps, and mutation-bind the step-level redirect case.

The unmodified exact head otherwise passes the policy self-test/live scan, Android policy self-test/live scan, full `actionlint`, Node syntax, and diff checks. I am continuing the exact-head review; this finding alone withholds CLEAN.

skyl:
Root detached final review — exact head `a547f5f65d9b1c373a7d0c8c325026e4ec241f9e`

No blocking findings.

Independent review covered the recursive required-workflow policy, exact gate and changes invocation checks, external reusable-workflow rejection, job/step fail-open controls, exact environment/shell/working-directory contracts, complete `needs` verdict handling, and the mutation corpus.

Re-executed at this exact head:
- Actions policy: 70 source cases, 6 gate-verdict cases, 41 exact-current mutations; live scan covered 141 action references across 14 workflow files
- Android policy: 20 cases plus live check
- Linux packaging image policy: 32 cases plus live check
- Rust toolchain, Zcash permissions, and public-boundary policies: self-tests plus live checks
- `actionlint` and `git diff --check`

All passed. Merge remains contingent on every hosted check, including the still-running macOS packaging job, completing successfully.

PR #519, opened 2026-08-23
docs(rust): decide Cargo workspace boundary

### LINKED ISSUE #341: Decide: single Cargo workspace vs. three standalone packages
## Question

There is **no root `Cargo.toml`** in this repo. Three first-party packages each act as their own workspace root:

- `wallet/plugins/tauri-plugin-zcash/`
- `wallet/zuuli/src-tauri/`
- `wallet/zuuallet/src-tauri/`

Consequences, all currently accepted deliberately (there are load-bearing comments in the manifests saying so):

- **Three separate `Cargo.lock` files.** CI comments note the app and standalone plugin "resolve some crates differently" — so we test one dependency graph and ship another.
- **The `[profile.*]` blocks are duplicated verbatim in all three manifests**, purely because there is no shared root.
- CI must pass `--manifest-path` per crate, and `rust_plugin` / `rust_app` run as separate 45-minute jobs partly *because* the lockfiles differ.
- Every new crate repeats all of the above.

## What to decide

Whether to consolidate into a single workspace, and if so what it costs.

**Arguments for a workspace:** one lockfile means one tested dependency graph; shared `[profile]` and `[workspace.dependencies]`; faster CI through a shared target dir; new crates cost nothing to add.

**Arguments against:** Tauri mobile builds have opinions about workspace layout; `zuuallet` and `zuuli` may legitimately want to resolve differently; a single lockfile means a bump for one app is a bump for both; and the `links = "sqlite3"` class of conflict becomes repo-wide rather than per-app — which is a real constraint, since two crates in one graph cannot use different `libsqlite3-sys` versions.

That last point cuts both ways: a workspace would *surface* such conflicts early rather than at integration time, but it also removes the ability to isolate them.

## Acceptance criteria

- [ ] A decision recorded with its rationale (ADR or `AGENTS.md`).
- [ ] If consolidating: one lockfile, deduplicated `[profile]` blocks, CI updated, all three crates and all mobile targets still building.
- [ ] If not consolidating: the reasons written down so this is not re-litigated, plus a note on how new crates should be added.

## Notes

Genuinely foundational and independent of any feature work. Worth deciding before the next Rust crate is added, not after.

skyl:
## Decision

Keep the three existing Cargo roots independent and treat the release train—not the repository—as the durable workspace boundary.

- Records the current lockfile, CI, Tauri mobile, cache, and release-path evidence behind the decision.
- Allows a product-scoped workspace if one application later splits into inseparable internal crates.
- Defines how new shared and app-internal crates must be added and integrated.
- Names explicit revisit triggers and the Android, iOS, desktop, generated-project, supply-chain, and release proofs required before a wallet-wide consolidation.
- Links the decision from `AGENTS.md` so the existing load-bearing profile comments are not the only record.

This is docs-only: it deliberately does not change Cargo resolution during the active ZUULI/TestFlight release train.

## Verification

- `scripts/check-rust-fmt.sh`
- Markdown local-link and final-newline check
- `git diff --check`

Closes #341

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -297,6 +297,10 @@ to catch these too, but catching them before the push is cheaper than a red run.
 
 ## Practical notes
 
+- The three first-party Cargo roots and lockfiles are deliberately independent.
+  The release-train boundary, new-crate rules, and explicit triggers for
+  reconsidering a wallet-wide workspace are recorded in
+  [docs/architecture/CARGO-WORKSPACE.md](docs/architecture/CARGO-WORKSPACE.md).
 - Submodules live at `z/{github-org}/{repo}` and track a branch (see
   `.gitmodules`). Update to latest with `git submodule update --remote`.
 - A submodule pin must be **fetchable from its configured `url`**. Verify a
--- a/docs/architecture/CARGO-WORKSPACE.md
+++ b/docs/architecture/CARGO-WORKSPACE.md
@@ -1,0 +1,144 @@
+# Cargo workspace boundary
+
+Status: accepted on 2026-08-23. This records the decision for
+[issue #341](https://github.com/free2z/zuu/issues/341).
+
+## Decision
+
+Keep the three existing Rust package roots independent:
+
+- `wallet/plugins/tauri-plugin-zcash/`
+- `wallet/zuuli/src-tauri/`
+- `wallet/zuuallet/src-tauri/`
+
+Each remains the root of its own Cargo resolution, lockfile, profiles, and
+target directory. Do not add a virtual workspace above all three merely to
+deduplicate configuration.
+
+The durable boundary is the **release train**, not the repository. ZUULI and
+Zuuallet are separately built applications, while `tauri-plugin-zcash` is a
+reusable package that must also work and test standalone. A product may use a
+product-scoped workspace later if it grows several inseparable internal crates;
+that does not imply a single workspace for every Rust package under `wallet/`.
+
+## Why
+
+### The committed lockfiles describe real, different states
+
+The three locks are not duplicate files waiting to be moved. At the time of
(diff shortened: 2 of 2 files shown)


skyl:
CLEAN at exact head `c2c8f696d86760c50deb528c052e95237d0974a5`.

Independent evidence audit:

- Repository/Cargo state matches the ADR: exactly three first-party manifests and locks exist under `wallet/`; `cargo metadata --locked --no-deps` reports each as a separate one-member workspace root with its own default target directory.
- The stated lock divergence is exact: ZUULI has Tauri `2.11.5` / `aws-lc-rs 1.17.3`; the plugin and Zuuallet have Tauri `2.10.2` / `aws-lc-rs 1.15.4`. All three locks contain `rusqlite 0.37.0` / `libsqlite3-sys 0.35.0` as claimed.
- The support matrix is real: the standalone plugin's lock is exercised on Linux plus native macOS/Windows and armv7 Android type-check; each application lock builds the plugin in that application's path-dependency graph.
- ZUULI packaging/release runs from `wallet/zuuli`, uses Tauri's locked Cargo passthrough, reads `src-tauri/Cargo.lock`, and collects the documented app-local target artifacts. The separate hosted jobs and scoped caches support the CI rationale.
- New-crate guidance is actionable and consistent with the manifest-discovering fmt/clippy/deny scripts. It permits an app-scoped workspace for inseparable internal crates, distinguishes standalone reusable contexts, and requires consumer lock updates plus explicit supported-platform coverage.
- Revisit triggers are concrete and the proof list targets the actual migration risks; this records the current release-train boundary without treating it as immutable architecture. The diff contains only public repository facts/paths.

Verification passed: `scripts/check-rust-fmt.sh` (all 3 roots, pinned Rust `1.97.1`), Markdown local-link/final-newline check, and `git diff --check`. Hosted exact-head `changes` and required `gate` are green; selector-controlled build jobs are legitimately skipped for this docs-only diff.

PR #520, opened 2026-08-23
feat(zuuli): preserve local article drafts

### LINKED ISSUE #251: ZUULI Articles: autosave drafts and guard against accidental loss of long-form work
## Problem

Every `/articles/new` field lives only in component state. A sidebar click, back gesture, refresh, app close, or crash silently destroys the entire draft. There is no autosave, dirty-navigation warning, recovery, or explicit discard action.

The reference Svelte editor already has bounded autosave and navigation protection; ZUULI should not make long-form authoring less safe.

## Acceptance criteria

- [ ] Draft content is autosaved locally and restored after refresh/reopen.
- [ ] In-app navigation and app/window close protect unsaved changes.
- [ ] Users can explicitly discard a draft.
- [ ] Successful publication clears only the published draft.
- [ ] Multiple drafts/users cannot overwrite one another.
- [ ] Autosave status and recovery are understandable without interrupting writing.

skyl:
Closes #251

## Summary

- autosave every article field to a bounded, versioned local store scoped by account and stable draft ID
- restore the most recent draft on reopen, expose a draft chooser, and flush pending edits before internal navigation, reload, backgrounding, or unmount
- detect stale cross-tab revisions instead of overwriting newer work, with clear saved/error/conflict and recovery states
- add explicit discard and clear only the successfully published draft

## Boundary

This is intentionally client-local. It adds no server draft endpoint or persistence assumption. A future storage schema is never overwritten, unavailable storage is surfaced to the author, and local write conflicts fail closed.

## Validation

- `npm run typecheck`
- `npm run build`
- `ZUULI_PW_PORT=1466 npm test` (all unit/contract checks and 116 Playwright tests passed)
- GitHub Actions pin, Zuuli Linux image, and Android gate policy checks plus self-tests


skyl:
BLOCKED at exact head `e15a39fede2da72544b4ba5976b4180ec95c1429` after independent adversarial review.

1. **Edits made while a publish request is in flight are silently deleted.** `publish()` snapshots render-state for the POST, but the title/subtitle/category/content controls remain editable. A later success calls `clearPublishedDraft()`, which sets `dirtyRef` false and unconditionally deletes the draft. Exact-head Playwright negative fixture: saved `Published version`, clicked Publish, typed `Edit typed while publishing and expected to survive` during the mock API delay, waited for success, then reopened the draft URL. The store was `null` and restored content was `""`. Disable all editing/draft actions while publishing or preserve any revision/dirty changes made after the submitted snapshot.

2. **The success/discard delete is not revision-aware and can erase a newer tab's work.** Exact-head two-page fixture: tab A published stored revision 1; before its response completed, tab B edited and synchronously flushed revision 2. Storage immediately before A's response contained revision 2/content `Newer second-tab revision that must survive`; after A's success it was `null`, and reopening restored `""`. `discardArticleDraft()` deletes solely by account/id, so both publish success and the trash action can remove a revision this component did not publish/display. Delete only when the stored revision still matches the submitted/current expected revision; preserve and surface conflicts otherwise.

3. **SPA back/history and programmatic draft switching bypass the loss guard.** The capture listener only guards `<a>` clicks. A browser POP/back gesture and the dropdown's `navigate(...)` proceed; cleanup calls `persistRef.current()` but ignores `false`. Exact-head fixture with `Storage.prototype.setItem` throwing: type work, wait for `Autosave unavailable`, SPA `goBack()`, then reopen the draft URL => zero confirmation dialogs and an empty title. This misses the issue's explicit back-gesture/in-app-navigation requirement. Route transitions (including POP and dropdown selection) need one shared save-or-confirm/rescue boundary.

4. **Malformed current-version storage is treated as a writable empty store.** `parseStore()` maps JSON parse errors and a non-array `drafts` member to `{version:1,drafts:[]}`; the next autosave overwrites the original bytes. Exact-head fixture seeded `{"version":1,"drafts":[`, typed once, and observed it replaced by a new one-draft store while the UI reported `Saved locally`. Reads may recover as empty, but writes must fail closed for corrupt/unrecognized data just as they do for a future version.

Focused checks themselves pass: `npm run typecheck`; 6/6 draft-store unit tests; 3/3 article-draft Playwright tests. The negative cases above are absent from the suite and reproduce at this exact SHA. Any push invalidates this review.

skyl:
Fix-forward pushed at `afad3ce` for all four reproduced blockers:

- publishing now locks every editor/draft action and submits an immutable field snapshot; handler-level guards also reject synthetic input during the request
- publish and explicit discard use compare-and-delete against the exact stored revision, preserving a newer tab revision and surfacing that outcome
- the app now uses React Router's data-router blocker so link clicks, imperative draft selection, and POP/back all share the same synchronous save-or-confirm boundary
- malformed current-version stores read as empty for recovery but are read-only for save/discard, preserving their original bytes

Regression coverage includes the injected in-flight edit, a two-tab revision-2 write before revision-1 publish completion, failed-storage dropdown and browser-back attempts, and byte-for-byte corrupt-store preservation. Fresh validation: production build; 8/8 draft unit tests; full 197/197 Node contracts; full 120/120 Playwright; workflow pin/Linux image/Android gate policies and self-tests.

skyl:
BLOCKED on re-review at exact head `afad3ceef73283e65d3e84ad7853312bde5c285f`.

The four prior blockers are fixed at this SHA: in-flight fields/actions are locked and the submitted snapshot is stable; a newer tab's revision survives publish success; the data-router blocker covers dropdown/imperative navigation and POP; corrupt v1 bytes remain byte-for-byte intact. Re-ran the exact focused coverage: typecheck PASS, production build PASS, 8/8 draft-store unit tests PASS, and 7/7 article-draft Playwright tests PASS (including all four negative fixtures).

One adjacent state transition still bypasses the protection boundary: **sign-out/account unmount can silently discard edits after persistence fails.** `useBlocker` only observes router transitions, while `TopBar` calls `logout()` directly; the session store sets `user: null` immediately, which unmounts `AuthenticatedAuthor`. Its cleanup retries `persistRef.current()` but ignores `false`, and there is no confirmation/rescue.

Exact-head Playwright reproduction:

1. Save revision 1 (`Stored revision`) and keep its draft URL.
2. Make only article-draft `localStorage.setItem` throw `QuotaExceededError`.
3. Edit the title to `Unsaved update before sign out`; wait for `Autosave unavailable`.
4. Account menu → Sign out.
5. Sign back into the same mock account and reopen the exact draft URL.

Observed: **0 confirmation dialogs**, restored title `Stored revision`; `Unsaved update before sign out` was lost. This is the same loss class the router blocker now prevents, reached through the account lifecycle rather than a location change. Route all explicit sign-out/account transitions through a shared save-or-confirm boundary (while retaining account-scoped writes), and add this fixture. Any push invalidates this review.

skyl:
Adjacent sign-out loss fixed at `b61b9f1`.

The authenticated article editor now registers its synchronous persistence callback with a small shared unsaved-transition boundary. The existing explicit `useSession.logout()` action runs that boundary before clearing account state: clean or successfully persisted drafts sign out normally; a failed write requires an explicit confirmation, and cancel leaves both the session and in-memory editor intact. The only other direct session clear is the involuntary expired-auth handling path and deliberately remains non-interactive.

The exact regression saves revision 1, makes only draft `setItem` fail, edits, selects Account → Sign out, verifies cancel preserves the account and edit, then verifies confirmation explicitly permits sign-out while leaving revision 1 durable.

Fresh validation: typecheck/build PASS; focused session tests 2/2; focused article-draft Playwright 8/8; complete Vitest and Node contract suites PASS (197/197 Node); complete Playwright 121/121; workflow pin/Linux image/Android gate policies and self-tests PASS.

skyl:
Root detached final review — exact head `f5a1da24b34781b4bdb5aa7b70e12296cae2db45`

No blocking findings.

I re-reviewed the account-scoped/revisioned persistence boundary, corrupt/future-store fail-closed behavior, publish snapshot locking, compare-and-delete semantics, router POP/imperative navigation guard, explicit sign-out guard registry, and the negative browser fixtures added for every earlier reproduced loss mode. The merge from trunk changes only the two already-reviewed CI-policy files; the product diff is unchanged.

Re-executed at this exact integrated head:
- GitHub Actions policy: 70 source-policy, 6 gate-verdict, and 41 current-workflow mutation cases plus the live 141-reference scan
- TypeScript typecheck
- Article draft store tests: 8/8
- Article draft Playwright tests: 8/8, including cancel/confirm sign-out under failed persistence
- `git diff --check`

All passed. Merge remains contingent on every hosted exact-head check completing successfully.

skyl:
Final trunk refresh review — exact head `eb3b06363d18c02141582157bcf2ed4a1b1a2e0f`

No blocking findings.

The article/session/Playwright tree is byte-for-byte identical to the previously clean reviewed head `f5a1da24b34781b4bdb5aa7b70e12296cae2db45`; this refresh only integrates merged AWS-LC and CODEOWNERS trunk changes. The PR diff against current main remains the same eight article-draft files.

Re-executed at this exact head:
- Actions policy: 70 source, 6 gate-verdict, 41 mutation cases and live 141-reference scan
- TypeScript typecheck
- draft-store tests 8/8
- session tests 2/2
- article-draft Playwright 8/8 on verified-free port 15432
- `git diff --check`

All passed. The first Playwright invocation encountered another agent's occupied default port and did not run any browser test; the supported `ZUULI_PW_PORT` override was then used without touching that process. Merge remains contingent on the new exact-head hosted matrix.

skyl:
BLOCKED on independent adversarial review at exact head `eb3b06363d18c02141582157bcf2ed4a1b1a2e0f`.

**The cross-tab revision contract is a non-atomic read/check/write over one shared blob, so a simultaneous save can silently erase another draft or account.** `saveArticleDraft()` reads `ARTICLE_DRAFT_STORAGE_KEY` at lines 227–233, mutates that snapshot, and writes the entire store at lines 246–252. `localStorage` makes each individual operation atomic, not that multi-operation sequence. Two tabs can therefore both read the same snapshot before either writes. Both return `saved`; whichever whole-store write lands last removes the other change. The `storage` listener cannot prevent this because it runs after the writes, and for the same draft both writers produce the same numeric revision, so lines 267–269 do not even recognize the surviving revision as foreign.

I reproduced both interleavings deterministically against the exported exact-head storage contract:

- Different accounts/IDs: Alice begins saving `draft-alice-01`; Bob completes `draft-bob-01` after Alice has read but before Alice writes. Both saves report success, but the final blob contains only Alice and `listArticleDrafts("bob")` is empty.
- Same draft/revision: two writers both read revision 1, both report `saved` revision 2, and the later write silently replaces the other body. The revision check never returns `conflict`.

This directly misses the acceptance criterion that multiple drafts/users cannot overwrite one another and the PR claim that stale cross-tab revisions fail closed. The sequential unit test and Playwright two-tab publish fixture do not exercise overlapping read/write windows. The persistence scheme needs a transactional/serialized update boundary (and a regression that forces both reads before either write); merely re-reading the same shared blob or relying on `storage` events does not make compare-and-set atomic.

Other exact-head evidence is green: typecheck; 25/25 focused draft/session/route-scroll unit tests; 8/8 article-draft Playwright; and 23/23 paid-action plus scroll-restoration Playwright tests. The BrowserRouter-to-data-router behavior passed POP, PUSH, active-tab, funding-history, draft-switch, and failed-save back-navigation coverage. Worktree remained detached and clean. Any push invalidates this review.

skyl:
Fixed the reproduced cross-tab overwrite class at exact head `3c9368aad309825aa94a018c325ebef895e3f123`.

Persistence no longer rewrites a shared collection. Each attempted body is synchronously stored once under a unique immutable claim; account/draft heads are disjoint compact references. A head advances only for the exact unchanged parent and sole visible same-parent claim. The winning claim remains until superseded, so even writers that miss each other during non-atomic head arbitration leave the overwritten body as a separately listable rescue; storage listeners surface that conflict. Legacy v1 data stays read-only, corrupt/future state remains fail-closed, compare-delete uses the same exact-parent arbitration, and the 100-draft/field-size bounds remain enforced.

Exact regressions include forced reentrant different-account/draft writers, forced same-parent siblings, and a parent-changing after-read/before-claim interleaving. Local evidence on the integrated branch: `13/13` focused Vitest, TypeScript typecheck, `8/8` article-draft Playwright, production build, plus full suite `202/202` Node/contract and `122/122` Playwright before the final unrelated SBOM-only main integration; focused/type/browser rerun after that integration is green. Hosted CI has been triggered for this exact head.

skyl:
Root exact-head re-review is clean at `3c9368aad309825aa94a018c325ebef895e3f123` on current main `c443ff8a053107691064ef16b20ea0d003b66a4e`.

I re-audited the replacement persistence protocol rather than carrying forward the earlier review. Draft bodies are now immutable unique claims, account/draft heads are disjoint compact records, a head can advance only from the exact unchanged parent with one visible same-parent claimant, and the winner remains until superseded. Forced overlapping different-account/draft writes therefore cannot rewrite one another; same-parent writers either produce one winner plus a listable rescue or both surface a conflict with every body recoverable. Legacy v1 stays read-only, corrupt/future records fail closed, tombstones and compare-delete remain revision-aware, and the UI lists/surfaces rescues through the storage listener.

I also rechecked the prior loss boundaries: immutable publish snapshot and disabled mutations, newer-tab preservation on publish, POP/imperative navigation blocking, explicit sign-out guard, close/reload flush, corrupt-store preservation, and explicit discard.

Exact local proof from the clean implementation worktree:
- TypeScript typecheck
- draft arbitration/unit suite: 13/13
- article-draft Playwright suite: 8/8 on isolated port 15437
- `git diff --check`

All passed. Merge remains contingent on every hosted check at this exact head becoming terminal green and no subsequent feedback/push.

Issue #368, opened 2026-08-09
security(send): bind ZEC confirmation to the immutable proposed payment


skyl:
## Summary

The ZEC send review dialog displays live form state, while execution uses an earlier proposal identified only by `proposalId`. The recipient and memo can change after proposal creation but before confirmation, causing the confirmation UI to describe a different payment than the one that will be signed and broadcast.

This is a transaction-signing TOCTOU defect.

## Evidence

- `wallet/zuuli/src/features/wallet/Send.tsx:184-205` creates a proposal from captured `to.trim()` and `memo` values.
- Recipient/memo controls remain mutable while preparation/review is active at `Send.tsx:331-339`, `:386-395`, and `:438-449`.
- The confirmation dialog renders the current recipient and memo at `Send.tsx:529-550`.
- Execution sends only `proposalId` at `Send.tsx:215-220`.
- `SendProposal` omits recipient and memo in `wallet/zuuli/src/features/wallet/types.ts:95-100`; the Rust response model similarly does not give the UI immutable payment details.
- Existing integrity comparison covers amounts, not recipient and memo.

One practical race is: create a proposal, edit recipient or memo while the async proposal/review flow is active, then approve a dialog whose displayed details no longer identify the stored proposal.

## Required remediation

- Return an immutable canonical payment snapshot from native proposal creation: network, recipient(s), amount(s), memo(s), fee, change policy, and proposal identifier/digest.
- Render confirmation exclusively from that snapshot, never from live form state.
- Bind execution to a digest or native confirmation token covering all signed payment details.
- Disable editing while proposing/reviewing, and invalidate any proposal whenever an input changes.
- Treat UI disabling as defense in depth, not the integrity boundary.

## Acceptance criteria

- Changing recipient, memo, amount, fee policy, or network after proposal creation invalidates execution.
- The confirmation view displays exactly the canonical details bound to native execution.
- Tests deterministically race edits against proposal completion and prove the reviewed and broadcast payment cannot differ.
- Multi-recipient and send-all flows receive the same binding.

skyl:
Implementation started in a fresh isolated worktree from exact `origin/main` `16f31d32c9f4bc8735974f42e38a4c0b6eee747b`. Scope is the approved immutable native review snapshot, versioned digest, opaque confirmation token, exact discard, and frontend generation fencing, with fail-first native/static/browser contracts. No real signing or funds will be used.

PR #523, opened 2026-08-23
fix(zuuli): externalize embeds from privileged native WebViews

### LINKED ISSUE #367: security(android): remote article iframes inherit privileged wallet IPC
## Summary

On Android, a YouTube or Vimeo iframe rendered inside an article receives ZUULI's Tauri initialization scripts and can invoke native commands as though it were the trusted local main frame.

This is a release-blocking authority-boundary failure. A compromised provider script or nested third-party frame can invoke wallet side effects without first compromising ZUULI's own JavaScript. Because the mobile main window receives `zcash:default`, the exposed surface includes proposal creation/execution, seed and key access, wallet deletion/switching, signing, unlock, and lightwalletd endpoint mutation.

## Evidence

- `wallet/zuuli/src/components/common/Markdown.tsx:261-290` creates unsandboxed remote YouTube/Vimeo iframes.
- `wallet/zuuli/src-tauri/tauri.conf.json:22-24` permits both providers in `frame-src`.
- `wallet/zuuli/src-tauri/capabilities/mobile.json:3-18` gives mobile `main` both `zcash:default` and `opener:default`.
- `wallet/plugins/tauri-plugin-zcash/permissions/default.toml:1-35` expands `zcash:default` to all sensitive wallet commands.
- Locked Tauri 2.11.5 marks core/invoke-key scripts main-frame-only, but locked Wry 0.55.1's Android path discards that flag and installs each script using `WebViewCompat.addDocumentStartJavaScript(..., setOf("*"))`.
- Wry's Android IPC reports the top-level WebView URL, not the requesting frame URL. Tauri therefore resolves the iframe's IPC request as the trusted local window.

This matches Tauri's documented platform limitation: on Android and Linux, the runtime cannot distinguish iframe requests from the window for capability matching: https://v2.tauri.app/security/capabilities/

The directly supported impact is blind command invocation. For example, remote frame JavaScript can create a send-all proposal and try the small/predictable proposal identifier space when invoking execution. If the seed is already cached, this can sign and broadcast without another custody prompt. Direct exfiltration of command return values from an iframe has not been established and is not required for the fund-loss path.

Linux is not confirmed by this exact path because WebKitGTK currently honors top-frame-only script injection. The confirmed vulnerability is Android.

## Required remediation

- Remove remote iframes from the privileged Android wallet WebView. Open embeds externally or in a genuinely separate, capability-free native view.
- Do not rely on the iframe `sandbox` attribute alone; Wry injects bridge scripts with a wildcard frame match.
- Replace `zcash:default` with named least-privilege capabilities. Seed/key reveal, wallet mutation, signing, endpoint mutation, proposal creation, and execution must not be ambient main-window authority.
- Require a native, expiring, one-use confirmation token bound to proposal ID, recipient, amount, and fee before execution.
- Report the upstream Wry defect: preserve main-frame-only injection and authenticate the actual requesting frame origin.

## Acceptance criteria

- Android instrumentation proves a cross-origin and nested iframe has no `__TAURI_INTERNALS__` and raw IPC cannot mutate a canary.
- Android article embeds externalize rather than creating an iframe in the privileged WebView.
- A capability contract test rejects `zcash:default` on the mobile main window.
- Execute-send rejects absent, stale, reused, or payment-mismatched native confirmation tokens.
- Tests cover both document-start-script support and the Android fallback path.

skyl:
## Summary

- render YouTube and Vimeo article/comment embeds as OS-external links in Tauri builds while preserving consent-gated browser embeds
- fail closed when the native opener fails and deny all framed content in the packaged Tauri CSP
- replace the mobile `zcash:default` grant with the exact named commands used by the current product
- add an executable capability/CSP allowlist contract plus renderer and Playwright regressions

## Security boundary

This is an immediately shippable defense-in-depth slice of #367. It prevents privileged packaged WebViews from retaining remote iframe authority both in the renderer and via `frame-src 'none'`, and makes expansion of the mobile main window wallet capability fail a committed test.

The native review-bound, single-use send authorization described by the issue already landed separately in #368/#441 and remains covered by the existing send-review contract tests.

This PR deliberately **does not close #367**. Remaining acceptance requires real Android instrumentation for cross-origin and nested iframe bridge denial, document-start versus fallback-path validation, and the corresponding Wry/upstream origin-authentication resolution. The Playwright case here supplies a deterministic renderer regression, not device evidence.

Refs #367

## Verification

- `npm test` — 202 Node tests and 114 Playwright tests passed; Vitest passed
- `npm run typecheck`
- `npm run build`
- `npm run test:mobile-webview-authority`
- `npm run release:verify`
- `node scripts/check-zcash-permissions.mjs --self-test`
- `git diff --check`


skyl:
Exact-head root review is clean at 573525166508f8191dd88b8d9c5f2f2dac21a5ab after integrating current main. I inspected the capability/CSP contract, confirmed no other capability applies to mobile main, checked the retained 23 Zcash permissions against current frontend bridge calls, and verified native external links synchronously prevent WebView navigation and fail closed if the opener rejects. Local exact-head evidence: npm run typecheck; production build; mobile authority contract 5/5; native embed Playwright 1/1 on an isolated port. This remains a progress slice of #367: physical Android cross-origin/nested-frame proof and upstream Wry/Tauri provenance remain explicitly open. Merge is appropriate once every hosted check for this exact head is terminal green.

Issue #379, opened 2026-08-09
security(release): generate SBOMs from the shipped platform artifacts


skyl:
## Summary

The release workflow labels and attests its SBOM as a platform/release inventory, but Syft scans the source tree while configuration excludes the actual built payloads and installed dependency trees.

The result is primarily a manifest/lockfile inventory. It can omit unexpected binaries, native libraries, bundled resources, build-script output, or dependencies that are present in the shipped AAB/IPA/app/DMG/AppImage/deb/rpm. This creates a false assurance boundary for vulnerability response and artifact attestation.

## Evidence

- `wallet/zuuli/syft.yaml` excludes `node_modules`, `target`, `dist`, and `release-artifacts`, and uses file selection that does not inventory arbitrary shipped files.
- Release jobs invoke Syft against `wallet/zuuli` (for example around `.github/workflows/zuuli-release.yml:302`) rather than the final packaged artifact/root filesystem.
- The workflow validates a component-count threshold, which proves a nonempty source inventory but not correspondence with the shipped bytes.

## Required remediation

- Generate a separate SBOM for each final platform artifact after build/sign/package.
- Unpack or mount AAB/APK, IPA/app, DMG/app, AppImage, deb, and rpm formats as needed and scan their effective shipped filesystem.
- Include native libraries, frameworks, embedded executables, web assets, licenses, and package metadata.
- Bind each SBOM and provenance attestation to the exact artifact digest.
- Keep source/lockfile SBOMs if useful, but label them distinctly and do not substitute them for artifact inventories.
- Add reconciliation checks between declared dependencies and shipped components.

## Acceptance criteria

- A test injects an undeclared canary library/file into a package and the artifact SBOM reports it.
- Each downloadable artifact has its own digest-bound SBOM.
- SBOM metadata identifies source inventory versus shipped-artifact inventory unambiguously.
- Release publication fails if an artifact lacks a successful artifact-level scan/attestation.

ed19:
Every day over 40 emails, is that normal? Or a glitch?

Angel E. Dominguez

skyl:
Claiming a bounded artifact-level SBOM slice in an isolated worktree. I am mapping the current release artifacts first, then will implement only formats that can be honestly unpacked/scanned and digest-bound in one orderly PR, with a canary mutation proving an undeclared shipped payload is inventoried or fails closed. Any unhandled formats will remain explicit follow-up scope; this PR will reference rather than close #379 unless every acceptance criterion is actually met.

[...]

PR #524, opened 2026-08-23
security: bind mobile SBOMs to shipped artifacts

skyl:
## Summary

- scan the actual unpacked Android AAB and iOS ZIP/IPA payloads in packaging smoke and protected release jobs
- inventory every shipped regular file and safe symlink, preserve Syft package discovery, and bind each SBOM to the exact artifact SHA-256 and byte count
- relabel remaining desktop inventories honestly as source-tree SBOMs and document the still-open AppImage/deb/rpm/DMG work
- enforce scan, binding, verification, manifest, and attestation order with release policy and mutation-sensitive canary tests

## Verification

- `npm test` (200 Node/Vitest tests and 113 Playwright tests)
- `npm run build`
- `npm run release:verify`
- `node scripts/verify-ci-cache-policy.mjs`
- `node scripts/apple-credential-boundary.mjs`
- `actionlint .github/workflows/zuuli-packaging.yml .github/workflows/zuuli-release.yml`
- GitHub Actions pin and Android gate policy checks plus self-tests
- `git diff --check`

Refs #379


skyl:
Adversarial review of exact head `e86e0a601b1f7bb6d9de7c60c30677a774d2d247`: **blocking correctness issue**.

`verifyArtifactSbom()` inventories the same mutable `--root` directory that `prepare` created and that Syft/finalization already consumed. It binds the archive digest and SBOM digest, but never proves that the inventoried root still corresponds to the archive bytes. A mutation of `artifact-sbom-work/*/root` after `prepare` and before `finalize-artifact` is incorporated into the SBOM; the subsequent `verify-artifact` sees the same mutated directory and succeeds while the bound AAB/IPA does not contain those bytes. This contradicts the shipped-artifact claim and the documentation's “independently re-inventories the payload” language.

Required fix: verification must derive its expected inventory independently from the exact artifact (for example, safely extract the archive into a fresh, newly-created verification root and compare that inventory to the SBOM), rather than trust/reuse the scan root. Keep the archive digest check, path/symlink safety checks, and binding check. Add a canary test that mutates the original scan root after `prepare`, finalizes against that mutation, and proves verification fails because a fresh artifact-derived inventory lacks it. The workflow contract should require the independent verification boundary.

I found no reason to merge this exact head until that invariant is fixed.

skyl:
Corrected the independent-verification blocker at exact head `74f87e32ae14bc1cf4ce9c288ada127fc5874e7c`.

The verifier no longer accepts a caller-supplied/root argument. It creates a fresh private temporary extraction directory, revalidates archive names and resource ceilings, extracts and inventories the exact digest-bound AAB/IPA, compares that inventory to the SBOM, rechecks artifact/SBOM stability, and removes the temporary root. All four packaging/protected-release invocations now use that root-independent interface before manifest/provenance creation.

Mutation evidence:
- prepare the fixture archive
- inject `post-prepare-injection.dylib` only into the original Syft root
- finalize an SBOM/binding that contains that injected file
- independent verification re-extracts the unchanged IPA and fails on the inventory mismatch
- deleting the original scan root entirely still leaves valid artifact verification green
- a commented valid invocation plus a live `--root=artifact-sbom-work/ios/root` substitution fails the exact named-step workflow contract

Verification:
- `npm run test:artifact-sbom` — 6/6
- `npm run release:verify` — green
- `actionlint` on both changed workflows — green
- full `npm test` — Vitest green, 203/203 Node policy tests, 113/113 Playwright tests
- focused Prettier for scripts and `git diff --check` — green

Normal push only; ready for fresh exact-head review. Do not merge until that review and new hosted CI are green.

skyl:
Root re-review is clean at exact refreshed head b4f5274198c16f13627d0cf112ef1b5a96d7ad5f on main ca912859ab58088be8b6495577eb5e0f0af5386e. The blocker is fixed: verify-artifact has no scan-root option, creates a fresh private temp root from the exact bound AAB/IPA, reapplies archive/member/resource/symlink checks, compares the complete independent inventory, rechecks artifact and SBOM stability, and cleans the root. The root-injection canary now fails as required; removal of the original scan root remains green; exact named workflow steps reject comments/same-root substitution. I also checked unsafe symlink extraction behavior and the post-extraction inventory fails closed. Exact-head local proof: artifact tests 6/6, merged mobile authority 5/5, typecheck, release identity, diff check, clean worktree. Merge only after every fresh hosted check on b4f5274 is terminal green.

PR #526, opened 2026-08-23
fix(live): complete private invite flow

skyl:
Closes #264

## What changed
- create private rooms through the public two-step contract, then join the returned UUID as the authenticated host
- carry the opaque UUID only in a URL fragment, display/copy it for the host, and resolve cold invite links directly without public discovery
- fail closed for malformed, wrong, expired, and ended-room secrets without rendering room metadata
- filter private meetings before the public discovery result is cached
- cover create/host join, invited guest, copy, refresh/deep-link, wrong-secret, cache exclusion, and end-room cleanup

## Public contract evidence
- `POST /api/dyte/{username}/private` returns the creator meeting secret UUID
- `POST /api/dyte/{username}/private/{secret}` grants the creator a host ticket and an invitee a participant ticket
- implementation is based only on the public OpenAPI and existing public React/Svelte clients; no private repository content was inspected

## Verification
- `npm run typecheck`
- focused Vitest: 7 passed
- focused Playwright: 2 passed
- full `npm test`: 197 Node policy tests + 115 Playwright tests passed (Vitest also green)
- `npm run build`
- `npm run release:verify`
- focused Prettier check and `git diff --check`

PR #525, opened 2026-08-23
fix(macos): fail closed on protected seed custody

skyl:
## Summary

- remove the macOS login-Keychain storage/read fallback: all normal seed writes and reads now target the Data Protection Keychain and propagate protected-custody errors
- expose old ordinary-Keychain records only through an explicit legacy boundary, then UFVK-validate, protected-write, protected-readback, and only finally delete the weaker record
- bind macOS packages to exact application/team/Keychain entitlements, require an authorizing Developer ID provisioning profile, and verify the claims before signing, after signing, and again during credential-free finalization
- add fault-injection tests for unavailable protected custody and every migration interruption boundary, plus mutation tests for missing/altered package entitlements and protected-workflow policy seals

## Release configuration

The `zuuli-app-stores` environment must provide a Developer ID provisioning profile as `APPLE_DEVELOPER_ID_PROVISIONING_PROFILE_BASE64`. The release job fails closed unless it authorizes `F9AV5HKF6N.cash.free2z.zuuli`, its Keychain access group, and the imported signing certificate.

## Verification

- `cargo +1.97.1 test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml` (111 passed)
- `cargo +1.97.1 clippy --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml -- -D warnings`
- `npm test` in `wallet/zuuli` (205 Node tests and 113 Playwright tests passed; Vitest passed)
- macOS entitlement policy/contract tests (141 passed)
- `actionlint .github/workflows/zuuli-release.yml`
- `./scripts/check-rust-toolchain.sh`
- local ad-hoc `codesign` extraction proved the packaged entitlement plist round-trips through the exact post-sign verification commands

## Remaining acceptance evidence

A signed installed-app run on a physical user-presence-capable Mac must still prove that every seed read prompts as intended. This PR therefore references the issue without closing it, and the existing `needs-device` status remains accurate.

Refs #372

PR #527, opened 2026-08-23
security: bind macOS SBOMs to shipped artifacts

skyl:
## Summary
- safely mount each macOS DMG read-only, clone its payload without following symlinks, and detach before scanning
- generate exact-artifact SBOMs and digest/byte bindings for both DMG and app ZIP in packaging smoke and protected release
- independently re-extract each exact artifact before manifests/provenance; extend workflow mutation policy and DMG adversarial fixtures

## Safety boundary
The canonical absolute `/Applications` link is recorded as link bytes but never traversed. A private mount is detached normally or with bounded forced cleanup; a live mount is never recursively removed. Verification uses a fresh private extraction and rejects a Syft-root-only injection.

## Verification
- artifact SBOM tests 7/7, including real local DMG create/mount/detach and mutation canary
- `actionlint` for both workflows
- `npm run release:verify`
- Prettier and `git diff --check`

Refs #379

This intentionally leaves Linux AppImage/deb/rpm extraction open.

### DIFF
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -427,19 +427,26 @@ jobs:
           [[ ${#zips[@]} -eq 1 && -f "${zips[0]}" ]] || { echo "expected exactly one macOS app ZIP" >&2; exit 1; }
           node scripts/artifact-sbom.mjs prepare --artifact="${dmgs[0]}" --root=artifact-sbom-work/macos-dmg/root
           node scripts/artifact-sbom.mjs prepare --artifact="${zips[0]}" --root=artifact-sbom-work/macos-zip/root
-      - name: Download Syft for macOS shipped artifacts
+      - name: Scan macOS DMG shipped artifact
         if: runner.os == 'macOS'
-        id: desktop_artifact_syft
-        uses: anchore/sbom-action/download-syft@fbfd9c6c189226748411491745178e0c2017392d # v0.20.10
+        uses: anchore/sbom-action@fbfd9c6c189226748411491745178e0c2017392d # v0.20.10
         with:
           syft-version: v1.50.0
-      - name: Scan macOS shipped artifacts
+          path: wallet/zuuli/artifact-sbom-work/macos-dmg/root
+          config: wallet/zuuli/syft-artifact.yaml
+          format: cyclonedx-json
+          output-file: wallet/zuuli/artifact-sbom-work/macos-dmg/syft.raw.sbom.cdx.json
+          upload-artifact: false
+      - name: Scan macOS app ZIP shipped artifact
         if: runner.os == 'macOS'
-        env:
-          SYFT: ${{ steps.desktop_artifact_syft.outputs.cmd }}
-        run: |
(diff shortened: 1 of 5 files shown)


skyl:
Hosted workflow startup is corrected at exact head `174779d01a649d4897807dc92fc786a452a465f2`.

The first two heads used Anchore’s nested `download-syft` action. Although actionlint and the upstream action definition accept it, this repository’s live selected-action policy rejected the workflow before any job existed. The refreshed head uses only the already-approved, exact-SHA-pinned root `anchore/sbom-action` for the DMG and app-ZIP roots. Packaging smoke now starts normally.

Fresh local policy evidence after that change: both workflows actionlint-clean; release identity/workflow mutation policy green; focused workflow mutation test green; diff check clean. The prior real-DMG 7/7 suite remains applicable to unchanged extraction/inventory code. Any independent review should use only `174779d` or a later explicitly reviewed head.

skyl:
Refreshed normally onto merged #526/current main `157d7d56189080f0987e3669447cb5f7d36dfd7d` at exact head `0923f2f868ffac43e579e12c080e1d26181807aa` (merge commit, no rewrite). The integration is conflict-free and adds only the already-merged private-live files; the macOS SBOM/workflow diff is unchanged.

Exact refresh checks: both workflows actionlint-clean; release identity and artifact workflow mutation policy green; focused archive/symlink/policy tests 3/3; diff check; clean worktree. The real-DMG extraction/mutation suite remains 7/7 on the unchanged implementation. Fresh hosted CI is required at this exact head.

skyl:
Full exact-head regression evidence for `0923f2f868ffac43e579e12c080e1d26181807aa` is green locally after the final trunk integration:
- complete Vitest suite
- 209/209 Node/policy/contract tests (including real DMG create/mount/copy/detach and mutation canary)
- Apple credential/finalization boundary verifier
- 125/125 Playwright tests

The worktree remained clean. Hosted packaging must still prove the exact macOS artifacts and root SBOM action on this head, and independent review is still pending.

skyl:
Review feedback is fixed at exact head `cf5606acbe97026810e60287b83b7dc6da309584`.

DMG cleanup now treats the mount as potentially live from the instant `hdiutil attach` starts, rather than only after a zero exit. An attach timeout/nonzero therefore always enters bounded normal detach then forced detach. Recursive cleanup occurs only after a detach succeeds; if both detach attempts fail, the private mount tree is deliberately preserved for runner cleanup and the SBOM boundary fails closed.

Two injected adversarial tests cover (1) nonzero attach after a partial mount followed by successful detach/removal, and (2) nonzero attach plus both detach failures, proving the potentially live tree is not recursively removed. The real UDZO DMG test still exercises `/Applications` symlink non-dereference and fresh artifact re-extraction.

Exact-head local evidence:
- artifact SBOM suite: 9/9, including the real DMG and both new cleanup races
- full `npm test`: Vitest green; 211/211 Node/policy/contract tests; Apple credential boundary; 125/125 Playwright
- `git diff --check`: clean

Fresh hosted CI is running for this exact head. Independent review will resume after the higher-priority #525 correction.

PR #530, opened 2026-08-23
ci(rust): lint macOS and Windows wallet targets

### LINKED ISSUE #361: ci(rust): clippy does not cover Windows or macOS-only code paths
## Problem

The clippy gate added in #354 / PR #359 runs on **Linux only**. The `custody-platforms` job that covers macOS and Windows runs `cargo test`, not clippy.

So any code behind `#[cfg(windows)]`, or macOS-only paths, is **never clippy-linted on the target where it actually compiles**.

This is not hypothetical. PR #359 found two platform-conditional defects in a single `#[cfg(unix)]` function, and one of them was invisible on macOS entirely:

- `cargo clippy --fix` deleted `stat.st_ino as u64`, which is redundant on Linux/Android (`st_ino` is `u64`) and **load-bearing on Apple** (`st_dev` is `i32`). Clippy sees one target at a time, so the fix compiles where it runs and breaks elsewhere.
- `u32::from(stat.st_mode)` is a genuine widening on macOS (`mode_t` is `u16`) and the reflexive identity on Linux — so `useless_conversion` fires on Linux only.

Both were caught because someone ran the gate in a clean `ubuntu:24.04` container. The symmetric case — a lint that only fires on macOS or Windows — has **no coverage at all** today.

## Why it wasn't done in #359

Extending clippy to those targets would surface an unmeasured backlog on two more platforms, which is a separate piece of work from enabling the gate. Correctly kept out of that diff.

## What to do

1. **Measure first.** Run `scripts/check-rust-clippy.sh` on macOS and on Windows and report the warning count per target. That number decides whether this is one small PR or a project.
2. Decide whether to extend the `custody-platforms` job or add clippy to a separate per-platform job.
3. If the backlog is small, fix and gate. If large, gate `-D warnings` only for the shared/Linux surface and file the platform-specific backlog separately rather than blocking.
4. Be careful with auto-fixes on platform-conditional code — #359 is the cautionary example. Every `--fix` on a `#[cfg]`-gated block needs to be read against **all** targets it compiles for, not just the one clippy ran on.

## Acceptance criteria

- [ ] Warning counts reported for macOS and Windows.
- [ ] A decision recorded on whether clippy gates those targets.
- [ ] If gated: any `#[allow]` is scoped with a comment, never crate-level (matching the convention set in #359).
- [ ] The gate is demonstrated able to fail on each newly-covered target, not just to pass.

## Notes

Dev is macOS and the shared gate is Linux — `AGENTS.md` already documents that skew as a real historical bug source. This issue is the remaining half of closing it for lints.

skyl:
## Summary

- run the complete pinned wallet Clippy gate on native macOS and Windows custody hosts
- cover the shared plugin plus both Tauri application crates, including ZUULI-only target code
- prove each native lane can fail by compiling a private known-warning fixture under `-D warnings` before accepting the repository verdict
- expand native cache coverage and trigger the matrix for either app backend

## Measured result

The exact existing entrypoint reports **0 macOS warnings** across all three manifests on `x86_64-apple-darwin`. Hosted Windows is the authoritative Windows measurement: this PR closes the issue only if that exact-head lane both rejects the negative control and reports all three manifests clean.

## Verification

- `scripts/check-rust-clippy.sh --self-test`
- `scripts/check-rust-clippy.sh` (all 3 macOS manifests clean)
- `actionlint .github/workflows/zuuallet.yml`
- `scripts/check-rust-toolchain.sh --self-test`

Closes #361


skyl:
Integrated merged #528/current main `8237454` and pushed replacement exact head `1db8bda`. Relative scope remains the same three CI/docs files. Exact integrated `--self-test` rejects the known clippy warning on x86_64-apple-darwin and `git diff --check` is clean; hosted macOS/Windows negative-control plus full all-three native clippy logs are being re-earned on this head.

skyl:
Replacement exact head `0976670` addresses the three independent-review blockers:

1. Native clippy moved into `zuuli.yml` as `rust_native_clippy` and is now an explicit dependency of the branch-protected `gate`; the gate verifier understands the union of ZUULI and Zuuallet-Rust selectors.
2. `wallet/rust-toolchain.toml`, `scripts/check-rust-toolchain.sh`, and `scripts/check-rust-clippy.sh` now select the native matrix, including Zuuallet-only Rust changes.
3. The negative fixture has a build script that compares Cargo`s actual `CARGO_CFG_TARGET_OS` with the matrix target before the lint verdict can count. On macOS the correct-target control passes; the exact previously demonstrated `CARGO_BUILD_TARGET=x86_64-unknown-linux-gnu` mutation now fails for target mismatch.

The required-gate policy now fail-closes on detaching the job, changing its native matrix/selector, or replacing either the target-bound self-test or lint command (46 current-workflow mutations). Android gate policy was re-bound to the updated canonical selector/gate and passes 20 mutations. Actionlint, toolchain 12-mutation policy, action-pins 123-mutation policy, and diff checks are green locally. Final review and CI must target `0976670` only.

skyl:
The exact `0976670` hosted failure was a real stale mutation fixture and is fixed at replacement head `8ce69e2`: `check-zuuli-linux-image.mjs` now mutates the updated schema selector containing `check-rust-clippy.sh`. Its full self-test (38 negative cases plus runtime cases) and live policy pass locally. All `0976670` CI is superseded; only `8ce69e2` counts.

skyl:
Adversarial review found that exact head `8ce69e2` could be weakened by adding step-level `if: false` to both native control steps while preserving their exact `run:` strings. That head and its CI are superseded.

Replacement head: `8c0acec1342eaa62958985fce473b07b7c91ed92`. The policy now requires exactly one copy of each native control with the exact `{name, shell: bash, run}` key set, so conditions, soft-fail decorators, duplicate named controls, and other execution modifiers fail closed. Two new live-workflow mutations prove both `if: false` bypasses are rejected. Exact local proof is green: 70 source-policy + 7 gate-verdict + 48 live-workflow mutations, current policy, Linux-image policy/self-test, Android gate policy/self-test, actionlint, and diff check.

Only checks and review for exact `8c0acec` count.

skyl:
Adversarial current-source finding confirmed: exact head `8c0acec` allowed the native checkout step to add `with.ref` for a prior clean SHA, so both exact lint commands could certify stale code. That head/CI are superseded.

Replacement head: `9eab5feabfd6fc1751eb38cface6111f9eb248dd`. The complete native job is now compared against an exact reviewed canonical contract: selector, matrix, current-head checkout with no decorators, submodule initialization, pinned toolchain resolution, pinned toolchain/cache inputs, and both unconditional controls. This also prevents an injected source-reset step between checkout and lint. New live mutations prove stale-checkout and intervening-reset attempts fail. Exact local proof: 70 source-policy + 7 gate-verdict + 50 live-workflow mutations, current policy, Linux-image and Android gate policies/self-tests, actionlint, and diff check all green.

Only checks/review for exact `9eab5fe` count.

skyl:
Integrated merged #532/current main `1c63d02fe756b0e72fa252c1916e68641b8e1cba` with a normal merge; new exact head is `4680fce43f96ec7416ad3931bddad4e864413b0e`. Combined relative diff remains the six intended CI/policy/docs files. Post-integration local proof is green: 70 source-policy + 7 gate-verdict + 50 live-workflow mutations, current action policy, Linux-image 38-case policy, Android gate 20-case policy, Rust toolchain policy, native macOS target negative control, actionlint, and diff check.

All pre-`4680fce` hosted checks/reviews are superseded for merge.

skyl:
The exact `4680fce` Windows lane proved the target-native gate and then exposed a real pre-existing Windows build defect: Tauri could not compile classic Zuuallet because `wallet/zuuallet/src-tauri/icons/icon.ico` was absent. The gate correctly failed rather than certifying only Linux/macOS paths. Its old summary incorrectly called every nonzero Cargo result a “clippy warning,” obscuring the actual target build error.

Replacement exact head `b5c7685ff50cf95fe218c016fa671135ed786e63` adds the required Windows icon, mechanically derived from the already-public Zuuallet PNG (32px and 16px ICO entries), and makes the gate summary distinguish lint, compile, and missing native-prerequisite failures. ImageMagick parses both ICO entries; the target-bound negative control, action policy, and diff checks are green locally. Existing ShellCheck notices are unchanged lines SC1007/SC2016. All earlier hosted checks/review are superseded; merge now requires the full exact-head matrix, especially Windows rejecting the negative control and reporting all three wallet crates clean.

skyl:
Selector/config blocker corrected at exact head b891785d2e1ac128b9e72c4b03b91233e75eac4a.

The change detector now selects target-native Clippy for root/shared Cargo manifests and config, wallet parent workspace/config surfaces, every current or future wallet Rust source, and future crate manifests/Clippy config. The current-source policy parses the actual named detector step and checks both real case arms; eight new mutations prove each selector class fails closed if removed. The Android detector digest and Linux schema-selector mutations were updated coherently.

Local evidence:
- check-github-actions-pins --self-test + live: PASS (70 source-policy, 7 gate-verdict, 58 current-workflow mutations; 148 refs / 14 files)
- check-zuuli-android-gate --self-test + live: PASS (20 mutations)
- check-zuuli-linux-image --self-test + live: PASS (38 negative cases + runtime)
- check-rust-toolchain --self-test + live: PASS (12 drift cases; 1.97.1)
- actionlint .github/workflows/zuuli.yml: PASS
- git diff --check: PASS

Fresh hosted checks are running for this exact head.

skyl:
Replacement exact head `11345155d369281ce920dfa1573c573f1359340a` normally integrates merged main `8f1cdb24825c810e4a64b1c421551d5923299604` and closes the last selector review gap.

Both change-detector arms now include a future wallet crate's own `.cargo/*` configuration, in addition to root/wallet ancestor configs, manifests, Clippy configs, and Rust sources. A 59th live-workflow mutation proves removing that nested config selector is rejected. The Android detector digest and merged Linux-image fixtures were rebound to the exact integrated detector.

Post-integration proof: 70 source-policy + 7 gate-verdict + 59 live-workflow mutations; 20 Android mutations; 43 Linux-image negatives plus runtime cases; 12 toolchain drift cases; target-native macOS negative control; actionlint across all four touched workflows; and diff check. All pre-`1134515` hosted checks and reviews are superseded.

skyl:
Exact hosted native evidence for head `11345155d369281ce920dfa1573c573f1359340a` is now complete:

- macOS (`aarch64-apple-darwin`): the target-bound `-D warnings` negative control failed as required, then all 3 wallet Rust crates passed pinned Clippy 1.97.1.
- Windows (`x86_64-pc-windows-msvc`): the target-bound `-D warnings` negative control failed as required, then all 3 wallet Rust crates passed pinned Clippy 1.97.1.
- Measured platform backlog: 0 warnings on both newly covered native targets.

All other hosted jobs are green; only the required aggregate gate is finishing. The verification invocation in the initial PR body should be read as `scripts/check-rust-clippy.sh --self-test <macos|windows>` on the corresponding native host.

PR #532, opened 2026-08-23
fix(deps): clear quick-xml advisories

skyl:
## Summary
- update the two stale Tauri build graphs from `plist 1.8.0` to `1.10.0`
- move their transitive `quick-xml` from `0.38.4` to `0.41.0`, matching ZUULI
- remove exactly RUSTSEC-2026-0194 and RUSTSEC-2026-0195 from the reviewed ignore list

## Verification
- Cargo precise update changed only `plist` and `quick-xml` in the plugin and Zuuallet locks
- `scripts/check-rust-deny.sh` passes advisories, bans, licenses, and sources across all three locked roots
- repository-wide locked clippy is running locally; hosted exact-head gates provide the authoritative full build proof

Refs #351


skyl:
Integrated merged #528/current main `8237454` and pushed replacement exact head `4289603`. Relative diff remains exactly `wallet/deny.toml` plus the plugin/Zuuallet locks; the lock movement is still only `plist 1.8.0 -> 1.10.0` and `quick-xml 0.38.4 -> 0.41.0`. Full advisory/bans/licenses/sources policy reran green across all three roots after integration; hosted exact-head build gates are restarting.

skyl:
Independent review is clean at exact integrated head `428960398ba4908e6ef76cdf8aa017df4db44222`: the relative diff is only the two intended ignore deletions and the two exact plist/quick-xml lock moves; ZUULI already resolves the same versions. All-root deny plus locked metadata/resolution pass. Negative mutation restoring the vulnerable versions under the no-ignore policy reproduces both advisories in exactly the plugin and Zuuallet roots while ZUULI stays clear. Merge remains gated on the full hosted exact-head matrix.

Issue #387, opened 2026-08-19
ZUULI distribution: source-control and publish polished App Store / Play listing media


skyl:
## Problem

ZUULI can reproducibly upload signed binaries, but its store presentation is not reproducible or auditable.

Evidence on current `origin/main` (`fb9f2ac29cd08cb4aed71c8c41655942ba4ce65d`):

- No ZUULI store metadata, screenshots, app previews, Play feature graphic, or locale directory is tracked in the repository.
- `scripts/mobile-release.sh` deliberately invokes `fastlane supply` with `--skip_upload_metadata true`, `--skip_upload_images true`, and `--skip_upload_screenshots true`.
- The Apple path uses `altool` only to validate/upload the IPA; it does not manage App Store Connect localizations, beta metadata, screenshots, or tester-facing presentation.
- Protected build `0.1.0+10` successfully reached TestFlight and Play internal, so binary delivery is working; listing quality is now the missing distribution layer.
- The public Apple lookup returns no released version and the public Play listing is unavailable, as expected for TestFlight/internal-only distribution. Current listing content therefore cannot be audited without authenticated read access.

This issue is intentionally separate from #301, which owns the app icon and Play feature graphic design, and from the UI/screenshot polish audit. It consumes their approved outputs.

## Scope

Create a reviewable, locale-ready store-media source of truth and a protected, fail-closed publish/read-back path for App Store Connect and Google Play.

## Acceptance criteria

- [ ] Add a documented `wallet/zuuli/store/` (or conventional Fastlane) hierarchy for App Store Connect and Play metadata/localizations; start with `en-US` while remaining compatible with the localization program in #322.
- [ ] Track reviewed app name/subtitle or short description, full description, keywords where applicable, support/privacy/marketing URLs, category/content-rating notes, and release/beta notes without embedding credentials.
- [ ] Track final screenshots for every phone/tablet class the current stores require. Capture them from signed or release-equivalent builds after the UI audit, with deterministic fixture state, no seed phrase/private data, correct safe areas, and no debug/mock disclosure unless the image is explicitly a demo.
- [ ] Consume the approved icon/feature-graphic outputs from #301; do not create a competing brand asset source.
- [ ] Add dimension/format/content validation that rejects missing, stale, transparent-when-forbidden, mis-sized, or duplicate media before any store call.
- [ ] Add an authenticated **read-only audit** mode that reports listing/localization/media completeness and build processing/tester state without printing credentials or sensitive tester identities.
[...]

PR #536, opened 2026-08-23
fix(zuuli): refresh deterministic store media

skyl:
## Summary

- pin the creator-profile capture to the authoritative explicit `page=1` catalog request and add a regression test that rejects missing-page, page-2, and page-size drift
- refresh all 20 deterministic ZUULI store candidates from exact public `main` source `bda85b9beefce2f1fb19639f3b67e5035b94f503`
- retain the publication gate (`publicationReady: false`) so final store-submission approval remains an explicit owner review

The stale capture was reproduced first: the creator profile timed out because its live request now includes `page=1` while the exact allowlist did not. With the corrected contract, the pinned capture completed two identical passes and exercised the exact public request set.

All 20 outputs were inspected at original resolution across the five declared device sets. They contain only the fictional Example Editorial fixture, match the current graphite/Plex/violet visual layer, and show no private account data, seed material, email addresses, or unexpected identifiers.

## Verification

- `npm run store:validate`
- `npm run test:store-listing` (56 tests)
- `npm run typecheck`
- `npm run release:verify`
- pinned `node scripts/store-screenshot-capture.mjs --write` (20 screenshots, two identical passes)
- exact-current-source capture-record validation
- `actionlint .github/workflows/*.yml`
- `git diff --check`

Refs #387


skyl:
Independent review at exact head `ff2816b6ec4d1416cb76e67e27d674d230ce548c`:

- inspected all 20 regenerated images as five labelled device-set contact sheets; dimensions/layout are coherent, content is the fictional Example Editorial fixture, and no private/tester identity or debug disclosure is visible
- reviewed the exact request-contract diff: creator capture now pins the sorted canonical `page=1&page_size=12` query, and the focused regression rejects omitted page, page 2, and changed page size
- reran the 19-test screenshot/provenance suite and the store manifest validator locally; both pass, with 20 unique entries and `publicationReady: false`
- verified the worktree is clean and the PR remains explicitly `Refs #387`, preserving owner/legal/store review as residual work

No review findings. Holding merge for the exact-head hosted gate and packaging matrix.

Issue #246, opened 2026-08-07
Wallet release gate: 24-word creation, verified recovery ceremony, and network-aware recipients


skyl:
## Goal

Make seed backup and first real-ZEC use safe and unambiguous.

## Requirements

- Generate new wallets with a 24-word BIP39 phrase; remove the 12-vs-24 user choice.
- Continue restoring all valid supported BIP39 phrase lengths for compatibility.
- Require an explicit backup acknowledgement before normal wallet use.
- Provide a safe reveal flow backed by native custody; never log, persist, or copy the phrase outside the explicit user action.
- Add a separate-profile restore drill that verifies the seed-derived wallet identity/address before declaring recovery proven.
- Make address validation network-aware; a syntactically valid Testnet address must not be presented as valid in the Mainnet app.
- Validate ZIP-321 requests and Unified Address receiver compatibility before proposal.
- Add tests for phrase lengths, wrong-network recipients, restore identity, cancellation, and legacy-native custody migration.
- Gate the first real-ZEC smoke test on successful recovery proof and merged crash-safe send support (#235).

Tracked by #227.

skyl:
The 2026-08-07 UI audit found two concrete auth paths that should be covered by this release gate:

- #267: a user can leave the Zcash login flow after local identity creation/while seed reveal is active; in-flight work is not invalidated and a later attempt can observe an initialized wallet and bypass the abandoned backup ceremony.
- #268: a returning user with no local wallet is offered only “Create a Zcash identity,” not restore/external signing, encouraging creation of a second identity.

Please include both in the recovery proof: mandatory backup state must survive method switching/app restart, and restore must verify the recovered identity before login continues.

skyl:
[...]

PR #538, opened 2026-08-23
fix(wallet): enforce 24-word wallet creation

skyl:
## Summary

- make the native wallet plugin the parameterless authority for new 24-word recovery phrases; reject stale or unknown create arguments
- remove phrase-length selection and count parameters from ZUULI and classic callers, including funding-login and link-identity creation paths
- preserve restore support for exactly 12, 15, 18, 21, and 24 words across wallet and identity flows
- bind the native, bridge, caller, restore, custody, and lease boundaries with fail-closed source policy and mutation tests; extend exact change selectors to every classic source the policy reads
- rebuild the tracked classic distribution and document the creation/restore contract

## Validation

- plugin tests: 123 passed; all-target clippy with warnings denied
- ZUULI: full Vitest and Node policy suites passed; exact-source Playwright 131/131 before prerequisite integration; focused post-integration mnemonic policy 3/3
- seed boundary: 16 Vitest assertions plus 84 policy/live mutations
- Actions policy: 70 source, 7 verdict, 99 workflow mutations; live scan 163 refs across 14 workflow files
- hosted-order changes replay: toolchain, public boundary, Linux image, Android, WASM, permission, and exact change detector all passed; detector selected zuuli=true and zuuallet_schema=true
- ZUULI production build verified a fresh 116-byte WASM artifact; classic build reproduced tracked dist exactly
- native restore fixtures cover valid 12/15/18/21/24-word BIP39 phrases

This is the independently shippable 24-word creation slice. The broader recovery-proof and wallet-management work remains in the parent issue.

Refs #246

PR #539, opened 2026-08-23
fix(zcash): validate network-aware payment requests

skyl:
## Summary

- require exactly one ZIP-321 payment instead of selecting the first entry
- validate URI recipients against the native wallet network and reject Unified Addresses with no receiver this wallet can pay
- decode only supported text/empty memos, clear stale renderer intent on parse failure, and retain proposal-boundary validation
- add Rust, mock, renderer, browser, and authority-policy regressions plus honest command documentation

## Proof

- `cargo test --locked --all-targets` — 129 passed
- `cargo clippy --locked --all-targets -- -D warnings`
- `npm test` — Vitest 692 passed / 5 skipped; Node 326 passed / 4 skipped; Playwright 132 passed
- `npm run typecheck`
- `npm run build`
- `actionlint`

Refs #246


skyl:
Exact tested head: `6a47a30967671e80f8af517de2072b520b2e0bb0`, normally integrated with current main `67d59dec7a53d44763870a4d564157a0f8573db4` after #538. Integration was conflict-free; the combined native suite proves both the 24-word creation policy and this independent payment classifier (129/129). Full local ZUULI proof: Vitest 692 pass / 5 expected skip, Node policy 326 pass / 4 expected skip, Playwright 132/132, typecheck and production build green. Rust clippy/fmt, actionlint, toolchain, permission-schema, immutable-action policy, and exact diff checks also pass.

Issue #326, opened 2026-08-09
ZUULI: let the user change the UI language


skyl:
## Problem

There is no way to change ZUULI's UI language, and no concept of a UI language at all. `index.html:2` is a hardcoded `<html lang="en" class="dark">` that is never mutated at runtime.

Even once strings are extracted, the app needs to (a) pick a sensible default from the system, (b) let the user override it, and (c) remember the choice.

### 1. No preference storage exists anywhere

The only `localStorage` use in the entire app is the auth token (`src/lib/api/http.ts:23,33,34`). `src/store/session.ts` and `src/store/wallet.ts` use **no Zustand `persist` middleware**. There is no preferences slice to extend — one has to be created.

### 2. There is no settings screen

`src/App.tsx:91-181` routes are `/login /search /creator /profile /kyc /wallet /ai /live /articles /buy`. **No `/settings`.** Grepping `settings|Settings|Preferences` across `App.tsx` and all of `src/components/layout/` returns nothing.

So this issue has to decide where the control lives:

1. **`src/components/layout/TopBar.tsx:140-167`** — the account dropdown. Add a `<Languages/> Language` item above the `Sign out` separator at `:163`. Least work, no new route, mirrors the existing Svelte selector. But it's inside an *account* menu, and language is a device preference that should work signed-out.
2. **`src/features/profile/index.tsx`** — "Edit profile" (`:98-101`). Closest thing to a settings screen, but it edits *public profile* data posted to the backend. A device-local language preference is a semantic mismatch.
3. **A new `/settings` route** — cleanest long-term, and the natural future home for theme, currency display, and node config. Most work.

Recommendation: option 3 if a settings screen is wanted anyway, option 1 as the pragmatic interim. Worth an explicit call rather than defaulting.

## Reading the system locale

Use **`navigator.languages[0]`**. It behaves identically in the Tauri webview and in browser dev/mock mode, needs no new Rust dependency, no Tauri capability, and no mock fallback — which matters given ZUULI's dual-runtime architecture (`src/lib/platform.ts`). This is what `ts/svelte/free2z/src/lib/i18n.ts:36` already does.

[...]

Issue #325, opened 2026-08-09
ZUULI i18n: adopt an i18n framework and extract all UI strings


skyl:
## Problem

ZUULI has no i18n framework and no message catalog. Every user-visible string is a hardcoded English literal in JSX. This issue covers picking a stack and extracting the strings.

**Prerequisites:** the text-minimization pass and the number-parsing fix should land first. Do not extract strings that are about to be deleted, and do not multiply a live parsing bug across locales.

## Scope

~781 user-visible literals across 118 files (~20,388 LOC), realistically **700-900 translatable messages** after minimization. Densest files:

| File | ~strings |
|---|---|
| `src/features/creator/index.tsx` | 46 |
| `src/features/wallet/Send.tsx` | 41 |
| `src/features/live/Room.tsx` | 32 |
| `src/features/ai/index.tsx` | 29 |
| `src/features/wallet/Onboarding.tsx` | 28 |
| `src/features/search/index.tsx` | 27 |
| `src/features/ai/PersonalityManager.tsx` | 27 |
| `src/features/profile/index.tsx` | 25 |
| `src/features/articles/pages/Feed.tsx` | 24 |
| `src/features/articles/pages/Author.tsx` | 24 |
| `src/features/buy/BuyTab.tsx` | 24 |
| `src/features/kyc/index.tsx` | 22 |

Plus toast-surfaced strings in `src/lib/api/free2z.ts` (~11 error strings) and `src/lib/oauth/transport.ts` (~9).

**Do not translate** `src/lib/api/mock-data.ts` (107 strings) — demo-mode fixture prose.

## Framework choice

Recommendation: **`i18next` + `react-i18next` + `i18next-icu`**. ICU MessageFormat gives real CLDR plurals, gender, and `select`, and lazy per-locale chunks fit an `App.tsx` that is already aggressively code-split. Lingui yields smaller bundles and compile-time extraction but adds a Babel/SWC step to what is today a plain Vite + tsc build.

Either way, lift `ts/svelte/free2z/src/lib/locale.ts` verbatim — it is stack-agnostic and already implements `normalizeLocale()` and the `resolveLocale()` precedence chain.

## Why extraction is not mechanical

Budget roughly two weeks. This cannot be done with a regex:

- Most literals are JSX with interpolation — e.g. `` `Sent ${formatTuzis(n)} to ${name}` `` at `src/features/creator/index.tsx:689`.
[...]

PR #540, opened 2026-08-23
feat(wallet): add safe legacy import preview

skyl:
## Summary

- add a parameterless native preview for only the fixed `com.2zinc.zuuli` sibling, with structured absent/ready/blocked/unsupported results
- inventory validated single- and multi-wallet layouts while redacting wallet IDs/names and returning only exact DB filenames, sidecar/custody presence, account counts, and domain-separated UFVK hashes
- inspect SQLite only through private DB/WAL/SHM snapshots whose copied hashes and source identities/hashes are verified; reject links, transient journals, corrupt/ambiguous databases, invalid/duplicate accounts, and source mutation
- generate command permissions and byte-identical tracked target schemas, grant the explicit mobile capability, and document the non-importing safety boundary

## Non-goals

This does not read/decrypt custody, call `SeedStore`, write a journal, mutate either app-data tree, publish runtime wallets, or add UI. Authorization, import, custody validation, crash-safe publication, and UI remain later slices.

## Local proof

- `cargo test --locked --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml`: 140 passed
- focused preview safety suite: 11 passed, including live WAL, zero-mutation, unreadable custody, source mutation/replacement, hard-link, journal, corrupt DB, missing/NULL/invalid/duplicate UFVK, and unlisted DB cases
- `cargo clippy --locked --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml --all-targets -- -D warnings`
- `cargo test --locked --manifest-path wallet/zuuli/src-tauri/Cargo.toml`: 14 passed
- nonce-forced Zuuallet schema regeneration plus permission negative controls/parity: 38 commands, 76 identifiers, 3 byte-identical tracked schemas, zero drift
- local Windows cross-check reached target dependencies but stopped before this crate because macOS lacks Windows SDK C headers required by `ring`; required hosted `Rust / native lints (windows)` is authoritative

Refs #272

PR #541, opened 2026-08-23
feat(zuuli): add lazy i18n kernel

skyl:
## Summary

- adopt `i18next`, `react-i18next`, and `i18next-icu`, port the framework-neutral locale resolver, and bootstrap one normalized locale before React mounts
- lazy-load exact-parity English, Spanish, and French catalogs; fail closed on missing, orphaned, invalid, or malformed ICU messages with mutation coverage
- migrate the common shell, navigation, root/not-found/retry UI, plus locale-bound `timeAgo` and explicit date formatting while preserving ASCII-only ZEC behavior
- record the ICU/lazy-catalog decision and keep language selection (#326) and feature-wide extraction out of this bounded slice

Refs #325

## Proof

- `npm ci`
- `npm run typecheck`
- `npx vitest run src/i18n src/lib/format.test.ts src/components/common/SectionLoadError.test.tsx src/components/layout/TopBar.test.tsx src/components/layout/TopBar.signed-in.test.tsx src/components/layout/navigation.test.ts` — 100/100
- full Vitest phase — 718 passed, 5 expected skipped
- full Node policy phase — 326 passed, 4 expected skipped
- `ZUULI_PW_PORT=1454 npx playwright test --timeout=120000` — 134/134
- `npm run build` — 3,286 modules; distinct `en` (1.55 kB), `es` (1.72 kB), and `fr` (1.83 kB) chunks; fresh WASM verified
- GitHub Actions pin policy, Rust toolchain policy, `actionlint`, and `git diff --check`

The default 60-second whole-test Playwright budget produced a rotating creator-catalog timeout under local aggregate load (133/134 twice); each affected viewport permutation passed alone in ~15 seconds, and the unchanged full suite passed 134/134 with a 120-second whole-test budget. No assertion failed in the final run.

## Remaining #325 scope

This does not claim full extraction: feature-local copy, remaining ICU conversions, and markup-bearing messages are intentionally left for subsequent reviewable slices. Mock fixture prose remains excluded. The issue stays open.


skyl:
Replacement head after #540 integration: `7289b46118cedd8a41b006bdf5b9112818ce3def`.

- normal merge parents: prior reviewed i18n head `e04b0da00a0eab346d063431831ac13be299d82f` + exact current main `ec8922e564c83ca567281c0e19186776023e81ef`
- conflict-free merge; the scoped PR diff against current main remains the same 30 i18n/shell files
- `npm run typecheck` — pass
- focused i18n/format/shell Vitest — 100/100
- `npm run test:mobile-webview-authority` — policy mutations, committed capability, and focused Playwright pass with #540's exact legacy-preview authority
- `npm run build` — pass; fresh WASM verified and distinct `en`/`es`/`fr` chunks retained
- GitHub Actions pin policy, Rust toolchain policy, `actionlint`, and `git diff --check` — pass
- clean worktree and exact remote head verified

Fresh hosted proof for this replacement head is now running. No merge performed.

skyl:
Mandatory mutation-gate replacement head: `3de7d9dc31fd105c898012fa6faf45f01360017f` (normal follow-up; no history rewrite). Prior `7289b46` CI is superseded.

The raw substring policy was comment-satisfiable. It is now a TypeScript-AST contract: locale loaders must be executable returned dynamic imports, no static import/top-level prefetch/extra import is accepted, every supported loader must exist, and catalog consumers must be real `MESSAGE_KEYS` member expressions rather than comments or strings.

Exact mutants killed locally:

1. missing catalog leaf → missing-key rejection
2. added catalog leaf → orphan-key rejection
3. invalid `bad-key` segment → invalid-key rejection
4. empty leaf → malformed-value rejection
5. broken plural syntax → ICU parse rejection
6. static English catalog import with a fake lazy comment → static-import rejection
7. top-level eagerly-started English import returned by the loader → returned-import rejection
8. eager French prefetch beside otherwise-valid loaders → global lazy-only rejection
9. unreachable import after an early return → executable-return rejection
10. removed French loader → missing-loader rejection
11. comments/string literals spelling `MESSAGE_KEYS.commonLoading` → no-consumer rejection
12. every real declared key consumer replaced with `undefined /* original member expression */` in syntactically valid production sources → all declared keys reported missing

The saved/browser fallback expectation now asserts independent literal `en`; Spanish/French Playwright expectations remain hardcoded independently from catalogs and use externally supplied locale/storage inputs.

Proof on exact tree:

- `npm run typecheck` — pass
- i18n mutation suite — 32/32
- full Vitest — 727 passed, 5 expected skipped
- focused i18n Playwright — 2/2
- #540 mobile authority policy mutations — 6/6; browser — 1/1
- production build — 3,286 modules, separate en/es/fr chunks, fresh WASM verified
- action pin/toolchain/actionlint/diff gates — pass
- exact current main remains `ec8922e564c83ca567281c0e19186776023e81ef`; worktree clean

Fresh hosted proof is required and now running. No merge performed.

skyl:
Reviewed adversarially on the owner's behalf. **Requesting changes** — one blocking defect (proven, not theorised), one live regression worth fixing in the same PR. The kernel itself is well-built and the catalog policy is genuinely fail-closed; it's the chosen failure *mode* that's the problem.

## Blocking: a catalog integrity failure blanks the wallet

`main.tsx:63-90` awaits `initializeAppI18n()` **before** `createRoot().render()`. If that rejects — a drifted catalog, a malformed ICU string, a failed dynamic `import()` of the locale chunk — nothing is ever rendered and the only artifact is a `console.error`.

Deleting `navigation.home` from `en.json` against the running dev server:

```json
{"root": {"rootHTMLLength": 0, "visibleText": ""},
 "logs": ["error: ZUULI locale bootstrap failed Error: en catalog key mismatch; missing=[navigation.home]"]}
```

`#root` innerHTML length **0**. No text, no error UI, no reload button. In a wallet.

The part that decided this for me: the PR edits the comment block at `main.tsx:9-17`, deleting the words *"dependency-free"* and *"we can't trust **any** higher-level component to be mountable"* — and then makes `RootFallback` depend on `useTranslation()`. The crash frame now depends on the exact subsystem whose failure it exists to survive, so it is unreachable in precisely the new failure class.

And the trigger isn't hypothetical. `CATALOG_LOADERS` (`i18n/index.ts:16-20`) is a code-split chunk fetch. On the browser build, a stale `index.html` served after a deploy that purged the old hashed chunk gives a 404 → blank app. Today zero chunks block first paint; after this, one does.

**Fix, keeping the fail-closed intent:** catch inside `mountApplication` and render `RootFallback` with **hardcoded English literals**. The crash frame is the one surface that must never be translated.

## Also blocking-ish: `formatDate`/`timeAgo` now discard the region subtag

`i18n/index.ts:47` passes the normalized **2-letter language** into `configureFormattingLocale`, so `format.ts:271,289` format with `"en"`/`"es"` instead of the user's actual `navigator.language`. Previously `toLocaleDateString(undefined, …)` used the system regional locale.

| user | before | after |
|---|---|---|
| en-GB | `31 Dec 2024, 20:02` | `Dec 31, 2024, 08:02 PM` |
| es-MX | `31 dic 2024, 08:02 p.m.` | `31 dic 2024, 20:02` |

That's a live regression on wallet transaction history (`History.tsx:167`, `components.tsx:197`) for every non-US user. One-line fix: normalize only for catalog selection, pass the full BCP-47 tag to `configureFormattingLocale`.

**And it's untested for the classic reason** — making `formatDate` ignore its `locale` argument entirely stays green, because the test computes its expected value with `"en-US"` and the runner is pinned to `en-US`. The bug cannot appear where it is measured.

## The guards: catalog policy is real, the build-boundary tests are not

`catalog-policy.ts` fails closed on everything it claims — deleting a key, adding an orphan, malformed ICU, collapsing a plural form all go red. Good.

The two `build-boundary.test.ts` tests are weak:

- **`:20-30`** is named *"keeps every production catalog behind an explicit dynamic import"* but only greps the raw text of `index.ts`. A static `import esEager from "@/i18n/locales/es.json"` in **any other module** pulls all three locales into the entry chunk with the test still green. The namespace form `import * as enEager` also slips past. The real proof already exists and is far better — `tests/i18n-kernel.pw.ts:11-31` counts actual network requests and asserts `["es"]`. Drop the source-grep and lean on that, or assert against `dist/` chunk contents.
- **`:32-40`** uses `source.includes("MESSAGE_KEYS." + property)`, so `navSearch` is satisfied by `navSearchAction`, `commonRetry` by `commonRetrying`, `navMore` by `navMoreDescription`, and so on. **Nine of 47 keys can silently lose their only consumer.** Fix with a word-boundary regex.

## Answers to what I asked

**Lazy — yes, verified in the build**, not just claimed: three separate chunks (`en` 1,550 B / `es` 1,722 B / `fr` 1,829 B), and `grep "Cargando"` hits only the `es` chunk, so `test-provider.tsx`'s static import of all three is correctly tree-shaken out.

**But the entry chunk grows and the body doesn't say so:**

```
main   554,492 B raw / 175,123 B gz
#541   649,164 B raw / 204,632 B gz
delta  +94,672 B (+17.1%) / +29,509 B (+16.9%)
```

That's `i18next` + `react-i18next` + `i18next-icu` + `intl-messageformat` in the entry chunk, bought to translate 47 shell strings. It amortizes as extraction proceeds — but report it, rather than only the 1.5 kB catalogs.

**Missing key/locale both throw**, with `fallbackLng: false` deliberately. So it never renders a raw key and never renders empty — right instinct, and the reason B1 hurts: the same strictness that prevents a half-translated UI also takes the whole app down.

**Scope is honest.** 47 keys against #325's 700–900 estimate is ~5-6%, and the body says `Refs #325` with an explicit remaining-scope section. No over-claim.

**English does render differently in three places.** `timeAgo` is the widest: `"3m ago"` → `"3 minutes ago"`, `"1d ago"` → `"yesterday"`, `"just now"` → `"now"` — roughly 2× width on article, comment, stream and activity cards. #325 asked for `Intl.RelativeTimeFormat` so this is requested rather than accidental, and English fit is audited at 320px; `{ style: "narrow" }` would give `"3 min. ago"` if you want the compact feel back. Also a copy downgrade: the More-sheet screen-reader description goes from `"Articles and account destinations"` to `"2 additional destinations"` (`Sidebar.tsx:240-244`) — a count is strictly less useful than naming what's in the sheet.

**RTL/bidi clean** — no `dir`, no bidi handling, only `documentElement.lang`. #327 correctly untouched, and #508's override characters aren't compounded.

**Persistence is read-only** — `localStorage` key `free2z-locale`, one `getItem` in try/catch, never written (the writer is #326). No contact with Zustand, the Knox token, or the paid-intent/checkout-return paths.

## Constraints

Truncation, money formatting, semantic tokens and fixture privacy all clean — `formatZec`/`parseZecToZatoshis`/`parseTuzis` untouched, all 66 `format.test.ts` cases pass.

**Mobile fit is only proven in English, and it's tighter than it looks.** Measured at 320px: fr `"Portefeuille"` needs **64px in a 64px cell** — it fits with literally zero headroom, `scrollWidth === 320`. The next locale breaks it: German gives `"Brieftasche"`/`"Startseite"`. `viewport.pw.ts` and `touch-targets.pw.ts` gained **no locale dimension**, and the new `i18n-kernel.pw.ts` opens es-MX at 320×568 but only asserts visibility, not overflow. Adding `es`/`fr` — and ideally a synthetic long-string pseudo-locale — to the viewport audit is the guard this needs.

## Worth filing after the blockers

`search-route.ts:1`'s `SEARCH_INPUT_LABEL` still holds the same string `TopBar.tsx:108` now translates, so after extraction the search page's aria-label stays English while the top bar's is localized. Mixed-language output today for es/fr users at `StreamCard.tsx:116` (`Started il y a 5 heures`) and `Room.tsx:318` — localizing `timeAgo` without its English wrappers. `i18n-kernel.pw.ts` uses `browser.newContext()`, which does **not** inherit `playwright.config.ts` `use`, silently dropping `timezoneId: "UTC"` — no flake today since neither test renders a date, but it's a hole in the #453/#455 convention. And `test-provider.tsx` omits `parseMissingKeyHandler`, so component tests are strictly more permissive than production: a missing key renders the raw key in tests and crashes the app in prod.

Fix B1 and B2 and I'll merge — the rest are follow-ups.

skyl:
Mutation-audit replacement head: `f8edcb5a959193ebc1da88b47a9e5e82fb89521f` (supersedes `3de7d9dc31fd105c898012fa6faf45f01360017f`).

The AST source-policy suite now has executable, exact-diagnostic negative controls for every reviewed branch:

- missing and duplicate `CATALOG_LOADERS` declarations
- non-object registry initializer
- spread entry and method entry
- duplicate locale and unsupported locale entries
- wrong locale-to-catalog mapping
- a second catalog import inside one loader
- parse-invalid production source

Each exact text mutation first proves it has exactly one live target; each policy rejection asserts the complete expected diagnostic. These extend the existing killed mutants for static imports, eagerly-started imports, eager prefetch, unreachable import, missing loader, and comment/string-only fake consumers.

Fresh local proof on exact head:

- `npm run typecheck`: PASS
- `npx vitest run src/i18n`: 42/42 PASS
- `npx vitest run`: 737 PASS, 5 expected skipped
- `npx playwright test tests/i18n-kernel.pw.ts`: 2/2 PASS
- `npm run test:mobile-webview-authority`: policy 6/6 + browser 1/1 PASS
- `npm run build`: PASS; 3,286 modules and separate `en`, `es`, `fr` production chunks; fresh WASM verified
- GitHub Actions pin policy: PASS (163 refs / 14 files)
- Rust toolchain policy: PASS (1.97.1 / MSRV 1.97)
- `actionlint`: PASS
- `git diff --check`: PASS

Exact origin/main remains `ec8922e564c83ca567281c0e19186776023e81ef`; worktree and remote head are clean/equal. Hosted replacement-head CI is now authoritative.

skyl:
Latest push is a rebase — **neither blocker is addressed**. Verified on the current head:

```
main.tsx:20    RootFallback() { const { t } = useTranslation(); ...   <- unchanged
main.tsx:82    .catch((error) => { console.error(...) })             <- unchanged, renders nothing
i18n/index.ts:54  configureFormattingLocale(locale, ...)             <- unchanged, still the 2-letter code
```

So a catalog integrity failure still leaves `#root` empty — I reproduced it by deleting `navigation.home` from `en.json`: `rootHTMLLength: 0`, no text, no reload button, only a console line. And `en-GB`/`es-MX` users still get the wrong date format on wallet history.

The two changes, concretely:

**B1** — catch inside `mountApplication` and render, rather than catching the promise and logging:

```tsx
async function mountApplication() {
  const root = ReactDOM.createRoot(document.getElementById("root")!);
  try {
    const i18n = await initializeAppI18n();
    root.render(/* … I18nextProvider … */);
  } catch (error) {
    console.error("ZUULI locale bootstrap failed", error);
    root.render(<RootFallback />);   // hardcoded English literals
  }
}
```

and revert `RootFallback` to literal strings. The comment this PR edited at `main.tsx:9-17` had it right before — the crash frame must be dependency-free, because it exists to survive exactly the subsystem it would otherwise depend on. Translating it is what makes it unreachable in the new failure class.

**B2** — one line: normalize only for catalog selection, pass the full BCP-47 tag to `configureFormattingLocale`. Worth adding a case that would fail today, i.e. asserting a non-`en-US` tag formats as that region expects; the current test can't catch it because it computes its expected value with `en-US` on an `en-US`-pinned runner.

Everything else in the review stands — the catalog-parity policy is genuinely fail-closed, laziness is verified in the build output, and the scope claim is honest.

skyl:
Train hold: please do not merge the current `f8edcb5` head even if its hosted checks finish green. `main` advanced through #542 while this run was active, and the mandatory 44-mutant librustzcash compatibility guard is landing as an immediate follow-up. I will normally merge the resulting current `main` into this published branch (no force-push), rerun combined local proof, and require a fresh exact-head hosted matrix before merge.

skyl:
Third push, still no fix — and I want to point at *what* is growing rather than restate the blockers.

`f8edcb5` adds **136 lines to `build-boundary.test.ts`**. Meanwhile, on that same head:

```
root.render(<RootFallback  ->  0 occurrences
RootFallback uses t()      ->  1 occurrence
```

And I grepped every test file under `src/i18n/` and `tests/` for anything touching the failure path — `rootHTML`, `innerHTML`, `bootstrap failed`, `RootFallback`, `mountApplication`:

```
(no matches)
```

**No test covers the blank-webview path**, so the 136 new lines cannot fail on it. This is the same shape as #510, where a policy self-test grew 151 → 288 lines across four commits while the job it guarded had never once compiled. Test coverage that grows around a known defect isn't progress on it.

The build-boundary coverage is worth having on its own merits — and it partly addresses my earlier note that the dynamic-import guard only greps `index.ts` — so I'm not asking you to drop it. I'm asking for the two changes that unblock, which are small:

1. Move the `try`/`catch` **inside** `mountApplication` and `root.render(<RootFallback />)` in the catch, with `RootFallback` reverted to literal English strings.
2. Pass the full BCP-47 tag to `configureFormattingLocale`; normalize only for catalog selection.

And one test that would have caught #1: break a catalog key, assert `#root` is **non-empty**. That is the assertion this PR is missing, and it is the one I'd weight above all 136 of the new lines.

skyl:
Sixth check against head `f8edcb5`. Both blockers are byte-for-byte unchanged since the first review:

```
$ git show f8edcb5:wallet/zuuli/src/main.tsx | grep -c 'root.render(<RootFallback'
0
$ git show f8edcb5:wallet/zuuli/src/main.tsx | sed -n '19,21p'
function RootFallback() {
  const { t } = useTranslation();
$ git show f8edcb5:wallet/zuuli/src/i18n/index.ts | sed -n '54p'
  configureFormattingLocale(locale, instance.t(MESSAGE_KEYS.commonPending));
```

`mountApplication` still has no internal `catch`, the outer `.catch` still only `console.error`s, and `RootFallback` — the crash frame — still depends on the subsystem whose failure it exists to survive. Meanwhile `build-boundary.test.ts` is 288 lines and `source-policy.ts` is 194. **The policy self-tests keep growing while the two guarded things stay broken.** That's the pattern, not the exception, and it's why I keep holding this.

I've been describing the fixes rather than writing them, which is my fault for the loop we're in. Here they are literally. Apply these two and I'll merge.

## B1 — `src/main.tsx`

```diff
-function RootFallback() {
-  const { t } = useTranslation();
+/**
+ * Copy is hardcoded English on purpose. This is the frame that renders when
+ * i18n itself failed to initialize, so it must not read from a catalog,
+ * a provider, or any hook that assumes one. Do not "extract" these strings.
+ */
+const CRASH_TITLE = "Something went wrong";
+const CRASH_BODY =
+  "ZUULI could not finish loading. Your funds are unaffected. Reload to try again.";
+const CRASH_ACTION = "Reload";
+
+function RootFallback() {
   return (
```

…and the three `{t(MESSAGE_KEYS.*)}` call sites become `{CRASH_TITLE}` / `{CRASH_BODY}` / `{CRASH_ACTION}`. Drop `useTranslation` and `MESSAGE_KEYS` from the imports.

Then the mount:

```diff
 async function mountApplication() {
+  const root = ReactDOM.createRoot(document.getElementById("root")!);
+  let i18n;
+  try {
+    i18n = await initializeAppI18n();
+  } catch (error: unknown) {
+    // A missing/malformed catalog is an integrity failure, not permission to
+    // render untranslated keys — but it is also not permission to leave a
+    // wallet as a blank webview. Render the untranslated crash frame.
+    console.error("ZUULI locale bootstrap failed", error);
+    root.render(<RootFallback />);
+    return;
+  }
-  const i18n = await initializeAppI18n();
-  ReactDOM.createRoot(document.getElementById("root")!).render(
+  root.render(
```

`createRoot` moves **above** the `try` deliberately: if the catalog `import()` 404s after a deploy, you still need a root to render into.

**Prove it fails first.** Delete `navigation.home` from `en.json`, run the dev server, and assert `#root`'s innerHTML is non-empty and contains "Reload". Today that assertion goes red at `rootHTMLLength: 0`. A Playwright case that stubs the `en` chunk request to a 404 and asserts the frame appears would be better still, and would cover the deploy-skew trigger directly.

## B2 — `src/i18n/locale.ts` + `index.ts`

```diff
+/**
+ * The catalog is selected by language ("en"), but dates and numbers must be
+ * formatted with the user's full tag ("en-GB") or every non-US user gets
+ * US conventions. Only honour the region when its language is the one we
+ * actually loaded a catalog for.
+ */
+export function resolveFormattingTag(
+  locale: SupportedLocale,
+  preferred: unknown,
+): string {
+  if (typeof preferred !== "string") return locale;
+  const tag = preferred.trim();
+  return normalizeLocale(tag) === locale ? tag : locale;
+}
```

```diff
-export async function createAppI18n(locale: SupportedLocale): Promise<i18n> {
+export async function createAppI18n(
+  locale: SupportedLocale,
+  formattingTag: string = locale,
+): Promise<i18n> {
@@
-  configureFormattingLocale(locale, instance.t(MESSAGE_KEYS.commonPending));
+  configureFormattingLocale(formattingTag, instance.t(MESSAGE_KEYS.commonPending));
@@
   const instance = await createAppI18n(locale);
+  const instance = await createAppI18n(
+    locale,
+    resolveFormattingTag(locale, window.navigator.language),
+  );
```

**And the test has to stop reading its expected value off the page.** `format.test.ts` currently computes the expectation with `"en-US"` on an `en-US`-pinned runner, so making `formatDate` ignore its `locale` argument entirely stays green — I checked. Assert against **hardcoded literals for a non-default region**, e.g. that `formatDate(d, "en-GB")` contains `"20:02"` and does **not** contain `"PM"`, while `formatDate(d, "en-US")` does contain `"PM"`. That case goes red on today's code and green after the fix, which is the whole point.

Everything else in my earlier review stands as follow-up, not blocking — the `includes()` prefix hole in `build-boundary.test.ts:32-40` (nine of 47 keys can lose their only consumer: `navSearch` satisfied by `navSearchAction`, `commonRetry` by `commonRetrying`), the +94 kB entry-chunk growth the body doesn't mention, and the missing locale dimension in `viewport.pw.ts` (fr `"Portefeuille"` fits 320px with `scrollWidth === 320` exactly — zero headroom, and German breaks it). Those can land after.

skyl:
Merge-train hold and correctness hold: this remote head is stale and does not yet contain the completed local follow-up that addresses the reviewer blockers. Keep draft until #556/#546 land, then integrate exact trunk, push the mutation-proved fallback/BCP47/import-boundary fixes, and rerun the full matrix.

skyl:

skyl:

skyl:

skyl:

skyl:

Issue #239, opened 2026-08-06
Dependencies: coordinated bleeding-edge upgrade train and fix-forward canaries


skyl:
## Goal

Keep ZUULI, every Tauri plugin backend, and the vendored Zcash source graph moving together across iOS, Android, macOS, Linux, and the Windows compile canary.

## Scope

- inventory direct/transitive Rust, npm, Tauri, Android Gradle/Kotlin/SDK/NDK, Apple/Xcode/SPM, GitHub Actions, and `z/` submodule versions
- machine-enforced single compatibility matrix and lockfile consistency
- daily upstream-HEAD canaries, not only weekly librustzcash checks
- automated isolated upgrade PRs grouped by ecosystem/compatibility boundary
- clean target-native builds and tests before merge
- fix forward for our API drift; upstream PR + reachable fork bridge for genuine regressions
- stale fork/pin detector that restores upstream immediately after merge
- generated dependency/SBOM/provenance report per release SHA
- rollback means reverting our release commit, never quietly parking an ecosystem dependency on an old pin

## Acceptance

- [ ] all direct dependencies audited and upgraded to current compatible HEAD/releases
- [ ] every plugin has compile/test coverage for every claimed backend
- [ ] iOS, Android, macOS, Linux native clean builds; Windows compile canary
- [ ] upstream canary failures open actionable issues/PRs automatically
- [ ] contributor runbook and agent instructions encode branch→PR→pin→resume
- [ ] Claude CLI adversarial review and green automation before founder merge

Tracked by #227.

skyl:
Claiming a narrow CI slice: make the existing upstream-HEAD canary daily and expand it to an explicit matrix for ZUULI, Zuuallet, and every first-party Rust plugin. I will bind the schedule, manifest coverage, source refresh, cache identity, and live build commands in mutation-tested policy with bounded timeouts. This is `Refs #239`; the umbrella issue remains open for the broader compatibility/upgrade program.

skyl:
The canary implementation exposed real current-upstream API drift before push. Splitting this into an orderly prerequisite PR: advance the librustzcash submodule to current upstream `main`, regenerate all three first-party locks, adapt the `propose_transfer` transaction-version argument and non-exhaustive birthday errors with focused tests, then prove plugin/ZUULI/Zuuallet builds. The daily matrix PR will follow only after all three legs are genuinely green. Both PRs use `Refs #239`; the umbrella remains open.

PR #542, opened 2026-08-23
fix(zcash): follow current librustzcash API

skyl:
Refs #239

Prerequisite for the daily all-first-party upstream-HEAD canary. This keeps the canary green on day one instead of recording a known-red baseline. The umbrella issue remains open for the canary follow-up.

## What changed

- advance `z/zcash/librustzcash` from `6cfb3857d248428cb53c7fb1f2b60ba15d5298a1` to exact upstream HEAD `330e4c0aa9e25199acdb93a56bf70126d0d3f2b9`
- regenerate the plugin, ZUULI, and Zuuallet locks; update Orchard narrowly from 0.15.0 to the compatible 0.15.3 after the stale lock failed current `zcash_primitives`
- pass the new `propose_transfer` `LockRequest` argument as `None`, preserving the existing process-local serialized send behavior until durable lock ownership and unlock recovery are designed together
- continue passing `None` for `TxVersion`, allowing librustzcash to choose it from the target height
- centralize `BirthdayError` formatting and retain a wildcard arm for its new `#[non_exhaustive]` contract
- add focused regression tests for both compatibility decisions and update the plugin API notes

## Local proof (Rust 1.97.1, macOS)

- plugin `cargo check --locked` and `cargo build --locked`
- ZUULI `cargo check --locked` and `cargo build --locked`
- Zuuallet `cargo check --locked` and `cargo build --locked`
- plugin `cargo test --locked --all-targets`: 131 passed
- plugin `cargo check --locked --target aarch64-apple-ios`
- repository rustfmt: 5/5 crates
- clippy negative control + live `-D warnings`: 5/5 crates
- cargo-deny: 5/5 crates (`advisories`, `bans`, `licenses`, `sources`)
- Rust toolchain live + 14 mutation cases
- Zcash permission parity live + negative controls
- GitHub Actions policy live + 70 source, 7 gate, and 99 workflow mutation cases
- public repository boundary live + self-test
- `git diff --check`

Android NDK is not installed on this machine; hosted CI supplies Android and the other platform-specific proof.

### DIFF
--- a/wallet/plugins/tauri-plugin-zcash/CLAUDE.md
+++ b/wallet/plugins/tauri-plugin-zcash/CLAUDE.md
@@ -31,12 +31,21 @@ pub type WalletProposal = zcash_client_backend::proposal::Proposal<
 
 ## librustzcash API gotchas
 
-- **`BirthdayError`** — has NO `Debug` or `Display` impl. Must pattern-match variants: `BirthdayError::HeightInvalid(e)` and `BirthdayError::Decode(e)`.
+- **`BirthdayError`** — is `#[non_exhaustive]`. Preserve actionable messages
+  for `HeightInvalid` and `Decode`, and retain a wildcard arm for future
+  upstream variants.
 - **`ConfirmationsPolicy`** — at `zcash_client_backend::data_api::wallet::ConfirmationsPolicy`, use `::default()`.
 - **`OvkPolicy`** — at `zcash_client_backend::wallet::OvkPolicy` (NOT `data_api::wallet`).
 - **`SingleOutputChangeStrategy::new`** — 4 args: `(FeeRule, Option<MemoBytes>, ShieldedProtocol, DustOutputPolicy)`.
 - **`Payment::new`** — returns `Result<Self, zip321::PaymentError>` (changed from `Option`), takes `Option<Zatoshis>` for amount. Six args: `(recipient, amount, memo, label, message, other_params)`. Use `.map_err(...)?`, not `.ok_or(...)?`.
-- **`propose_transfer`** — takes **9** args and has phantom `CommitmentTreeErrT`. Use turbofish: `propose_transfer::<_, _, _, _, zcash_client_sqlite::wallet::commitment_tree::Error>(...)`. The last two args are `spend_policy: &input_selection::SpendPolicy` (use `SpendPolicy::default()` to preserve fully-shielded behavior) and `proposed_version: Option<TxVersion>` (use `None` to let target height decide).
+- **`propose_transfer`** — takes **10** args and has phantom
+  `CommitmentTreeErrT`. Use turbofish:
+  `propose_transfer::<_, _, _, _, zcash_client_sqlite::wallet::commitment_tree::Error>(...)`.
+  The last three args are `spend_policy: &input_selection::SpendPolicy` (use
+  `SpendPolicy::default()` to preserve fully-shielded behavior),
+  `lock_inputs: Option<LockRequest>` (currently `None`; durable lock ownership
+  and unlock recovery must be designed together), and
+  `proposed_version: Option<TxVersion>` (`None` lets target height decide).
 - **`create_proposed_transactions`** — takes **8** args and has phantom `InputsErrT` + `ChangeErrT`. Use `std::convert::Infallible` for both. The 8th arg is `expiry_height: Option<BlockHeight>` (use `None` to keep the builder-derived expiry). The transaction version is now carried on the proposal itself (`proposal.proposed_version()`), not passed here.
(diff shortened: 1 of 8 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Approving on substance — merging once `Rust / native lints (windows)` and `gate` are green.**

This is the money crate, so I had the review read the actual upstream source at both submodule commits rather than trust the description. Every claim in the body checked out.

### The submodule move is doctrine-compliant and verifiable

`6cfb3857` → `330e4c0a`, 741 commits — and `330e4c0a` is **byte-identical to `zcash/librustzcash` `main` HEAD** right now (`gh api .../compare/330e4c0a...main` → `identical`), reachable from the configured remote and branch. No fork, no transient pin.

### Two API breaks, both compile-forced — no silent semantic swap

| Call site | Change | Preserved |
|---|---|---|
| `send.rs:1403`, `:1560` `propose_transfer` | 9 → 10 args: `spend_policy, lock_inputs: Option<LockRequest>, proposed_version`; `DbT: WalletRead` → `WalletWrite` | **yes** |
| `create_proposed_transactions` `send.rs:1796` | byte-identical upstream at new HEAD | untouched, correctly |
| `BirthdayError` matches | now `#[non_exhaustive]` + `Debug`/`Display`; wildcard arm required | **yes**, exact message strings preserved |

The argument-reordering risk I was most worried about can't occur here: `Option<LockRequest>` and `Option<TxVersion>` are distinct concrete types, and each is passed through a named helper with an explicit return type, so a swap is a type error rather than a behaviour change.

**`lock_inputs: None` is genuinely behaviour-preserving** — upstream only touches the DB when it's `Some` (`wallet.rs:947-950`), so `None` performs no lock acquisition and no lock writes, exactly the pre-locking behaviour. And the `WalletRead → WalletWrite` tightening is already satisfied: both call sites already pass the write handle.

### Every load-bearing default verified unchanged across the 741-commit window

`SpendPolicy::default()` — same pool set `{Sapling, Orchard, Ironwood}`, still no transparent; its two new fields default to `LockedInputPolicy::Exclude` (inert, nothing is ever locked) and `NoteSelection::Accumulate` (the historical behaviour). `ConfirmationsPolicy::default()` byte-identical. `SingleOutputChangeStrategy::new` still 4 args, same order, `ShieldChange` still default — upstream *dropped* `unpadded_orchard_pool_bundles`, which was already our value. Fee rule still `zip317::standard()`, change pool still Orchard, `OvkPolicy::Sender` unchanged.

Two shifts arrive from upstream rather than this diff: `propose_transfer` now calls `check_transaction_size()` (fails at propose time instead of build time — an improvement), and a ZIP 318 canonical-crossing path that is **inert until NU6.3 activation**.

Nice detail on `format_birthday_error`: de-duplicating the two byte-identical private copies while keeping the original message strings, rather than falling through to upstream's new `Display`. User-facing text is unchanged.

### The finding that matters most

**No test anywhere in the repo exercises `propose_transfer` or `create_proposed_transactions`. Coverage on the send path is compile-only.** All 42 tests in `send.rs` cover the surrounding state machine — confirmation tokens, digest binding, broadcast status, recovery-journal fail-closed paths — and stop at the librustzcash boundary.

That predates this PR and isn't a reason to block it, but it means these four mutations would all ship silently:

```
&SpendPolicy::default() -> shielded_pools([Sapling])   stops spending Orchard/Ironwood notes
ShieldedPool::Orchard   -> Sapling                      change lands in the wrong pool
ConfirmationsPolicy::default() -> ::minimum()           spends 1-conf notes
OvkPolicy::Sender       -> ::Discard                    wallet loses its own outgoing-view records
```

Filing that separately — it's the highest-value follow-up on this crate.

The new `BirthdayError` test is a real guard and would fail on a plausible wrong adaptation. The `assert!(proposal_lock_request().is_none())` test is close to tautological — two constant functions returning their constants — and shouldn't be read as coverage of the send path, though it does document intent for free.

### Clean elsewhere

`Rust / Android 32-bit` **passes** (1m38s), so the build-7 `S_IFMT` class stays closed; no new width-dependent arithmetic, and the one remaining `as usize` on a file length is bounded two lines earlier. All 58 deletions accounted for line-by-line — every one is part of the adaptation, no `Cargo.toml` dependency declarations changed, and the `orchard` feature stays explicitly declared rather than relying on upstream's new default-on.

New transitive crates (`pczt`, `zcash_pool_migration`, `postcard`, `cobs`, `embedded-io`, `libm`) all cleared `Rust / supply chain`.

### Honest limits

The mutation matrix above was **reasoned from verified signatures, not executed** — a full librustzcash graph build wasn't run. And worth stating plainly: the reviewable surface here is 6 source lines against ~200k lines of upstream churn, with CI compile plus `cargo-deny` standing behind it. That's inherent to the stay-on-HEAD doctrine, not a flaw in this PR — but it's the argument for landing the daily upstream-HEAD canary from #239, so the next jump is one commit wide instead of 741.

Two smaller follow-ups worth doing whenever this crate is next open: adopt `LockRequest` for the propose→confirm→execute window (the 2-minute confirmation TTL is exactly what `LockRequest::for_blocks()` is for, and balance currently doesn't reflect value committed to a pending confirmation), and switch `{e:?}` to `{e}` at `send.rs:1416`/`:1813` now that every upstream error enum has `Display` — `ProposalError::TransactionTooLarge` carries per-pool note counts specifically so a wallet can explain itself in notes rather than bytes.

Issue #544, opened 2026-08-23
Spike: can we adopt akd (WhatsApp-lineage key transparency) instead of hand-rolling the KT log?


skyl:
Part of #305 (Phase 1 implementation wave). **Timebox: 2–3 days. Throwaway branch, no production code.**

## Question

`docs/e2ee/ARCHITECTURE.md` §9.2 specifies a CONIKS/SEEMless/Parakeet-lineage key-transparency directory with privacy-preserving lookups. The decision we need is **not** which construction we prefer — it is whether we can adopt a production implementation instead of hand-rolling one.

[`akd`](https://github.com/facebook/akd) (Auditable Key Directory) is Meta's Rust implementation of exactly this lineage — append-only zero-knowledge set, VRF labels, membership / non-membership / append-only proofs — and it backs WhatsApp Key Transparency.

## Why this decides the design

The failure modes are asymmetric. A bug in a simple cleartext log leaks metadata: bad, but loud and discoverable by inspection. A **soundness bug in an append-only ZKS lets the log substitute a key undetectably** — server swaps Alice's key, serves it to Bob, rolls it back, and Alice's self-audit sees nothing. That does not weaken the guarantee, it inverts it: #133's MITM restored, with a transparency log on top asserting it cannot happen.

That asymmetry argues for adopting a production implementation rather than hand-rolling something simpler — the same rule ADR 0001 already applied when it chose OpenMLS over writing our own MLS.

## What to determine

1. **Does `akd` build for `wasm32-unknown-unknown`?** `f2z-kt-core` will be linked by the log server, the witness **and** the client, native and WASM. If `akd` cannot reach the browser, the web client needs a second verifier implementation — exactly the divergence ADR 0001 forbids. **This is the likeliest hard blocker.**
2. **Has `akd` been independently audited?** WhatsApp's deployment is real evidence but is not a published assessment. We held OpenMLS to a named auditor and a version floor (`>= 0.8.1`); apply the same standard and report a concrete floor.
3. **Does its dependency graph fit?** Specifically against the repo-wide `rusqlite 0.37` / `libsqlite3-sys` `links = "sqlite3"` singleton documented at `AGENTS.md:147`, which already disqualified `openmls_sqlite_storage` 0.2.0. Also check it against the pinned toolchain in `wallet/rust-toolchain.toml`.
4. Rough size/perf of a proof verification in WASM, if it builds.

## Outcome

Report findings as a comment on this issue.

[...]

PR #548, opened 2026-08-23
docs(e2ee): freeze the Phase 1 relay wire specification

### LINKED ISSUE #545: E2EE Phase 1: freeze the wire specification — nothing can be built against placeholders
Part of #305. **Docs only — no code ships from this issue.** Prerequisite for every implementation PR in the Phase 1 wave.

## Problem

Phase 0 (#318, #346) is merged and is a genuinely good design document. But it deliberately marked every constant, label and wire encoding a *proposed placeholder*:

> "Constants, labels and wire encodings marked *proposed* are placeholders to be fixed in the Phase 1 specification; they are written down so the structure can be reviewed, not so they can be implemented as-is."

A verification pass against the tree confirms the practical consequence. **None of the following exists anywhere in `docs/e2ee/`:**

- a relay command grammar (only the capability *names* `APPEND`/`READ`/`ACK`/`DELETE` appear, in prose)
- a canonical serialization choice
- any anti-replay construction for signed relay commands
- queue creation / send-key binding / rotation mechanics
- a first-contact path — **there is no way for Alice to send Bob a `Welcome`, because Bob has no queue for Alice yet**
- any quota or rate-limit mechanism (§13-I is explicitly deferred and unspecified)
- a `SignedTreeHead` format, epoch cadence, or `DirectoryEntry` structure
- a key-rotation rule for the directory
- a witness-cosignature message format

Nothing can be implemented against placeholders. This issue closes those gaps **on the record**, before code exists, honoring the repo's own "listed rather than invented" doctrine.

## Deliverables

- `docs/e2ee/WIRE.md` — the relay protocol, specified concretely enough for a second implementer to work from: transport, framing, canonical serialization, the full command set with request/response shapes and stable error codes, the signing transcript and what exactly is covered, anti-replay, queue lifecycle, ACK semantics, TTL, padding enforcement, and the capability document.
- ADRs **0008–0012** and **0014**, one per invented decision, each stating what the merged docs left open and why this answer was chosen.
- Updates to `ARCHITECTURE.md` §13 moving genuinely-closed questions into §13.1 (letters are stable identifiers and are retired, never reused).

**Deliberately deferred to a follow-up PR:** `docs/e2ee/KT.md` and ADR 0013 (the key-transparency log construction), which are blocked on the outcome of #544.

## Blocking pre-work

Both must be resolved **in this PR** and recorded in the docs:

1. **Handle charset.** A restricted charset (proposed: `[a-z0-9_]{1,30}`, ASCII, NFC) makes an entire class of homograph/mixed-script impersonation not exist. **Check it against free2z's actual username rules first.** If real usernames are broader, some existing users cannot use messaging in v1 — a product cost to accept deliberately, not to discover later.
2. **SimpleX licensing (§13-E).** Verify the server licence from *published repository metadata only*. Record the clean-room posture explicitly: nobody on the project reads `simplexmq` source; the design derives solely from the public protocol document. Record who performed the check and that they did not read source.

## Acceptance

- [ ] Every gap listed above is closed with a concrete, implementable answer.
- [ ] Each invented decision has an ADR stating the alternatives considered and the reason.
- [ ] Where a property does **not** hold, it is stated as not holding — consistent with the existing docs' posture. In particular the honest limits on first-contact anti-abuse (proof-of-work taxes phones far more than rented GPUs) and on capability separation vs. the relay operator (§4.9) are stated, not glossed.
- [ ] No code, no crates, no manifests.

skyl:
Closes #545. **Docs only — no code, no crates, no manifests, no workflow changes.** Everything is under `docs/e2ee/`.

Phase 0 deliberately marked every constant, label and wire encoding a *proposed placeholder*. Nothing can be implemented against placeholders, so this closes those gaps on the record, before code exists, in the voice the merged docs established: written for a third-party auditor, properties that do not hold stated as not holding, open questions listed rather than answered by invention.

## What landed

**`docs/e2ee/WIRE.md`** — the relay protocol, specified concretely enough that an independent implementer could build an interoperable relay from it alone.

- **Transport:** WebSocket over TLS 1.3 only. Justified on ADR 0001 (the WASM client can only do WebSocket, and there must be one implementation) and on the plainer point that a second transport is a second parser and a second fuzz target on an unauthenticated front door. The relay refuses to bind a non-loopback address without TLS unless explicitly overridden, and the override is published in the signed capability document. Server-side Ping every ~25 s, with the load-balancer idle-timeout trap called out as the most common way a working relay behaves badly in production.
- **Serialization: `tls_codec`.** Canonical by construction (no encoding options exist), already in the client's graph via OpenMLS, and an auditor reading an MLS system reads TLS presentation syntax without translation. CBOR rejected — its canonical form is a *profile*, i.e. a rule a decoder will let you skip. **Mandatory re-encode-equality rule:** decode, re-encode, byte-compare, sign over the re-encoded bytes; a mismatch is fatal. Stated honestly as compensating for implementation slack in decoders rather than for format ambiguity.
- **Framing:** one binary frame = one message; text frames fatally rejected; bounded in-flight window with pipelining and out-of-order responses correlated by request id. Errors carry a code and no free-text field.
- **Command set** with full request/response shapes: unsigned `HELLO` / `GET_CAPABILITIES` / `GET_CHALLENGE` / `PING`; recv-key-signed `CREATE_QUEUE` / `SUBSCRIBE` / `UNSUBSCRIBE` / `READ` / `ACK` / `DELETE_QUEUE`; send-key-signed `BIND_SEND` / `APPEND`. **`APPEND` returns no index and no depth** — and the same rule is extended to *errors*, because otherwise a bound sender learns queue state by filling it. All send-side refusals collapse to one code, and the residual bit a sender can still obtain is stated rather than glossed.
- **The signing transcript** covers a domain-separation label, the version, the request id, a timestamp, a client nonce — plus:
  - a **relay identity binding**, closing a live attack: with ADR 0005 publishing addresses on *k* relays, relay A can otherwise replay a signed `ACK` to relay B and **B deletes ciphertext the recipient never received**. The relay must prove possession of the identity key in `HELLO`, and the expected `relay_id` travels with the address in the in-band advert.
  - a **TLS exporter channel binding** (RFC 8446 §7.5), with the honest caveat that a relay behind a TLS-terminating proxy cannot compute it; in that mode it is zeroed, the capability document says so, and a client may refuse.
- **Anti-replay:** bounded timestamp window plus a fail-closed `(key, nonce)` seen-set. Why it need not survive restart is explained — a restart changes the TLS session, so the channel binding already invalidated the corpus — **and the exception is stated**: in proxied mode that argument does not hold, so such a relay must persist the seen-set or declare `antireplay_persistence: "volatile"`.
- **Queue lifecycle:** the recipient creates; **the relay generates both addresses from its own CSPRNG** (client-chosen addresses allow squatting and collisions); the send side starts unbound; the peer learns `send_addr` from an authenticated advert inside MLS; `BIND_SEND` is **once-only and irreversible**. The consequence is said plainly: an operator that binds first causes the legitimate sender to get `ERR_ALREADY_BOUND`, which the client MUST treat as a loud non-dismissible failure — **this does not prevent the theft of the write capability, it makes it noisy.**
- **Rotation** needs no new command. It also **narrows `ARCHITECTURE.md` §5.4**, which claimed the exporter "derives the next queue addresses without a round trip": it derives the next queue *keys*.
- **ACK:** cumulative, monotone, idempotent, never selective; acking beyond the head is an error, because pre-acking would let a reader silently black-hole its own queue while senders still see `accepted`. Tied to §6.3's four states, with the durable-write MUST restated where an implementer will meet it.
- **TTLs:** per-queue message TTL clamped to the 30-day ceiling, **plus a queue idle TTL** the merged docs had nothing for — without it abandoned queues grow the queue table without bound. Its failure mode (a user offline past the TTL loses their queues) is stated.
- **Padding:** relay rejects non-conforming sizes, and the accepted set lives in the **capability document**, not the protocol, so §13-F's traffic study becomes a config change rather than a flag day. The relay's list is called what it is: a filter, not the source of truth.
- **Stable numeric error codes.** "No such queue" and "exists but your key does not authorize it" return the **same** code, or the relay is an existence oracle. And the timing side channel that remains — the two paths do different amounts of database work — is stated, with a dummy-verification mitigation and **no claim of constant time**.
- **The capability document:** limits, TTLs, padding sizes, channel-binding mode, durability mode, anti-abuse mode, operator contact and jurisdiction, source commit and build digest. Served over the protocol *and* over plain HTTPS at a well-known path so a human can read a relay's policy before connecting — with the honest note that nothing but the operator forces the two to agree.

**First contact — a genuine hole in the merged design.** §6.2 only describes queues advertised *inside* an existing group, so there was **no path for the `Welcome` that creates the group**. Closed with a distinctly-flavoured **contact queue**: send side never bound, published in the owner's directory entry, unsigned appends, hard-capped (pending cap, byte cap, per-source bucket, required PoW stamp over a relay-issued single-use challenge). Limits stated: PoW taxes phones far more than rented GPUs; disk exhaustion via mass queue creation is **made expensive, not closed** (§13-I is the real answer, still deferred); and a contact queue can be filled to deny first contact.

**Anti-abuse v1:** connection limits, per-queue quotas, queue-creation control (open / PoW / token, defaulting to PoW, with `token` labelled as the identifier it is), and global backpressure. **Never delete un-acked messages to make room** — refusing writes is correct, because silently dropping accepted messages breaks the other half of delete-on-ack through the tail-truncation blind spot. What is not closed is named: distributed creation flooding with valid stamps; an abusive but legitimately bound sender; per-source limits harming Tor and CGNAT users, with no good answer absent identity the relay deliberately lacks.

**ADRs 0008–0012 and 0014**, in the existing format. **0013 is deliberately absent** — the KT log construction is blocked on #544 — and its absence is noted in `WIRE.md` §1.2, in `ARCHITECTURE.md` §9.2, in the README ADR table, and in ADR 0014, so the gap is visibly intentional.

**§13 housekeeping.** Only **§13-E** was genuinely closed and moved to §13.1. §13-B, F, G, I, K, L, M were **not** closed — several are cited by this spec as deliberately open. Two new letters opened: **§13-N** (PoW function and calibration) and **§13-O** (messaging-handle eligibility). Letters retired, never reused.

## The two blocking pre-checks

**1. Handle charset — checked, and real usernames are broader.** Proposed `[a-z0-9_]{1,30}`, ASCII, so homograph and mixed-script impersonation does not exist. Against the tree:

- `py/dj/free2z/openapi/f2z.yaml` publishes `username` as `pattern: ^[\w.@+-]+$`, `maxLength: 150` on `Registration`, `CreatorDetail`, `CreatorList` and `CommentAuthor` — Django's `AbstractUser` default. Uppercase, `.`, `@`, `+`, `-` and 150 characters are all allowed today.
- **Whether non-ASCII is allowed cannot be determined from this repository.** `AUTH_USER_MODEL = 'g12f.Creator'` (`py/dj/free2z/settings.py:29`) but the model module is not vendored here, and the OpenAPI pattern is emitted identically for Django's Unicode and ASCII username validators. Recorded as a MUST-confirm, because under the Unicode validator the platform's existing handle space already contains the homograph surface.
- Uniqueness is enforced case-insensitively in the serializers (`username__iexact`), and reads use `__iexact`. The front ends narrow nothing — `@{creator.username}` is rendered verbatim and URL-encoded into paths; the one narrower guard (`/^[a-zA-Z0-9_]{1,64}$/` in the billing `+page.server.ts`) validates a route parameter and is already narrower than registration.

**Decision and cost accepted:** eligibility is `lowercase(username)` matching `[a-z0-9_]{1,30}`. Case folding is included because platform uniqueness is already case-insensitive, so it cannot collide — **with the caveat that this must be confirmed with a query**, since the model is not here and a database-level constraint cannot be assumed. Accounts with `.`, `@`, `+`, `-`, non-ASCII, or >30 characters are **not eligible for messaging in v1**. **How many is unknown** — no user data in this repo — and §13-O says it must be measured before the rule ships. Aggressive normalization was rejected (it creates collisions and relocates the homograph surface into our own code); a separate opt-in messaging handle is deferred to directory design.

**2. SimpleX licensing (§13-E) — verified, no source read.** Performed 2026-08-23 from **published repository metadata only** (GitHub's REST repository endpoint and a top-level contents listing). `simplex-chat/simplexmq` — the reference SMP server — reports **AGPL-3.0**; `simplex-chat/simplex-chat` likewise. No source file was fetched, opened or read. The clean-room posture is recorded as a binding rule in `WIRE.md` §15: **nobody on this project reads `simplexmq` source**; the design derives solely from the public protocol description. The nuance is recorded rather than glossed — the protocol document lives in that same AGPL repository, so the posture rests on the distinction between the expression of a specification and the protocol it describes, stated as our position and explicitly not as legal advice.

## Notes for review

- Every invented decision says it is invented, gives the reasoning, and lists the rejected alternatives. Nothing is presented as though the merged docs already said it.
- `scripts/check-public-repo-boundary.sh` passes locally (exit 0).
- `ARCHITECTURE.md` edits go slightly beyond §13: the §5.4 exporter claim is corrected, §6.2's circularity is flagged, and §6's SimpleX action item is closed. Leaving a now-wrong statement uncorrected seemed worse than a small, clearly-marked edit — happy to split those out if the reviewer prefers.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -6,7 +6,7 @@
 [#132](https://github.com/free2z/zuu/issues/132),
 [#133](https://github.com/free2z/zuu/issues/133),
 [#134](https://github.com/free2z/zuu/issues/134).
-**Companion:** [`THREAT-MODEL.md`](./THREAT-MODEL.md),
+**Companion:** [`THREAT-MODEL.md`](./THREAT-MODEL.md), [`WIRE.md`](./WIRE.md),
 [`decisions/`](./decisions/).
 
 This document is written to be read by a third-party security auditor. Where
@@ -329,7 +329,7 @@ is the mechanism we intend to use for component-scoped exports as it stabilizes.
 |---|---|---|---|
 | FROST/DKG (§11) | `free2z/frost/v1` | `ceremony_id` | Session domain separator, transcript binding, outer AEAD for part-2 shares |
 | WebRTC binding (§10) | `free2z/webrtc/v1` | `session_id` | Binds DTLS fingerprints to the group |
-| Queue rotation (§6.2) | `free2z/queue/v1` | `peer_leaf_index` | Derives the next queue addresses without a round trip |
+| Queue rotation (§6.2) | `free2z/queue/v1` | `peer_leaf_index` | Derives the next queue **signing keys** and the rotation schedule — see the correction below |
 | Local history wrap | `free2z/history/v1` | `conversation_id` | At-rest key for retained plaintext |
 
 Because these are exporter outputs, they inherit the group's forward secrecy and
@@ -338,6 +338,17 @@ precisely the "confidential, authenticated, replay-bound channel between an
 agreed participant list" that
 [#132](https://github.com/free2z/zuu/issues/132) requires, obtained from a
 standard instead of invented.
+
+> **Correction (2026-08-23).** An earlier revision of the table above said the
+> queue exporter *"derives the next queue addresses without a round trip."* That
+> is no longer accurate and should not be read as though it were.
+> [ADR 0009](./decisions/0009-queue-addressing-and-binding.md) has the **relay**
+> generate both queue addresses from its own CSPRNG, because client-chosen
+> addresses permit squatting and collisions. The exporter therefore derives the
+> rotation *schedule* and the next queue *signing keys*; the new address comes
+> back from the relay and still has to reach the peer in an in-band advert. **A
+> rotation costs one relay round trip and one in-band message** —
+> [`WIRE.md` §7.5](./WIRE.md#75-rotation-needs-no-new-command).
 
 ### 5.5 What hybrid PQ does and does not buy
 
@@ -380,9 +391,19 @@ Its design follows SimpleX's SMP: pairwise, unidirectional, opaque queues with
 no user identifiers anywhere. We adopt the *protocol design*; the implementation
 is clean-room.
 
-> **Action item (§13-E):** verify SimpleX's server licensing **before** anyone
-> reads their server source. Parts of that project are copyleft and we need a
-> clean-room posture.
+> **§13-E, closed 2026-08-23.** The licence was verified from published
+> repository metadata only, without reading any source file: `simplex-chat/simplexmq`
+> — the reference SMP server — is **AGPL-3.0**. The binding rule and the full
+> clean-room posture are recorded in
+> [`WIRE.md` §15](./WIRE.md#15-clean-room-posture-toward-simplex--blocking-pre-check-resolved):
+> **nobody on this project reads `simplexmq` source**; the design derives solely
+> from the public protocol description.
+
+The relay's wire protocol — transport, framing, canonical serialization, the
+command set, the signing transcript, anti-replay, queue lifecycle, ACK semantics,
+TTLs, padding enforcement, error codes, the capability document, first contact and
+anti-abuse — is specified in [`WIRE.md`](./WIRE.md). Where that document narrows
+or corrects something below, it says so at the point of use.
 
 ### 6.1 What the relay is
 
@@ -443,6 +464,16 @@ via the server. Addresses rotate on a schedule using the
 `free2z/queue/v1` exporter so a long-lived relationship does not present a
 long-lived identifier.
 
+> **This paragraph is circular for a pair that has never spoken**, and the gap is
+> real rather than a detail: there is no path for the `Welcome` that creates the
+> group in the first place, because the recipient has no queue the sender is
+> authorized to write to.
+> [ADR 0011](./decisions/0011-first-contact-contact-queues.md) and
+> [`WIRE.md` §12](./WIRE.md#12-first-contact) close it with a hard-capped
+> **contact queue** whose send side is never bound and whose address is published
+> in the owner's directory entry — together with the honest limits on what its
+> anti-abuse actually buys.
+
 **Fan-out (D2):** a message to a recipient with *d* registered devices is sent
 to *d* queues, one per device, as *d* independently encrypted MLS messages
 (each device is its own MLS leaf). Group messages to a group with *m* member
@@ -476,10 +507,14 @@ Three properties follow, and each is deliberate:
    themselves MLS application messages. The relay cannot forge, suppress
    undetectably (§7), or observe it.
 
-Undelivered ciphertext has a **TTL ceiling** (proposed: 30 days) after which the
-relay drops it and the sender learns only from the continued absence of a
-receipt. Quota and TTL are relay policy and are published in the relay's
-capability document so clients can choose.
+Undelivered ciphertext has a **TTL ceiling** (30 days) after which the relay
+drops it and the sender learns only from the continued absence of a receipt.
+Quota and TTL are relay policy and are published in the relay's capability
+document so clients can choose. The wire-level rules — a per-queue message TTL
+clamped to that 30-day ceiling, a **queue idle TTL** that the merged docs did not
+have and without which the queue table grows without bound, and the exact ACK
+semantics that make delete-on-ack safe — are
+[`WIRE.md` §7.7](./WIRE.md#77-ttls) and [§8](./WIRE.md#8-ack-semantics).
 
 ### 6.4 Delete-on-ack and lost acknowledgements
 
@@ -502,7 +537,13 @@ colluding relay. Real text is ~100 bytes; the padding dominates, which is the
 intent.
 
 > **Open question (§13-F):** bucket sizes need a traffic study. The values above
-> are placeholders chosen for arithmetic convenience, not measurement.
+> are placeholders chosen for arithmetic convenience, not measurement. Because of
+> that, [`WIRE.md` §9](./WIRE.md#9-padding-enforcement) puts the accepted size set
+> in the relay's **capability document** rather than in the protocol, so that
+> adopting the measured answer is an operator config change rather than a
+> federation-wide flag day — and states plainly that the relay's list is a filter,
+> not the source of truth: the property comes from the client padding to its own
+> set before the ciphertext leaves the device.
 
 ### 6.6 Delivery receipts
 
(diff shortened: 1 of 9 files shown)


skyl:
Reviewed after merge — the review finished a few minutes after this landed. Recording the outcome here so it isn't lost.

**This is unusually good spec work.** ADR-format compliant across all six records, 189 internal links with all 91 anchors resolving, the 0013 gap deliberate and flagged in six places, and it does **not** over-claim implementation — `Closes #545` only, with #305/#311/#313 correctly left open, and both `WIRE.md:3` and the README saying plainly that nothing is implemented.

Three things I want to name as genuinely right, because they're the ones most specs get wrong:

- **The signing transcript is unambiguous.** `CommandTranscript` is a fixed-shape `tls_codec` struct with a domain label, so the classic concat-without-length-prefixes bug is actually avoided. The re-encode-equality rule is correctly described as compensating for decoder slack rather than format ambiguity.
- **The cross-relay `ACK` replay attack in §5.2 is real, and the `relay_id` binding closes it.** Good catch.
- **Delete-on-ack creates no deletion oracle and no forced-deletion path** — recv-key-signed, transcript-bound to one relay and one session, cumulative and monotone, with `ERR_ACK_TOO_HIGH` handled correctly.

Findings are filed as three issues rather than comments, since the spec is now frozen on `main` and implementations will be held to it:

- https://github.com/free2z/zuu/issues/550 — four defects that would produce divergent implementations: the `ERR_NO_ACCESS`/`ERR_UNAVAILABLE` contradiction (which reopens the existence oracle §10 exists to close), `BIND_SEND` being unable to pass §5.1 step 5 *and* acting as an unauthenticated **destructive** existence probe, a chosen-difficulty claim with no difficulty anywhere in the tree, and §14.3's unspecified `lowercase()` letting U+212A KELVIN collide with handle `k`.
- https://github.com/free2z/zuu/issues/551 — ADR 0014 introduces a platform-held offline reset authority under a heading that says these are *not* owner decisions. The ADR is admirably honest that such a recovery is cryptographically indistinguishable from an attack, which is exactly why it's the owner's call — and `THREAT-MODEL.md:71` still claims identity keys cannot be substituted undetectably.
- https://github.com/free2z/zuu/issues/552 — over-claims and undefined behaviour, including that ADR 0008's "already in the dependency graph / already ships to WASM" reason is false today (`tls_codec` is in zero of 43 `Cargo.lock` files), and that `relay_proof` covers none of `HelloResponse` — so under `transport_security: none` an active attacker can downgrade `channel_binding_mode` and lie about `capabilities_digest` with the proof still verifying.

One line I'd strike outright: `WIRE.md:1178-1182` describes the padding size check as a defence "against a malicious client trying to use length as a covert channel to a colluding relay." A colluding relay simply doesn't enforce it. The sloppy-client half is fine; that half can't hold.

PR #549, opened 2026-08-23
ci(rust): let the Rust hygiene gates police a second tree

### LINKED ISSUE #543: ci(rust): hygiene scripts are hardcoded to wallet/ and would not cover a second Rust tree
Part of #305 (Phase 1 implementation wave).

## Problem

`scripts/check-rust-fmt.sh`, `check-rust-clippy.sh` and `check-rust-deny.sh` all hardcode `RUST_ROOT=$REPO_ROOT/wallet` and discover crates by `find` from there. `scripts/check-rust-toolchain.sh` carries a literal `MANIFESTS=(wallet/...)` list.

E2EE messaging introduces a new top-level Rust namespace (`rs/`). As written, **none of these gates would cover it** — it would ship un-formatted, un-linted and outside the supply-chain policy, silently, from its first commit.

Two of them also hard-require the `z/zcash/librustzcash` submodule checkout, which `rs/` does not need.

## Scope

- Add `--root <dir>` (default `wallet`, so current behaviour is unchanged) and `--config <deny.toml>` to the three check scripts.
- Make the `z/zcash/librustzcash` requirement **per-root** rather than unconditional.
- Teach `check-rust-toolchain.sh` to accept additional toolchain files and manifests so a second tree's pin is verified against the same single source of truth.
- Keep every script's `--self-test` working, and extend it to cover the new flag.

## Acceptance

- [ ] `wallet/` behaviour is byte-identical — the existing `zuuli.yml` gate passes unchanged.
- [ ] Each script's `--self-test` still passes and exercises `--root`.
- [ ] No new crates or trees are added by this PR. It is purely the capability to police a second tree.

Blocks the `rs/` skeleton PR.

skyl:
Closes #543. Part of #305.

The three crate gates hardcoded `RUST_ROOT=$REPO_ROOT/wallet`, and
`check-rust-toolchain.sh` carried a literal `MANIFESTS=(wallet/...)` list plus
`TOOLCHAIN_FILE=wallet/rust-toolchain.toml`. The `rs/` namespace would have
landed with **zero** coverage from all four. This PR is **purely the
capability** — no `rs/` directory, no new crate, no new tree.

## What changed

**`--root <dir>` on `check-rust-fmt.sh`, `check-rust-clippy.sh`,
`check-rust-deny.sh`**, defaulting to `wallet`. An absolute `--root` is accepted
so the self-tests can point at a scratch tree. Every message that said `wallet/`
now says the root it actually judged, which for the default is the same string.

**`--config <path>` on `check-rust-deny.sh`**, defaulting to `<root>/deny.toml`.
Which advisories we accept is a judgement about a particular dependency graph,
and a server graph is not the wallet's — so a second tree gets its own policy
rather than an `ignore` list that means two different things at once.

**clippy's `z/zcash/librustzcash` requirement is now derived, not assumed.** The
crates that reach into the submodule say so themselves, by path dependency
(`path = "../../../z/zcash/librustzcash/..."`), so the requirement is read out
of the discovered manifests instead of being demanded unconditionally.
`wallet/` is still held to it; a tree that never points into it is not. The
self-test asserts the derivation **both ways**, so it cannot go silent for
`wallet` without the gate saying so.

**`check-rust-toolchain.sh` gains `--toolchain-file` and `--manifest`.** A second
tree needs its own `rust-toolchain.toml`, because Cargo picks the toolchain from
the directory it runs in — but that file is a **restatement**, not a second
decision. Registered files must repeat `wallet/rust-toolchain.toml`'s channel
exactly, and naming the source of truth as a restatement of itself is an error
rather than a tautological pass. `--print-channel` is untouched;
`.github/workflows/zuuli.yml` depends on it and still gets `1.97.1`.

**Self-tests.** `check-rust-toolchain.sh --self-test` keeps all 14 existing
drift cases and gains a synthetic second tree; `check-rust-clippy.sh --self-test
<target-os>` keeps its target-native negative control and gains four more.
`check-rust-fmt.sh` and `check-rust-deny.sh` had no self-test at all and now
have one each, negative control included. The gate runs both new ones — a check
nobody has watched fail is not a check.

## Verification (macOS, pinned 1.97.1, librustzcash checked out)

Old = `scripts/check-rust-*.sh` as it stands on `origin/main`, run from this same
worktree so `REPO_ROOT` resolves identically.

### `check-rust-fmt.sh` — byte-identical

```
diff old-no-arg new-no-arg          -> identical
diff old-no-arg new "--root wallet" -> identical

==> checking wallet/plugins/tauri-plugin-zcash
==> checking wallet/zuuallet/src-tauri
==> checking wallet/zuuli/crypto-target-spike
==> checking wallet/zuuli/src-tauri
==> checking wallet/zuuli/wasm-spike
All 5 Rust crate(s) under wallet/ are formatted (rustfmt from 1.97.1).
```

```
$ scripts/check-rust-fmt.sh --self-test
self-test: --root refuses a tree with no crates to judge.
self-test: --root accepts a formatted second tree.
self-test: --root rejects an unformatted crate.        <- negative control
self-test: --fix formats crates under --root.
self-test: 4 case(s) passed; rustfmt from 1.97.1.
```

### `check-rust-deny.sh` — byte-identical

`cargo-deny` interleaves its own warning blocks nondeterministically across
runs (old-vs-old differs too), so the script's own output is compared on stdout,
and the full stream is compared sorted.

```
diff old-stdout new-stdout                          -> identical
diff old-stdout new-stdout "--root wallet"          -> identical
diff <(sort old) <(sort new)                        -> identical
diff <(sort old) <(sort new --root wallet)          -> identical
diff <(sort old) <(sort new --root wallet --config wallet/deny.toml) -> identical

==> cargo-deny check all -- wallet/plugins/tauri-plugin-zcash
... (5 crates) ...
All 5 Rust crate(s) under wallet/ satisfy wallet/deny.toml.

$ scripts/check-rust-deny.sh --print-version
0.20.2                     # old prints the same
```

```
$ scripts/check-rust-deny.sh --self-test
self-test: --root refuses a tree whose deny.toml is missing.
self-test: --root enforces <root>/deny.toml.
self-test: --config decides the verdict (a forbidden licence fails).   <- negative control
self-test: --config refuses a policy file that is not there.
self-test: 4 case(s) passed; cargo-deny 0.20.2.
```

The negative control is one scratch crate judged under two policies that
disagree about its licence. Passing both ways would mean `--config` is
decorative, which is the whole mechanism a second tree's policy rests on.

### `check-rust-clippy.sh`

The `wallet/` verdict is the ~45-minute job this PR deliberately triggers; the
required `rust_clippy` and `rust_native_clippy` gates are the proof and this PR
sits inside `zuuli.yml`'s `changes` glob so all of them run. Locally verified:

```
$ scripts/check-rust-clippy.sh                     # submodule deliberately absent
z/zcash/librustzcash is not checked out; clippy resolves and compiles the whole graph.
Run: git submodule update --init z/zcash/librustzcash
exit=1
$ scripts/check-rust-clippy.sh --root wallet       # identical, same exit
```

That is the point of item 3: making the requirement conditional did **not**
loosen it for `wallet`.

```
$ scripts/check-rust-clippy.sh --self-test macos
Clippy -D warnings rejected the target-native negative control for macos on x86_64-apple-darwin.
self-test: wallet/ still requires the z/zcash/librustzcash checkout.
self-test: a tree that never points into z/zcash/librustzcash is not held to it.
self-test: --root rejects a crate with a known clippy warning.     <- negative control
self-test: --root accepts a clippy-clean second tree.
```

And from this PR's own `Rust / native lints (windows)` job — the extended
self-test and the full `wallet/` verdict, on the required gate:

```
Run scripts/check-rust-clippy.sh --self-test "windows"
Clippy -D warnings rejected the target-native negative control for windows on x86_64-pc-windows-msvc.
self-test: wallet/ still requires the z/zcash/librustzcash checkout.
self-test: a tree that never points into z/zcash/librustzcash is not held to it.
self-test: --root rejects a crate with a known clippy warning.
self-test: --root accepts a clippy-clean second tree.
...
All 5 Rust crate(s) under wallet/ are clippy-clean (clippy from 1.97.1).
```

### `check-rust-toolchain.sh` — byte-identical

```
diff old-no-arg new-no-arg -> identical

note: 1 Rust compiler selection(s) intentionally follow the stable channel (upstream canaries).

$ scripts/check-rust-toolchain.sh --print-channel   ->  1.97.1   (old: 1.97.1)
$ scripts/check-rust-toolchain.sh --print-msrv      ->  1.97
$ scripts/check-rust-toolchain.sh --print-targets   ->  wasm32-unknown-unknown
```

`--self-test` keeps all 14 existing cases verbatim and adds five:

```
self-test: the unmodified tree passes.
... 14 existing drift cases, unchanged ...
self-test: a second Rust tree that restates the pin passes.
self-test: caught drift when a second tree pins a channel of its own.     <- negative control
self-test: caught drift when a second tree's crate MSRV drifts.           <- negative control
self-test: caught drift when a second tree loses its toolchain file.      <- negative control
self-test: the source of truth is rejected as a restatement of itself.    <- negative control
```

And the flags on a real tree:

```
$ scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml
drift: rs/rust-toolchain.toml is missing

$ scripts/check-rust-toolchain.sh --toolchain-file wallet/rust-toolchain.toml
drift: wallet/rust-toolchain.toml is the source of truth, not a restatement of it

$ scripts/check-rust-toolchain.sh --manifest wallet/zuuli/src-tauri/Cargo.toml

$ scripts/check-rust-toolchain.sh --manifest wallet/rust-toolchain.toml
drift: wallet/rust-toolchain.toml does not declare rust-version
```

### Everything else in `scripts/`

`check-github-actions-pins.mjs`, `check-public-repo-boundary.sh`,
`check-zcash-permissions.mjs`, `check-zuuli-android-gate.mjs` and
`check-zuuli-linux-image.mjs` all pass. `shellcheck` finding counts are
unchanged from `origin/main` for all four scripts (3 / 2 / 5 / 3).

## Notes for review

- The two new gate steps (`rust_fmt` and `rust_deny` self-tests) add seconds to
  jobs that already install the toolchain and, for deny, already fetch the
  pinned `cargo-deny`. They are added only in `zuuli.yml`: the self-test is
  about the script, not the tree, so running it once in the required gate is
  enough — the `zuuallet.yml` mirrors keep running the real checks only.
- The clippy submodule check moved after manifest discovery, since that is what
  it is now derived from. It still runs before any crate is linted.

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -239,6 +239,12 @@ jobs:
           toolchain: ${{ steps.rust_toolchain.outputs.version }}
           components: rustfmt
 
+      # A check nobody has watched fail is not a check: this builds a scratch
+      # tree under --root and proves the verdict tracks it, unformatted case
+      # included. Seconds, and it is what keeps --root honest for a second tree.
+      - name: Prove the formatting gate still fails
+        run: scripts/check-rust-fmt.sh --self-test
+
       - name: Check formatting of every Rust crate under wallet/
         run: scripts/check-rust-fmt.sh
 
@@ -278,6 +284,12 @@ jobs:
       # the version and digests are pinned in the script. Deliberately not a
       # third-party installer action — the tool that polices our supply chain
       # should not arrive through an unpinned one.
+      # Judges one scratch crate under two policies that disagree about its
+      # licence. Passing both ways would mean --config is decorative, which is
+      # the whole mechanism a second tree's own policy rests on.
+      - name: Prove the policy still decides the verdict
+        run: scripts/check-rust-deny.sh --self-test
+
       - name: Check advisories, licences and sources
         run: scripts/check-rust-deny.sh
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,7 +173,13 @@ integrate, and move it forward alongside everything else.
   and a check that ran on whatever was installed would report version skew as a
   code defect. Each has a mirrored job in `zuuallet.yml`, because the required
   `gate` lives in `zuuli.yml` and its change detector does not select
-  `wallet/zuuallet/**`.
+  `wallet/zuuallet/**`. Which tree they judge is `--root`'s to say (default
+  `wallet`, so the wallet gate is unchanged), `--config` chooses the policy
+  `check-rust-deny.sh` enforces, and clippy's `z/zcash/librustzcash`
+  requirement is read out of the manifests rather than assumed — so a second
+  top-level Rust namespace is policed by the same three gates without
+  inheriting the wallet's path-dependency baggage. All three carry a
+  `--self-test` that proves they still fail on the thing they exist to catch.
 - **Target-native clippy is part of the required gate.** Linux cannot compile
   macOS- and Windows-only Keychain, Credential Manager, filesystem, and OAuth
   branches. The `rust_native_clippy` matrix in `.github/workflows/zuuli.yml`
@@ -209,6 +215,13 @@ read a TOML file at the moment they need the value:
 | `dtolnay/rust-toolchain@<sha> # <version>` in packaging/release and target-native Zuuallet jobs | The action's **version branches hardcode the compiler in `action.yml` and do not declare a `toolchain` input at all** — a commit-pinned ref *is* the version pin, and the trailing comment is its only readable record |
 | `dtolnay/rust-toolchain@<sha> # stable` in source-derived gate/Zuuallet jobs | `uses:` cannot contain an expression, so the generic action implementation is commit-pinned while its `toolchain:` input still reads the version from `wallet/rust-toolchain.toml`; the upstream canary deliberately omits that input so only its compiler selection follows `stable` |
 | MSRV/`cargo +<version>` lines in the wallet READMEs, the plugin `CLAUDE.md`, and `wallet/zuuli/docs/releasing.md` | Prose |
+
+A second top-level Rust tree does not get a decision of its own either. It needs
+its own `rust-toolchain.toml`, because Cargo picks the toolchain from the
+directory it runs in, and its crates carry their own `rust-version` — both are
+restatements. Register them with `scripts/check-rust-toolchain.sh
+--toolchain-file <path>` and `--manifest <path>` and they are held to
+`wallet/rust-toolchain.toml` exactly like every row above.
 
 The `wallet/zuuli` gate reads the channel out of the file (`--print-channel`) and
 feeds it to the toolchain action, so the required CI jobs hold **no literal at
--- a/scripts/check-rust-clippy.sh
+++ b/scripts/check-rust-clippy.sh
@@ -1,25 +1,32 @@
 #!/usr/bin/env bash
-# Hold every Rust crate under wallet/ to `cargo clippy -D warnings`.
+# Hold every Rust crate under a Rust tree to `cargo clippy -D warnings`.
 #
 # Clippy's lint set is not stable across compiler versions: lints are added,
 # promoted between groups, and taught new patterns with every release, so the
 # same source can be clean on one toolchain and warn on the next. A check that
 # ran on whatever compiler happened to be installed would report that version
 # skew as if it were a code defect. So, exactly like scripts/check-rust-fmt.sh,
-# this renders a verdict only on the toolchain named by
-# wallet/rust-toolchain.toml — it runs cargo from the directory holding that
-# file so rustup selects the pin, and refuses to continue if a different
-# compiler answers. Bumping the pin is therefore the one moment new lints can
-# appear, and they appear in the bump's own pull request.
+# this renders a verdict only on the toolchain named by the tree's
+# rust-toolchain.toml — it runs cargo from the directory holding that file so
+# rustup selects the pin, and refuses to continue if a different compiler
+# answers. Bumping the pin is therefore the one moment new lints can appear, and
+# they appear in the bump's own pull request.
 #
-# Crates are discovered rather than listed. A new crate under wallet/ is gated
+# The tree defaults to wallet/. A second top-level Rust namespace is gated by
+# pointing --root at it; its rust-toolchain.toml is a restatement of
+# wallet/rust-toolchain.toml, not a second decision, and
+# scripts/check-rust-toolchain.sh is what holds it to that.
+#
+# Crates are discovered rather than listed. A new crate under the tree is gated
 # from its first commit, which is the cheapest moment and never gets cheaper.
 #
 # Unlike scripts/check-rust-fmt.sh this genuinely compiles: clippy is a rustc
-# driver and needs the whole dependency graph, so z/zcash/librustzcash must be
-# checked out — the wallet crates reach the Zcash crates by path into it — and
-# the Tauri system libraries must be installed. Budget it alongside the real
-# builds, not alongside the formatting check.
+# driver and needs the whole dependency graph. The wallet crates reach the Zcash
+# crates by path into z/zcash/librustzcash, so for that tree the submodule must
+# be checked out and the Tauri system libraries installed — budget it alongside
+# the real builds, not alongside the formatting check. That requirement is read
+# out of the manifests rather than assumed, so a tree whose crates never point
+# into the submodule is not held to a checkout it does not use.
 #
 # --all-targets so tests, benches and examples are linted too. Test code that
 # nobody lints is where the next warning backlog comes from.
@@ -27,6 +34,7 @@
 # Usage:
 #   scripts/check-rust-clippy.sh          fail on any clippy warning
 #   scripts/check-rust-clippy.sh --self-test <target-os>
+#   scripts/check-rust-clippy.sh --root DIR   lint a different Rust tree (default: wallet)
 #   scripts/check-rust-clippy.sh --fix    apply clippy's machine-applicable fixes
 #
 # After --fix, read every edit before committing it and run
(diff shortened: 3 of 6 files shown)


skyl:
Train coordination: this PR may land before the #239 compatibility-mutant recovery. The recovery also edits `.github/workflows/zuuli.yml`; its unpublished branch will rebase onto the resulting `main` and rerun all workflow mutation suites, so please merge #549 first once its exact CI is green rather than merging both concurrently.

skyl:
Independent mutation audit on exact head `281a7433b4d93e60e55232e0e2e0f563a7e0923e` passed under the host's actual GNU bash 3.2.57 (the compatibility environment that invalidated earlier shell guards).

Executed, not inferred:

- `/bin/bash scripts/check-rust-fmt.sh --self-test`: 4/4, including unformatted second-tree red and restored/fixed green
- `/bin/bash scripts/check-rust-toolchain.sh --self-test`: all 14 original drift mutants plus second-tree channel/MSRV/missing-toolchain/self-restatement mutants killed
- `/bin/bash scripts/check-rust-deny.sh --self-test`: 4/4, including the same fixture under allow/deny policies proving `--config` changes the verdict
- `/bin/bash scripts/check-rust-clippy.sh --self-test macos`: target-native warning red, wallet submodule requirement preserved, no-dependency tree exempted, second-tree warning red and restored green
- `git diff --check a24d60a...HEAD`: clean

The negative controls exercise `--root`/`--config` against scratch trees; none are comment- or source-string-satisfiable. I found no blocking defect in the capability diff. Exact-head hosted matrix remains authoritative before admin squash merge.

skyl:
Reviewed adversarially on the owner's behalf. **Approving** — merging once `gate` and `Rust / native lints (macos)` go green (both still running).

This is the cleanest large refactor I've reviewed here. ~600 changed shell lines, 131 deletions, and **not one control vanished** — I had the review account for every removed line rather than read the diff.

### It blocks, and I checked the mechanism

The `+12` in `zuuli.yml` is **two `run:` steps inside jobs that already feed the gate** (`rust_fmt:245-246`, `rust_deny:290-291`), not new jobs. `gate.needs` already lists both, and the gate now consumes `REQUIRED_JOBS_JSON: ${{ toJSON(needs) }}` via `--verify-gate-results`, so the #515/#517/#518 "new job must be exposed and consumed" policy isn't even engaged. A `--self-test` failure fails the step → the job → the gate. The third self-test was already gate-bound in `rust_native_clippy`.

### Every script actually asserts — mutation-tested against a real synthetic tree

Not against the scripts' own fixtures, which is the distinction that matters:

```
fmt --root rs       badly-formatted fn        -> exit 1, names rs/relay, prints the diff
clippy --root rs    clippy::needless_bool     -> exit 1, "1 crate(s) failed the clippy gate: rs/relay"
deny --root rs      licence policy violation  -> exit 1
deny --config …     wallet policy vs GPL crate-> exit 1   (--config is load-bearing, not decorative)
deny --root rs      [bans] deny anyhow        -> exit 1
toolchain           channel 1.96.0            -> exit 1, "drift: pins 1.96.0, expected 1.97.1"
toolchain           rust-version 1.90         -> exit 1
toolchain           file deleted              -> exit 1, "drift: … is missing"
```

No script merely walks a second directory.

### Zero bare `[[ ]]` across ~600 shell lines

Given #477, this was the first thing I wanted counted. Every added `[[ ]]` and `(( ))` is in an enforcing position — `if`, `while`, `|| fail`, `|| continue`, `|| return`. **Bare statements introduced: 0. Inert occurrences: 0.**

That matters more than usual here because two of these scripts run on macOS: `check-rust-clippy.sh` and `check-rust-toolchain.sh` are both invoked from `rust_native_clippy` (`macos-latest` + `windows-latest`). Both macOS-exposed paths were exercised on bash 3.2.57 directly, and `Rust / native lints (windows)` is already green with the new step.

The two `(( … )) && arr=(…)` guard idioms at `check-rust-toolchain.sh:612,662` are intentionally non-fatal, and a false left side was verified not to abort or leak a nonzero status past the list.

### The deletions

`-63`/`-25`/`-33`/`-10` are `case`→`while` parser conversions, message parameterisation, and one `find` block hoisted into `discover_manifests()` **byte-for-byte**. Old and new default output are byte-identical for fmt, and for deny (same stream, same exit). `--print-version` still `0.20.2`, `-h` still exit 0. Net *stricter* in two places: `check-rust-deny.sh --bogus` now exits 2 with a message instead of being handed to cargo-deny as a check name.

One genuine relaxation, tested at the real condition rather than reasoned about:

```diff
-if [[ ! -f $SUBMODULE/Cargo.toml ]]; then
+if manifests_need_submodule "${manifests[@]}" && [[ ! -f $SUBMODULE/Cargo.toml ]]; then
```

In a worktree with `z/zcash/librustzcash` genuinely absent, old and new both exit 1 with the same message, with and without `--root wallet`. **Not weakened for `wallet/`**, and the derivation is now asserted both ways by the gate-bound self-test.

And `check-rust-toolchain.sh:516`'s new `[[ -e … ]] || continue` isn't a lost control — a registered-but-absent path is now reported by `run_checks` as `drift: <path> is missing` rather than killing the staging loop, which is what makes `--manifest`/`--toolchain-file` able to report a missing file at all.

### The money crate stays covered

Verified explicitly rather than assumed: `wallet/plugins/tauri-plugin-zcash` appears in every discovery pass, clippy shares the identical `find` expression, and it remains a literal `MANIFESTS` entry. `git ls-files 'wallet/**Cargo.toml'` returns exactly the 5 crates the scripts discover. `zuuallet.yml` isn't touched at all.

### Cost: ~+8s per PR, no new runner slots

Measured per step from this PR's own run: `changes` +3s, `rust_fmt` +3s, `rust_deny` +1s, `rust_native_clippy` +5s. The deny self-test is effectively free — `install_cargo_deny` caches per version, so it moves the download earlier rather than repeating it.

### Three nits worth fixing, one before merge if it's cheap

1. **`AGENTS.md:180` overstates.** "so a second top-level Rust namespace **is policed** by the same three gates" — present tense reads as accomplished, but nothing is policed until a workflow passes `--root`. The PR body is scrupulously honest ("purely the capability"); the doctrine file should match. Suggest "*can be* policed … once its root is registered in `.github/workflows/zuuli.yml`". Worth adding the CI-wiring half next to the CLI recipe at `:219-225` too — that's exactly where an unregistered `rs/` slips through.
2. **`check-rust-toolchain.sh:596` prints a stale count** — `${#SELF_TEST_MUTATIONS[@]}` says "14 drift case(s) caught" while the run now catches **18**. A recorded number that no longer describes what ran is the rot this repo keeps catching elsewhere.
3. **`check-rust-fmt.sh:110-114` mislabels its first self-test case.** It prints `--root refuses a tree with no crates to judge`, but the fixture has no `rust-toolchain.toml` at that point, so the sub-invocation dies on *that* branch instead. The "gone blind" branch is never exercised. One `printf '[toolchain]\nchannel=…'` before the assertion fixes it.

### The follow-up I'd most want

Right now a registered-but-absent root is fine, but an **unregistered** root is silently invisible: with `rs/` present and not registered, `check-rust-toolchain.sh` prints "Every Rust toolchain pin agrees…" and exits 0. Registration is explicit by design, like `MANIFESTS`.

So #543's own stated failure mode — "it would ship un-formatted, un-linted and outside the supply-chain policy, silently, from its first commit" — is still *reachable*; this PR makes it fixable, not impossible. Add an assertion that every tracked `Cargo.toml` outside `z/` lives under a policed root. Today `git ls-files '*Cargo.toml' | grep -v '^z/'` is exactly the 5 wallet crates, so it's cheap, and it would fail the instant `rs/relay/Cargo.toml` lands unregistered — turning registration from a discipline into a control. Filing that.

Issue #553, opened 2026-08-23
ci(rust): an unregistered Rust root is silently unpoliced — make registration a control, not a discipline


skyl:
## Problem

#549 gave the four Rust hygiene gates a `--root` / `--config` / `--toolchain-file` / `--manifest` capability so a second tree (`rs/`, the E2EE relay namespace from #305) *can* be policed. Every script genuinely asserts — mutation-tested against a synthetic `rs/` tree, all four catch real violations.

But **registration is explicit**, so an *unregistered* root is silently invisible. With `rs/` present and not wired into `zuuli.yml`:

```
$ scripts/check-rust-toolchain.sh
Every Rust toolchain pin agrees…
exit 0
```

That means #543's own stated failure mode is still reachable:

> it would ship un-formatted, un-linted and outside the supply-chain policy, silently, from its first commit

#549 makes that **fixable**, not **impossible**. The gap closes only if someone remembers to pass `--root rs` when the skeleton lands — and "someone remembers" is the failure mode this repo keeps designing out.

## Suggested fix

Assert that every tracked `Cargo.toml` outside `z/` lives under a policed root.

```
git ls-files '*Cargo.toml' | grep -v '^z/'
```

Today that returns exactly the 5 wallet crates the scripts already discover, so the check is cheap and passes immediately. It would fail the instant `rs/relay/Cargo.toml` lands without its root being registered — turning registration from a discipline into a control.

Natural home is `scripts/check-rust-toolchain.sh`'s existing `MANIFESTS`/`DOC_PINS` registry logic, since that is already the place that knows what is registered.

## Acceptance

- Adding a tracked `Cargo.toml` outside `z/` and outside every policed root fails CI.
- The check is mutation-proven: create such a manifest, watch it go red.
- The existing 5-crate wallet set still passes unchanged.

## Two smaller items from the same review

1. **`AGENTS.md:180` overstates.** "so a second top-level Rust namespace **is policed** by the same three gates" — present tense, but nothing is policed until a workflow passes `--root`. Should read "*can be* policed … once its root is registered in `.github/workflows/zuuli.yml`". The CLI recipe at `AGENTS.md:219-225` should also carry the CI-wiring half, since that omission is exactly where an unregistered tree slips through.
[...]

Issue #552, opened 2026-08-23
spec(e2ee): WIRE.md over-claims, undefined behaviour, and a relay_proof downgrade


skyl:
## Two claims the tree does not support

**1. ADR 0008's second decisive reason is factually false today.**

`WIRE.md:199-203` / `decisions/0008`:
> **It is already in the client's dependency graph.** OpenMLS is built on `tls_codec`… The crypto core already contains this decoder, **already ships it to WASM**, and **already has it in whatever audit scope the core has**.

and `decisions/0008:109-110`:
> it is exercised by every MLS implementation built on it, **including ours**.

There is no OpenMLS in this repo, no crypto core, no WASM crypto bundle, no audit scope, and `tls_codec` appears in **zero of the 43 `Cargo.lock` files** in the tree, submodules included. The only crypto-adjacent Rust outside the wallet is `wallet/zuuli/crypto-target-spike/`, an explicitly isolated capability probe with fixed test bytes.

The *decision* is still right — this is a tense problem, not a wrong choice. But it is load-bearing (reason 2 of 3), and a frozen spec shouldn't assert a dependency-graph fact that isn't. Should read "will be, once ADR 0001's core lands."

**2. `WIRE.md:1834-1835`** — "no source file from that project has been read **by anyone on this project**" is an unverifiable claim about third parties, in a document whose whole value is that it doesn't assert what it can't support. Also #545 asked to record *who* performed the check; "an agent working on #545" is thin attribution for a licensing posture.

## Undefined behaviour in a document that claims to be frozen

- **`send_retry_max` is used and never defined.** `WIRE.md:839` makes it the normative backoff bound; it appears nowhere else — not in `Capabilities`, no default.
- **`ERR_QUOTA` is unimplementable.** `:1220` defines code 14 as "queue count, creation rate" — neither quota is specified or published anywhere.
- **§11.1's central claim is false.** `:1266` calls `Capabilities` "A relay's entire externally-relevant policy, in one signed structure," and `decisions/0012:80` says "**Everything is published.**" The struct omits every Layer-1 limit (`:1595-1599`: per-source concurrent connections, new-connection rate, per-connection command rate), the per-queue append rate, the max queue count, and the Layer-4 published high-water marks (`:1641-1642`). Either publish them or drop the "everything"/"entire" claims.
[...]

Issue #550, opened 2026-08-23
spec(e2ee): four WIRE.md defects that would produce divergent implementations


skyl:
## Problem

`docs/e2ee/WIRE.md` (frozen in #548) contains four defects that would produce **divergent implementations** or reopen an oracle the spec explicitly closes elsewhere. All four should be fixed before any implementation PR starts, because a frozen contract is much cheaper to correct now than after two clients disagree.

The spec is otherwise unusually good — the signing transcript is genuinely unambiguous, the cross-relay `ACK` replay binding is a real catch, and delete-on-ack does not create a deletion oracle. These are the exceptions.

## 1. `ERR_NO_ACCESS` vs `ERR_UNAVAILABLE` contradict, and the contradiction reopens the existence oracle

`WIRE.md:431-432` (normative verification order):
> 5. Look up the address and check that `signer_key` is the key registered for it. Fail → `ERR_NO_ACCESS` (§10 — the same code as "no such queue").

`WIRE.md:830-833`, `:1221`:
> every send-side refusal that would distinguish queue state collapses to the single code `ERR_UNAVAILABLE`
> | 15 | `ERR_UNAVAILABLE` | **Send side only.** Absent, deleted, expired…

So `APPEND` to an absent address is code 10 under §5.1 and code 15 under §6.3/§10.

Worse, §6.3's collapse list omits *"exists but your key doesn't authorize it"*. Under §5.1 a wrong-key `APPEND` returns 10 while a full or absent queue returns 15 — so an attacker holding a leaked `send_addr` can distinguish a live bound queue from everything else. That is exactly what §10's existence-oracle rule exists to prevent.

## 2. `BIND_SEND` cannot pass §5.1 step 5, and it is an unauthenticated destructive existence oracle

`WIRE.md:929` states the send side is unbound — no key is registered. Step 5 (`:431`) requires `signer_key` to be *the key registered for the address*, so **every legitimate first `BIND_SEND` fails**. `CREATE_QUEUE` has an explicit carve-out (`:713-715`); `BIND_SEND` and `CREATE_CONTACT_QUEUE` have none.

Then the oracle: `ERR_ALREADY_BOUND` (11) is distinguishable from `ERR_NO_ACCESS` (10), and `BIND_SEND` requires no prior authorization — you sign with a key you just generated. Anyone holding a candidate `send_addr` (the "leaked log line, partially-observed advert, database from a decommissioned relay" the spec itself names at `:1236-1239`) gets a free existence probe.

[...]

PR #560, opened 2026-08-24
feat(rs): add the rs/ workspace, f2z-codec, and its own required gate

### LINKED ISSUE #555: E2EE Phase 1: rs/ workspace skeleton, f2z-codec, and its own CI gate
Part of #305. Unblocked by #543 (the hygiene scripts can now police a second tree). Prerequisite for every `rs/` crate that follows.

## Scope

**1. The `rs/` namespace and workspace.**

A new top-level `rs/`, alongside the existing `ts/`, `py/`, `z/`, `wallet/`. One `[workspace]`, one `rs/Cargo.lock`.

Named `rs/` and **not** `relay/` or `srv/` deliberately: ADR 0001 requires one Rust crypto core compiled natively for ZUULI *and* to WASM for the web, so `rs/` will hold client crates beside the servers. A namespace named for one role would be wrong within a quarter.

`rs/README.md` must state that this is a greenfield workspace decision and **explicitly does not decide #341** — every argument against consolidation there is wallet-specific (Tauri mobile layout, deliberate divergent resolution, the `links = "sqlite3"` singleton) and none applies to a fresh tree. If #341 later chooses a single repo-wide workspace, `rs/` folds in with one edit.

`rs/rust-toolchain.toml` restates `channel = "1.97.1"` and is registered with `scripts/check-rust-toolchain.sh --toolchain-file` so it is verified as a **restatement** of the single source of truth at `wallet/rust-toolchain.toml`, never as a second source.

**2. `rs/crates/f2z-codec`.**

The canonical encoding layer every other crate sits on: `tls_codec` wrappers, the domain-separated signing-transcript builder, the newtypes for addresses/payloads/keys with **redacting `Debug`**, and the padding-bucket validator. Per `docs/e2ee/WIRE.md`.

Non-negotiable properties, and they are the reason this crate exists separately:

- `#![forbid(unsafe_code)]`, `no_std` + `alloc`, **no I/O, no tokio**.
- Builds for `wasm32-unknown-unknown`. This is enforced by a CI job, not by convention — it is the boundary that keeps the web client possible at all.
- The **re-encode-equality** property from `WIRE.md` §4 is implemented here once and property-tested, so there is no parse-versus-verify gap anywhere in the system.
- A test asserts `format!("{:?}", …)` on the address/payload/key newtypes contains no hex. `--log-level trace` must never become a ciphertext archive.

**Licence: permissive**, matching the repo. Per the owner decision on #305, the *server binaries* are AGPL-3.0 but the shared crates stay permissive so the ZUULI client and third-party implementations can link them. Set `license` explicitly in every `Cargo.toml` from the first commit — the boundary must never be implicit.

**3. `.github/workflows/rs.yml`, with its own `gate`.**

A separate workflow, not an extension of `zuuli.yml`. Mixing a 45-minute wallet build into the feedback loop of a messaging PR is exactly what this avoids, and `zuuli.yml`'s change detector already correctly yields `zuuli=false` for an `rs/`-only diff.

Jobs: `changes` (same shell-`case` idiom, **failing open to true**), `rs_fmt`, `rs_clippy`, `rs_deny`, `rs_test`, and **`rs_wasm`** (`cargo build --target wasm32-unknown-unknown`), plus `gate`.

**Every job in `gate`'s `needs:` must also appear in its inline `results` string.** `zuuli.yml:469` already carries the comment explaining why — a job that is awaited but not inspected makes the gate decorative for that job.

**4. `scripts/check-workflow-gates.mjs`.**

Make that rule mechanical instead of a comment. Parse every workflow, find each `gate` job, and assert the set of names in `needs:` (minus `changes`) equals the set bound to `needs.<job>.result` env vars **and** referenced in the `results=` line. Run it from `changes` in both workflows, with a `--self-test` carrying a negative control, matching the convention the other scripts follow.

This closes the decorative-gate hazard for `zuuli.yml` too, which is the larger win.

**5. `rs/deny.toml`.**

Separate from `wallet/deny.toml`: different graph, different policy. **Start with an empty `ignore` list** — a greenfield tree has no measured backlog to grandfather, and #351 exists because that backlog is real work to pay down.

Include a licence assertion enforcing the permissive/AGPL split described above, so the boundary cannot drift silently.

## Acceptance

- [ ] `rs/` builds, tests, and passes fmt/clippy/deny under the shared scripts via `--root rs`.
- [ ] `f2z-codec` compiles for `wasm32-unknown-unknown` in CI.
- [ ] `check-workflow-gates.mjs` passes, and its `--self-test` demonstrably fails on an injected mismatch.
- [ ] **An `rs/**`-only diff yields `zuuli=false`** (no 45-minute wallet builds on messaging PRs) and a `wallet/**`-only diff yields `rs=false`. Assert both in the PR description with evidence.
- [ ] The `rs.yml` gate must be added as a required status check in branch protection — call this out explicitly in the PR body as a repo-settings ask, since it cannot be done in a file.

skyl:
Opens the E2EE Phase 1 tree: a top-level `rs/` workspace, `f2z-codec` implementing the encoding layer of `docs/e2ee/WIRE.md`, and a separate required gate for it.

Closes #555

## What is here

**`rs/`** — one `[workspace]`, one `Cargo.lock`, `rust-toolchain.toml` restating `channel = "1.97.1"`, `deny.toml`, `README.md`. The README states plainly that this is a greenfield workspace decision and **does not decide #341**, and says why: every argument against consolidation there is wallet-specific (Tauri's mobile layout, the deliberately divergent lockfile resolutions, the `links = "sqlite3"` singleton) and none of it applies to a tree with no Tauri, no SQLite and no path dependencies into `z/`.

`rs/rust-toolchain.toml` and `rs/crates/f2z-codec/Cargo.toml` are registered with `scripts/check-rust-toolchain.sh` through the `--toolchain-file` / `--manifest` flags from #549, so both are held to `wallet/rust-toolchain.toml` as restatements. Neither inherits from `[workspace.package]`: that check reads the literal `rust-version` line out of each crate's own manifest, and the licence is worth stating in the file a reader of that crate opens.

**`rs/crates/f2z-codec`** — `#![forbid(unsafe_code)]`, `no_std` + `alloc`, no I/O, no tokio, builds for `wasm32-unknown-unknown`. Contents: `tls_codec` wrappers for every structure `WIRE.md` defines in §4 (framing), §5 (`CommandAuth` / `SignedAuth` / `CommandTranscript`), §6 (the full command set and push bodies), §11.1 (`Capabilities` / `SignedCapabilities`), §12.2 (contact queues) and §13.1 (PoW); the domain-separated transcript builder; the redacting newtypes; the padding-bucket validator.

**`.github/workflows/rs.yml`** — its own `changes` / `rs_fmt` / `rs_clippy` / `rs_deny` / `rs_test` / `rs_wasm` / `gate`, reusing the shared scripts via `--root rs`. Every action is SHA-pinned with a version comment; the toolchain version is read from `scripts/check-rust-toolchain.sh --print-channel` and never restated in YAML.

**`scripts/check-workflow-gates.mjs`** — wired into the `changes` job of both `zuuli.yml` and `rs.yml`.

## Evidence

All run locally on the pinned 1.97.1 toolchain.

**The five checks under `--root rs`:**

```
$ scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
                                  --manifest rs/crates/f2z-codec/Cargo.toml

$ scripts/check-rust-fmt.sh --root rs
All 2 Rust crate(s) under rs/ are formatted (rustfmt from 1.97.1).

$ scripts/check-rust-clippy.sh --root rs
All 2 Rust crate(s) under rs/ are clippy-clean (clippy from 1.97.1).

$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml
All 2 Rust crate(s) under rs/ satisfy rs/deny.toml.

$ cd rs && cargo test --locked --all-targets && cargo test --locked --doc
test result: ok. 49 passed   (unit)
test result: ok.  7 passed   (tests/padding.rs)
test result: ok.  5 passed   (tests/redaction.rs)
test result: ok.  7 passed   (tests/reencode_equality.rs)
test result: ok.  1 passed   (doc)
```

Each script's own `--self-test` also passes (`check-rust-fmt.sh --self-test`, `check-rust-clippy.sh --self-test macos` locally / `linux` in CI, `check-rust-deny.sh --self-test`), and `rs.yml` runs each self-test immediately before the verdict it guards.

**WASM:**

```
$ cd rs && cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec
    Finished `dev` profile [unoptimized + debuginfo] target(s)
```

**`check-workflow-gates.mjs`, against the tree as it exists:**

```
Every gate inspects every job it awaits: 2 gate(s) across 14 workflow file(s).

self-test: an explicitly wired gate passes.
self-test: an exhaustively wired gate passes.
self-test: a job awaited but never bound is detected.
self-test: a job bound but never reported is detected.
self-test: a name reported that is not awaited is detected.
self-test: a result bound for a job that is not awaited is detected.
self-test: a needs object handed to nothing that reads it is detected.
self-test: a gate that awaits only the change detector is detected.
self-test: a gate that does not await the change detector is detected.
self-test: a gate whose needs is a block sequence passes.
check-workflow-gates self-test: 10 case(s) passed.
```

Seven of those ten are negative controls: each is a one-line mutation of a fixture that passes, so the detection is demonstrably the thing detecting.

**The change detectors are disjoint.** Both `case` patterns were extracted from the workflow files themselves — not retyped — and run over synthetic file lists:

```
== an rs/-only diff ==
   rs/Cargo.toml, rs/Cargo.lock, rs/rust-toolchain.toml, rs/deny.toml, rs/README.md,
   rs/crates/f2z-codec/Cargo.toml, rs/crates/f2z-codec/src/lib.rs,
   rs/crates/f2z-codec/src/frame.rs, rs/crates/f2z-codec/tests/reencode_equality.rs,
   .github/workflows/rs.yml
  zuuli.yml detector -> zuuli=false      <-- no 45-minute wallet build
  rs.yml    detector -> rs=true

== a wallet/-only diff ==
   wallet/zuuli/src-tauri/src/main.rs, wallet/zuuli/src-tauri/Cargo.toml,
   wallet/zuuli/src-tauri/Cargo.lock, wallet/plugins/tauri-plugin-zcash/src/lib.rs,
   wallet/zuuallet/src/pages/Settings.tsx, wallet/deny.toml, .github/workflows/zuuli.yml
  zuuli.yml detector -> zuuli=true
  rs.yml    detector -> rs=false         <-- no rs jobs on wallet PRs

== a shared-script diff (must select BOTH) ==
   scripts/check-rust-clippy.sh
  zuuli.yml detector -> zuuli=true
  rs.yml    detector -> rs=true
```

`node scripts/check-github-actions-pins.mjs` and `scripts/check-public-repo-boundary.sh` also pass.

## :warning: A repo-settings ask that cannot be made in a file

**`rs / gate` must be added as a required status check in branch protection on `main`.** Until it is, this workflow runs and reports but nothing blocks a merge on it. Everything else in this PR is enforced by files; this one line of repository settings is not.

## A real finding: the issue's premise about `zuuli.yml`'s gate is out of date

#555 says *"`zuuli.yml:469` already carries the comment explaining why"* a gate must inspect every job it awaits, and specifies the new script as asserting that the `needs:` set equals both the `needs.<job>.result` env bindings **and** the names in a `results=` line.

**`zuuli.yml` has no `results=` line and no per-job `needs.<job>.result` bindings today**, and line 469 is an unrelated `Swatinem/rust-cache` step. `zuuli.yml`'s gate now does this instead (lines 748+):

```yaml
      - name: Verify required jobs succeeded or legitimately skipped
        env:
          POLICY_OUTCOME: ${{ steps.policy.outcome }}
          REQUIRED_JOBS_JSON: ${{ toJSON(needs) }}
        run: |
          node scripts/check-github-actions-pins.mjs --verify-gate-results
```

I read `verifyGateResults` in `scripts/check-github-actions-pins.mjs`: it iterates `Object.entries(needs)` and throws on any entry whose `result` is not the expected one. That is **strictly stronger** than an inline `results=` string — there is no per-job wiring to drift, because there is no per-job wiring.

So I did not change `zuuli.yml`'s gate semantics to make my script pass. The script accepts both idioms and names them:

- **Explicit** — `needs.<job>.result` bindings plus a `results=` line, sets must be equal. This is what `rs.yml` uses, because `verifyGateResults` hardcodes zuuli's selector outputs (`changes.outputs.zuuli`, `zuuallet_schema`) and cannot judge an `rs` gate.
- **Exhaustive** — `toJSON(needs)` handed to a verifier from a short, deliberately short allowlist (`EXHAUSTIVE_VERIFIERS`, currently just `--verify-gate-results`). This is what `zuuli.yml` uses.

The honest limit is written into the script's header: for an exhaustive gate it proves the whole `needs` object reaches a recognized verifier; it does not read that verifier and prove it loops. Binding `toJSON(needs)` and then running something *not* on the allowlist is detected and is one of the negative controls.

## `WIRE.md` readings I made where it is ambiguous — flagged, not silently chosen

1. **`command`, `status` and `event` are raw `u16` on the wire, not enums.** §3.5 requires a relay to answer an unknown command code with a **non-fatal** `ERR_UNKNOWN_COMMAND`; decoding it as an enum would make it a fatal `ERR_MALFORMED` and would break re-encode equality for a frame that is, at the framing layer, perfectly well-formed. `Command::from_code` returns `Option`. The same applies to error codes (§10 keeps retired codes readable) and to push events (an unknown event is ignored, not fatal). By contrast `FrameKind` and `CommandAuth.present` **do** fail to decode on an unknown value, because they select a body — §3.3 names "an unknown-variant byte silently mapped to a default" as exactly the bug it exists to kill.

2. **`PowStamp stamp; /* empty when queue_creation_mode = open */` (§6.2).** `PowStamp` is fixed-width with no variable-length field, and §3.3 forbids omitting bytes, so "empty" can only mean a zero-valued stamp. Implemented as `PowStamp::empty()` (all-zero challenge, all-zero salt, counter 0).

3. **`Response.body` "MUST be empty when status != 0" is a validity rule, not an encoding rule.** The decoder does not enforce it; `Response::validate()` does. Enforcing it in the decoder would make a non-conforming peer response unrepresentable rather than rejectable, and it needs to travel the ordinary `ERR_MALFORMED` path.

4. **`request_id` nonzero on requests/responses, zero on pushes (§4.3) is likewise `RelayFrame::validate()`, not decoding.** A relay answers a bad `request_id` with `ERR_MALFORMED` — which it can only do if it decoded far enough to know which `request_id` to answer on.

5. **`opaque scope<0..255>` in `ChallengeRequest` (§6.1) carries a `contact_addr` for `contact_append`, but operator fields in §11.1 use the same `<0..255>` shape and exist to be read.** Rather than two types, `ShortBytes`'s `Debug` decides on content: printable ASCII renders (escaped, so a newline in an operator name cannot forge a log line); anything else redacts. Noted because it is a judgement, not a spec statement.

6. **`RelayFrame`'s `select` is hand-encoded.** `kind` *is* the discriminant; a derived `tls_codec` enum would put the tag on the wire twice.

7. **`WIRE.md` writes `T x<0..2^k-1>` for vectors of non-byte elements** (`ReadResponse.messages`, `Capabilities.protocol_versions`, `Capabilities.padding_sizes`). `tls_codec` ships `TlsVecU8/U16/U24` for this, but only against the `std` `Serialize`/`Deserialize` traits — a `no_std` crate uses the `SerializeBytes`/`DeserializeBytes` halves, which `tls_codec` implements only for its *byte* vectors. `rs/crates/f2z-codec/src/vec.rs` fills that gap with the identical encoding, and tests both §3.3-relevant cases: a declared length longer than the contents is an error, and bytes stranded inside the declared length after the last element are an error.

## Two things found while building this

**`tls_codec`'s byte vectors leak through `Debug`.** `TlsByteVecU24`'s derived `Debug` prints `TlsByteVecU24 { vec: [222, 222, 222, …] }` — a complete dump containing no hex at all, so a hex-only redaction test would have passed over it. The frame bodies are therefore a `Body` newtype rather than a bare `tls_codec` vector, and `tests/redaction.rs` now also checks decimal byte lists. Worth knowing before any other crate in this tree holds a `tls_codec` vector.

**`rs/deny.toml`'s allow list is exactly the licences the graph carries** (MIT, Apache-2.0, BSD-3-Clause, Unicode-3.0) and nothing more. cargo-deny reports every unmatched allowance as a warning, and a policy that emits warnings on a clean run trains people to ignore its output. Adding a permissive licence when a dependency needs one is a one-line reviewed edit — the intended cost.

## Workspace lints

`rs/Cargo.toml` denies `unwrap_used`, `expect_used`, `panic`, `indexing_slicing` and `arithmetic_side_effects` for every member. The relay is a public, unauthenticated network listener and a panic in its parser is a remote denial of service. Integration tests relax them with a note saying why.


skyl:
Reviewed adversarially on the owner's behalf — this comment covers **supply chain, toolchain registration, workspace layout, and whether `Closes #555` is earned**. Two more reviews (the `f2z-codec` crate vs `WIRE.md`, and the gate/CI wiring) are in flight and will follow.

**Requesting changes: two blockers.** The evidence blocks in the body are, unusually for this repo, entirely honest — I re-ran every one of them against this exact tree and they reproduce verbatim, numeric test counts included. Both defects are claims about things *outside* those blocks.

---

## BLOCKING 1 — `rs/deny.toml` does not carry ADR 0013's `akd` floor, so `Closes #555` must not stand

`rs/deny.toml:105-107` is the entire `[bans]` section:

```toml
[bans]
multiple-versions = "warn"
wildcards = "deny"
```

No `[[bans.deny]]` anywhere. The string `akd` does not appear in this PR at all:

```
$ grep -rn "akd" rs/ scripts/ .github/ ; echo "(end)"
(end)
```

**I ran the mutation test #555's amended criteria ask for.** Scratch member with `akd = "0.12.0"`, resolved into `rs/Cargo.lock` (`akd v0.12.0`, `akd_core v0.12.0`, both crates.io):

```
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml advisories
advisories ok
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml bans
bans ok
```

**Both green.** Not the asymmetry — no control at all. A stale `0.12.0` pin passes the full `rs` supply-chain gate today, exactly as ADR 0013 predicted when nothing enforces the floor. That matters because I verified separately that OSV returns `{}` for `akd` at every version, `rustsec/advisory-db` has no `crates/akd` entry, and `/repos/facebook/akd/security-advisories` returns `[]` — `cargo audit` and `cargo deny check advisories` are **genuinely silent** on the auditor append-only bypass (`facebook/akd#495`). The `[[bans.deny]]` entry is the only belt.

Pasting ADR 0013's entry (`decisions/0013-key-transparency-log.md:96-116`) into `rs/deny.toml` and re-running:

```
```

…and with `akd = "0.13"`: `bans ok`. So the fix is a verified ~15-line paste, it is not a blanket ban, and the advisories-green/bans-red asymmetry is now demonstrated on real output rather than asserted. ADR 0013 bans `akd_mysql` outright too — include that third entry.

**Why it happened:** the branch is 1 commit behind. `git merge-base HEAD origin/main` = `2486a34`; `origin/main` = `ece2cc0` (#557), which added ADR 0013 and `KT.md`. Neither exists in this tree. That explains it but doesn't repair it — `Closes #555` would close the issue whose amended criteria exist *specifically* to create the one control `KT.md:1180-1184` says has no second line of defence, leaving that document making a present-tense claim about an entry nobody wrote and no open issue tracking it.

**Either** add the three entries here (preferred — `rs/deny.toml` is being created in this PR; there will never be a cheaper moment; rebase onto `ece2cc0` first so the inline comment can point at a file that exists), **or** narrow to `Refs #555` and leave #555 open. Please paste the mutation evidence above into the body either way, so the next reviewer can't delete the entry as redundant-with-`cargo audit` — which is precisely the mistake it exists to survive.

---

## BLOCKING 2 — `rs.yml`'s gate publishes the status context `gate`, colliding with the only required check on `main`

I'd already spotted this independently before the review started; the reviewer confirmed it with live evidence.

- `.github/workflows/rs.yml:301` — `gate:` with **no `name:`**. Every other job in the file has one (`rs / rustfmt`, `rs / clippy`, `rs / cargo-deny`, `rs / tests`, `rs / wasm32`). So its context is the job id: `gate`.
- `.github/workflows/zuuli.yml:758` — `gate:`, also no `name:`. Same context.
- `gh api repos/free2z/zuu/branches/main/protection/required_status_checks` → `{"contexts":["gate"],"checks":[{"context":"gate","app_id":15368}]}`.

**Live on this PR right now:**

```
gate                    pass     10s   .../runs/32677847000/job/97289250868
Rust / ZUULI backend    pending  0     .../runs/32677847023/...
changes                 pass     17s   .../runs/32677847000/...
changes                 pass     20s   .../runs/32677847023/...
```

Run `32677847000` is **rs.yml** — `gh api repos/free2z/zuu/actions/runs/32677847000 --jq .path` returns `.github/workflows/rs.yml`. Run `…023` is `wallet/zuuli`. **The required context `gate` is currently reading green off rs.yml's 10-second gate while zuuli.yml's full-stack gate hasn't reported.** `changes` collides identically. Which workflow answers the required check is decided by report order, not by policy.

The body asks for "`rs / gate`" to be added as a required check — **there is no such context**. Fix: `name: rs / gate` at `rs.yml:301` and `name: rs / changes` at `:38`. Then the repo-settings ask becomes coherent, non-colliding, and I can action it.

I'm deliberately not asserting *which* run GitHub resolves the requirement against when two check runs share the name — the third reviewer is establishing that empirically. It doesn't change the fix: two workflows must not compete for the one name protecting `main`.

---

## BLOCKING 3 (one line) — `rs/README.md:117` states the thing #553 exists to falsify

> "All four discover crates rather than listing them, so a new crate under `rs/` is gated from its first commit."

Three of the four discover. Mutation-proved by adding `rs/crates/scratch/Cargo.toml` with no `rust-version` and no `license`:

| Script (unchanged flags) | Sees it? | Result |
|---|---|---|
| `check-rust-fmt.sh --root rs` | yes | RED — `3 crate(s) are not formatted` |
| `check-rust-clippy.sh --root rs` | yes | RED — `3 crate(s) failed the clippy gate` |
| `check-rust-deny.sh --root rs --config rs/deny.toml` | yes | RED — `licenses FAILED` |
| `check-rust-toolchain.sh --toolchain-file … --manifest …f2z-codec/Cargo.toml` | **no** | **GREEN** |

`check-rust-toolchain.sh:381` greps `rust-version` out of a hardcoded `MANIFESTS` array plus explicit `--manifest` flags. A second crate under `rs/crates/` ships with a wrong MSRV, or none, and passes. That's #553, not this PR's bug — but this is the first tree where it bites, and the sentence reads as a solved problem. Please point it at #553 instead.

---

## `Closes #555`, criterion by criterion

| # | Criterion | Verdict |
|---|---|---|
| A1 | `rs/` passes fmt/clippy/deny/test via `--root rs` | **MET** — all four reproduce verbatim; 49 / 7 / 5 / 7 / 1 doc exactly as claimed |
| A2 | `f2z-codec` compiles for `wasm32-unknown-unknown` | **MET** |
| A3 | `check-workflow-gates.mjs` passes; self-test fails on injected mismatch | **MET** — `2 gate(s) across 14 workflow file(s)`, `10 case(s) passed`, 7 named negative controls |
| A4 | `rs/**`-only → `zuuli=false`; `wallet/**`-only → `rs=false` | **MET** — verified independently by extracting the `case` patterns from `zuuli.yml:128` / `rs.yml:109` and running them over synthetic file lists |
| A5 | Gate called out as a repo-settings ask | **PARTIAL** — names a context that doesn't exist (BLOCKING 2) |
| §5 | `rs/deny.toml` starts with an empty `ignore` | **MET** — `:46` `ignore = []` |
| §5 | Licence assertion enforcing the permissive/AGPL split | **MET, mutation-verified** — `:72` `private = { ignore = false }` is load-bearing: a member with no `license` field → `licenses FAILED`; adding `license = "MIT"` clears it. `exceptions = []` at `:95` means an AGPL binary fails until named. |
| §1 | Toolchain files registered as restatements | **MET** |
| §1 | README states it does not decide #341 | **PARTIAL** — #341 is closed; see below |
| **amended AC** | `akd` floor in `rs/deny.toml`, mutation-tested, with an ADR 0013 / `akd#495` comment | **NOT MET** — BLOCKING 1 |

---

## What checks out, and is worth saying plainly

**`rs/deny.toml` is stricter than `wallet/deny.toml` in every single dimension and more permissive in none:**

| | `wallet/` | `rs/` |
|---|---|---|
| `advisories.ignore` | 22 entries (#351 backlog) | `[]` |
| `licenses.private.ignore` | `true` | **`false`** |
| `licenses.allow` | 14 licences | 4 |
| `licenses.exceptions` | `aws-lc-sys`/OpenSSL | `[]` |
| `bans.multiple-versions` | `allow` | `warn` |
| `bans.wildcards` | unset (= allow) | `deny` |

No drift toward permissive, and the header at `:11-25` explicitly forbids copying wallet entries across. That's the right instinct for a new tree. The deny run is green with **zero** warnings — no `license-not-encountered` — so the claim that the allow list is exactly what the graph carries holds.

**The dependency surface is clean.** `rs/Cargo.lock` holds 52 packages; the *library* graph is 15, and `cargo tree -e build` is empty — no build scripts. `blake2` → `digest`, and `tls_codec` → `tls_codec_derive`/`zeroize`, with the `syn`/`quote`/`proc-macro2` chain host-only and absent from the wasm artifact. No `serde`, no `tokio`, no network crates, nothing forcing `std`. The remaining ~37 entries are the `proptest` dev tail, correctly excluded from `--lib`.

**The toolchain restatement discipline is real and the stated reason holds.** `rust-version.workspace = true` does *not* match `check-rust-toolchain.sh:381`'s regex (`.workspace` sits between the key and the `=`), so inheriting would hit `:383` `fail "$rel does not declare rust-version"`. Verified. The `license` half is different — cargo-deny reads resolved metadata, so `license.workspace = true` would pass — and the manifest is honest that its reason there is readability, not mechanism.

---

## Follow-ups (not blocking)

1. **`rs/README.md:22-41` argues against a premise that's no longer true.** #341 is **CLOSED / COMPLETED** as of `2026-08-23T09:18:55Z` via PR #519, which landed `docs/architecture/CARGO-WORKSPACE.md` — present in this PR's own tree, header `Status: accepted on 2026-08-23. This records the decision for issue #341.` The README discusses it as open ("**asks whether**", "**If #341 later chooses**") and reconstructs the counter-arguments from the issue body rather than the accepted decision.

   The substance survives and gets *stronger*: the ADR says *"Do not create a repository-root workspace… If a wallet-wide workspace is ever adopted, its candidate root is `wallet/`."* The decided consolidation target was never repo-wide, so a fourth root at `rs/` doesn't make it harder — the two trees would never have merged under the accepted decision. And `rs/` satisfies the ADR's own "Adding Rust crates" step 5. Cite the ADR, mark #341 decided, and drop the `:40-41` hedge (which also understates the cost: folding two workspaces means one reviewed resolution diff across both lockfiles, not deleting a table).

2. **`rs/deny.toml:27-31` — the comment describes a directive that isn't there.** It explains evaluating every target, but annotates `all-features = false`, which is a *feature* knob. The target knob is `[graph] targets`, which is absent — and absent means all targets, so the described behaviour comes from cargo-deny's default, not the line beneath the comment. Both settings are defaults, making the whole block a no-op that documents itself as doing something. Set `targets = []` explicitly with that comment, or move it. This repo polices "a comment claiming a control" everywhere else.

3. On **#553**: `wallet/zuuli/crypto-target-spike/Cargo.toml:5` declares `rust-version = "1.97"` and nothing checks it — a second known unregistered manifest. Worth adding there, with the mutation result above as evidence that the gap is now reachable in two trees rather than one. Full census on this head: `rs/Cargo.toml` (virtual root, correctly unregistered), `f2z-codec` ✔, the four `MANIFESTS` entries ✔, `crypto-target-spike` ✗.

---

**Everything else in the evidence blocks re-ran verbatim**, including `check-rust-deny.sh --self-test` (`4 case(s) passed; cargo-deny 0.20.2`), `check-github-actions-pins.mjs` (`178 external reference(s), 15 file(s)`), `check-public-repo-boundary.sh` (silent, exit 0), and the three-scenario change-detector matrix. The one numeric slip is cosmetic: "line 469 is an unrelated `Swatinem/rust-cache` step" is off by three — on `origin/main`, `zuuli.yml:466` is the `uses:` and `:469` is its `key: zuuli-plugin` — still inside the step it names. Not a finding.

skyl:
Second review — **`rs/crates/f2z-codec` against the frozen `WIRE.md`**, and whether its tests are behavioural. (Supply chain / `Closes #555` is my earlier comment; the gate-wiring review follows.) Run on 1.97.1 in a clean worktree at `f7a17e9`.

**The encoding is correct — I checked all 33 spec structures field-for-field and found zero divergences. But the test suite does not test the encoding.** 11 of 20 real mutations to the wire format survive all 69 tests green, including a vector length-prefix width change. Plus one real ciphertext leak.

---

## BLOCKING 1 — `Canonicalized<T>`'s derived `Debug` dumps the received frame as a decimal byte list

`rs/crates/f2z-codec/src/canonical.rs:41-45`:

```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct Canonicalized<T> {
    value: T,
    encoded: Vec<u8>,   // <- derived Debug renders this
}
```

`Canonicalized` is the type **every receive path holds** — it wraps the whole decoded frame *and* its bytes. A probe test:

```
FIRST 200 CHARS: Canonicalized { value: AppendRequest { payload: Payload(<redacted; 1024 bytes>) },
                 encoded: [0, 4, 0, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, ...
LEN: 5223
panicked: Canonicalized Debug dumped the ciphertext bytes
```

A 1 KiB `APPEND` payload renders as 5,223 characters of ciphertext. This is *precisely* the hazard `tests/redaction.rs:14-17` documents and credits itself with having closed:

> `tls_codec`'s own byte vectors derive `Debug` and print `TlsByteVecU24 { vec: [222, 222, 222, …] }` — a complete dump that contains no hex at all. That is why every body and payload in this crate is a newtype rather than a bare `tls_codec` vector.

The `Body`/`Payload` newtypes fix it on the inside, and `Canonicalized` reintroduces it one level up. `tests/redaction.rs` never formats a `Canonicalized`, so all 5 redaction tests pass. This defeats stated crate property #2 (`lib.rs:17-19` — "`--log-level trace` must not become a ciphertext archive") for the most-held type in the crate.

Fix: hand-write `Debug` for `Canonicalized` (`encoded: <redacted; N bytes>`) and add it to `tests/redaction.rs`'s case list.

## BLOCKING 2 — nothing in the repository asserts what any structure's bytes actually are

All 7 tests in `tests/reencode_equality.rs` compare the code's output to the code's own output: encode→decode→assert-equal (`:105`), decode→encode fixpoint (`:116`, `:131`), mutated-input non-equality (`:146`), and `:171`, where `sender_tag` and `receiver_tag` both come from `hash(LABEL_BODY, builder.signing_bytes_for_auth(...))` **on the same builder** — both directions share every bug.

That is the correct test for §3.3, which is a decoder-slack property rather than a format property, and I have no complaint about it as one. The problem is that it's the *only* integration test of the encoding. The sole hardcoded-byte assertions in the crate are `src/vec.rs:178,191,194` (prefix encodings), `src/frame.rs:546,550,557` (`CommandAuth` present byte, `SignedAuth` total length), and `src/transcript.rs:339-341` (`TRANSCRIPT_LEN` and the label prefix). Three structures out of thirty-three.

### Mutation results — 20 real mutations, 11 not caught

Each applied to the source, then `cargo test --locked --all-targets && --doc` (baseline 49+7+5+7+1 = 69 green), then reverted.

| # | Mutation | Site | Result |
|---|---|---|---|
| M1 | `ReadResponse.messages`: `VecU24` → `VecU16` — **the length-prefix interop break** | `commands.rs:438` | **NOT CAUGHT** |
| M2 | `CreateQueueResponse`: swap `recv_addr` / `send_addr` | `commands.rs:396` | **NOT CAUGHT** |
| M3 | `HelloResponse`: swap `channel_binding_mode` / `transport_security` | `commands.rs:245` | **NOT CAUGHT** |
| M4 | `LABEL_BODY` `…/v1/body` → `…/v1/bodx` | `hash.rs:25` | **NOT CAUGHT** |
| M7 | `CommandTranscript`: swap `address` / `signer_key` | `transcript.rs:71` | **NOT CAUGHT** |
| M8 | `Capabilities.protocol_versions`: `VecU8` → `VecU16` | `commands.rs:590` | **NOT CAUGHT** |
| M10 | `SignedAuth`: swap `address` / `signer_key` | `frame.rs:424` | **NOT CAUGHT** |
| M13 | `LABEL_RELAY_ID` one byte | `hash.rs:29` | **NOT CAUGHT** |
| M17 | `PowStamp` digest: `counter.to_be_bytes()` → `to_le_bytes()` | `pow.rs:145` | **NOT CAUGHT** |
| M19 | `ChallengeResponse`: swap `challenge` / `expires_at_ms` | `commands.rs:329` | **NOT CAUGHT** |
| M21 | `H(label, x)` → `H(x, label)` in **both** `hash` and `hash2` | `hash.rs:62` | **NOT CAUGHT** |
| M5 | `MAX_PLAUSIBLE_BUCKETS` 16 → 17 | `padding.rs:44` | caught |
| M6 | fixed-newtype `Debug` prints inner bytes | `types.rs:97` | caught |
| M9 | `ErrorCode::Quota` 14 → 24 | `error.rs:114` | caught |
| M11 | `Command::Read` `0x0013` → `0x0016` | `commands.rs:113` | caught |
| M12 | `PROPOSED_BUCKETS` `65_536` → `32_768` | `padding.rs:37` | caught |
| M14 | `ShortBytes` prefix `U8` → `U16` | `types.rs:337` | caught (incidentally — via `TRANSCRIPT_LEN`) |
| M15 | `FrameKind::Push` 3 → 4 | `frame.rs:62` | caught |
| M16 | `RelayFrame.request_id` BE → LE | `frame.rs:367` | caught |

Three clusters, each a distinct missing test:

1. **Field order (M2, M3, M7, M10, M19, M23).** The `TlsSerializeBytes`/`TlsDeserializeBytes` derive pair moves together, so a swap is invisible to any round-trip. M7 and M10 are the serious ones: swapping `address` and `signer_key` in `CommandTranscript` and `SignedAuth` **changes what every Ed25519 signature in the protocol commits to**, and 69 tests stay green.
2. **Vector prefix width (M1, M8).** M1 is the exact defect I asked the review to hunt: `WIRE.md:756` declares `QueuedMessage messages<0..2^24-1>`, the code correctly writes `VecU24`, and changing it to `VecU16` — silently capping a `READ` response at 64 KiB and desynchronising every other implementation — passes everything. M14 was caught only because `ShortBytes` happens to sit inside `TRANSCRIPT_LEN`; the eight `ShortBytes` operator fields of `Capabilities` have no byte pinning at all.
3. **Hash construction and labels (M4, M13, M21).** There is not one known-answer vector for `H` anywhere. `hash.rs:153` only asserts two labels differ; `:159` is self-referential. M21 is the alarming one — turning `H(label, x)` into `H(x, label)` inverts the construction from prefix to suffix, which makes `labels_are_prefix_free` (`hash.rs:132`) assert **the wrong invariant**, and nothing notices. M17 is the same shape: the PoW digest is the one number a relay and client must agree on byte-for-byte, and its endianness is untested.

**The fix is cheap:** one `tests/wire_vectors.rs` with a hardcoded `&[u8]` per structure (encode-side assertion, not round-trip) plus five hardcoded BLAKE2b digests for the `LABEL_*` constants. That single file kills all 11 surviving mutants — and it's also the artefact a third-party implementer needs, and the thing `WIRE.md` currently cannot be independently checked against.

---

## What I verified positively

**Encoding, field-for-field, all 33 structures — zero divergences.** Framing (`WIRE.md:316-343` → `frame.rs`), including the hand-encoded `select` at `frame.rs:363-405` writing the tag exactly once. `CommandAuth`/`SignedAuth`/`CommandTranscript` (`:414-444`) with all 11 transcript fields in order. §6.1-§6.4 (`:643-870` → `commands.rs:222-527`). §11.1 `Capabilities` (`:1275-1333`) — **all 38 fields**, in order, correct widths. §12.2 contact queues. §13.1 `PowParams`/`PowStamp`.

**Vector bounds specifically** — `WIRE.md` uses TLS byte-counting semantics and `src/vec.rs` implements it correctly: the prefix counts **bytes**, not elements, proven at `vec.rs:178` (`vec![1u16,2,3]` → `[6, 0,1, 0,2, 0,3]`). `protocol_versions<1..255>`→`VecU8`, `padding_sizes<1..2^16-1>`→`VecU16`, `messages<0..2^24-1>`→`VecU24`, `opaque body/payload<0..2^24-1>`→`TlsByteVecU24`, `opaque scope/operator_*<0..255>`→`TlsByteVecU8`. All ✓. `vec.rs:127-143` errors on both a declared length exceeding contents and bytes stranded after the last element decodes — the two §3.3 cases by hand — and refuses zero-width elements rather than looping.

**Domain separation.** The six labels are the exact ASCII from `WIRE.md:437, 476, 447, 683, 496, 1638`, the set is genuinely prefix-free, and `hash.rs:132` asserts it **mechanically over `LABELS`** rather than by inspection, so a seventh label can't be added without the check running. No label not in `WIRE.md`. That's the right shape for an unstated invariant — my only complaint is M21: the invariant isn't tied to the construction it protects.

**Padding validator — behavioural.** `PaddingBuckets::new` (`:71-91`) enforces strictly-ascending, non-empty, no zero, and the `len > u16::MAX/4` overflow bound (correct: 4 bytes per `uint32`, 16,383 max). `accepts` is exact set membership via `binary_search`. `tests/padding.rs:81` cross-checks acceptance against the constant array over `0..200_000`, and `padding.rs:217` hits ±1 around every bucket. M12 confirms the set is pinned.

**Redacting newtypes — behavioural.** All nine fixed-width newtypes have a hand-written `Debug` (`types.rs:95-99`), none derived. `Payload`/`Body` report length only. `ShortBytes` decides on content and escapes (`:384-397`). No `Display` impls on any of them, so there's no second formatter to leak through. `tests/redaction.rs` checks lowercase hex, uppercase hex, base64url, decimal byte lists, a 4-byte decimal prefix, and any 8+ char hex-looking run — genuinely paranoid, and M6 confirms it fires. B1 is a coverage gap, not a weakness in these tests.

**Build properties — verified, not trusted.** `#![no_std]` + `#![forbid(unsafe_code)]` at `lib.rs:72-73`. The only `std::` in the source is a comment at `error.rs:202`. Two runtime deps, both `default-features = false`: `blake2` and `tls_codec`. `cargo tree -e build` empty. `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec` → `Finished` in 16.41s. No I/O, no tokio, no clocks, no randomness. Test counts match the body exactly.

---

## Follow-up (non-blocking)

1. **`<1..N>` minimum bounds are not enforced.** `WIRE.md:1276` declares `protocol_versions<1..255>` and `:1301` `padding_sizes<1..2^16-1>`, both minimum 1. `VecU8`/`VecU16` accept empty; an empty `protocol_versions` encodes as `0x00` and round-trips clean. `PaddingBuckets::new` rejects empty, but `Capabilities.padding_sizes` is a raw `VecU16<u32>` that never passes through it.
2. **`MAX_MESSAGE_TTL_SECONDS` is dead.** `lib.rs:117` declares it, `commands.rs:623` mentions it in a doc comment, nothing calls it, and there is no `Capabilities::validate()`. §11.3 step 5's "MUST refuse if `max_message_ttl_seconds > 2592000`" has no code and no test.
3. **The showcase doc example builds a spec-invalid frame.** `lib.rs:44-48` constructs `Command::Append` with `CommandAuth::Unsigned`; `WIRE.md:626` requires `APPEND` to be signed by the send key. Fine as a framing illustration, but it's the crate's front door.
4. `tests/redaction.rs:313` — `assert!(!format!("{hostile:?}").contains('\n'))` runs against `b"ok"`, which has no newline to begin with. The escaping claim above it isn't exercised.

## Spec ambiguities the crate silently resolved

None wrong; all now normative-by-implementation, so they belong on the open issues.

**#550:** `PowParams.algorithm == 0` now means "no PoW required" (`pow.rs:67-79`) — `WIRE.md:1621` names only `1 = blake2b-leading-zero-bits` and leaves `0` undefined; the crate infers it from §6.1's `/* zeroed when no PoW is required */` at `:697`. Reasonable, but it permanently reserves algorithm 0, and that's a new statement. `PowStamp::empty()` = all-zero likewise resolves `:733` — I agree with the fixed-width-struct reasoning, but it should land in `WIRE.md` rather than a Rust doc comment. The crate correctly does **not** pick a side on `ERR_NO_ACCESS` vs `ERR_UNAVAILABLE` (`error.rs` defines both with the spec's own text and has no per-command selection logic), and has no handle code at all, so it carries no exposure to the U+212A problem.

**#552:** the crate is **direct evidence that §11.1's "entire externally-relevant policy" claim is false** — `Capabilities` compiles exactly as specified and therefore has no field for `send_retry_max` (`WIRE.md:839`), any Layer-1 connection limit, the per-queue append rate (`:1601`), a max queue count, or any Layer-4 high-water mark (`:1644`). A relay implementing this struct literally cannot publish them. And `hash::hello_proof_message` (`hash.rs:110-124`) is now the reference implementation of the `relay_proof` gap: it signs `"free2z/relay/v1/hello" || channel_binding || client_nonce` and nothing else, so under `transport_security: none` an active attacker can still rewrite `protocol_version`, `channel_binding_mode`, `transport_security` and `capabilities_digest` in the same `HelloResponse` and the proof verifies. Faithful to the spec, weakness and all.

**New, on neither issue:** `PaddingBuckets::new` is **stricter than §11.1** — it rejects `0` and duplicates (`padding.rs:82-88`) where the spec says only "ascending", so a relay publishing `[0, 1024]` is spec-legal and this client refuses it. The strictness is right; the spec should say so. And `has_more` / `contact_queues_enabled` are read as nonzero-is-true (`commands.rs:440, 675`), which `WIRE.md` never specifies — making `0x01` and `0x02` semantically identical yet distinct canonical byte strings, a mild tension with §3.3's spirit if not its letter.

---

**Bottom line: the codec is right, and I couldn't find a test that would have told me if it weren't.** Fix B1, add the byte-vector file, and this is solid work.

Issue #568, opened 2026-08-24
ci: the ZUULI gate does not re-run `check-workflow-gates.mjs`, unlike the rs gate


skyl:
Raised in the #562 review and deferred out of that PR.

`rs.yml`'s gate re-runs both policy scripts inside the gate job itself, so a `continue-on-error` or a tampered `changes` job cannot launder a policy failure into a `success` conclusion:

```yaml
      - name: Recheck the workflow policy
        id: policy
        run: |
          node scripts/check-github-actions-pins.mjs --self-test
          node scripts/check-github-actions-pins.mjs
          node scripts/check-workflow-gates.mjs --self-test
          node scripts/check-workflow-gates.mjs
```

`zuuli.yml`'s gate re-runs only the pins script:

```yaml
      - name: Recheck immutable actions and fail-closed required jobs
        id: policy
        run: |
          node scripts/check-github-actions-pins.mjs --self-test
          node scripts/check-github-actions-pins.mjs
```

Both workflows run `check-workflow-gates.mjs` in their `changes` job, so the check is not absent from the ZUULI suite — it just lacks the second belt inside the gate that the rs gate has. The wallet gate is the one branch protection actually requires, which makes it the one where the belt matters most.

## Why this was not just added

`scripts/check-github-actions-pins.mjs` asserts the ZUULI gate's policy step contains an exact, ordered, exactly-two-command list:

```js
    policyCommands.length !== 2 ||
    policyCommands[0] !== GATE_POLICY_SELF_TEST_COMMAND ||
    policyCommands[1] !== GATE_POLICY_COMMAND ||
```

Adding two commands to the step means editing that assertion, its constants, and whichever of its 99 current-workflow mutation cases depend on the shape. That is the exact-length assertion hazard that has #556 blocked, and it is a materially larger change than the rename #562 needed. Doing it in the same PR would have put a required-gate structural assertion into a PR about a job name.

## What to do

- Add `node scripts/check-workflow-gates.mjs --self-test` and `node scripts/check-workflow-gates.mjs` to the ZUULI gate's policy recheck step.
- Widen `requiredGateControlFailures` in `check-github-actions-pins.mjs` to assert the new exact four-command list rather than loosening it to "contains".
- Add a mutation case proving the widened assertion fails when a command is dropped or reordered — a longer exact list is only worth having if dropping an entry is caught.

## Acceptance

- [ ] The ZUULI gate re-runs `check-workflow-gates.mjs` with its self-test first.
[...]

Issue #567, opened 2026-08-24
ci: two more duplicated status-check contexts — `build` (3 workflows) and `frontend` (2)


skyl:
Found while fixing #562. The `gate`/`changes` collision that issue describes was not the only one — a full scan of every job's published check-run name (`name:` when declared, job id otherwise) across all 14 workflow files turns up **four** duplicated contexts, not two.

Run against `origin/main` at 241b714, with the tolerated list emptied:

```console
$ node scripts/check-workflow-gates.mjs   # with TOLERATED_CONTEXT_COLLISIONS emptied
jobs scanned: 56 across 14 workflow file(s)
- .github/workflows/docs-about-free2z.yml, .github/workflows/ts-react-free2z.yml, .github/workflows/ts-svelte-free2z.yml: 3 job(s) publish the status-check context "build" (...#build, ...#build, ...#build)
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "changes" (...)
- .github/workflows/zuuallet.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "frontend" (...)
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "gate" (...)
```

`gate` and `changes` are fixed by the #562 PR. `build` and `frontend` are left as-is there, deliberately, and recorded in `TOLERATED_CONTEXT_COLLISIONS` in `scripts/check-workflow-gates.mjs`.

## Why they were tolerated rather than fixed in that PR

Branch protection on `main` requires exactly one context:

```console
$ gh api repos/free2z/zuu/branches/main/protection --jq .required_status_checks
{"checks":[{"app_id":15368,"context":"gate"}],"contexts":["gate"],"strict":true}
```

Neither `build` nor `frontend` is required, so neither can currently decide a merge — the ambiguity is cosmetic today, unlike `gate`, where it was load-bearing. Renaming five jobs across five workflows is also unrelated churn in a PR whose point is the required context.

But "not required *yet*" is the whole risk. All five colliding jobs run on `pull_request`, so they already produce duplicate rows in the PR checks list, and the day anyone adds `build` to branch protection they inherit #562's exact defect.

## What to do

- `docs-about-free2z.yml#build` → `docs / build` (or similar)
- `ts-react-free2z.yml#build` → `ts-react / build`
- `ts-svelte-free2z.yml#build` → `ts-svelte / build`
[...]

PR #569, opened 2026-08-24
ci: give rs.yml's gate a unique status-check context, and make the collision a control

### LINKED ISSUE #562: ci: two workflows now publish the required `gate` context — branch protection resolves by report order
## Problem

`main` now has **two workflows publishing a status check named `gate`**, and `gate` is the only required status check protecting the branch.

```
$ gh api repos/free2z/zuu/branches/main/protection/required_status_checks
{"contexts":["gate"],"checks":[{"context":"gate","app_id":15368}],"strict":true}
```

- `.github/workflows/zuuli.yml:758` — `gate:`, no `name:` → context `gate`
- `.github/workflows/rs.yml:301` — `gate:`, no `name:` → context `gate`

Every other job in `rs.yml` declares a display name (`rs / rustfmt`, `rs / clippy`, `rs / cargo-deny`, `rs / tests`, `rs / wasm32`). The gate does not, so it inherits the job id — the same job id `zuuli.yml`'s gate uses. `changes` collides identically in both files.

Landed in #560.

## This fires on every PR, not just `rs/` ones

`.github/workflows/rs.yml:17-22` has **no path filter at the workflow level**:

```yaml
on:
  pull_request:
  merge_group:
  push:
    branches: [main]
  workflow_dispatch:
```

and its gate is `if: always()` (`rs.yml:303`) so that "a non-rs skip reports success rather than remaining pending" (its own comment, `:294-295`). That is the right design for a path-filtered gate in isolation — but combined with the name collision it means **every pull request in this repository now produces two check runs named `gate`**, one of which reports success in about ten seconds regardless of what the PR touched.

## Observed live, before the merge

On PR #560's own head `f7a17e9`:

```
$ gh api repos/free2z/zuu/commits/f7a17e9.../check-runs --jq '.check_runs[] | select(.name=="gate")'
gate  completed  success  app_id 15368  → runs/32677847000/job/97289250868

$ gh api repos/free2z/zuu/actions/runs/32677847000 --jq .path
.github/workflows/rs.yml
```

At that moment `wallet/zuuli` (run `32677847023`, which owns the other `gate`) was still `in_progress`. **The required context was reading green off rs.yml's ten-second gate while the full-stack gate had not reported.** `changes` showed the same duplication — two entries, one from each run.

## Why this is a branch-protection risk, and what I have not established

GitHub resolves a required context to the **most recent** check run bearing that name. With two producers the resolution is decided by **report order**, which is a function of runner contention, not policy.

The dangerous ordering is concrete: `zuuli.yml`'s gate waits on macOS universal builds, Android AAB, iOS, and Windows/macOS custody jobs, so it lands late and can land red. `rs.yml`'s gate waits on five fast Linux jobs. If the `rs` jobs are ever queued behind runner contention and report **after** a red zuuli gate, the required check reads the green one.

I am deliberately not asserting that a red `gate` has already been overridden — I have not observed that, and I would rather name the mechanism precisely than overstate it. What is certain and sufficient on its own:

- Two independent workflows now answer to the single name that branch protection checks.
- Which one answers is not deterministic.
- The repo's whole CI doctrine (#512/#515/#516/#517/#518, `check-workflow-gates.mjs`) is built on `gate` being *the* fail-closed verdict. That property cannot survive two producers, whatever the resolution order turns out to be.

There is also a straightforward correctness problem independent of ordering: `rs.yml`'s gate has no knowledge of the wallet build and `zuuli.yml`'s has none of `rs/`, so neither is a verdict on the whole PR, yet each presents itself under the name that means exactly that.

## Fix

Two lines plus a settings change:

```diff
   changes:
+    name: rs / changes
     runs-on: ubuntu-latest
```
```diff
   gate:
+    name: rs / gate
     needs: [changes, rs_fmt, rs_clippy, rs_deny, rs_test, rs_wasm]
```

Then add `rs / gate` to branch protection as a second required context, so `main` requires `["gate", "rs / gate"]`. #560's own body asks for exactly this — it names `rs / gate` as the context to add — but no job produces that name, so the ask was unactionable as written.

## Make it a control rather than a discipline

`scripts/check-workflow-gates.mjs` (added in the same PR) exists to stop a gate becoming decorative, and it verifies that every job a gate awaits is bound into `env:` and named on the `results=` line. It did not catch this, because a duplicate context name across two files is outside what it inspects.

Add an assertion: **no two workflows may publish the same status-check context**, computed as `job.name ?? job_id` across every workflow file. That is a handful of lines given the script already parses all 14 workflows, it fails on today's tree until the rename lands, and it converts "remember not to reuse a job id" into something enforced. Mutation-test it in the usual way — confirm it goes red on the current tree and green after the rename.

## Acceptance

- `rs.yml`'s `gate` and `changes` jobs declare distinct `name:` values.
- Branch protection requires both gates; neither can be satisfied by the other.
- `check-workflow-gates.mjs` fails when two workflows publish the same context name, demonstrated red-then-green.

Found while reviewing #560, which merged before the review landed.

skyl:
Closes #562.

`rs.yml`'s `gate` and `changes` jobs declared no `name:`, so they published their job ids as check-run names — the same names `zuuli.yml`'s jobs publish. `rs.yml` triggers on every `pull_request` with no path filter and its gate is `if: always()`, so **every PR produced two check runs named `gate`**, one of which reports success in ten seconds regardless of what the PR touched. `gate` is the only context branch protection requires, and GitHub resolves a required context by name, so which run satisfied the protected context was a function of report order rather than policy.

Two lines fix it. The rest of this PR makes the defect impossible to reintroduce silently, because the script whose entire purpose is "the required gate is not decorative" counted three same-named gates and exited 0.

## 1. The fix

`.github/workflows/rs.yml` — `name: rs / gate` and `name: rs / changes`, matching the display names the file's other five jobs (`rs / rustfmt`, `rs / clippy`, `rs / cargo-deny`, `rs / tests`, `rs / wasm32`) already carry.

The file's header comment claimed "the `gate` job at the end is what branch protection should require", which named a context that did not exist unambiguously. It now names `rs / gate` and says why the display name is load-bearing rather than cosmetic.

## 2. `check-workflow-gates.mjs` gains two assertions

**(a) No two workflows may publish the same check-run context**, computed as `name ?? job_id` across every workflow file.

**(b) A gate's `needs:` must cover every job in its own file.** The existing rules prove `needs` / `env:` bindings / the `results=` line agree *with each other*; a job added to the file and never wired into `needs` satisfies all three and is invisible to the gate.

---

# Mutation testing

Every assertion below was run. Nothing here is predicted.

## (a) RED before the rename, GREEN after

New assertion code present, rename **not** applied — the tree exactly as `origin/main` has it:

```console
Required-gate wiring failed:
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "changes" (.github/workflows/rs.yml#changes, .github/workflows/zuuli.yml#changes); branch protection matches contexts by name, so a duplicate resolves by report order rather than policy
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "gate" (.github/workflows/rs.yml#gate, .github/workflows/zuuli.yml#gate); branch protection matches contexts by name, so a duplicate resolves by report order rather than policy
2 failure(s) across 2 gate(s) in 14 workflow file(s).
EXIT=1
```

With the rename applied — this branch:

```console
Every gate inspects every job it awaits and covers every job in its file, and no two workflows publish one status-check context: 2 gate(s), 56 job(s), 14 workflow file(s).
EXIT=0
```

## (b) RED with a real unawaited job

Not a fixture — an actual seventh job injected into `rs.yml`, well-formed, `needs: changes`, gated on `needs.changes.outputs.rs == 'true'`, with a real checkout and a real `cargo audit` step, and simply absent from `gate.needs`:

```console
Required-gate wiring failed:
- .github/workflows/rs.yml:326: gate does not await rs_audit, defined in the same workflow; a job outside the gate's needs is a job the gate cannot fail on
1 failure(s) across 2 gate(s) in 14 workflow file(s).
EXIT=1
```

Reverted; the tree in this PR contains no `rs_audit` (`grep -c rs_audit .github/workflows/rs.yml` → `0`) and is green again.

## The new self-test controls actually fire

Five cases added (10 → 15). Each was then proven to be doing the detecting, by breaking the production code path it exercises and confirming the self-test names that exact case:

| Mutation to `check-workflow-gates.mjs` | Self-test result |
|---|---|
| coverage assertion disabled (`unawaited` forced empty) | **EXIT=1** — `self-test FAILED: a job in the file that the gate does not await was not detected. Failures: (none)` |
| collision assertion disabled (`contextCollisionFailures` returns `[]`) | **EXIT=1** — `self-test FAILED: two workflows publishing the same status-check context was not detected. Failures: (none)` |
| tolerated map treated as a wildcard (file list not compared) | **EXIT=1** — `self-test FAILED: a third producer of a tolerated duplicate context was not detected.` (the `gate` group passed under the wildcard while `alpha`/`beta`/`changes` still failed — the control is specific to `gate`, not incidentally satisfied) |
| per-file opt-out ignored (`exempt` forced empty) | **EXIT=1** — `self-test FAILED: a job excused by the per-file opt-out should pass, got: .github/workflows/fixture.yml:33: gate does not await gamma, defined in the same workflow; …` |

Restored:

```console
self-test: an explicitly wired gate passes.
self-test: an exhaustively wired gate passes.
self-test: a job awaited but never bound is detected.
self-test: a job bound but never reported is detected.
self-test: a name reported that is not awaited is detected.
self-test: a result bound for a job that is not awaited is detected.
self-test: a needs object handed to nothing that reads it is detected.
self-test: a gate that awaits only the change detector is detected.
self-test: a gate that does not await the change detector is detected.
self-test: a gate whose needs is a block sequence passes.
self-test: a job in the file that the gate does not await is detected.
self-test: a job excused by the per-file opt-out passes.
self-test: two workflows publishing the same status-check context is detected.
self-test: two gates in one tree with distinct display names passes.
self-test: a third producer of a tolerated duplicate context is detected.
check-workflow-gates self-test: 15 case(s) passed.
EXIT=0
```

---

# Pre-existing collisions found by the scan

Assertion (a) was run against the whole tree with `TOLERATED_CONTEXT_COLLISIONS` emptied, before deciding what to tolerate. There are **four** duplicated contexts on `main`, not two:

```console
- docs-about-free2z.yml, ts-react-free2z.yml, ts-svelte-free2z.yml: 3 job(s) publish "build"
- rs.yml, zuuli.yml: 2 job(s) publish "changes"
- zuuallet.yml, zuuli.yml: 2 job(s) publish "frontend"
- rs.yml, zuuli.yml: 2 job(s) publish "gate"
```

`gate` and `changes` are fixed here. `build` and `frontend` are **not renamed in this PR** — they are recorded in `TOLERATED_CONTEXT_COLLISIONS` with the reason, and filed as #567.

The reason for the split: branch protection requires exactly one context (`{"contexts":["gate"]}`), so neither `build` nor `frontend` can decide a merge today — the ambiguity is cosmetic for them and was load-bearing for `gate`. Renaming five jobs across five workflows is also unrelated churn in a PR about the required context.

The tolerated map is **keyed by the exact set of producing files, not by the name**, so it is not a category exemption: a fourth `build` producer still fails. That property has its own negative control (row 3 of the table above).

# Was the opt-out list needed for (b)?

**No — it is implemented and tested, but empty in production.** Both gates already await every job in their own file:

- `rs.yml`: 6 non-gate jobs (`changes`, `rs_fmt`, `rs_clippy`, `rs_deny`, `rs_test`, `rs_wasm`), all 6 in `gate.needs`.
- `zuuli.yml`: 10 non-gate jobs, all 10 in `gate.needs`.

`GATE_EXEMPT_JOBS` exists anyway because the excuse has to be written next to the rule when someone eventually needs one — a scheduled canary that is not part of a PR verdict is the foreseeable case, and `zuuallet.yml#upstream-canary` is exactly that shape should a gate ever be added there. To keep it from being untested code, `scanRepository` accepts an injected map and the self-test exercises both directions (excused → passes; ignored → the coverage failure returns).

# Other checks

```console
$ node scripts/check-github-actions-pins.mjs --self-test
self-test: 70 source-policy, 7 gate-verdict, and 99 current-workflow mutation case(s) passed.

$ node scripts/check-github-actions-pins.mjs
GitHub Actions policy passed: every external action/reusable workflow is immutably pinned, and every required-gate dependency is bound to the enforcing verdict (178 external reference(s), 15 file(s)).

$ node scripts/check-zuuli-android-gate.mjs --self-test
Android gate policy self-test passed (31 mutations).

$ scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml --manifest rs/crates/f2z-codec/Cargo.toml
```

The exact-length assertion hazard (#556) was avoided: nothing in `check-github-actions-pins.mjs` is touched, and its gate-structure assertions are scoped to `REQUIRED_WORKFLOW_PATH = ".github/workflows/zuuli.yml"`, which this PR does not modify. Adding a `name:` key changes no job id, so every `needs:` and `needs.<job>.result` binding is unaffected.

**#477 shell hygiene: 0 bare `[[ ]]` or `(( ))` added.** This PR adds no shell at all — the one `+ run:` line in the diff is `- run: echo gamma` inside a JavaScript fixture string.

---

# Required follow-up: branch protection

**After this merges, branch protection must add `rs / gate` as a second required context**, so `main` requires `["gate", "rs / gate"]`. That is a repository-settings change and cannot be made from a workflow file.

**The sequencing is not arbitrary — rename first, settings second.** Renaming first makes the currently-protected `gate` context unambiguous immediately: it becomes the wallet's full-stack gate and nothing else, which is the property that is broken right now. Adding `rs / gate` to protection first would change nothing, because no job publishes that name yet, and would leave the ambiguity on `gate` in place for the whole interval.

In the window between this merge and the settings change, `rs` is reporting-only — which is what #560's own body says it expects — and `gate` means the wallet suite, correctly.

Also deferred, from #562's acceptance list: `zuuli.yml`'s gate does not re-run `check-workflow-gates.mjs` the way `rs.yml`'s does. Adding it means changing an exact ordered two-command assertion in `check-github-actions-pins.mjs` and its mutation cases — the #556 hazard — so it is filed as #568 rather than folded in here.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -10,9 +10,13 @@ name: rs
 # workflow demonstrated both directions: an rs/-only diff yields zuuli=false,
 # and a wallet/-only diff yields rs=false.
 #
-# The `gate` job at the end is what branch protection should require. Adding it
-# as a required status check is a repository-settings change and cannot be made
-# in this file.
+# The `gate` job at the end publishes the context `rs / gate`, and that is what
+# branch protection should require alongside the wallet's `gate`. Every job here
+# carries an `rs / ` display name so none of them collides with a job id in
+# another workflow: check-run names are the only thing branch protection matches
+# on, and a name two workflows both publish resolves by report order rather than
+# by policy. Adding `rs / gate` as a required status check is a
+# repository-settings change and cannot be made in this file.
 
 on:
   pull_request:
@@ -36,6 +40,7 @@ env:
 
 jobs:
   changes:
+    name: rs / changes
     runs-on: ubuntu-latest
     timeout-minutes: 5
     outputs:
@@ -290,15 +295,23 @@ jobs:
             --target "${{ steps.rust_toolchain.outputs.target }}" \
             -p f2z-codec
 
-  # Branch protection should require this stable context. It succeeds only when
-  # the change detector and every applicable rs job succeeded; a legitimate
-  # non-rs skip reports success rather than remaining pending.
+  # Branch protection should require this stable context, `rs / gate`. It
+  # succeeds only when the change detector and every applicable rs job
+  # succeeded; a legitimate non-rs skip reports success rather than remaining
+  # pending.
+  #
+  # The `name:` is not cosmetic. Without it this job would publish its job id,
+  # `gate`, which is the context zuuli.yml's full-stack gate already publishes
+  # and the one branch protection requires — two producers for one name, with
+  # the winner decided by report order (#562).
   #
   # Every job in `needs:` below is bound into `env:` AND named on the `results=`
   # line. scripts/check-workflow-gates.mjs enforces that the three sets are
   # equal, so a job that is awaited but never inspected fails review instead of
-  # quietly making this gate decorative.
+  # quietly making this gate decorative. It also requires this `needs:` to cover
+  # every job in this file, so adding a job without wiring it here fails.
   gate:
+    name: rs / gate
     needs: [changes, rs_fmt, rs_clippy, rs_deny, rs_test, rs_wasm]
     if: always()
     runs-on: ubuntu-latest
--- a/scripts/check-workflow-gates.mjs
+++ b/scripts/check-workflow-gates.mjs
@@ -29,6 +29,26 @@
 // list is therefore a deliberate act, and EXHAUSTIVE_VERIFIERS is short on
 // purpose.
 //
+// Two further properties are checked, because a gate can be perfectly wired to
+// the jobs it awaits and still not be the verdict branch protection thinks it
+// is:
+//
+//   3. UNIQUE CONTEXT. A job publishes its check-run name — `name:` when
+//      present, otherwise the job id — and branch protection matches required
+//      checks by that name. GitHub's own documentation says to "make sure that
+//      job names are unique across all workflows. Using the same job name in
+//      multiple workflows can cause ambiguous status check results and block
+//      pull requests from being merged." Two workflows answering to one name
+//      makes which run satisfies the context a function of report order rather
+//      than policy, so this rejects duplicates across the whole tree.
+//
+//   4. TOTAL COVERAGE. A gate's `needs:` must cover every job in its own file.
+//      The wiring rules above only prove `needs` / `env:` / `results=` agree
+//      *with each other*; a job added to the file and never added to `needs`
+//      satisfies all three and is invisible to the gate. That is the likeliest
+//      future regression, so `needs ⊇ {jobs in file} − {gate}` is asserted,
+//      with an explicit per-file opt-out for jobs deliberately left outside.
+//
 // Usage:
 //   node scripts/check-workflow-gates.mjs             judge every workflow
 //   node scripts/check-workflow-gates.mjs --self-test negative controls first
@@ -51,6 +71,40 @@ const CHANGE_DETECTOR = "changes";
 /// Commands that consume `toJSON(needs)` by iterating every entry. Each has been
 /// read and does. Adding one means reading the new verifier first.
 const EXHAUSTIVE_VERIFIERS = ["--verify-gate-results"];
+
+/// Check-run names that more than one workflow is knowingly allowed to publish,
+/// mapped to the exact set of files allowed to publish them. Every entry is a
+/// documented exception, not a category: a *new* producer of one of these names
+/// still fails, because the value is the full file list rather than a wildcard.
+///
+/// Neither name below is a required status check, so neither can currently
+/// decide a merge — that is the whole reason they are tolerated rather than
+/// treated as the `gate` collision was. Renaming them is a follow-up, tracked
+/// in #567; adding to this map instead of fixing a collision needs the same.
+const TOLERATED_CONTEXT_COLLISIONS = new Map([
+  [
+    "build",
+    [
+      ".github/workflows/docs-about-free2z.yml",
+      ".github/workflows/ts-react-free2z.yml",
+      ".github/workflows/ts-svelte-free2z.yml",
+    ],
+  ],
+  [
+    "frontend",
+    [".github/workflows/zuuallet.yml", ".github/workflows/zuuli.yml"],
+  ],
+]);
+
+/// Jobs a gate is deliberately not required to await, per workflow file.
+///
+/// Empty on purpose: today both gates await every job in their own file, so
+/// nothing needs excusing. It exists because the excuse must be written down
+/// next to the rule when one is eventually needed — a scheduled canary that is
+/// not part of a PR verdict is the foreseeable case — rather than the rule
+/// being weakened to accommodate it. The self-test exercises this path with an
+/// injected map so the opt-out is not untested code.
+const GATE_EXEMPT_JOBS = new Map();
 
 const NEEDS_RESULT = /needs\.([A-Za-z0-9_-]+)\.result/g;
 const TO_JSON_NEEDS = /toJSON\(\s*needs\s*\)/;
@@ -105,6 +159,26 @@ function jobsIn(lines) {
   }
   if (current) current.end = lines.length;
   return jobs;
+}
+
+/// The check-run name a job publishes, which is what branch protection matches
+/// a required context against: the job's `name:` when it declares one, and the
+/// job id otherwise.
+///
+/// Conservative about matrix jobs on purpose. A job with a `strategy.matrix`
+/// publishes `<name> (<values>)` per leg, so two files sharing a base name are
+/// not literally the same context — but the resolution is still to give them
+/// distinct names, so the base name is what is compared.
+function publishedContext(lines, job) {
+  for (let index = job.start + 1; index < job.end; index += 1) {
+    const line = lines[index];
+    if (isBlank(line)) continue;
+    if (indentOf(line) <= 2) break;
+    if (indentOf(line) !== 4) continue;
+    const match = /^ {4}name:\s*(.+?)\s*$/.exec(line);
+    if (match) return match[1].replace(/^['"]|['"]$/g, "");
+  }
+  return job.name;
 }
 
 /// The job ids in a job's `needs:`, in any of the three shapes GitHub accepts.
@@ -182,7 +256,16 @@ function difference(left, right) {
 }
 
 /// Judge one `gate` job. Returns an array of human-readable failures.
-export function gateFailures(relativeFile, lines, job) {
+///
+/// `siblings` is every job id defined in the same file, and `exempt` the ids
+/// that file is allowed to leave outside the gate.
+export function gateFailures(
+  relativeFile,
+  lines,
+  job,
+  siblings = [],
+  exempt = new Set(),
+) {
   const failures = [];
   const needs = parseNeeds(lines, job);
   if (!needs) {
@@ -195,6 +278,21 @@ export function gateFailures(relativeFile, lines, job) {
       `${relativeFile}:${needs.index + 1}: gate must await ${CHANGE_DETECTOR}`,
     );
   }
+
+  // Coverage. Wiring the three sets to each other proves nothing about a job
+  // that was never wired at all, so require the gate to await every job its own
+  // workflow defines.
+  const unawaited = difference(
+    difference(new Set(siblings.filter((id) => id !== job.name)), exempt),
+    awaited,
+  );
+  if (unawaited.size) {
+    failures.push(
+      `${relativeFile}:${needs.index + 1}: gate does not await ${sorted(unawaited)}, ` +
+        "defined in the same workflow; a job outside the gate's needs is a job the gate cannot fail on",
+    );
+  }
+
   const judged = difference(awaited, new Set([CHANGE_DETECTOR]));
   if (judged.size === 0) {
     failures.push(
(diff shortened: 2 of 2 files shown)


skyl:
Measured on this PR's own head, after every suite completed — the duplicate is gone in the place that matters:

```console
$ gh pr view 569 --json statusCheckRollup \
    --jq '[.statusCheckRollup[].name]|group_by(.)|map(select(length>1))|map({name:.[0],count:length})'
[]
```

That is the exact query #562's comment ran against #560, where it returned `[{"count":2,"name":"changes"},{"count":2,"name":"gate"}]`.

Full rollup, all green:

```
changes                      COMPLETED  SUCCESS     <- zuuli.yml, the protected context
gate                         COMPLETED  SUCCESS     <- zuuli.yml, the protected context
rs / changes                 COMPLETED  SUCCESS
rs / gate                    COMPLETED  SUCCESS     <- new, distinct
rs / cargo-deny              COMPLETED  SUCCESS
rs / clippy                  COMPLETED  SUCCESS
rs / rustfmt                 COMPLETED  SUCCESS
rs / tests                   COMPLETED  SUCCESS
rs / wasm32                  COMPLETED  SUCCESS
frontend                     COMPLETED  SKIPPED
Rust / …  (8 jobs)           COMPLETED  SKIPPED
Zuuallet / generated schema  COMPLETED  SKIPPED
```

The wallet jobs skip because this diff touches only `.github/workflows/rs.yml` and `scripts/`, which its detector does not select; `gate` reports success for that legitimate skip, which is its designed behaviour. `gate` here is unambiguously zuuli.yml's — there is now exactly one check run answering to that name.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — `gate` and `rs / gate` both green and now distinct. I re-ran every mutation myself in a throwaway worktree rather than reading them off the body.

## The guard is real, and specific

```
MUT A  remove `name: rs / gate` (the exact pre-PR state)
  -> EXIT=1  2 job(s) publish the status-check context "gate"
             (.github/workflows/rs.yml#gate, .github/workflows/zuuli.yml#gate);
             branch protection matches contexts by name, so a duplicate resolves
             by report order rather than policy

MUT B  also remove `name: rs / changes`
  -> EXIT=1  two failures, one per colliding context — not one generic message

MUT D  inject a real 7th job `rs_audit` (needs: changes, if: rs == 'true')
  -> EXIT=1  rs.yml:324: gate does not await rs_audit, defined in the same
             workflow; a job outside the gate's needs is a job the gate cannot fail on

MUT E  drop rs_wasm from gate.needs
  -> EXIT=1  three stacked failures: the new coverage rule AND both pre-existing
             binding/results rules. The old assertions still work.
```

Baseline on the PR head is green, and the rollup is clean:

```
$ gh pr view 569 --json statusCheckRollup \
    --jq '[.statusCheckRollup[].name]|group_by(.)|map(select(length>1))'
[]
```

versus `[{"count":2,"name":"changes"},{"count":2,"name":"gate"}]` on #560. That's the defect closed, measured the same way it was found.

Self-test is 10 → 15, and the new controls are the right ones — including **"a third producer of a tolerated duplicate context is detected"**, which is what stops `TOLERATED_CONTEXT_COLLISIONS` becoming an open door for *new* producers. Keying the map by the exact producing file list rather than by name is the correct design.

## The pre-existing-collision scan is the best thing in this PR

Running it with the tolerated map emptied found **four** duplicated contexts on `main`, not the two I knew about:

| Context | Producers |
|---|---|
| `gate` | rs.yml, zuuli.yml — fixed here |
| `changes` | rs.yml, zuuli.yml — fixed here |
| `build` | docs-about-free2z.yml, ts-react-free2z.yml, ts-svelte-free2z.yml |
| `frontend` | zuuallet.yml, zuuli.yml |

Neither `build` nor `frontend` is a required context, so neither can decide a merge — tolerating them with the reason written down and filing #567 is the right call rather than renaming four workflows in a P0 fix.

`GATE_EXEMPT_JOBS` being implemented-but-empty is also right: both gates already await every job in their own file (rs.yml 6/6, zuuli.yml 10/10), and having `scanRepository` accept an injected map means the opt-out path is exercised by the self-test instead of being untested code waiting for its first user.

## One follow-up, not blocking

The tolerated map can silence the protected context itself:

```
MUT C  add ["gate", [rs.yml, zuuli.yml]] to TOLERATED_CONTEXT_COLLISIONS,
       then remove `name: rs / gate`
  -> EXIT=0
```

Two deliberate steps, so it's not an accident waiting to happen — but it's the same self-describing-scope shape I flagged on #556, where dropping entries from `EXPECTED_PACKAGES` leaves the checker green at a lower mutant count. The fix is small: a `PROTECTED_CONTEXTS` constant (`gate`, and `rs / gate` once branch protection has it) that `TOLERATED_CONTEXT_COLLISIONS` may never cover, with its own negative control. Filing it.

## Conventions

- `check-github-actions-pins.mjs` untouched, and correctly so — its gate-structure assertions are scoped to `REQUIRED_WORKFLOW_PATH = ".github/workflows/zuuli.yml"`, so the exact-ordered-command hazard currently blocking #556 was sidestepped entirely rather than worked around.
- Adding `name:` changes no job id, so every `needs:` reference is unaffected.
- **#477: zero bare `[[ ]]`/`(( ))` added** — this PR adds no shell at all; the one `+ run:` line in the diff is inside a JS fixture string.
- The header comment at `rs.yml:13-15` was updated too. It previously pointed branch protection at `gate`, naming a context that didn't unambiguously exist; it now names `rs / gate`. That was worth catching.

## Next, in this order

1. This merges.
2. **Branch protection gets `rs / gate` as a second required context** — `main` → `["gate", "rs / gate"]`. I'll make that change directly.

The sequencing in the PR body is right and worth restating: renaming first makes the currently-protected `gate` unambiguous *immediately*, whereas adding the context first would have changed nothing (no job published that name) and left the ambiguity standing in the meantime.

#568 (zuuli.yml's gate not re-running `check-workflow-gates.mjs` the way rs.yml's does) is correctly deferred — closing it means editing the exact ordered two-command assertion that #556 is currently stuck on, and doing both at once would collide.

PR #571, opened 2026-08-24
supply-chain(rs): add the akd >= 0.13.0 floor; correct three rs/ docs

### LINKED ISSUE #565: supply-chain(rs): the akd >= 0.13.0 floor is missing — a 0.12.0 pin passes both advisories and bans
## Problem

#555 closed as COMPLETED via #560, which created `rs/deny.toml` — but **without ADR 0013's `akd` version floor**, which was the one criterion I added to #555 specifically because nothing else can enforce it.

`rs/deny.toml:105-107` is the entire `[bans]` section:

```toml
[bans]
multiple-versions = "warn"
wildcards = "deny"
```

No `[[bans.deny]]`. The string `akd` does not appear anywhere in that PR:

```console
$ grep -rn "akd" rs/ scripts/ .github/ ; echo "(end)"
(end)
```

## Why nothing else catches it

ADR 0013 (`docs/e2ee/decisions/0013-key-transparency-log.md:93-94`) sets a hard floor at `akd 0.13.0` — the first release containing [`facebook/akd#495`](https://github.com/facebook/akd/pull/495), *"Fix auditor append-only bypass in batch node insertion"*, whose own PR body states that **"a malicious server can exploit this to rewrite an existing label's value while still producing a valid append-only proof."** Merged `2026-08-12T17:45:01Z`; `0.13.0` published `2026-08-13T16:29:07Z` and is `ahead_by: 1` from the merge commit.

There is no advisory for it, anywhere. Verified across all four sources:

- OSV API returns literal `{}` for `akd` unversioned, at `0.12.0`, and at `0.9.0`
- `rustsec/advisory-db` has no `crates/akd` directory (404); code search returns `total_count: 0`
- `gh api /repos/facebook/akd/security-advisories` → `[]`
- GitHub advisory DB filtered to Rust / `akd` → `[]`

So `cargo audit` and `cargo deny check advisories` are **genuinely silent** on a stale `0.12.0` pin. `KT.md:1180-1184` says so plainly, and is right: *"A reviewer who removes that entry has removed the floor; **there is no second line of defence**."*

## Demonstrated, not asserted

Scratch member under `rs/` depending on `akd = "0.12.0"`, resolved into `rs/Cargo.lock` (`akd v0.12.0`, `akd_core v0.12.0`, both from crates.io):

```console
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml advisories
advisories ok
All 3 Rust crate(s) under rs/ satisfy rs/deny.toml.

$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml bans
bans ok
All 3 Rust crate(s) under rs/ satisfy rs/deny.toml.
```

**Both green.** Not the asymmetry the entry exists to create — no control at all.

After pasting ADR 0013's entry into `rs/deny.toml`:

```console
    ├ akd v0.12.0
    ├ akd_core v0.12.0
```

…and with `akd = "0.13"` (resolves `akd v0.13.0`): `bans ok`. So it is a ~15-line paste, it is not a blanket ban, and the fix is verified in both directions.

## `KT.md` currently describes a control that does not exist

`docs/e2ee/KT.md:1180-1184` states the floor in the **present tense** — "`akd >= 0.13.0`, `akd_core >= 0.13.0`, hard, **enforced by** a `[[bans.deny]]` entry" — and closes with a sentence that presupposes an entry there is to remove. Until this lands, that paragraph is a claim about something nobody wrote. (Tracked alongside the other ADR 0013 / KT.md accuracy fixes in #559, but it resolves itself the moment the entry exists.)

## Fix

1. Add ADR 0013's three `[[bans.deny]]` entries to `rs/deny.toml` verbatim — `akd`, `akd_core` (both floored at `0.13.0`) and `akd_mysql` (banned outright) — including the inline comment naming `facebook/akd#495` and the "no advisory exists" reasoning, so the next reviewer cannot delete the entry as redundant with `cargo audit`. That is precisely the mistake it exists to survive.
2. Mutation-test it in the PR and paste the output: `bans` **red** at `0.12.0`, `advisories` **green** at `0.12.0`, both green at `0.13.0`. The asymmetry is the whole justification; demonstrating it is what makes the entry defensible later.

## Acceptance

- `rs/deny.toml` carries the three entries with the ADR 0013 / `akd#495` comment.
- The red/green/green mutation evidence is in the PR body.
- `KT.md:1180-1184`'s present tense becomes true.

Found while reviewing #560, which merged before the review landed. Closes the gap #555 was amended to cover.

### LINKED ISSUE #566: docs(rs): README claims a discovery guarantee that doesn't hold, and argues a decided question as open
Three documentation defects landed with the `rs/` tree in #560 (merged as `241b714`). None affects behaviour; all three describe a control or a decision inaccurately, in files whose whole purpose is to keep the record straight.

## 1. `rs/README.md:117` states the thing #553 exists to falsify

> "All four discover crates rather than listing them, so a new crate under `rs/` is gated from its first commit."

Three of the four discover. Mutation-proved by adding `rs/crates/scratch/Cargo.toml` with no `rust-version` and no `license`:

| Script (unchanged flags) | Sees it? | Result |
|---|---|---|
| `check-rust-fmt.sh --root rs` | yes | RED — `3 crate(s) are not formatted: rs/. rs/crates/f2z-codec rs/crates/scratch` |
| `check-rust-clippy.sh --root rs` | yes | RED — `3 crate(s) failed the clippy gate: … rs/crates/scratch` |
| `check-rust-deny.sh --root rs --config rs/deny.toml` | yes | RED — `licenses FAILED` |
| `check-rust-toolchain.sh --toolchain-file … --manifest …/f2z-codec/Cargo.toml` | **no** | **GREEN** |

`scripts/check-rust-toolchain.sh:381` greps the literal `rust-version` line out of a hardcoded `MANIFESTS` array plus explicit `--manifest` flags (`.github/workflows/rs.yml:72`). A second crate under `rs/crates/` ships with a wrong MSRV, or none, and passes.

That is #553, not a bug introduced here — but this is the first tree where it bites, and the sentence reads as a solved problem to the next person. Point it at #553 instead.

**Also for #553:** `wallet/zuuli/crypto-target-spike/Cargo.toml:5` declares `rust-version = "1.97"` and nothing checks it — a second known unregistered manifest. Full census on `main`:

```
rs/Cargo.toml                                  virtual root, no [package] — correctly unregistered
rs/crates/f2z-codec/Cargo.toml                 registered via --manifest (rs.yml:72)         ok
wallet/plugins/tauri-plugin-zcash/Cargo.toml   MANIFESTS[3]                                  ok
wallet/zuuallet/src-tauri/Cargo.toml           MANIFESTS[2]                                  ok
wallet/zuuli/crypto-target-spike/Cargo.toml    NOT in MANIFESTS, no --manifest anywhere      GAP
wallet/zuuli/src-tauri/Cargo.toml              MANIFESTS[1]                                  ok
wallet/zuuli/wasm-spike/Cargo.toml             MANIFESTS[0]                                  ok
```

## 2. `rs/README.md:22-41` argues against a premise that is no longer true

The README discusses #341 as an open question — "**asks whether** the *repository* should have a single Rust workspace", "**If #341 later chooses** one repo-wide workspace" — and reconstructs the counter-arguments from the **issue body's** "Arguments against" section.

#341 is **CLOSED / COMPLETED** as of `2026-08-23T09:18:55Z`, via PR #519, which landed `docs/architecture/CARGO-WORKSPACE.md` — present in `rs/`'s own tree, header: `Status: accepted on 2026-08-23. This records the decision for issue #341.`

**The substance survives and gets stronger.** The ADR says:

> Do not create a repository-root workspace. Cargo automatically enrolls path dependencies below a workspace root, and this synthetic monorepo contains upstream Cargo projects under `z/` that are not ours to govern as workspace members. **If a wallet-wide workspace is ever adopted, its candidate root is `wallet/`**, with explicit members and exclusions.

The decided consolidation target was never repo-wide — it is `wallet/`. So a fourth root at `rs/` does **not** make the accepted consolidation harder; the two trees would never have merged under it. `rs/` also satisfies the ADR's own "Adding Rust crates" step 5 (*"Run the repository's manifest-discovering format, clippy, and cargo-deny checks. Add explicit locked build/test coverage…"*) — that is exactly what `rs.yml` does.

Two one-line fixes:
- Cite `docs/architecture/CARGO-WORKSPACE.md` rather than the closed issue, and say #341 is **decided**.
- Drop the `:40-41` hedge — "If #341 later chooses one repo-wide workspace, `rs/` folds into it by moving these members and deleting one `[workspace]` table" — which guards against a scenario the accepted ADR rules out, and understates the cost anyway (folding two workspaces means one reviewed dependency-resolution diff across both lockfiles, not deleting a table). Replace with the ADR's revisit-trigger framing.

## 3. `rs/deny.toml:27-31` — the comment describes a directive that isn't there

```toml
[graph]
# Evaluate every target. This tree compiles for wasm32-unknown-unknown as well
# as for the servers' hosts, and a host-only view would miss whatever the wasm
# half pulls in.
all-features = false
```

`all-features` is a **feature** knob. The target knob is `[graph] targets`, which is absent — and absent means all targets, so the behaviour the comment describes is delivered by cargo-deny's default, not by the line beneath it. Both settings here are also cargo-deny's defaults, making the whole `[graph]` block a no-op that documents itself as doing something.

Either set `targets = []` explicitly with that comment attached, or move the comment to where the behaviour actually comes from. This repo polices "a comment claiming a control" everywhere else, and the intent — evaluate the wasm target too — is worth stating in a way that survives a future cargo-deny default change.

## Worth stating: the rest checks out

So these three don't get mixed up with doubts about the tree. `rs/deny.toml` is **stricter than `wallet/deny.toml` in every dimension and more permissive in none**: empty `advisories.ignore` (vs 22 entries for the #351 backlog), `licenses.private.ignore = false` (vs `true`), 4 allowed licences (vs 14), no exceptions, `multiple-versions = "warn"` (vs `allow`), `wildcards = "deny"` (vs unset). The deny run is green with **zero** warnings, so the claim that the allow list is exactly what the graph carries holds. And the toolchain restatement discipline has a real mechanical reason: `rust-version.workspace = true` does not match `check-rust-toolchain.sh:381`'s regex, so inheriting would fail the check outright.

Found while reviewing #560. Related: #553, #562, #563, #564, #565.

skyl:
Closes #565
Closes #566

## The akd floor, mutation-tested

`rs/deny.toml` now carries ADR 0013's three `[[bans.deny]]` entries verbatim —
`akd` and `akd_core` floored at `0.13.0`, `akd_mysql` banned outright — with the
inline comment naming [`facebook/akd#495`](https://github.com/facebook/akd/pull/495)
and stating that no RustSec/OSV advisory exists, so `cargo audit` and
`cargo deny check advisories` will never catch a downgrade. That comment is the
load-bearing part: it is what stops the next reviewer deleting the entry as
redundant with `cargo audit`, which is exactly the mistake it exists to survive.

**Method.** Scratch workspace member `rs/crates/akd-floor-probe` depending on
`akd = "0.12.0"`, `rs/Cargo.lock` resolved from crates.io (`akd v0.12.0`,
`akd_core v0.12.0`), then each check run separately. The scratch member and the
regenerated lockfile are **not** in this diff — `git status` is clean of them and
`rs/Cargo.lock` is byte-identical to `main`.

### `akd = "0.12.0"` — advisories **GREEN**

```console
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml advisories

==> cargo-deny check advisories -- rs/.
advisories ok

==> cargo-deny check advisories -- rs/crates/akd-floor-probe
advisories ok

==> cargo-deny check advisories -- rs/crates/f2z-codec
advisories ok

All 3 Rust crate(s) under rs/ satisfy rs/deny.toml.
```

### `akd = "0.12.0"` — bans **RED**

```console
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml bans

==> cargo-deny check bans -- rs/.
    │         ━━━ banned here
    │         ━━━━━━━━ banned here
bans FAILED

==> cargo-deny check bans -- rs/crates/akd-floor-probe
bans FAILED

==> cargo-deny check bans -- rs/crates/f2z-codec
bans ok

2 crate(s) failed the supply-chain policy: rs/. rs/crates/akd-floor-probe
```

**Advisories green while bans is red, on the same lockfile, at the same commit.**
That asymmetry is the entire justification for the entry: there is nothing in the
advisory databases to find, so the ban is the only control.

### `akd = "0.13"` (resolves `akd v0.13.0`, `akd_core v0.13.0`) — both **GREEN**

```console
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml advisories
advisories ok  (×3)
All 3 Rust crate(s) under rs/ satisfy rs/deny.toml.

$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml bans
bans ok  (×3)
All 3 Rust crate(s) under rs/ satisfy rs/deny.toml.
```

Not a blanket ban: `0.13.0` passes, `0.12.0` does not, and the failure names the
crate and the version.

### `docs/e2ee/KT.md:1180-1184`

Unchanged, deliberately. It stated the floor in the present tense — "hard,
enforced by a `[[bans.deny]]` entry" — and describes a control that now exists.
Re-read against the landed entry, every clause is true as written, including
"there is no second line of defence". Nothing to correct.

## #566/1 — `rs/README.md` claimed a discovery guarantee that does not hold

Verified here, not taken on trust. Throwaway `rs/crates/scratch/Cargo.toml` with
no `rust-version` and no `license`, all four scripts run with their registered
flags unchanged:

| Script (registered flags) | Sees it? | Verdict |
|---|---|---|
| `check-rust-fmt.sh --root rs` | yes | **RED** — `3 crate(s) are not formatted: rs/. rs/crates/f2z-codec rs/crates/scratch` |
| `check-rust-clippy.sh --root rs` | yes | **RED** — `2 crate(s) failed the clippy gate: rs/. rs/crates/scratch` |
| `check-rust-deny.sh --root rs --config rs/deny.toml` | yes | **RED** — `error[unlicensed]: scratch = 0.0.0 is unlicensed` |
| `check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml --manifest rs/crates/f2z-codec/Cargo.toml` | **no** | **GREEN** |

Three of four discover; the toolchain check greps a hardcoded `MANIFESTS` array
plus explicit `--manifest` flags, so a second crate under `rs/crates/` ships with
a wrong MSRV, or none, and passes. The README now says so, tables the evidence,
tells the next author to add a `--manifest` flag in the same commit that creates
the crate, and points at #553 for the real fix. The throwaway crate is removed
and `rs/Cargo.lock` restored.

One honest note on the clippy row: the throwaway is only clippy-red once it opts
into `[lints] workspace = true`. Discovery — the claim under test — happened
either way; `==> cargo clippy -- rs/crates/scratch` appears in both runs.

## #566/2 — `rs/README.md` argued a decided question as open

`docs/architecture/CARGO-WORKSPACE.md` (accepted 2026-08-23, PR #519) is in the
tree. The section is rewritten to cite the ADR rather than the closed issue, and
the `:40-41` hedge — "If #341 later chooses one repo-wide workspace, `rs/` folds
into it by … deleting one `[workspace]` table" — is gone, because the ADR rules
that scenario out and the hedge understated the cost anyway.

The argument got stronger rather than shorter. The ADR's decided consolidation
target is **`wallet/`**, not repo-wide ("If a wallet-wide workspace is ever
adopted, its candidate root is `wallet/`"), and `rs/` is not under `wallet/` — so
a fourth root here does not obstruct the accepted consolidation; the two trees
were never candidates to merge. The wallet-specific reasons are re-sourced from
the ADR (release tooling reads `src-tauri/Cargo.lock` and `src-tauri/target`;
the locks resolve Tauri 2.11.5 / `aws-lc-rs` 1.17.3 against 2.10.2 / 1.15.4; the
`links = "sqlite3"` singleton), and the hedge is replaced with the ADR's own
revisit-trigger framing. It also records that `rs.yml` is how this tree satisfies
the ADR's "Adding Rust crates" step 5.

## #566/3 — `rs/deny.toml`'s `[graph]` comment described a directive that was not there

Chose **`targets = []` explicit, comment attached to it**, rather than moving the
comment. The comment states an intent — evaluate the wasm target too — that we
actually depend on; leaving it to a default means an absent key is
indistinguishable from an unconsidered one, and an upstream default change would
narrow the graph silently between two CI runs with no commit of ours in between.
`all-features = false` keeps a separate, accurate comment saying it is a
*feature* knob. Neither line changes today's behaviour; the point is that the
file now says which line delivers which property.

## `crypto-target-spike` — registered

`wallet/zuuli/crypto-target-spike/Cargo.toml:5` declared `rust-version = "1.97"`
and nothing checked it. Registered as one line in `check-rust-toolchain.sh`'s
`MANIFESTS` array rather than a workflow `--manifest` flag, because `zuuli.yml`
runs the script bare — so the array is where a *wallet* manifest belongs, and no
workflow needed editing. Green both ways:

```console

```

`AGENTS.md`'s restatement table said "the three `Cargo.toml` manifests"; it is now
"every registered `Cargo.toml` manifest", which does not rot with the count.

## Checks

```
scripts/check-rust-deny.sh --root rs --config rs/deny.toml      advisories ok, bans ok, licenses ok, sources ok (2 crates, zero warnings)
scripts/check-rust-deny.sh --self-test                          4 case(s) passed; cargo-deny 0.20.2
scripts/check-rust-fmt.sh --root rs                             2 crate(s) formatted (rustfmt 1.97.1)
scripts/check-rust-clippy.sh --root rs                          2 crate(s) clippy-clean (clippy 1.97.1)
scripts/check-rust-toolchain.sh --toolchain-file rs/... --manifest rs/crates/f2z-codec/Cargo.toml   pass
scripts/check-rust-toolchain.sh                                 pass
scripts/check-rust-toolchain.sh --self-test                     14 drift case(s) caught
node scripts/check-workflow-gates.mjs [--self-test]             15 case(s); 2 gates, 56 jobs, 14 workflow files
node scripts/check-github-actions-pins.mjs [--self-test]         176 case(s); 178 external refs, 15 files
```

No workflow file is touched, so `.github/workflows/zuuli.yml`'s `changes` policy
step and its ordered command assertion are untouched (#556's hazard).

**#477 shell hygiene: 0 bare `[[ ]]` / `(( ))` statements added.** The only shell
change in the diff is one array element; the three `[[bans.deny]]` lines a naive
grep flags are TOML array-of-tables in `rs/deny.toml`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -210,7 +210,7 @@ read a TOML file at the moment they need the value:
 
 | Restatement | Why it cannot just read the file |
 |---|---|
-| `rust-version` in the three `Cargo.toml` manifests | Cargo needs a literal, and it is the **two-component MSRV floor** (`X.Y`) of the three-component channel (`X.Y.Z`) — deliberately a different form, compared as such |
+| `rust-version` in every registered `Cargo.toml` manifest | Cargo needs a literal, and it is the **two-component MSRV floor** (`X.Y`) of the three-component channel (`X.Y.Z`) — deliberately a different form, compared as such |
 | `ZUULI_RUST_VERSION` in `zuuli-packaging.yml` and `zuuli-release.yml` | A workflow-level `env:` cannot be computed from a file, and the release jobs verify the installed compiler with `rustc --version \| grep -F "rustc $ZUULI_RUST_VERSION "` |
 | `dtolnay/rust-toolchain@<sha> # <version>` in packaging/release and target-native Zuuallet jobs | The action's **version branches hardcode the compiler in `action.yml` and do not declare a `toolchain` input at all** — a commit-pinned ref *is* the version pin, and the trailing comment is its only readable record |
 | `dtolnay/rust-toolchain@<sha> # stable` in source-derived gate/Zuuallet jobs | `uses:` cannot contain an expression, so the generic action implementation is commit-pinned while its `toolchain:` input still reads the version from `wallet/rust-toolchain.toml`; the upstream canary deliberately omits that input so only its compiler selection follows `stable` |
--- a/rs/Cargo.toml
+++ b/rs/Cargo.toml
@@ -1,7 +1,8 @@
 # The `rs/` workspace — one [workspace], one Cargo.lock.
 #
-# See rs/README.md for why this tree exists, why it is named `rs/`, and why the
-# decision recorded here deliberately does not decide #341.
+# See rs/README.md for why this tree exists, why it is named `rs/`, and why a
+# fourth package root here sits inside the workspace boundary decided for #341
+# in docs/architecture/CARGO-WORKSPACE.md rather than cutting across it.
 
 [workspace]
 resolver = "3"
--- a/rs/README.md
+++ b/rs/README.md
@@ -19,26 +19,52 @@ a namespace named for one role would be wrong within a quarter.
 `rs/` is the role-neutral name. `f2z-codec` is the first crate in it and is
 already both: the relay parses frames with it, and so does the browser.
 
-## This does not decide #341
+## #341 is decided, and a root here is inside the decision
 
-[#341](https://github.com/free2z/zuu/issues/341) asks whether the *repository*
-should have a single Rust workspace. **Nothing here answers that**, and this
-file exists partly to say so out loud, because "a second workspace was added
-after #341 was filed" is exactly the kind of fact that later gets read as a
-decision.
+[#341](https://github.com/free2z/zuu/issues/341) asked whether the *repository*
+should have a single Rust workspace. It is **decided**, and the answer is in
+[`docs/architecture/CARGO-WORKSPACE.md`](../docs/architecture/CARGO-WORKSPACE.md)
+— accepted 2026-08-23 — not in the issue thread. Read the ADR, not this section,
+when the question comes up again.
 
-Every argument against consolidation in #341 is wallet-specific:
+The ADR keeps the wallet's three Rust package roots independent and rules out a
+repository-root workspace outright, because Cargo enrols path dependencies below
+a workspace root and `z/` holds upstream Cargo projects that are not ours to
+govern as members. Its own words:
 
-- Tauri's mobile project layout expects the app crate where it is.
+> If a wallet-wide workspace is ever adopted, its candidate root is `wallet/`,
+> with explicit members and exclusions.
+
+**So the consolidation the ADR leaves open is `wallet/`-scoped, and `rs/` is not
+under `wallet/`.** A fourth package root here does not make that consolidation
+harder; the two trees were never candidates to merge under it. The ADR's reasons
+for the boundary are wallet-specific in exactly the same way:
+
+- The release system runs the Tauri CLI from the app directories and reads
+  `src-tauri/Cargo.lock` and `src-tauri/target`; Tauri's generated mobile
+  projects expect the app crate where it is.
 - The wallet's three lockfiles resolve the Zcash and Tauri graphs **differently
-  on purpose**, and unifying them would silently change what ships.
+  on purpose** — Tauri 2.11.5 / `aws-lc-rs` 1.17.3 on one side, 2.10.2 / 1.15.4
+  on the other — so unifying them starts with a coordinated dependency update,
+  not a relocation.
 - `libsqlite3-sys` declares `links = "sqlite3"`, so Cargo hard-errors on a graph
   containing two versions — which makes `rusqlite 0.37` a repo-wide singleton and
   makes any workspace merge a resolution question, not a layout question.
 
-None of those applies to a fresh tree with no Tauri, no SQLite and no path
-dependencies into `z/`. If #341 later chooses one repo-wide workspace, `rs/`
-folds into it by moving these members and deleting one `[workspace]` table.
+None of those engages a tree with no Tauri, no SQLite and no path dependencies
+into `z/`. What `rs/` does owe the ADR is its "Adding Rust crates" step 5 — run
+the repository's manifest-discovering format, clippy and cargo-deny checks, and
+add explicit locked build/test coverage for the context the crate promises. That
+is what `.github/workflows/rs.yml` is.
+
+The ADR does not freeze the question; it lists **revisit triggers** — a shared
+release cadence, two or more further shared crates, a divergence that escapes
+the per-app locked builds, measured CI data, or release tooling made independent
+of the app-local lock and target paths — and it accepts consolidation only after
+a migration proves the whole list of builds and tests. None of those triggers is
+about `rs/`. If one ever fires and this tree is in scope, folding it in is one
+reviewed dependency-resolution diff across both lockfiles — not the deletion of
+a `[workspace]` table.
 
 ## The toolchain pin is a restatement
 
@@ -114,9 +140,29 @@ scripts/check-rust-clippy.sh --root rs
 scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
 ```
 
-All four discover crates rather than listing them, so a new crate under `rs/` is
-gated from its first commit. `.github/workflows/rs.yml` runs them, plus
-`cargo test` and the WASM build, behind its own `gate`.
+**Three of the four discover crates; the toolchain check does not.**
+`check-rust-fmt.sh`, `check-rust-clippy.sh` and `check-rust-deny.sh` find every
+`Cargo.toml` under `--root`, so a new crate under `rs/` is formatted, linted and
+supply-chain-gated from its first commit. `check-rust-toolchain.sh` reads a
+hardcoded manifest list plus the explicit `--manifest` flags above, so **a second
+crate under `rs/crates/` would ship with a wrong MSRV, or none, and still pass.**
+
+Proved by adding a throwaway `rs/crates/scratch/Cargo.toml` with no
+`rust-version` and no `license`, then running all four with the flags above
+unchanged:
+
+| Script | Sees it? | Verdict |
+|---|---|---|
+| `check-rust-fmt.sh --root rs` | yes | RED — `3 crate(s) are not formatted: … rs/crates/scratch` |
+| `check-rust-clippy.sh --root rs` | yes | RED — `2 crate(s) failed the clippy gate: … rs/crates/scratch` |
+| `check-rust-deny.sh --root rs --config rs/deny.toml` | yes | RED — `error[unlicensed]: scratch = 0.0.0 is unlicensed` |
+| `check-rust-toolchain.sh --toolchain-file … --manifest …/f2z-codec/Cargo.toml` | **no** | **GREEN** |
+
+So **every new crate in this tree needs another `--manifest` flag added to
+`rs.yml`** in the same commit that creates it, until
+[#553](https://github.com/free2z/zuu/issues/553) makes that check discover
+manifests too. `.github/workflows/rs.yml` runs all four, plus `cargo test` and
+the WASM build, behind its own `gate`.
 
 ### Workspace lints
 
--- a/rs/deny.toml
+++ b/rs/deny.toml
@@ -28,6 +28,18 @@
 # Evaluate every target. This tree compiles for wasm32-unknown-unknown as well
 # as for the servers' hosts, and a host-only view would miss whatever the wasm
 # half pulls in.
+#
+# `targets` is the knob that decides that, and an empty list is cargo-deny's
+# "every target". It is stated rather than left absent because this is a
+# property we rely on: an absent key is indistinguishable from an unconsidered
+# one, and a default that changes upstream would narrow the graph silently,
+# between two CI runs, with no commit of ours in between.
+targets = []
+
+# A *feature* knob, not a target one. `--all-features` would judge a graph
+# nothing in this tree compiles; the crates here are checked with the features
+# they actually build with. Restated rather than inherited so that turning it on
+# is a visible, reviewed edit.
 all-features = false
 
 # =============================================================================
@@ -105,6 +117,26 @@ exceptions = []
 [bans]
 multiple-versions = "warn"
 wildcards = "deny"
+
+[[bans.deny]]
+# akd < 0.13.0 contains an auditor append-only bypass, facebook/akd#495, merged
+# 2026-08-12 and released in 0.13.0 on 2026-08-13: a malicious log can rewrite an
+# existing label's value while still producing a VALID append-only proof, which
+# defeats the witness role entirely (see docs/e2ee/KT.md §7).
+# There is no RustSec/OSV advisory — an OSV query for `akd` returns {} — so
+# `cargo audit` and `cargo deny advisories` will NOT catch a downgrade. This entry
+# is the floor. Do not remove it, and do not assume tooling is holding it.
+name = "akd"
+version = "<0.13.0"
+
+[[bans.deny]]
+name = "akd_core"
+version = "<0.13.0"
+
+[[bans.deny]]
+# akd_mysql's last release is 0.8.9 (2023-03-15) and it is not maintained.
+# The durable Database impl is ours (KT.md §11.2).
+name = "akd_mysql"
 
 # =============================================================================
 # Sources. Everything resolves to crates.io. Making that a rule rather than a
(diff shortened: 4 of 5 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Merging once `gate` reports** — `rs / gate` is already green. I ran the asymmetry test myself in a throwaway worktree rather than reading it off the body, because that asymmetry *is* the entire argument for the entry existing.

## The floor works, and it works for the stated reason

Scratch member pinning `akd = "0.12.0"`, resolved into `rs/Cargo.lock` (`Adding akd v0.12.0 (available: v0.13.0)`):

```console
$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml advisories
==> cargo-deny check advisories -- rs/.
advisories ok
EXIT=0

$ scripts/check-rust-deny.sh --root rs --config rs/deny.toml bans
==> cargo-deny check bans -- rs/.
    │         ━━━ banned here
    │         ━━━━━━━━ banned here
bans FAILED
EXIT=1
```

Then bumping the same member to `akd = "0.13"` (resolves `0.13.0`): `bans ok`, `EXIT=0`.

**Advisories green, bans red, at the same commit, on the same lockfile.** That is the demonstration `KT.md:1180-1184` needed in order to be true, and it's now on the record where the next reviewer will find it. A stale `0.12.0` pin is caught by exactly one thing in this repository, and this PR is that thing.

The comment at `rs/deny.toml:122-128` is doing the important work:

> There is no RustSec/OSV advisory — an OSV query for `akd` returns `{}` — so `cargo audit` and `cargo deny advisories` will **NOT** catch a downgrade. This entry is the floor. Do not remove it, and do not assume tooling is holding it.

That is precisely the sentence that stops a future reviewer deleting the entry as redundant, which is the failure mode it exists to survive. Keeping `akd_core` as a bare entry without repeating the prose, and giving `akd_mysql` its own one-line reason (unmaintained since 2023-03-15, our `Database` impl is the durable one), is the right amount of comment — enough to justify, not so much that it rots.

**No scratch member shipped** — `git ls-tree -r --name-only` on the PR head returns nothing matching `scratch`, and the baseline `check-rust-deny.sh --root rs --config rs/deny.toml` is clean at `EXIT=0`, "All 2 Rust crate(s)". I checked because a leftover fixture is the easy mistake here.

## The README correction is right, and better than what I asked for

`rs/README.md:143` now reads **"Three of the four discover crates; the toolchain check does not."** and `:163` points at #553 for making it discover. That replaces a sentence which stated the exact guarantee #553 exists to falsify — a new crate under `rs/crates/` goes red on fmt, clippy and deny, and **green** on the toolchain check, because `check-rust-toolchain.sh:381` greps a hardcoded `MANIFESTS` array plus explicit `--manifest` flags.

Correcting it toward the issue rather than toward a vaguer claim is the right instinct: the README now tells the next person where the gap is instead of assuring them there isn't one.

## AGENTS.md and #556 don't collide

Worth stating since I blocked #556 earlier for exactly this class of problem. #571's AGENTS.md change is a single table-cell edit at ~`:213` (`the three Cargo.toml manifests` → `every registered Cargo.toml manifest`, correct given `rs/` is now registered); #556 adds a new guardrail section at ~`:159`. Different regions, no shared enumerated list, no exact-length assertion over either. They merge cleanly in whichever order they land.

## The #341 rewrite

Citing `docs/architecture/CARGO-WORKSPACE.md` instead of the closed issue is the fix, and the argument does get stronger for it: the ADR's decided consolidation target is `wallet/`, not repo-wide, so a separate `rs/` root doesn't obstruct anything that was actually decided. Dropping the "if #341 later chooses one repo-wide workspace" hedge is right — it guarded against a scenario the accepted ADR rules out.

## Everything else

`rs/deny.toml`'s `[graph]` comment now sits with the directive it describes rather than above an unrelated feature knob. `rs/Cargo.toml` and `scripts/check-rust-toolchain.sh` changes are consistent with registering the new member. Docs-only otherwise, no workflows touched, and `zuuli.yml`'s `changes` policy step correctly left alone.

`Closes #565` and `Closes #566` both stand — #565's amended acceptance asked for the entries, the ADR/`akd#495` comment, and the mutation evidence, and all three are here.

skyl:
Merge-train hold: please queue this behind #556. The required-gate source-contract PR has already integrated current trunk twice and had two full hosted matrices superseded by unrelated rs/docs merges. Do not admin-merge #571 until #556 reaches exact-current-trunk green and lands; then integrate its squash normally and rerun this PR.

Issue #564, opened 2026-08-24
test(f2z-codec): no byte-level wire vectors — 11 of 20 encoding mutations survive all 69 tests


skyl:
## Problem

`f2z-codec` (landed in #560, merged as `241b714`) encodes all 33 `WIRE.md` structures **correctly** — I checked every one field-for-field against its spec line and found zero divergences. But **nothing in the repository asserts what any structure's bytes actually are**, so 11 of 20 real mutations to the wire format pass the entire 69-test suite green.

That includes changes that would silently desynchronise every other implementation, and changes to what Ed25519 signatures commit to.

## Mutation results

Each mutation applied to the crate source, then `cargo test --locked --all-targets && cargo test --locked --doc` (baseline 49 + 7 + 5 + 7 + 1 = 69, all green), then reverted.

| # | Mutation | Site | Result |
|---|---|---|---|
| M1 | `ReadResponse.messages`: `VecU24` → `VecU16` | `commands.rs:438` | **NOT CAUGHT** |
| M2 | `CreateQueueResponse`: swap `recv_addr` / `send_addr` | `commands.rs:396` | **NOT CAUGHT** |
| M3 | `HelloResponse`: swap `channel_binding_mode` / `transport_security` | `commands.rs:245` | **NOT CAUGHT** |
| M4 | `LABEL_BODY` `…/v1/body` → `…/v1/bodx` | `hash.rs:25` | **NOT CAUGHT** |
| M7 | `CommandTranscript`: swap `address` / `signer_key` | `transcript.rs:71` | **NOT CAUGHT** |
| M8 | `Capabilities.protocol_versions`: `VecU8` → `VecU16` | `commands.rs:590` | **NOT CAUGHT** |
| M10 | `SignedAuth`: swap `address` / `signer_key` | `frame.rs:424` | **NOT CAUGHT** |
| M13 | `LABEL_RELAY_ID` one byte | `hash.rs:29` | **NOT CAUGHT** |
| M17 | `PowStamp` digest: `counter.to_be_bytes()` → `to_le_bytes()` | `pow.rs:145` | **NOT CAUGHT** |
| M19 | `ChallengeResponse`: swap `challenge` / `expires_at_ms` | `commands.rs:329` | **NOT CAUGHT** |
| M21 | `H(label, x)` → `H(x, label)` in **both** `hash` and `hash2` | `hash.rs:62` | **NOT CAUGHT** |
[...]

PR #572, opened 2026-08-24
security(f2z-codec): stop Canonicalized's Debug dumping received bytes

### LINKED ISSUE #563: security(f2z-codec): Canonicalized's derived Debug dumps the received frame as plaintext bytes
## Problem

`rs/crates/f2z-codec/src/canonical.rs:41-45`:

```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct Canonicalized<T> {
    value: T,
    encoded: Vec<u8>,   // <- the derived Debug renders this
}
```

`Canonicalized` is the type **every receive path holds** — it wraps the decoded frame *and* the bytes it arrived as. The derived `Debug` prints `encoded` as a decimal byte list.

Measured with a probe test against a 1 KiB `APPEND` payload:

```
FIRST 200 CHARS: Canonicalized { value: AppendRequest { payload: Payload(<redacted; 1024 bytes>) },
                 encoded: [0, 4, 0, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, ...
LEN: 5223
panicked: Canonicalized Debug dumped the ciphertext bytes
```

5,223 characters of ciphertext from one debug-format of one frame.

## Why this is the crate's own documented hazard

`rs/crates/f2z-codec/tests/redaction.rs:14-17` explains the exact mechanism and credits the design with having closed it:

> `tls_codec`'s own byte vectors derive `Debug` and print `TlsByteVecU24 { vec: [222, 222, 222, …] }` — a complete dump that contains no hex at all. That is why every body and payload in this crate is a newtype rather than a bare `tls_codec` vector.

The newtypes do work. `Payload` and `Body` report length only (`types.rs:231, 305`), all nine fixed-width newtypes have hand-written `Debug` impls (`types.rs:95-99`, none derived), and there are no `Display` impls to leak through a second formatter. Note the probe output above: `Payload(<redacted; 1024 bytes>)` is correctly redacted **inside** `value`.

`Canonicalized` then reintroduces the identical leak one level up, holding a raw `Vec<u8>` with a derived `Debug`.

This defeats stated crate property #2 (`rs/crates/f2z-codec/src/lib.rs:17-19`): *"`--log-level trace` must not become a ciphertext archive."*

## Why the tests miss it

`tests/redaction.rs` is genuinely paranoid where it looks — it checks lowercase hex, uppercase hex, base64url, decimal byte lists, a 4-byte decimal prefix, and any 8+ character hex-looking run, and mutation-testing confirms it fires (making a fixed newtype's `Debug` print its inner bytes goes red on `types::tests::debug_never_renders_bytes`).

It simply **never formats a `Canonicalized`**. All 5 redaction tests pass. This is the coverage-gap form of the defect class: the guard is real, the fixture just doesn't contain the shape that would violate the invariant.

## Fix

Hand-write `Debug` for `Canonicalized`, matching the convention the rest of the crate already uses:

```rust
impl<T: fmt::Debug> fmt::Debug for Canonicalized<T> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("Canonicalized")
            .field("value", &self.value)
            .field("encoded", &format_args!("<redacted; {} bytes>", self.encoded.len()))
            .finish()
    }
}
```

## Acceptance

- `Canonicalized`'s `Debug` is hand-written and reports `encoded` by length only.
- `tests/redaction.rs` formats a `Canonicalized` wrapping a real payload and runs it through the existing detector battery — red before the fix, green after.
- Ideally: a test asserting no type in the crate's public API derives `Debug` while holding raw bytes, so the next wrapper doesn't repeat this. The newtype discipline is currently a convention; this is the second place it had to be applied by hand.

Found while reviewing #560 (merged as `241b714`).

skyl:
`Canonicalized` is the type **every receive path holds** — the decoded value plus the canonical bytes §5 hashes. Its derived `Debug` rendered the raw `Vec<u8>` as a list of decimal integers.

`Closes #563`

## Red before

Probe: format a `Canonicalized<AppendRequest>` wrapping a 1 KiB payload of `0xde`.

```
LEN: 5223
FIRST 200: Canonicalized { value: AppendRequest { payload: Payload(<redacted; 1024 bytes>) }, encoded: [0, 4, 0, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222, 222
```

5,223 characters of ciphertext from one debug-format of one frame. Note the payload is correctly redacted *inside* `value`; the leak is entirely the sibling field.

## A second defect: the battery could not see it

The first thing I did was add the new case to `tests/redaction.rs`'s detector battery and run it against unmodified `main`. **It passed.** The leak above is real and the battery did not fire.

Every decimal detector in `assert_no_leak` was anchored on the opening bracket — `decimal_list()` builds `"[222, 222, …]"`. A real dump is almost never the whole list: a wire buffer opens with a length prefix, so the secret starts partway through, as `[0, 4, 0, 222, 222, …`. The `[` never lines up. The detector was blind to precisely the shape it existed to catch, and the hex and base64url detectors are irrelevant to a decimal dump.

So this PR also unanchors it: `decimal_run()` matches the joined run without brackets, and the four-byte-prefix assertion uses that. The other five redaction tests still pass unchanged under the stricter detector.

Without this change the new test is ornamental — it goes green against the unfixed code.

## Green after

```
running 7 tests
test a_challenge_scope_is_an_address_and_redacts_but_operator_text_does_not ... ok
test every_command_carrying_a_secret_redacts_it ... ok
test every_opaque_newtype_redacts ... ok
test a_canonicalized_value_never_dumps_the_bytes_it_arrived_as ... ok
test a_payload_reports_its_length_and_nothing_else ... ok
test no_type_derives_debug_while_holding_raw_bytes ... ok
test a_whole_frame_renders_without_a_single_secret_byte ... ok

test result: ok. 7 passed; 0 failed
```

With the detector unanchored but `canonical.rs` untouched, the new test fails as it should:

```
Canonicalized<AppendRequest> leaked a decimal byte run: Canonicalized { value: AppendRequest {
payload: Payload(<redacted; 1024 bytes>) }, encoded: [0, 4, 0, 222, 222, 222, 222, 222, ...
```

## Mutation tests

Every guard here was reverted and confirmed to go red.

**A — revert `Debug` to derived on `Canonicalized`** (run against the final, formatted tree):

```
test a_canonicalized_value_never_dumps_the_bytes_it_arrived_as ... FAILED
test no_type_derives_debug_while_holding_raw_bytes ... FAILED
test result: FAILED. 5 passed; 2 failed
```

```
no_type_derives_debug_while_holding_raw_bytes:
  canonical.rs: `Canonicalized` derives Debug while holding `Vec<u8>` — a derived Debug
  renders that as a decimal byte dump.
```

Both guards fire independently.

**B — add a new byte-holding type with no fixture anywhere.** This is the case the fixtures structurally cannot cover.

```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct SessionTicket { pub issued_at_ms: u64, pub secret: [u8; 32] }
```

```
pow.rs: `SessionTicket` derives Debug while holding `[u8;` — ...
test result: FAILED. 0 passed; 1 failed
```

**C — the tuple-struct branch of the scanner.** No type in the crate currently exercises it, so I verified it is not dead code rather than assuming:

```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct RawTicket(pub alloc::vec::Vec<u8>);
```

```
pow.rs: `RawTicket` derives Debug while holding `Vec<u8>` — ...
test result: FAILED. 0 passed; 1 failed
```

All three mutations were reverted; the tree is clean.

## Sibling audit — every byte-holding type in the crate

Grepped every occurrence of `Vec<u8>`, `TlsByteVec*`, `[u8; N]` and `&[u8]` in a field position across all 11 source files.

| Type | File | Byte storage | `Debug` | Verdict |
|---|---|---|---|---|
| `Canonicalized<T>` | `canonical.rs` | `Vec<u8>` | **was derived** | **the defect — fixed here** |
| `QueueAddress`, `PublicKey`, `RelayId`, `ChannelBinding`, `Challenge`, `Digest` | `types.rs` | `[u8; 32]` | hand-written (macro, `types.rs:95`) | ok — `<redacted>` |
| `Signature` | `types.rs` | `[u8; 64]` | hand-written (same macro) | ok |
| `Nonce`, `Salt` | `types.rs` | `[u8; 16]` | hand-written (same macro) | ok |
| `Payload` | `types.rs:231` | `TlsByteVecU24` | hand-written | ok — length only |
| `Body` | `types.rs:305` | `TlsByteVecU24` | hand-written | ok — length only |
| `ShortBytes` | `types.rs:375` | `TlsByteVecU8` | hand-written | ok — printable ASCII escapes, non-printable redacts |
| `VecU8/U16/U24<T>` | `vec.rs:79` | `Vec<T>`, not bytes | hand-written, delegates to `T` | ok — elements are wire structs, each redacting |
| `PaddingBuckets` | `padding.rs:56` | `Vec<u32>` | derived | ok — published relay policy (§11.1), meant to be read |

`Canonicalized` was the **only** leak. No second fix was needed. Nine fixed-width newtypes, three variable-length ones, and the generic vectors are all correctly hand-written; the one remaining derived `Debug` over a collection holds public policy numbers, not bytes.

There are still no `Display` impls in the crate, so nothing leaks through a second formatter.

## Why I made it a control

The issue asks for this and I think it earns its place: the discipline had now been applied by hand twice, which makes it a convention rather than a control, and the failure mode is silence — a fixture-based guard only covers types someone remembered to write a fixture for. Mutation B is the whole argument: a new byte-holding type goes green forever under the fixtures and red immediately under the scan.

`no_type_derives_debug_while_holding_raw_bytes` walks `src/` (resolved from `CARGO_MANIFEST_DIR`, not the process CWD), finds every `#[derive(...)]` containing `Debug`, captures the struct or enum it applies to, and fails if the body mentions `Vec<u8>`, `[u8;`, `&[u8]`, `TlsByteVec` or `Box<[u8]>`.

It is a textual scan, so I want to be straight about the limits rather than oversell it:

- It handles multi-line derive lists (`PowParams`), attributes and doc comments between the derive and the item, braced and tuple/unit items, and strips line comments before counting braces so a brace in a field doc cannot desync the depth count.
- A syn-based scan would be more precise, but it would mean a dev-dependency and a lot of machinery for a crate of 11 rustfmt-formatted files. The textual scan catches the shape that actually occurs.
- The real risk with a scanner is that it silently stops matching and passes forever, so it is anchored: it asserts it found `RelayFrame` and more than 20 derive-`Debug` types, and it hard-fails if a `derive(Debug)` sits on an item it cannot parse rather than skipping it quietly. A parser regression is loud, not green.
- It checks all types, not only `pub` ones — a private type still renders inside a public `Debug`.

## Checks

| Check | Result |
|---|---|
| `cargo test --locked --all-targets` | 49 + 7 + 7 + 7 = **70** (was 49 + 7 + 5 + 7 = 68) |
| `cargo test --locked --doc` | 1 |
| total | **71** (was 69), +2 new tests |
| `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec` | ok |
| `scripts/check-rust-fmt.sh --root rs` | clean |
| `scripts/check-rust-clippy.sh --root rs` | clean |
| `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` | advisories/bans/licenses/sources ok |

The `Debug` impl uses `core::fmt` and `format_args!` only — no `std`, no `unsafe`, `no_std` + `alloc` intact, and the wasm build confirms it.

Scoped to this defect: #564 (byte-level wire vectors) is untouched, and `.github/workflows/zuuli.yml` is untouched.

### DIFF
--- a/rs/crates/f2z-codec/src/canonical.rs
+++ b/rs/crates/f2z-codec/src/canonical.rs
@@ -26,6 +26,7 @@
 //! a default all become the same loud, immediate, testable failure.
 
 use alloc::vec::Vec;
+use core::fmt;
 
 use tls_codec::{DeserializeBytes, SerializeBytes};
 
@@ -38,10 +39,32 @@ use crate::error::CodecError;
 /// agree. [`Canonicalized::bytes`] is therefore always safe to hash or sign:
 /// there is no path by which a caller can obtain the *received* bytes from
 /// here.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct Canonicalized<T> {
     value: T,
     encoded: Vec<u8>,
+}
+
+/// Hand-written, like every other `Debug` in this crate that stands between a
+/// log line and a ciphertext.
+///
+/// `value` renders through its own `Debug`, which the newtypes of
+/// [`crate::types`] have already redacted. `encoded` is a bare `Vec<u8>` — the
+/// entire received frame, payload included — and a derived `Debug` would print
+/// it as a list of decimal integers. A decimal dump is a dump: 1 KiB of
+/// ciphertext becomes 5 KiB of log. This is the crate's stated property #2, and
+/// `Canonicalized` is the type every receive path holds, so it is the single
+/// worst place in the crate to leave that derived.
+impl<T: fmt::Debug> fmt::Debug for Canonicalized<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Canonicalized")
+            .field("value", &self.value)
+            .field(
+                "encoded",
+                &format_args!("<redacted; {} bytes>", self.encoded.len()),
+            )
+            .finish()
+    }
 }
 
 impl<T> Canonicalized<T> {
--- a/rs/crates/f2z-codec/tests/redaction.rs
+++ b/rs/crates/f2z-codec/tests/redaction.rs
@@ -45,10 +45,21 @@ fn upper_hex(bytes: &[u8]) -> String {
     bytes.iter().map(|byte| format!("{byte:02X}")).collect()
 }
 
+/// `222, 222, 222, …` — the body of a derived `Debug` on a byte slice, with no
+/// surrounding brackets.
+///
+/// Unbracketed on purpose. A real dump is very often *not* the whole list: a
+/// wire buffer opens with a length prefix, so the secret starts partway through
+/// as `[0, 4, 0, 222, 222, …`. Anchoring the pattern on `[` makes the check
+/// miss exactly the shape it exists to catch.
+fn decimal_run(bytes: &[u8]) -> String {
+    let joined: Vec<String> = bytes.iter().map(|byte| byte.to_string()).collect();
+    joined.join(", ")
+}
+
 /// `[222, 222, 222, …]` — a derived `Debug` on a byte slice.
 fn decimal_list(bytes: &[u8]) -> String {
-    let joined: Vec<String> = bytes.iter().map(|byte| byte.to_string()).collect();
-    format!("[{}]", joined.join(", "))
+    format!("[{}]", decimal_run(bytes))
 }
 
 fn base64url(bytes: &[u8]) -> String {
@@ -117,10 +128,12 @@ fn assert_no_leak(label: &str, rendered: &str, secret: &[u8]) {
         "{label} leaked a decimal byte list: {rendered}"
     );
     // A prefix of a decimal dump is enough to convict: no correct output of this
-    // crate ever prints four consecutive byte values.
+    // crate ever prints four consecutive byte values. Matched without brackets,
+    // so a dump that begins after a length prefix — `[0, 4, 0, 222, 222, …` —
+    // is caught as readily as one that begins at the opening bracket.
     assert!(
-        !rendered.contains(&decimal_list(&secret[..secret.len().min(4)])),
-        "{label} leaked a decimal byte list prefix: {rendered}"
+        !rendered.contains(&decimal_run(&secret[..secret.len().min(4)])),
+        "{label} leaked a decimal byte run: {rendered}"
     );
     // Every byte of the secret is 0xde, so even a two-byte prefix would show as
     // a repeated `de`. Catch a partial dump too.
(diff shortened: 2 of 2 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I re-ran every mutation myself in a throwaway worktree rather than reading them off the body.

## The finding underneath the finding is the important one

#563 was a leak. What this PR also found is that **the redaction battery could not see it** — and that is the more valuable half.

`tests/redaction.rs` builds its decimal detector as `decimal_list()`, anchored on the opening bracket: `"[222, 222, 222, 222]"`. A real dump starts partway through a buffer, after the length prefix — `encoded: [0, 4, 0, 222, 222, …` — so the `[` never lines up with the secret's first byte and the detector slides straight past. The battery I described in #563 as "genuinely paranoid" was blind to precisely the shape it existed to catch.

I verified that directly. Re-anchoring the detector to the old bracket form **and** reverting `Canonicalized` to a derived `Debug`:

```
$ cargo test --locked --test redaction
test a_canonicalized_value_never_dumps_the_bytes_it_arrived_as ... ok
test result: FAILED. 6 passed; 1 failed
```

The leak test reports **ok** with the leak present. Without the `decimal_run()` unanchoring in this PR, the new test would have been ornamental — green against broken code, added to a file that already congratulated itself on catching this class. That is the defect pattern this repo keeps producing, caught before it shipped rather than after.

(The single failure in that run is the structural scan, which catches it by a different route — see below.)

## Mutation results, all mine

```
MUT A  revert Canonicalized's Debug to derived (and drop the hand-written impl)
  -> EXIT=101
     test a_canonicalized_value_never_dumps_the_bytes_it_arrived_as ... FAILED
     test no_type_derives_debug_while_holding_raw_bytes           ... FAILED
     Both guards fail INDEPENDENTLY — neither is carrying the other.

MUT B  old anchored detector + derived Debug
  -> leak test reports `ok`. The unanchoring is load-bearing, not cosmetic.

MUT C  add a brand-new byte-holding type with a derived Debug and no fixture
       (`pub struct SessionTicket { pub secret: [u8; 32] }`)
  -> EXIT=101
     canonical.rs: `SessionTicket` derives Debug while holding `[u8;` — a derived
     Debug renders that as a decimal byte dump. Hand-write Debug and report the
     field by length, as `Payload` and `Canonicalized` do.
```

MUT C is the one that matters for the long run. Fixtures only cover types someone remembered to write a fixture for, which is exactly how this got in — the newtype discipline was a convention, and this makes it a control. The error message tells the next person what to do instead of just what failed, which is the right shape.

I also checked the scanner isn't quietly inert: it is anchored (asserts it found `RelayFrame` and >20 types, hard-fails on an item it can't parse), so a textual-parse regression is loud rather than permanently green. Being a textual scan rather than `syn` is the right call for a `no_std` crate with two dependencies, and the PR states that limit plainly instead of overselling it.

## Sibling sweep — I agree with the conclusion

`Canonicalized` was the only leak. `PaddingBuckets`' derived `Debug` over `Vec<u32>` is correct to leave alone — that is published relay policy meant to be read, not received bytes. Still no `Display` impls anywhere, so there is no second formatter to leak through.

## Numbers

Tests **69 → 71** (49 + 7 + 7 + 7 + 1 doc), verified locally, matching CI. wasm32 still builds and the impl uses only `core::fmt` + `format_args!` — no `std` crept in via the new formatting code, which was the thing to watch on a `#![no_std]` crate. fmt, clippy and cargo-deny clean.

`#564` and `zuuli.yml` correctly untouched — the latter matters, since #556 is blocked on that exact-ordered-command assertion and a second PR touching it would have collided.

## Follow-up, not blocking

The detector fix raises a question worth a look when #564 is done: `decimal_list()` still exists and is still used elsewhere in the file. If any other assertion relies on the anchored form, it has the same blind spot. Worth a sweep — every place a detector anchors on a delimiter that a real dump would not begin with.

Issue #559, opened 2026-08-24
docs(e2ee): ADR 0013's audit-scope claim is disprovable — and the truth is a stronger argument


skyl:
## Problem

ADR 0013 (added in #557, merged as `ece2cc0`) makes one claim about the NCC audit's scope that is **disprovable from the audit PDF and the `akd` git history in about ten minutes** — in the one document whose entire value is that it doesn't assert what it can't support.

Everything else in the ADR checked out. I verified independently: the NCC audit metadata (Aug 2023, 3 consultants / 20 person-days, 40 pages, scope v0.9.0 commit `be1055e`, 1 Medium / 8 Low / 6 Info, fixed as of v0.11.0 commit `85b3b07`) is verbatim-correct; `facebook/akd#495`'s title, merge date `2026-08-12T17:45:01Z` and security characterization are verbatim-correct; `0.13.0` published `2026-08-13T16:29:07Z` is the first release containing it (`compare/2e99b03...v0.13.0` → `ahead_by: 1`); the OSV/RustSec/GHSA silence on `akd` is real across all four sources, so the load-bearing conclusion — that `cargo audit` and `cargo deny advisories` are genuinely silent on a stale `0.12.0` pin — holds. `AppendOnlyProof` really is O(entries added), not O(log n) (`akd_core/src/types/mod.rs:515-547`: `SingleAppendOnlyProof { inserted: Vec<AzksElement>, unchanged_nodes: Vec<AzksElement> }`, one per epoch, unbounded).

The one error is the audit-scope bullet.

Found while reviewing #557 (merged). Companion to #558.

## The claim

`decisions/0013:206-211` (and, in wire form, `KT.md:1050`):

> The audit does not cover `HistoryProofV2` / the batch-insert rewrite (where #495 lived).

Two separate problems.

### `HistoryProofV2` does not exist

Zero occurrences in the **entire git history** of `facebook/akd` — `git log --all -S 'HistoryProofV2'`, case-insensitive pickaxe, GitHub code search and commit search all return empty. The type is and always has been `HistoryProof`.

There *is* real post-audit work called "history proof v2": commit `6998ca6`, titled *"PR433: add history proof v2 (#433)"*, **2024-05-24**, shipped in `0.12.0`, renaming `until_marker_vrf_proofs` → `past_marker_vrf_proofs` and friends. So the *substance* — that it postdates the Aug 2023 audit, by nine months — is right. The identifier is invented.

[...]

Issue #558, opened 2026-08-24
spec(e2ee): KT.md defects that would produce divergent implementations


skyl:
## Problem

`docs/e2ee/KT.md` (added in #557, merged as `ece2cc0`) is the most implementable spec in the set — but it carries defects of exactly the class #550 catalogued in `WIRE.md`: normative rules that contradict each other, encoded structures with no verification rule, and a MUST whose parameter is explicitly undecided. Two of them also contradict ADR 0014, which #557 amended to assert that *"nothing in this ADR changed to accommodate them."*

Found while reviewing #557 (merged). Related: #305, #311, #551.

## 1. P0 — the steady-state poll fails the document's own verification order

`KT.md:568-582` (§6.3): *"A `SignedTreeHead` is acceptable… **only if all of the following hold**"*, then

- rule 3: `epoch > last.epoch`
- rule 5: `published_at_ms > last.published_at_ms`
- rule 8: `If epoch == last.epoch: root_hash, tree_size and published_at_ms must all be identical.`

Under "all must hold", rule 3 makes rule 8's antecedent unreachable — any STH with `epoch == last.epoch` is already rejected by rules 3 and 5 before rule 8 is consulted. What depends on that:

- **The witness loop breaks.** `KT.md:659-663` (§7.1) polls `GET /kt/v1/sth` and applies §6.3. Between epochs — most polls, and *every* poll under §5.1's 600 s heartbeat — the log correctly returns the STH the witness already holds. §6.3 says it is not acceptable. §7.1 step 3 names only "rollback, fork, chain break, or a changed `vrf_public_key`" as halt triggers, and this is none of them. **A conforming witness's normal steady-state behaviour is undefined.**
- **The client case is worse.** `KT.md:861-863` (§8.1 step 2): *"Apply §6.3's monotonicity checks… **Any failure is fatal and is fork evidence**."* A client resolving two handles inside one epoch — entirely ordinary — receives the same STH twice and by the letter of the spec must file a `FaultReport`.

**Fix:** state that a re-presented identical STH is a no-op; scope rules 3/5 to the `epoch > last.epoch` branch and make rule 8 the explicit `epoch == last.epoch` branch.

## 2. P0 — KT.md removes ADR 0014's only user-facing mitigation, while the ADR claims nothing changed

[...]

Issue #551, opened 2026-08-23
spec(e2ee): ADR 0014 creates a platform reset authority under a heading disclaiming owner decisions


skyl:
## The decision that needs your explicit acceptance

`docs/e2ee/README.md` prefaces the Phase 1 decision table with:

> Not owner decisions — these are **invented in the specification**

But **ADR 0014 creates a platform-held offline reset authority key** that can take over any handle after a cooldown. That is a product-and-trust decision of the same weight as ADRs 0001-0007, not a wire encoding, and it currently sits under a heading disclaiming owner-decision status.

The ADR is admirably honest about what it is (`decisions/0014:90-98`):

> A platform-assisted recovery is **cryptographically indistinguishable from a platform-assisted attack**

That candour is exactly right, and it's why the decision belongs to you rather than to the spec.

**It also leaves the security-claims document wrong.** `docs/e2ee/THREAT-MODEL.md:71` still asserts:

> Cannot substitute an identity key undetectably

With a platform reset authority, that is no longer true as stated. The PR defers the fix to "when the KT documents land," which means the auditor-facing threat model currently describes a system without a platform reset authority. That annotation should not wait.

## Three concrete holes in ADR 0014, independent of the governance question

1. **Messages in flight are not addressed.** The ADR covers directory identity-key authorization and says nothing about existing MLS groups, queued ciphertext, or credentials issued under the old key at rotation time. `decisions/0014:73`'s phrasing — "clients keep using the old key and refuse to encrypt to the new one" — is also imprecise: in MLS you encrypt to a *group*, not to an identity key.

2. **`no_reset` has no enforcement rule.** `decisions/0014:79-82` offers the flag as "the strongest available posture," but the ADR never states that the log **MUST** reject a reset entry for a handle whose current entry sets it. Without that MUST, the flag is decoration — a modified relay simply ignores it.

3. **"Cancel the reset with a `RotationProof`"** (`:73`) — a `RotationProof` binds old→new; the cancel message is undefined. And `:72` says the cooldown is "(**proposed**: 7 days)", reintroducing exactly the placeholder language #545 existed to eliminate.

## Naming collision worth fixing while you're there

`WIRE.md` §7.5 "rotation" (queue keys) and ADR 0014 "rotation" (identity keys) are different operations sharing a word, with no cross-reference between them.

## Acceptance

- The platform reset authority is either explicitly accepted by the owner and moved into the owner-decision set, or removed.
[...]

PR #573, opened 2026-08-24
docs(e2ee): replace §14 handle-charset unknowns with measured production facts

skyl:
`WIRE.md` §14 specified the messaging handle charset and flagged three things as MUST-confirm-before-relying-on, because they could not be determined from this public repository. **All three have now been determined** — the username validator from the `Creator` model's migration history, and the eligibility and collision measurements against the production database on **2026-08-24**. This PR replaces the uncertainty with the facts and fixes what the facts contradict.

Docs only. **No code, crates, or workflows.**

## On what is and is not published here

`zuu` is public and public history is a one-way ratchet, so **free2z's absolute account metrics are deliberately not in this diff, this description, or the commit message.** Account totals are business data, not protocol data. Findings are given as **proportions and categories**, which is the form the design decision actually rests on — what matters is *what fraction* of existing accounts the rule excludes, not how many accounts exist. Note also that a percentage plus any single absolute count reconstructs a total exactly, so neither form is used anywhere. The precise figures are recorded in the deployment repository. No username or other user-identifying value appears anywhere.

### `WIRE.md` §14.2 — now a measurement, not a conditional

- The validator is **`UnicodeUsernameValidator`** (`Creator(AbstractUser)` does not override `username`). So **non-ASCII usernames are legal on free2z today, and a small but non-zero number of accounts carry one.** The old conditional ("*if* the Unicode validator is in use, the platform's handle space already contains the homograph attack surface") is now unconditional and stated in the present tense as a finding about the shipped product.
- Adds the measured eligibility proportions: **~90% of existing accounts eligible after case-folding, ~10% not**, with a qualitative prevalence table showing that punctuation (`.@+-`) is the dominant cause of ineligibility, non-ASCII and over-length names are rare, and uppercase — the most common deviation of all — costs nobody a handle because the mapping folds case.
- The material point: **every ineligible account has logged in at least once, and all but one is still active.** Real users, not abandoned rows.

### `WIRE.md` §14.3 — the case-folding argument was false; dated correction added

§14.3 argued that folding case *cannot* collide two distinct existing accounts **because** platform uniqueness is enforced case-insensitively, and correctly caveated that this holds only if the property is real in the database.

**It is not real, and the query settles it the other way.** The column has a plain case-**sensitive** `unique=True`; no migration adds a `Lower("username")` functional index, a `UniqueConstraint`, or `citext`. The check exists only in `RegistrationSerializer.validate_username` and `CreatorProfileUpdateSerializer.validate_username`, so any path bypassing them (admin, data migration, management command, social login) can create a case-variant duplicate — **and production holds such duplicates today, in more than one group.** The existence is the finding; a single colliding pair is enough to falsify the argument, so the count adds nothing and is not published.

Handled with a dated correction block in the style of `ARCHITECTURE.md` §4.9 / §5.4 and ADR 0007 — **the original claim is left standing with a pointer, not silently rewritten**. The *decision* (fold case) survives; the claim that it is free does not. Every colliding account must be resolved by hand before the mapping ships, and the clean end state is a DB-level `UniqueConstraint(Lower("username"))`, tracked in the deployment repository.

### `ARCHITECTURE.md` §13-O — measured, and **still open**

Restructured with the measurement. Two sub-questions are marked **closed** (what fraction; which validator). The question itself **stays open**, and the bullet says why: the *policy* question is unanswered and is a product decision — is that ~10% of real users simply excluded from messaging discovery in v1, or served by a separate opt-in messaging handle (§13-K)? A fourth sub-bullet records the case-collision blocker surfaced by the measurement.

### `THREAT-MODEL.md` §4.10 — new known limit (consistency)

The threat model did not mention homographs at all, which now reads as an omission: the surface is present in the shipped product. New §4.10 states it in the present tense, notes the messaging charset makes the attack not exist *inside* the messenger, and is explicit that this **does not fix the platform** — a homograph username can still appear on a profile page or comment byline. The mitigation is narrow and stated exactly (a non-conforming string does not resolve, so silent impersonation becomes a lookup failure), and no property of free2z's own username space is claimed.

## Checks

- `scripts/check-public-repo-boundary.sh` passes; private tracking is referred to obliquely ("the deployment repository"), no private repo named, no private issue linked.
- All new cross-document anchors verified against their target headings.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -1148,14 +1148,37 @@ deliberate.
   Messaging handles are `[a-z0-9_]{1,30}`, ASCII, so that homograph and
   mixed-script impersonation does not exist. free2z usernames are broader
   (`^[\w.@+-]+$`, up to 150 characters), so some existing accounts are not
-  eligible. **How many is not known** — this repository holds no user data — and
-  it must be measured before the rule ships, because it is the difference between
-  a rounding error and a migration project. Two sub-questions are also open:
-  whether the `Creator` model uses Django's Unicode or ASCII username validator
-  (undeterminable from this repository, and it decides whether the platform's
-  existing handle space already contains the homograph surface); and whether the
-  excluded population is served by a separate opt-in messaging handle, which is
-  directory design and touches §13-K.
+  eligible. **Measured 2026-08-24 — and still open**, because measurement was
+  never the whole question. What the measurement settled:
+  - **How many, closed.** **Approximately 90% of existing accounts are eligible
+    after case-folding and approximately 10% are not** — overwhelmingly because
+    the username contains `.` `@` `+` or `-`; non-ASCII characters and
+    over-length names account for very few. It is neither the rounding error nor
+    the migration project the question was framed around, and it is not a set of
+    dead rows: **every ineligible account has logged in at least once and all but
+    one is still active.** Absolute counts are business data and are deliberately
+    not published in this repository; they are recorded in the deployment
+    repository.
+  - **Which validator, closed.** `Creator(AbstractUser)` does not override
+    `username`; the field is recorded with
+    `UnicodeUsernameValidator`. **Non-ASCII usernames are legal on free2z today
+    and a small, non-zero number of accounts carry one** — so the platform's
+    existing handle space already contains the homograph surface, in the present
+    tense, as a property of the shipped product rather than a risk of a future one
+    ([`WIRE.md` §14.2](./WIRE.md#142-checked-against-free2zs-real-username-rules--measured),
+    [`THREAT-MODEL.md` §4.10](./THREAT-MODEL.md#410-the-platform-username-space-contains-homographs-messaging-handles-do-not)).
+  - **Still open — and it is a product decision, not a measurement.** Is that
+    ~10% of real users simply excluded from messaging discovery in v1, or are
+    they served by a separate opt-in messaging handle? The latter is directory
+    design and touches §13-K. Nobody has decided, so this stays open.
+  - **Surfaced by the measurement, and blocking.** Case-insensitive username
+    uniqueness is enforced only in two serializers, not in the database, and
+    **production holds case-variant duplicate accounts today, in more than one
+    group**. Those must
+    be resolved before the handle mapping ships; the correction and the intended
+    end state (a database-level `UniqueConstraint(Lower("username"))`) are in
+    [`WIRE.md` §14.3](./WIRE.md#143-the-decision-and-the-cost-accepted). The
+    backend work is tracked in the deployment repository.
 - **P. KT epoch cadence and maximum merge delay.** Opened 2026-08-23 by
   [`KT.md` §5](./KT.md#5-epochs-and-the-merge-promise). The values there — a
   600-second epoch published whether or not there is work, and a 3,600-second
--- a/docs/e2ee/THREAT-MODEL.md
+++ b/docs/e2ee/THREAT-MODEL.md
@@ -555,6 +555,36 @@ read, ACK or delete, and a sender-side key that cannot drain the queue. It is
 simply not unlinkability, and the earlier text in this document that said
 otherwise was wrong.
 
+### 4.10 The platform username space contains homographs; messaging handles do not
+
+Stated in the present tense because it is a property of the product as it ships
+today, not a risk of a future one. free2z usernames are validated by Django's
+`UnicodeUsernameValidator`, confirmed against the model's migration history, so
+non-ASCII usernames are legal and — measured on production 2026-08-24 — **a
+small, non-zero number of accounts carry one today**
+([`WIRE.md` §14.2](./WIRE.md#142-checked-against-free2zs-real-username-rules--measured)).
+For a threat model the count is beside the point; that it is not zero is the
+point.
+`@аlice` with a Cyrillic а and `@alice` with a Latin a are two different,
+equally valid, equally registerable free2z accounts right now. **The homograph
+attack surface in the platform's handle space is present, not hypothetical.**
+
+Messaging handles are `[a-z0-9_]{1,30}` compared as bytes
+([`WIRE.md` §14](./WIRE.md#14-handle-charset-for-messaging--blocking-pre-check-resolved)),
+so the KT directory's label space contains no confusable pairs and homograph
+impersonation **does not exist inside the messenger** — that is why the charset
+was restricted, and it is what makes the log's labels not homograph-attackable
+([`KT.md` §1.3](./KT.md#13-conventions)).
+
+What that does **not** do is fix the platform. A profile page, a comment byline
+or a link on free2z can still carry a homograph username, and a user who copies
+a handle from one of those surfaces into a contact search is relying on the
+platform's charset, not on ours. The mitigation is narrow and worth stating
+exactly: such a string does not match the messaging charset, so the directory
+refuses to resolve it at all, which turns a silent impersonation into a lookup
+failure. Making free2z's own username space safe is backend and product work,
+out of scope for this design, and **no property of it is claimed here**.
+
 ## 5. Summary: what we claim, precisely
 
 | Property | ZUULI (native) | Web (WASM) |
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -1735,35 +1735,85 @@ which? is `ﬁ` `fi`? is `ı` `i`?), and a charset with no case, no scripts and 
 confusable pairs makes an entire class of homograph and mixed-script impersonation
 **not exist** rather than be defended against.
 
-### 14.2 Checked against free2z's real username rules
+### 14.2 Checked against free2z's real username rules — measured
 
 **This was checked before being written down**, and the answer is that real
-usernames are considerably broader.
+usernames are considerably broader. The first revision of this section flagged
+three things it could not settle from this repository — which username validator
+is attached, whether case-insensitive uniqueness is a database property, and how
+many existing accounts the rule excludes. **All three have since been measured**:
+the model and its migration history were read out of band, and the counts were
+run against the production database on **2026-08-24**. The findings below are
+facts, not conditionals.
+
+**On what this section reports.** No username or other user-identifying value
+appears here, and neither do free2z's absolute user metrics: this is a public
+repository, and account totals are business data rather than protocol data.
+Findings are given as **proportions and categories** — which is the form the
+design decision actually rests on, since what matters is *what fraction* of
+existing accounts the rule excludes, not how many accounts exist. The precise
+figures behind every proportion below are recorded in the deployment repository.
 
 | Source | Finding |
 |---|---|
 | `py/dj/free2z/openapi/f2z.yaml` — `Registration`, `CreatorDetail`, `CreatorList`, `CommentAuthor` | `username` is `type: string`, **`pattern: ^[\w.@+-]+$`**, **`maxLength: 150`**, described as *"Letters, digits and @/./+/-/_ only."* This is Django's `AbstractUser.username` contract. |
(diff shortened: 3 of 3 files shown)


skyl:
**Holding this one for an explicit owner decision before merge, on the disclosure question rather than the substance.** A full correctness review is running separately and I'll post it when it lands. This comment is only about what this PR makes public.

I checked the obvious worry first and it's clean: **no username or user-identifying value appears anywhere in the diff** — I grepped it. And the private-ish file citations aren't new. `py/dj/` and `ts/svelte/free2z/` are tracked *in this repo*, and `py/dj/apps/g12f/serializers/creator.py` was already cited at `WIRE.md:1747` on `main` from #548. So `scripts/check-public-repo-boundary.sh` passing is real, not a false negative.

What *is* new is a category the boundary script cannot see, because it isn't about paths.

## 1. A live, unpatched data-integrity weakness, described with its bypass paths

The §14.3 correction states that case-insensitive username uniqueness is enforced **only** in `RegistrationSerializer.validate_username` and `CreatorProfileUpdateSerializer.validate_username`, that the column carries a plain case-**sensitive** `unique=True`, that no migration adds `Lower("username")`, a `UniqueConstraint`, or `citext` — and then names the paths that bypass it: *admin, data migration, management command, social login*. It also confirms the weakness is already realised in production: **2 collision groups covering 4 accounts.**

That is an accurate and genuinely valuable finding. It is also, on a public repository, a description of how to create a case-variant duplicate of an existing account on a live platform — published **before** the `UniqueConstraint(Lower("username"))` that would close it exists. The PR is explicit that the constraint is still only "tracked in the deployment repository."

The disclosure is yours to make and I'm not going to pretend otherwise — it's your product and your repo, and the honesty is the point of these documents. But merging it is a decision, not a formality, and it shouldn't happen because a reviewer waved it through on CI being green.

The cheap de-risking, if you want the finding without the recipe: keep the *conclusion* (DB-level uniqueness is not enforced; 4 accounts already collide; folding is therefore not free) and drop the enumeration of bypass paths, which adds nothing a reader needs in order to act. Or land the constraint first and publish this as history.

## 2. Aggregate production figures

`1,749` total accounts, `173` ineligible, `172` still active, `8` non-ASCII, `2` collision groups. No individual is identifiable, and every number is load-bearing for the §13-O policy question — you can't argue "are these real users or abandoned rows?" without them.

Worth naming anyway: this publishes free2z's registered-account count. That's a commercial fact, not a security one, and it's irreversible in a way a doc edit isn't. If the argument survives on percentages (`9.9%`, `90.1%`) plus the activity finding, the absolute totals are optional. If you want the totals, keep them — just make it a choice.

## 3. The standing instruction I'm working under

You told me earlier: *"Don't bleed tuzi (private) into zuu (public) right now."* I've applied that to paths, issue numbers, and backend internals all night. Production database measurements are the same category by a different route — the facts were "determined out of band" precisely because they aren't in this repo — so I'd rather surface it and be told it's fine than quietly decide it is.

## What I'm not saying

I'm not saying don't merge. The measurement is the right work: it converts three MUST-confirm-before-relying-on caveats into facts, and it found a real blocker (the 4 accounts) that would otherwise have surfaced during a migration. The §14.3 correction reversing the PR's own earlier argument, rather than quietly dropping it, is the honesty pattern working.

**Give me a yes on the disclosure and I'll merge on green.** If you'd rather trim, say which of the two items above and I'll dispatch it.

One thing that stands regardless: with `UnicodeUsernameValidator` now **confirmed** in use and non-ASCII usernames confirmed to exist, #550's item 4 gets sharper, not softer — the U+212A KELVIN → `k` homograph path is no longer hypothetical on this platform. Whether this PR closes that item, narrows it, or leaves it open is what the correctness review is determining.

skyl:
Noting for the record: this merged at `02:14:52Z`, about a minute after I posted the disclosure question above, so that question went unanswered rather than answered. Not a complaint — it's your repo and your call, and I'd have merged it too on a yes. But I'd rather say plainly that the decision was made by the merge than let it look like it was cleared on review.

**One consequence worth acting on.** The weakness described in §14.3 is now public and permanent in git history, so trimming the text later recovers little. What *does* change is the ordering: the `UniqueConstraint(Lower("username"))` and the resolution of the 4 colliding accounts were "tracked in the deployment repository" as pre-migration cleanup, and they're now the fix for a publicly documented gap. That moves them from *before the handle mapping ships* to *soon*, independently of the messaging timeline.

Everything in this comment is already in the merged document, so nothing here adds exposure.

**What I'd suggest, in order:**

1. Land the DB constraint and resolve the 4 accounts. Once that's done, `WIRE.md` §14.3's correction block should get a second dated line saying the gap is closed — the correction pattern works in both directions, and a reader arriving later should find the resolution attached to the finding rather than having to go looking.
2. `ARCHITECTURE.md` §13-O stays open on the policy question, which is right — ~173 real users (172 still active) either lose messaging discovery in v1 or get a separate opt-in handle per §13-K. That's a product decision and nothing in this PR pretends otherwise. Good.
3. The measurement made **#550 item 4 sharper, not softer**, and that should be reflected there: with `UnicodeUsernameValidator` now confirmed in use and 8 non-ASCII usernames confirmed to exist, the U+212A KELVIN → `k` path is reachable on the live platform rather than hypothetical. I'll update #550 once the correctness review lands, since that review is establishing whether §14 now names the case-mapping function normatively — which is the half of item 4 the measurement doesn't answer by itself.

A full correctness review is still running and I'll post it here when it does, even though the PR is merged — same as #557, where the findings became #558 and #559. If it turns up anything material I'll convert it to issues rather than comments, since there's no longer a PR to change.

For what it's worth, the substance of the work is good: converting three MUST-confirm-before-relying-on caveats into measurements is exactly right, and the §14.3 block reversing this project's own earlier argument — leaving the wrong claim standing with a pointer instead of quietly deleting it — is the honesty pattern doing its job.

skyl:
Correctness review, as promised — posting it here even though this is merged, same as #557. The material findings become issues below.

Reviewed head `6851473`. **The craft is sound and every in-repo claim checks out. But this must not be read as closing #550 item 4 — and the measurement makes that defect demonstrably worse, not better.**

First, credit where it's due: **the counts were redacted to proportions** before merge, and I verified the redaction is faithful rather than lossy — ~90%←90.1%, ~10%←9.9%, "a little over half"←54.3%, "just under a tenth"←9.43%, "overlap in only a single account"←174 memberships/173 distinct, "all but one still active"←172/173. No absolute count leaks anywhere in the tree at this head, and no paraphrase is misleading. The prior head's arithmetic also reconciled exactly (1,576+173=1,749; 165+8+1=174 over 173 distinct → exactly one double), so nothing was quoted from a different query than the one described.

## The blocking finding: #550 item 4 is untouched, and this PR makes it live

§14.2 now asserts **unconditionally** that `UnicodeUsernameValidator` is in use (`WIRE.md:1766`, `ARCHITECTURE.md:1162`). Checked against that validator's actual regex:

```
KELVIN SIGN (U+212A)
  validator matches ^[\w.@+-]+\Z         : True
  'K'.lower() -> 'k'   'K'.casefold() -> 'k'
  eligible under WIRE.md:1824
    lowercase(username) matches /^[a-z0-9_]{1,30}$/ : True
```

A platform account `K` (U+212A) is **registerable and eligible** for messaging handle `k`, colliding with account `k`. And this PR's own §14.3 correction is what makes it reachable end-to-end: it establishes that case-insensitivity is enforced only in two serializers, that admin / data migration / management command / social login bypass them, and that production **already contains rows created that way**. The KELVIN path travels those exact bypasses.

That is the same "two distinct accounts map to one messaging handle, the directory has to pick a winner" failure this PR itself calls **blocking** at `WIRE.md:1868-1872` — except this instance is invisible to the measurement that found the other one, because an `iexact`/`LOWER()` duplicate-group query is not the same predicate as "lowercase collides after ASCII filtering." Nothing in the PR mentions it.

`grep` across `docs/e2ee/` at this head: no occurrence of `casefold`, `str.lower`, `to_lowercase`, or `U+212A`, and no normative definition of `lowercase()`.

## Two more, in the same family

**The document now carries two inconsistent normative definitions of eligibility.** `WIRE.md:1824` is post-mapping (`lowercase(username) matches /^[a-z0-9_]{1,30}$/`); `:1803` and `:1884-1886` are raw-string for the non-ASCII clause. For the KELVIN class they disagree — `:1824` says eligible, `:1803`/`:1885` say ineligible. That is exactly the contradiction #550 cited by line number in the frozen text, surviving verbatim.

**`lowercase()` and "case-folding" are used interchangeably for the same normative step**, and in Unicode they are different functions on precisely the characters at issue:

```
'ſ' (U+017F)  .lower() -> 'ſ'  (ineligible)   .casefold() -> 's'   (eligible)
'ẞ' (U+1E9E)  .lower() -> 'ß'  (ineligible)   .casefold() -> 'ss'  (eligible)
```

The eligibility split was computed by one of them and the doc doesn't say which, so the proportions aren't reproducible and two implementations will disagree.

## `THREAT-MODEL.md` §4.10 over-claims

`:574` says the KT directory's label space "contains **no confusable pairs**". `[a-z0-9_]` contains `1`/`l` and `0`/`o` — the canonical ASCII homoglyph pairs. What the charset removes is the **cross-script** class, which `WIRE.md:1735-1737` scopes carefully as *"an entire class of homograph and mixed-script impersonation"*; §4.10 drops the scoping and asserts the absolute, two paragraphs above the "what we claim, precisely" table. This is the #551/#558 pattern a third time.

Also `:586` says "no property of it is claimed here" while `:562-566` claims two properties of free2z's username space; presumably means no *safety* property, which in this document should be written. And §4.10 doesn't cross-reference the case-collision blocker this PR just found, despite being the directory-integrity section.

## Medium — the redaction removed the evidence for a conclusion that stayed

`ARCHITECTURE.md:1153-1161` and `WIRE.md:1890-1897` keep "neither a rounding error nor a migration project", but the earlier head's qualifier — "~10% of a **small user base**" — was deleted with the counts. A bare proportion can't distinguish those: 10% of 1,700 and 10% of 10,000,000 are different projects. Either restore an order-of-magnitude qualifier or drop the conclusion. Relatedly `ARCHITECTURE.md:1153` is headed **"How many, closed."** and then declines to say how many — "What share" would match the content.

## Minor

- **Date stamps are a day ahead of this repo's convention.** Everything is `2026-08-24` but the commit is `2026-08-23 22:05 -0400`. #557 set the precedent: merged `2026-08-24T00:31Z`, corrections dated `2026-08-23` — local date. `ARCHITECTURE.md:1146` reads "Opened 2026-08-23" three lines above a correction dated 2026-08-24 for the same evening's work.
- `WIRE.md:1803` breaks the table's unit — three rows give shares, that one gives "A small number of accounts", under a column headed "Prevalence".
- `WIRE.md:1806` "the punctuation rule is almost the whole of the exclusion" reads as *all* of it and quietly contradicts "and not zero" at `:1803`.
- `WIRE.md:1740-1743` says "*this section* flagged three things"; two of the three were flagged in §14.3.

## What I verified as correct

**The house-style correction is genuinely compliant** — the original argument stands at `:1827-1830` with the inline pointer, the original caveat paragraph is intact at `:1832-1839`, and the correction is the dated blockquote at `:1842`, matching `ARCHITECTURE.md:332`/`:342`/`:780` and `THREAT-MODEL.md:93`/`:115` exactly. Nothing silently rewritten. That was the property I most wanted checked, and it holds.

**§13-O stays honestly open** — `ARCHITECTURE.md:1152` "Measured — and still open", the policy sub-bullet ends "Nobody has decided, so this stays open", and the case-collision blocker is recorded as a fourth, blocking sub-bullet.

**All anchors resolve** — every relative link and `#anchor` across `docs/e2ee/` at this head: 0 failures, including the four new cross-document links.

**Every in-repo claim in the §14.2 table is accurate**, checked file by file: `settings.py:29`, the absence of a model or migrations under `py/dj/apps/g12f/`, both `validate_username` methods with `username__iexact`, `FORBIDDEN_USERNAMES:32`, `views/creator.py:35,127`, and the Svelte `USERNAME_PATTERN`. No contradiction with `KT.md:90-92`.

**Explicitly unverifiable from this repo**, and correctly marked "Determined out of band" in the doc: the `Creator` model and its migration, the absence of a `Lower("username")` constraint, and every production figure.

## Issues filed

#550 updated with the KELVIN evidence and the two eligibility contradictions. The `THREAT-MODEL.md` §4.10 over-claim and the post-redaction items are filed separately.

PR #574, opened 2026-08-24
docs(e2ee): add CLIENT-CONTRACT.md — the frontend interface for E2EE messaging

skyl:
## What this is

`docs/e2ee/CLIENT-CONTRACT.md` — the document a ZUULI frontend dev, a web dev,
or an agent implements against **without reading any Rust**.

The E2EE design is fully specified across `ARCHITECTURE.md`, `THREAT-MODEL.md`,
`WIRE.md`, `KT.md` and ADRs 0001–0014. All of it is written for a protocol
implementer or a security auditor. None of it tells a frontend developer what
to call, what comes back, what the events are, or what they must not do — so
frontend work cannot start until the backend lands. This document is the
interface, published ahead of the implementation, so the two can proceed in
parallel.

**Docs only.** No TypeScript, no Rust, no crates. It decides nothing about the
protocol: every property is a restatement of the documents above, cited at the
point of use.

## What it covers

- **§1 Scope and trust statement.** ZUULI native gets the full guarantees; the
  browser does not — against a coerced server serving targeted code there is
  **no** defense (`THREAT-MODEL.md` §3.6, ADR 0001).
- **§2 Where each piece lives**, including why **enrollment is an app-crate
  command, not a plugin command**: it needs the wallet seed, and the seed must
  never cross IPC into a garbage-collected JS heap that nothing can zeroize.
  Precedent is `src-tauri/src/oauth.rs`.
- **§3 The command surface.** `plugin:f2zmsg|<cmd>`, payloads nested under a
  single `args` key, camelCase both sides — mirroring `src/lib/wallet/bridge.ts`
  and `tauri-plugin-zcash` exactly. Engine lifecycle, conversations and first
  contact, sending, message listing, delivery state and receipts, gap repair,
  local retention, ephemeral hints, purge requests, directory lookup, safety
  numbers, self-audit and alarms, device info, relays and witnesses. Plus the
  four-places permission/capability rule the Zcash plugin already enforces.
- **§4 The TypeScript contract** — `types.ts` / `bridge.ts` / `mock.ts` /
  `events.ts` — and **§4.1 the mechanical mock-covers-bridge parity test the
  wallet contract lacks.** Today only `tsc --noEmit` guards that boundary, and
  `invoke<T>()` is an unchecked cast, so a mock returning the wrong *values*
  and a backend renaming a field both ship silently. The proposal: schemas as
  the single source of truth in `types.ts` with types derived from them, a
  `RESULTS` manifest in `bridge.ts` asserted key-for-key against the bridge's
  own methods, a test that drives every mock method and parses the result, and
  the same schemas run at the real boundary under `import.meta.env.DEV`.
- **§5 Events**, their payloads and ordering guarantees. This is the app's
  **first real `listen()` consumer** — `zcash://sync-progress` has zero
  listeners in ZUULI today and the app polls. **§5.4 requires mock mode to
  drive inbound traffic via an echo peer**, with a `zuuli.mock.f2zmsg-scenario`
  switch mirroring `zuuli.mock.wallet-scenario` and a 15-row scenario matrix —
  otherwise the inbox, transcript, delivery indicators, gap marker, alarm
  banner and contact-request flow all get built blind.
- **§6 State machines.** Engine state, and the four delivery states with a
  table of what each is *actually* evidence of and what it is explicitly not.
  Conflating them is how delete-on-ack loses messages, and single-device v1 is
  exactly the coincidence that tempts the collapse.
- **§7 Ordering.** The DAG's partial order, total-ordered by
  `(epoch, sender_leaf_index, msg_id)`. `sent_at` is advisory and never orders,
  filters, dedups or decides anything.
- **§8 A closed `ErrorCode` union** — 30 members — with, per code, what the UI
  must do and whether it is retryable.
- **§9 THE RULES YOU MUST NOT BREAK.** Never ACK before the durable write.
  Never use `sent_at` for anything load-bearing. Never claim web parity, never
  call a hint enforcement, never call a purge a deletion. Never make the
  key-change alarm dismissible and never use a toast (`sonner` auto-dismisses).
  Never proceed silently below *t* independent witnesses. Never copy
  `tauri-plugin-zcash`'s `on_event` blur handler — it clears the seed on
  `Focused(false)` and desktop windows lose focus constantly. Never render a
  gap as a silent hole. Never offer FROST/DKG in a browser.
- **§10 What is deliberately not in v1**, each with its ADR or open-question
  letter: multi-device registration, ceremonies, attachments, push, Nostr
  fallback, groups > 2, rename/transfer, gossip, WebRTC transport.
- **§11 The browser client.** IndexedDB is evictable and
  `navigator.storage.persist()` may be refused, so **a client that cannot
  promise durability must not ACK** — a specified no-ACK mode, because
  ACK-then-evict loses the message from every copy that ever existed. Plus the
  `[a-z0-9_]{1,30}` messaging-handle charset and the first-class
  ineligible-account state it forces.
- **§12 What is provisional, and what is open.**

## On honesty about what is not settled

Per the brief, shapes the backend has not settled are marked **PROVISIONAL**
rather than invented, and §12.1 tables them. Where the temptation to be precise
was declined:

- **`WitnessConfig.independent`.** Marked provisional: the rule that computes it
  does not exist ([§13-Q](https://github.com/free2z/zuu/blob/main/docs/e2ee/ARCHITECTURE.md#13-open-questions)).
  It is in the contract now so nobody builds a screen assuming the *configured*
  count is the meaningful one.
- **`ContactRequest.bodyPreview`.** May become `null` always; showing any part
  of an unsolicited first-contact payload is an unanswered safety question.
- **Retry budgets, backoff constants, enrollment merge latency.** Named in the
  protocol docs without values.

### Rebased onto #573 — and one thing this branch got wrong, corrected

#573 landed while this was being written and replaced `WIRE.md` §14's unknowns
with **measurements taken against production on 2026-08-24**. The first commit
here was written against the previous state and said the ineligible-account
population "has not been measured", treating ~10% as an unverified planning
figure. That was true when written and is now false. The second commit fixes it,
because a contract that is stale about its own cited source is exactly the
failure this document claims to avoid.

§11.3 now carries the measured facts and the two client consequences they force:

- **~90% eligible after case-folding, ~10% not** — overwhelmingly `.@+-`;
  non-ASCII and over-length very few; uppercase disqualifies nobody. **Every
  ineligible account has logged in at least once and all but one is still
  active.** That is roughly a tenth of real, current users, so the ineligible
  screen is specified as a state a tenth of users will see.
- **It is not a temporary gap awaiting a migration.** Whether that ~10% is
  excluded in v1 or served by a separate opt-in handle is an undecided *product*
  question (§13-O), so no copy may promise either outcome. §10's table now flags
  that row as the one that may still move before v1.
- **`lowercase(username)` is not a unique key.** Case-insensitive uniqueness
  lives in two serializers, not the database, and production holds case-variant
  duplicates today. So a client must never derive a messaging handle locally and
  must always ask `check_handle_eligibility` — a real API rule that falls out of
  a database fact.
- **`THREAT-MODEL.md` §4.10 is recorded**: the platform's own username space
  already contains the homograph surface, in the present tense. The messaging
  charset does not fix that. Its mitigation is narrow — a homograph string fails
  to resolve — which means a user pasting a handle copied from a free2z profile
  page can get a lookup failure for a username that visibly exists, and the UI
  must explain that rather than look broken.
- `HandleEligibility.reason` split into `punctuation` / `non-ascii` /
  `too-long`, because the causes differ by orders of magnitude in prevalence and
  one "invalid handle" string would be accurate and useless.

Absolute counts are deliberately absent here, as in #573: this is a public
repository and account totals are business data.

## Proposed follow-up: `scripts/check-client-contract.mjs`

Not written here, because a checker with nothing to check gets deleted — but
proposed in §12.3 and worth filing as an issue against the plugin's landing:

`scripts/check-client-contract.mjs` should extend the mechanical drift check
that `scripts/check-zcash-permissions.mjs` already performs, to this document.
Sketch:

1. Parse `docs/e2ee/CLIENT-CONTRACT.md` §3's command tables — every row's first
   column is a backticked command name — into a set. Fail if the parse yields
   zero commands (so a heading rename cannot silently disable the check).
2. Parse `wallet/plugins/tauri-plugin-f2zmsg/build.rs`'s
   `const COMMANDS: &[&str]` into a set, exactly as `check-zcash-permissions.mjs`
   already reads the Zcash plugin's list.
3. `assertSameSet(documented, declared, "client contract command table")` —
   reusing that script's existing bidirectional missing/extra reporter, so a
   command added to the plugin without a doc row, **and** a doc row for a
   command that does not exist, both fail.
4. Assert the same set against `src/lib.rs`'s `invoke_handler`,
   `permissions/default.toml`'s `allow-*` entries, and
   `capabilities/default.json` — which is the check that already exists for
   `zcash` and needs only a second plugin id.
5. Assert §2.2's enrollment commands are **absent** from `COMMANDS` — they are
   app-crate commands and must never become plugin commands, which is the one
   invariant in this document that a well-meaning refactor would break first.

Wire it into the `zuuli.yml` gate's changed-path filter alongside
`scripts/check-zcash-permissions.mjs`, adding `docs/e2ee/CLIENT-CONTRACT.md`
and `wallet/plugins/tauri-plugin-f2zmsg/*` to the trigger list.

## Also

- `docs/e2ee/README.md`: contents-table row, a Phase 2 bullet, and a Status
  note stating that the interface this document describes **does not exist
  yet** — there is no `tauri-plugin-f2zmsg` and no frontend messaging module.

## Verification

- `scripts/check-public-repo-boundary.sh` — passes.
- All 57 cross-document anchor links and every relative file link resolved
  against the actual headings in `ARCHITECTURE.md`, `THREAT-MODEL.md`,
  `WIRE.md`, `KT.md` and `decisions/`.

### DIFF
--- a/docs/e2ee/CLIENT-CONTRACT.md
+++ b/docs/e2ee/CLIENT-CONTRACT.md
@@ -274,11 +274,18 @@ interface HandleEligibility {
   candidate: string | null;        // lowercase(username), if it matches
   reason:
     | null
-    | "charset"                    // contains . @ + - or non-ASCII
+    | "punctuation"                // contains . @ + or - — the dominant cause (§11.3)
+    | "non-ascii"                  // contains a non-ASCII character
     | "too-long"                   // > 30 chars after lowercasing
     | "not-signed-in";
 }
 ```
+
+`reason` distinguishes the causes because they are wildly different in
+prevalence and the UI should say the true specific thing: punctuation accounts
+for almost the whole of the ineligible population, non-ASCII and over-length for
+very little (§11.3). A single "invalid handle" string would be accurate and
+useless.
 
 `f2zmsg_unenroll` takes a typed confirmation string, in the shape
 `discard_unrecoverable_send` already uses (`"I CHECKED WALLET HISTORY"`), because
@@ -1327,7 +1334,7 @@ should be designed around speculatively.
 | **Groups larger than 2** | 1:1 is one MLS group like any other, so the model generalizes; fan-out is O(members × devices) and the scale limit is unanswered ([§13-M](./ARCHITECTURE.md#13-open-questions)). `Conversation` deliberately carries `peerHandle` rather than a member list — that is the v1 shape, and it will change. |
 | **Handle rename and transfer** | A change of *owner* rather than of *key*, creating a MITM window of a different shape. Still open ([§13-K](./ARCHITECTURE.md#13-open-questions)); [ADR 0014](./decisions/0014-directory-key-rotation.md) settles key rotation only. |
 | **Client gossip** | The only anti-equivocation that functions before independent witnesses exist, and **its wire format does not exist** ([`KT.md` §8.4](./KT.md#84-gossip), [§13-R](./ARCHITECTURE.md#13-open-questions)). No UI for it in v1. |
-| **A separate opt-in messaging handle** | The clean answer for accounts excluded by the charset rule, deferred as directory design ([`WIRE.md` §14.4](./WIRE.md#144-alternatives-rejected)). §11.3 is what v1 does instead. |
+| **A separate opt-in messaging handle** | The clean answer for the ~10% of real, active accounts the charset rule excludes (§11.3), deferred as directory design ([`WIRE.md` §14.4](./WIRE.md#144-alternatives-rejected)). Whether v1 simply excludes them or serves them this way is **an undecided product question**, not a settled omission ([§13-O](./ARCHITECTURE.md#13-open-questions)) — the one row in this table that may still move before v1. |
 | **P2P (WebRTC) transport for messages** | An optimization, not a dependency ([`ARCHITECTURE.md` §10](./ARCHITECTURE.md#10-p2p-webrtc--an-optimization-not-a-dependency)). Not in the v1 command surface. |
 
 ---
@@ -1422,30 +1429,73 @@ characters — so **existing accounts containing `.`, `@`, `+`, `-`, any non-ASC
 character, or exceeding 30 characters after lowercasing are not eligible for a
 messaging handle in v1.**
 
-> **How many accounts that is has not been measured.** `WIRE.md` §14.3 says so
-> explicitly and [§13-O](./ARCHITECTURE.md#13-open-questions) tracks it as a
-> blocking pre-check: the count must be measured before the rule ships, because
-> it is the difference between a rounding error and a migration project. A
-> working planning figure of *roughly ten percent* has been used in discussion;
-> **it is an estimate, not a measurement**, and no UI copy, metric, or capacity
-> plan may present it as a number we know. Design the ineligible state as though
-> it were common.
+**How many accounts that is, is now measured, and the answer is the reason this
+screen matters.** Measured against production on 2026-08-24
+([`WIRE.md` §14.2](./WIRE.md#142-checked-against-free2zs-real-username-rules--measured),
+[§13-O](./ARCHITECTURE.md#13-open-questions)):
+
+| Of all existing accounts | Share |
+|---|---:|
+| Eligible after case-folding | **~90%** |
+| **Ineligible** | **~10%** |
+
+Ineligibility is overwhelmingly caused by `.` `@` `+` or `-` in the username —
+just under a tenth of all accounts. Non-ASCII characters and over-length names
+account for very few, and uppercase disqualifies nobody because the mapping
+folds case. **Every ineligible account has logged in at least once and all but
+one is still active**: this is not a set of dead rows, it is roughly a tenth of
+real, current users. Design the ineligible state as a state a tenth of users
+will see, because it is one.
+
+Two things the frontend must not infer from that number:
+
+- **It is not a temporary gap awaiting a migration.** Whether that ~10% is
+  excluded from messaging discovery in v1 or served by a separate opt-in
+  messaging handle is an **undecided product question**
+  ([§13-O](./ARCHITECTURE.md#13-open-questions), §13-K). Do not write copy that
+  promises either outcome.
+- **`lowercase(username)` is not yet a unique key.** Case-insensitive username
+  uniqueness lives in two serializers and **not in the database**, and
+  production holds case-variant duplicate accounts today, in more than one
+  group ([`WIRE.md` §14.3](./WIRE.md#143-the-decision-and-the-cost-accepted)).
+  Those must be resolved before the mapping ships — backend work, tracked
+  elsewhere — but the client consequence is now: never derive a messaging handle
+  from a username locally, and never assume two accounts that fold to the same
+  string are the same account. Always ask `check_handle_eligibility`.
+
+Note also that the platform's *own* username space already contains the
+homograph surface — `UnicodeUsernameValidator` is the validator actually
+attached, non-ASCII usernames are legal on free2z today, and a small non-zero
+number of accounts carry one
+([`THREAT-MODEL.md` §4.10](./THREAT-MODEL.md#410-the-platform-username-space-contains-homographs-messaging-handles-do-not)).
+The messaging charset does not fix that and must not be presented as fixing it.
+Its mitigation is narrow and exact: a homograph string does not match the
+messaging charset, so the directory refuses to resolve it at all, turning a
+silent impersonation into a **lookup failure**. That means a user pasting a
+handle copied from a free2z profile page or a comment byline may get
+`handle-ineligible` or `handle-not-found` for a username that visibly exists —
+and the UI must explain that rather than look broken.
 
 So the ineligible state is **a first-class screen, not an error dialog**:
 
 - `check_handle_eligibility` is callable **before** enrollment and before the
   engine runs, so the UI can decide what to render without provoking a failure.
 - `EngineState: "ineligible"` renders an explanatory state that says what
-  *specifically* disqualifies the username (`reason` distinguishes `charset`
-  from `too-long`), and that the account works exactly as it does today —
+  *specifically* disqualifies the username (`reason` distinguishes
+  `punctuation` from `non-ascii` from `too-long`), and that the account works
+  exactly as it does today —
   it simply cannot be discovered by handle in the messenger yet.
 - It must **not** be apologetic-and-empty, must not offer a rename flow (handle
   rename and transfer are unsolved — [§13-K](./ARCHITECTURE.md#13-open-questions)),
-  and must not imply the restriction is temporary or arbitrary. Say what it buys:
-  nobody can register `@аlice` with a Cyrillic а.
-- Ineligible users must still be able to **receive** nothing and **see** nothing
-  broken: no empty inbox that looks like a bug, no perpetual spinner, no
-  notification badge.
+  and must not imply the restriction is temporary or arbitrary. Say what it buys,
+  precisely: `@аlice` with a Cyrillic а cannot exist as a **messaging handle**.
+  It can and does exist as a free2z username — that is exactly the surface the
+  charset excludes from the messenger, and overstating it into "nobody can
+  register that" is false (§11.3 above,
+  [`THREAT-MODEL.md` §4.10](./THREAT-MODEL.md#410-the-platform-username-space-contains-homographs-messaging-handles-do-not)).
+- Nothing about the ineligible state may look broken: no empty inbox that reads
+  as a bug, no perpetual spinner, no notification badge, no navigation entry
+  that leads somewhere dead.
 
 The mock's `ineligible-handle` scenario (§5.4) exists so this screen is built
 and reviewed rather than discovered at launch.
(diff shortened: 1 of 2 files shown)


skyl:
Merge-train hold: please queue this behind #556. The required-gate source-contract PR has already integrated current trunk twice and had two full hosted matrices superseded by unrelated rs/docs merges. Do not admin-merge #574 until #556 reaches exact-current-trunk green and lands; then integrate its squash normally and rerun this PR.

skyl:
Adversarial review, posted post-merge — findings become issues, per how this queue runs.

**The document is substantially sound and I want to say that first.** I checked its claims about existing code rather than taking the body's word, and they hold on every point except B2: the `zcash://sync-progress` orphan (6 emits in `sync.rs`, zero `listen()` consumers in `wallet/zuuli/`, `store/wallet.ts:105-130` polling at 6s/30s instead), the seed-clear on `WindowEvent::Focused(false)` at `lib.rs:104-127`, sonner's 4s default with no override, both mock localStorage keys, the `"I CHECKED WALLET HISTORY"` marker at `bridge.ts:220`/`send.rs:1966`, and the `args`-nesting convention at `bridge.ts:32-38`. All 12 `§13-<letter>` citations exist. `ErrorCode` type ↔ table parity is 29/29 with zero drift. Every relative link and anchor resolves. The PROVISIONAL marking is real and used honestly.

## Blocking — worth a small fast-follow rather than issues

### B1. §11.3 restates a claim `WIRE.md` formally retracted, and contradicts itself 32 lines later

`:1423-1425`:

> An account is eligible iff `lowercase(username)` matches the pattern; **case folding is safe because platform uniqueness is already enforced case-insensitively.**

That is verbatim the claim `WIRE.md:1827-1830` makes and that `WIRE.md:1843-1866` **explicitly retracts** in its dated correction — *"the case-folding argument above is false… Case-insensitive uniqueness is not a database property… Folding case *does* collide two distinct existing accounts."*

And `:1457-1460`, thirty-two lines below, says the opposite of `:1425`:

> **`lowercase(username)` is not yet a unique key.** Case-insensitive username uniqueness lives in two serializers and **not in the database**, and production holds case-variant duplicate accounts today.

So the document asserts P at `:1425` and ¬P at `:1457`, and P is the proposition its own cited source retired 24 hours earlier. The second commit went to real trouble folding #573's corrections into §11.3 and left the retracted sentence standing in the paragraph above the corrected one. A PR whose body argues *"a contract that is stale about its own cited source is exactly the failure this document claims to avoid"* shipped stale about its own cited source.

**One sentence to fix.** The leave-standing-with-a-dated-blockquote convention does *not* apply — the file is new, so there is no prior published claim to preserve. Just delete or invert `:1424-1425`.

### B2. §2.3 and §12.3 assert a mechanical check that does not exist

`:149-152` says every plugin command must appear in four places including `permissions/default.toml` and `capabilities/*.json`, enforced "**by the same mechanism `scripts/check-zcash-permissions.mjs` already enforces**".

That script **never opens `permissions/default.toml` and never opens any `capabilities/*.json`** — `grep -n "default.toml\|capabilities" scripts/check-zcash-permissions.mjs` returns nothing. What it enforces is a different six-way parity (`build.rs` `COMMANDS` ↔ `lib.rs` `invoke_handler` ↔ `permissions/autogenerated/commands/*.toml` ↔ `permissions/schemas/schema.json` ↔ `permissions/autogenerated/reference.md` ↔ zuuallet's tracked generated schemas, `:121-211`). `default.toml` is read only by `seed-capture-boundary.mjs:759`; `mobile.json` only by `mobile-webview-authority.mjs:156`. And `git ls-files -- 'wallet/zuuli/src-tauri/gen/schemas/*'` is empty.

This propagates into the follow-up: `:1568` and the PR body's sketch step 4 both describe the capability-file check as *"the check that already exists for `zcash` and needs only a second plugin id"*. It does not exist and cannot be obtained by parameterizing a plugin id — it has to be written. That mis-scopes proposed work, which is why I'd fix it now rather than file it.

Correcting `:149-161` and `:1568` to state the real parity, and marking places 3/4 as unenforced today, arguably **strengthens** §12.3 — it turns a "just add an id" into a justified reason to write the checker.

## The boundary, since I asked for it specifically

`scripts/check-public-repo-boundary.sh` passes, but that means little: it `git grep`s the literal string `free2z/tuzi` (`:10-12`) and is structurally incapable of seeing anything else.

§11.3 does carry, in this public repo, the ~90%/~10% production split, "all but one is still active" (a count in prose form), "production holds case-variant duplicate accounts today", and the `@аlice` Cyrillic worked example against a live product.

**None of it is new.** I diffed each statement against `main`: every one is a verbatim or near-verbatim restatement of what #573 already made public in `WIRE.md` §14.2/§14.3, `THREAT-MODEL.md` §4.10 and `ARCHITECTURE.md` §13-O. This PR is not a new disclosure — it **widens the surface from three files to four**, with no signal to a reader that this copy is a copy. If that content is ever trimmed, the fix now has to touch this file too. No private repo name, no private issue number, no infra detail, no absolute counts — matching #573's posture.

## Should fix — filing as issues

- **C1** `DeliveryState` is used at `:388` and `:488` and **declared nowhere**. Every other union in the file is declared inline. The doc calls `types.ts` "the single source of truth for shapes" (`:766`), yet the one type §6.2 spends a page insisting must not be collapsed can only be transcribed by inferring it from a table.
- **C2** The "closed union" isn't closed over the codes it claims to map. `ERR_STALE_TIMESTAMP`, `ERR_EPOCH_UNAVAILABLE`, `ERR_NO_ACCESS`, `ERR_ACK_TOO_HIGH`, `ERR_NOT_PERMITTED`, `ERR_RANGE_TOO_WIDE` have no target and no stated fallback. The clock-skew case is the sharp one: its nearest member is `relay-protocol-violation`, defined as *"a bug in one of the two implementations"* — so a wrong device clock routes into a defect-report affordance that blames the relay. It also silently resolves #550 item 1 in favour of the collapse, which is probably the right call but is presented as settled.
- **C3** `directory-cooldown` (`:1232`) tells the UI to say the old key "can still cancel the reset" — there is no `cancel_reset` command in §3, and per #551 the cancel message is undefined at the protocol layer too. A UI obligation whose referent exists at neither layer. Relatedly ADR 0014's `no_reset` appears nowhere in the contract, so a client built to this document cannot let a user set it; an explicit "not in v1" row would be more honest than silence.
- **C4 — the one I care most about.** `KT.md` §6.3 rule 7 is normative on clients and forbids skipping intervening tree heads; #558 measured that at ~4,320 sequential fetches for a one-month-offline client with no batch endpoint. This document mentions monotonicity twice, both times only as failure → fork evidence. No command, no progress event, no engine sub-state, no error code, no cost warning. Meanwhile §11.4 leads with *"`lookup_verify` is ~1.11 ms"*. Both true; together they leave a frontend developer thinking directory verification is a millisecond operation. Cold-start latency, progress UI and cancellability are exactly frontend concerns, and this is the only document that would raise them. **This is the one place the doc under-claims a cost rather than over-claiming a guarantee.**
- **C5** `set_retention_policy` (`:518`) requires `conversationId` and carries no `scope`, so the global policy is readable via `get_retention_policy` but not settable — while §3.7's prose reads like global is the primary case.
- **C6** Non-membership has two contradictory representations: `DirectoryResolution.found: false` (`:614`, "a PROVED non-membership, not an error") and an `ErrorCode` member `handle-not-found` (`:1234`). The contract never says which `resolve_handle` does. `KT.md` §9.5 has no unknown-handle code, which argues for `found: false` and deleting the member.
- **C7** `RelayWarning: "capability-digest-mismatch"` (`:738`) is unreachable — `:774-777`, `:1229` and `WIRE.md` §11.2 all say the client **refuses** on mismatch, so there is no `RelayConfig` left to carry the warning.

**C8** goes on the existing issues rather than a new one: "no confusable pairs" (`:1418-1419`) is now the third file carrying that over-claim (#575), and "any non-ASCII character… not eligible" (`:1428-1430`) is restated as settled while #550 item 4 shows it isn't — and `:290` hardens it into a type as `HandleEligibility.reason: "non-ascii"`. Real mitigating factor: `:1462-1463` forbids the client from deriving handles locally and mandates `check_handle_eligibility`, so the client-side blast radius is contained.

**Credit where due on C9:** `independentWitnesses` inherits `KT.md:942`'s unimplementable MUST, and the document *says so* — `:759` and `:790-793` mark it PROVISIONAL and §12.1 tables it. That's the right handling of an inherited defect. The only residual is that §9 rule 5 states the MUST without the PROVISIONAL pointer 550 lines away.

## Minor

`:128-133` says the seed is reachable "**only**" through `get_seed_phrase(token)` — there are two (`get_seed_phrase` at `commands.rs:1058`, `get_backup_seed_phrase` at `:1085`), and it's the *backup* one that's granted on mobile while `get_seed_phrase` deliberately isn't. Both are gated by `consume_sensitive_display`, so §2.2's security argument survives — but the "only" is wrong. `:134-136` cites `oauth_loopback_start` as the precedent for the args convention, but `oauth/transport.ts` is inconsistent about it (`:351-354`/`:317`/`:625` nest, `:203-206`/`:222` don't), so the cited precedent is a counterexample. `:975` says "try/catch"; `mock.ts:59` uses optional chaining. The PR body says the union has 30 members; it has 29. `MessageReceivedEvent` (`:946`) and `set_witness_set`'s element shape (`:707`) are referenced but never defined. §6.1's diagram draws a `stopped → faulted` edge the prose doesn't describe.

PR #580, opened 2026-08-24
ci(gates): forbid tolerating a collision on a protected context

### LINKED ISSUE #570: ci: check-workflow-gates' allowlist can silence a collision on a protected context
## Problem

`scripts/check-workflow-gates.mjs` (extended in #569 to close #562) now correctly rejects two workflows publishing the same status-check context. But its allowlist can be used to silence a collision on the **protected** contexts themselves.

Verified against the merged tree at `572163a`:

```console
# add ["gate", [".github/workflows/rs.yml", ".github/workflows/zuuli.yml"]]
# to TOLERATED_CONTEXT_COLLISIONS, then remove `name: rs / gate` from rs.yml
Every gate inspects every job it awaits and covers every job in its file, and
no two workflows publish one status-check context: 2 gate(s), 56 job(s), 14 workflow file(s).
EXIT=0
```

For contrast, the same removal **without** the allowlist entry fails correctly:

```console
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the
  status-check context "gate" (...); branch protection matches contexts by name,
  so a duplicate resolves by report order rather than policy
EXIT=1
```

## Why it's worth closing even though it takes two deliberate steps

This is not an accident waiting to happen — it needs both an allowlist edit and a `name:` removal, and a reviewer would see the allowlist entry in the diff. It's the same shape as the hole I flagged on #556, where dropping entries from `EXPECTED_PACKAGES` leaves that checker green at a lower mutant count: **the guarded scope is self-describing, so it can be narrowed by editing the thing that defines it.**

The difference here is what's behind the guard. `TOLERATED_CONTEXT_COLLISIONS` exists for contexts that *cannot decide a merge* — that is the stated and correct reason `build` and `frontend` are in it (#567). A required context is categorically different: tolerating a collision there re-creates the exact ambiguity #562 was filed for, in the one place it matters.

## Fix

A `PROTECTED_CONTEXTS` constant the allowlist may never cover:

```js
/// Contexts branch protection requires on `main`. A collision on one of these
/// is the #562 defect and can never be tolerated: two producers means the
/// required check resolves by report order rather than policy. Keep in sync
/// with `gh api repos/free2z/zuu/branches/main/protection/required_status_checks`.
const PROTECTED_CONTEXTS = new Set(["gate", "rs / gate"]);
```

and a check that fails if any key of `TOLERATED_CONTEXT_COLLISIONS` is in it — so the allowlist edit itself goes red, before the `name:` removal is even reached.

Current state, for the constant:

```console
$ gh api repos/free2z/zuu/branches/main/protection/required_status_checks --jq '{contexts,strict}'
{"contexts":["gate","rs / gate"],"strict":true}
```

## Acceptance

- `PROTECTED_CONTEXTS` is declared with a comment pointing at the API call that is its source of truth.
- Adding a protected context to `TOLERATED_CONTEXT_COLLISIONS` fails the check — demonstrated red, with real output, in the PR body.
- A negative control in `--self-test` covers it, in the same style as the existing 15.
- The two are kept honest by hand today. If it's cheap, consider having the check assert that every name in `PROTECTED_CONTEXTS` is actually published by exactly one job — that catches the reverse drift, where branch protection gains a context no workflow emits and PRs hang pending forever.

Found while verifying #569 before merging it. Related: #562, #567, #556.

skyl:
Closes #570.

`TOLERATED_CONTEXT_COLLISIONS` could be used to silence a collision on the
contexts branch protection actually requires. This adds `PROTECTED_CONTEXTS`,
which the allowlist may never cover, plus the reverse-drift assertion #570
asks about.

## The hole, before and after

The exploit takes two edits — the allowlist entry, then removing
`name: rs / gate` from `rs.yml`. Same tree, both checkers:

```console
$ node scripts/_baseline.mjs     # origin/main's checker, verbatim
Every gate inspects every job it awaits and covers every job in its file, and no two workflows publish one status-check context: 2 gate(s), 56 job(s), 14 workflow file(s).
EXIT=0

$ node scripts/check-workflow-gates.mjs     # this PR
Required-gate wiring failed:
- TOLERATED_CONTEXT_COLLISIONS names "gate", which branch protection requires: a duplicate producer of a required context resolves by report order rather than policy, so it can never be tolerated — remove the entry and give the colliding job a distinct name:
- branch protection requires the status-check context "gate", which 2 jobs publish (.github/workflows/rs.yml#gate, .github/workflows/zuuli.yml#gate); a required context must have exactly one producer
- branch protection requires the status-check context "rs / gate", which no job publishes; a required context nothing reports leaves every pull request pending forever (fix the job's name: or update PROTECTED_CONTEXTS to match branch protection)
3 failure(s) across 2 gate(s) in 14 workflow file(s).
EXIT=1
```

**The ordering the issue asks for.** The allowlist edit *alone* is red — the
`name:` removal is never reached, so the tree is never briefly ambiguous:

```console
# only change: ["gate", [".github/workflows/rs.yml", ".github/workflows/zuuli.yml"]]
# added to TOLERATED_CONTEXT_COLLISIONS. rs.yml untouched.
$ git diff --stat
 scripts/check-workflow-gates.mjs | 179 +++++++++++++++++++++++++++++++++++++--
 1 file changed, 174 insertions(+), 5 deletions(-)

$ grep -n 'name: rs / gate' .github/workflows/rs.yml
314:    name: rs / gate

Required-gate wiring failed:
- TOLERATED_CONTEXT_COLLISIONS names "gate", which branch protection requires: a duplicate producer of a required context resolves by report order rather than policy, so it can never be tolerated — remove the entry and give the colliding job a distinct name:
1 failure(s) across 2 gate(s) in 14 workflow file(s).
EXIT=1
```

Green on the unmodified tree:

```console
Every gate inspects every job it awaits and covers every job in its file, no two workflows publish one status-check context, and each protected context (gate, rs / gate) has exactly one producer and no allowlist entry: 2 gate(s), 56 job(s), 14 workflow file(s).
EXIT=0
```

## Reverse drift: implemented

#570 leaves this optional; it is clean here and produces no false positives,
so it ships. Both required contexts are Actions check runs from workflow files
this script already parses (`zuuli.yml#gate`, `rs.yml#gate`), so "published by
exactly one job" is decidable from the tree with nothing to work around. The
one case that *would* false-positive — a required context from a third-party
app or a workflow in another repository, which this script cannot see — is
named in the constant's comment as not belonging in the set. There is no such
context today.

Real-tree mutation, `name: rs / gate` removed from `rs.yml`, script otherwise
unmodified:

```console
Required-gate wiring failed:
- .github/workflows/rs.yml, .github/workflows/zuuli.yml: 2 job(s) publish the status-check context "gate" (.github/workflows/rs.yml#gate, .github/workflows/zuuli.yml#gate); branch protection matches contexts by name, so a duplicate resolves by report order rather than policy
- branch protection requires the status-check context "gate", which 2 jobs publish (.github/workflows/rs.yml#gate, .github/workflows/zuuli.yml#gate); a required context must have exactly one producer
- branch protection requires the status-check context "rs / gate", which no job publishes; a required context nothing reports leaves every pull request pending forever (fix the job's name: or update PROTECTED_CONTEXTS to match branch protection)
3 failure(s) across 2 gate(s) in 14 workflow file(s).
EXIT=1
```

That is the failure mode worth having: it is silent today — the check just sits
pending and nothing goes red — and it is what renaming a gate job without
updating branch protection looks like.

## Self-test: 15 -> 20 cases, each one broken on purpose

```console
...
self-test: a protected context added to the tolerated-collision allowlist is detected.
self-test: a tolerated collision on a context that is not protected passes.
self-test: two protected contexts, each published by exactly one job passes.
self-test: a protected context that no job publishes is detected.
self-test: a protected context published by two jobs is detected.
check-workflow-gates self-test: 20 case(s) passed.
EXIT=0
```

A control nobody has watched fail is not a control. Each new case, mutated in
turn:

```console
--- M1: drop the protectedToleranceFailures call ---
Error: self-test FAILED: a protected context added to the tolerated-collision allowlist was not detected. Failures: (none)
EXIT=1

--- M2: tolerance rule fires on every allowlist key, protected or not ---
Error: self-test FAILED: a tolerated collision on a context that is not protected should pass, got: TOLERATED_CONTEXT_COLLISIONS names "build", which branch protection requires: ...
EXIT=1

--- M4: producer rule ignores the zero-producer case (=== 1 -> <= 1) ---
Error: self-test FAILED: a protected context that no job publishes was not detected. Failures: ...
EXIT=1

--- M5: producer rule ignores the many-producer case (=== 1 -> >= 1) ---
Error: self-test FAILED: a protected context published by two jobs was not detected. Failures: ...
EXIT=1
```

**M3, stated exactly.** The mutation for the remaining case — matching
`entry.job` instead of `entry.context`, i.e. reading job ids instead of
published check-run names — makes the suite red, but the pre-existing case
"two gates in one tree with distinct display names" catches it first:

```console
--- M3: producer rule matches the job id instead of the published context ---
Error: self-test FAILED: two gates in one tree with distinct display names should pass, got: branch protection requires the status-check context "gate", which 2 jobs publish (.github/workflows/fixture.yml#gate, .github/workflows/other.yml#gate); a required context must have exactly one producer
EXIT=1
```

To show the new case is load-bearing on its own, the same mutation with that
one earlier case deleted:

```console
--- M3, isolated ---
Error: self-test FAILED: two protected contexts, each published by exactly one job should pass, got: branch protection requires the status-check context "gate", which 2 jobs publish (...); branch protection requires the status-check context "other / gate", which no job publishes; ...
EXIT=1
```

So that case is the green baseline its paired detected case ("a protected
context that no job publishes") is one line away from, in the style the file
already uses — not an independent detector. Saying so rather than picking a
mutation that flatters it.

One harness change came out of this: fixtures now scan with fixture-local
`tolerated` and `protectedContexts` (`fixtureOptions`). Previously a fixture
inherited the repository's live `TOLERATED_CONTEXT_COLLISIONS`, which meant an
allowlist entry added for a real workflow could change what a self-test case
means — and it made M2 unlocalizable.

## Other checks

```console
$ node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs
check-workflow-gates self-test: 20 case(s) passed.
Every gate inspects every job it awaits and covers every job in its file, no two workflows publish one status-check context, and each protected context (gate, rs / gate) has exactly one producer and no allowlist entry: 2 gate(s), 56 job(s), 14 workflow file(s).

$ node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs
self-test: 70 source-policy, 7 gate-verdict, and 99 current-workflow mutation case(s) passed.
GitHub Actions policy passed: every external action/reusable workflow is immutably pinned, and every required-gate dependency is bound to the enforcing verdict (178 external reference(s), 15 file(s)).
```

- **No workflow file is touched.** `zuuli.yml`'s `changes` policy step is
  untouched, so nothing collides with #556; no job is renamed, so #567 is
  unaffected. `git diff --stat` is one file: `scripts/check-workflow-gates.mjs`.
- **#477 shell hygiene:** 0 lines of shell added — the change is JavaScript and
  comments only.

`PROTECTED_CONTEXTS` is hand-maintained against the API, same as it was before
this PR; the difference is that drift in either direction now fails loudly
instead of being discovered at a merge.

### DIFF
--- a/scripts/check-workflow-gates.mjs
+++ b/scripts/check-workflow-gates.mjs
@@ -49,6 +49,15 @@
 //      future regression, so `needs ⊇ {jobs in file} − {gate}` is asserted,
 //      with an explicit per-file opt-out for jobs deliberately left outside.
 //
+//   5. PROTECTED CONTEXTS. The names branch protection actually requires get a
+//      stricter reading of both rules above. The tolerated-collision allowlist
+//      may not name one of them, so narrowing the guard to admit a duplicate
+//      `gate` fails on the allowlist edit itself rather than on the `name:`
+//      change that would follow it. And each required name must be published by
+//      exactly one job, which catches the opposite drift: a required context no
+//      workflow emits never reports at all, so PRs hang pending forever with
+//      nothing anywhere going red.
+//
 // Usage:
 //   node scripts/check-workflow-gates.mjs             judge every workflow
 //   node scripts/check-workflow-gates.mjs --self-test negative controls first
@@ -72,6 +81,35 @@ const CHANGE_DETECTOR = "changes";
 /// read and does. Adding one means reading the new verifier first.
 const EXHAUSTIVE_VERIFIERS = ["--verify-gate-results"];
 
+/// The check-run names branch protection requires on `main`. Source of truth:
+///
+///   gh api repos/free2z/zuu/branches/main/protection/required_status_checks \
+///     --jq '{contexts,strict}'
+///   {"contexts":["gate","rs / gate"],"strict":true}
+///
+/// These are the contexts that can decide a merge, which makes them different in
+/// kind from every other name in the tree, in two directions:
+///
+///   * A collision on one of them is exactly the #562 defect — two producers
+///     means the required check resolves by report order rather than by policy —
+///     so it can never be tolerated. `TOLERATED_CONTEXT_COLLISIONS` may not
+///     contain any of these names, and saying so makes the *allowlist edit*
+///     fail rather than the `name:` removal that would follow it.
+///
+///   * The reverse drift is just as bad and quieter: a required context that no
+///     job publishes never reports, so every PR waits on it forever. Nothing in
+///     the tree goes red — the check simply sits pending — and the first person
+///     to notice is whoever tries to merge. So each of these must be published
+///     by exactly one job.
+///
+/// The second rule assumes every required context comes from a job in this
+/// repository's `.github/workflows`, which is true today: both are Actions
+/// check runs from `zuuli.yml` and `rs.yml`. If branch protection ever requires
+/// a context from something this script cannot see — a third-party app, or a
+/// workflow living in another repository — that entry does not belong in this
+/// set, and the comment here is the place to record why.
+const PROTECTED_CONTEXTS = new Set(["gate", "rs / gate"]);
+
 /// Check-run names that more than one workflow is knowingly allowed to publish,
 /// mapped to the exact set of files allowed to publish them. Every entry is a
 /// documented exception, not a category: a *new* producer of one of these names
@@ -81,6 +119,10 @@ const EXHAUSTIVE_VERIFIERS = ["--verify-gate-results"];
 /// decide a merge — that is the whole reason they are tolerated rather than
 /// treated as the `gate` collision was. Renaming them is a follow-up, tracked
 /// in #567; adding to this map instead of fixing a collision needs the same.
+///
+/// "Cannot decide a merge" is not left to the reader's judgement: a key that is
+/// in `PROTECTED_CONTEXTS` is rejected outright, so this map can never be used
+/// to excuse the one collision that matters.
 const TOLERATED_CONTEXT_COLLISIONS = new Map([
   [
     "build",
@@ -389,8 +431,59 @@ export function contextCollisionFailures(published, tolerated) {
   return failures;
 }
 
+/// Reject an allowlist entry that covers a context branch protection requires.
+///
+/// This is a check on the checker's own configuration rather than on the tree,
+/// so it fails with no workflow needing to be wrong yet. That is the point:
+/// silencing the protected `gate` collision takes two edits — the allowlist
+/// entry and then the `name:` removal — and this makes the first one red, so
+/// the second is never reached and the tree is never briefly ambiguous.
+export function protectedToleranceFailures(tolerated, protectedContexts) {
+  const failures = [];
+  for (const context of [...tolerated.keys()].sort()) {
+    if (!protectedContexts.has(context)) continue;
+    failures.push(
+      `TOLERATED_CONTEXT_COLLISIONS names "${context}", which branch protection requires: ` +
+        "a duplicate producer of a required context resolves by report order rather than policy, " +
+        "so it can never be tolerated — remove the entry and give the colliding job a distinct name:",
+    );
+  }
+  return failures;
+}
+
+/// Require each protected context to be published by exactly one job.
+///
+/// The zero case is the silent one — branch protection waiting on a name no job
+/// emits blocks every PR with nothing to look at — and it is what a rename of a
+/// gate job without the matching branch-protection update looks like. The
+/// many case is already reported as a collision; it is repeated here because
+/// "exactly one" is the property being asserted, and a required context with two
+/// producers is worth saying twice.
+export function protectedProducerFailures(published, protectedContexts) {
+  const failures = [];
+  for (const context of [...protectedContexts].sort()) {
+    const producers = published.filter((entry) => entry.context === context);
+    if (producers.length === 1) continue;
+    if (producers.length === 0) {
+      failures.push(
+        `branch protection requires the status-check context "${context}", which no job publishes; ` +
+          "a required context nothing reports leaves every pull request pending forever " +
+          "(fix the job's name: or update PROTECTED_CONTEXTS to match branch protection)",
+      );
+      continue;
+    }
+    failures.push(
+      `branch protection requires the status-check context "${context}", which ${producers.length} jobs publish ` +
+        `(${producers.map((entry) => `${entry.file}#${entry.job}`).join(", ")}); ` +
+        "a required context must have exactly one producer",
+    );
+  }
+  return failures;
+}
+
 export function scanRepository(root, options = {}) {
   const tolerated = options.tolerated ?? TOLERATED_CONTEXT_COLLISIONS;
+  const protectedContexts = options.protectedContexts ?? PROTECTED_CONTEXTS;
   const exemptions = options.exempt ?? GATE_EXEMPT_JOBS;
   const failures = [];
   const published = [];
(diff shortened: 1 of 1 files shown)


skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I re-ran the mutations myself in a throwaway worktree.

## The ordering property is the thing, and it holds

#570's whole point was that the guard must fire on the **first** of the two steps, not the second. Verified with `name: rs / gate` still present at `rs.yml:314` — i.e. the allowlist edit *alone*, no `name:` removal:

```console
Required-gate wiring failed:
- TOLERATED_CONTEXT_COLLISIONS names "gate", which branch protection requires: a duplicate
  producer of a required context resolves by report order rather than policy, so it can never
  be tolerated — remove the entry and give the colliding job a distinct name
EXIT=1
```

So the exploit dies at step one. The message also tells you the fix rather than only the violation, which is the right shape for a policy checker somebody will hit at 2am.

Baseline is green and the summary line now states what was actually checked rather than a generic pass:

> Every gate inspects every job it awaits and covers every job in its file, no two workflows publish one status-check context, and each protected context (gate, rs / gate) has exactly one producer and no allowlist entry: 2 gate(s), 56 job(s), 14 workflow file(s).

## The reverse-drift check earns its place

I was ambivalent about asking for this and it turned out to be the better half. Removing `name: rs / gate`:

```
- 2 job(s) publish the status-check context "gate" … resolves by report order rather than policy
- branch protection requires "gate", which 2 jobs publish; a required context must have exactly one producer
- branch protection requires "rs / gate", which no job publishes; a required context nothing reports
  leaves every pull request pending forever (fix the job's name: or update PROTECTED_CONTEXTS
  to match branch protection)
```

Three independent failures from one mutation. The third is a failure mode nothing else in the repo could catch: branch protection naming a context no workflow emits is **silent until someone tries to merge**, and then it looks like a hung PR rather than a misconfiguration. That's exactly the class this repo keeps getting bitten by.

The false-positive risk is handled honestly rather than assumed away: both required contexts are Actions check runs from workflow files this script already parses, and the case that *would* false-positive — a required context from a third-party app or another repo — is named in the constant's comment as not belonging in the set. That's the right place for that caveat.

## Self-test 15 → 20, and the caveat is stated rather than buried

Four of the five new controls name their own case when broken. The fifth is caught first by the pre-existing "two gates in one tree with distinct display names", and only fires alone once that case is deleted — **and the PR body says so**, correctly describing it as the green baseline for its paired detected case rather than an independent detector.

That disclosure is worth more than the case itself. "20 cases pass" with one of them quietly redundant is precisely the policy-self-tests-growing-while-the-guarded-thing-stays-broken pattern; saying which one isn't load-bearing is what stops the count becoming a number nobody can interpret.

The harness change that fell out is a real improvement too: fixtures now scan with fixture-local `tolerated`/`protectedContexts`, so a future real allowlist entry can no longer silently change what a self-test case *means*. Fixtures inheriting live config is a subtle way for a test suite to stop testing what it says.

## Constraints

No workflow file touched at all — so `zuuli.yml`'s `changes` policy step is untouched and there's no collision with #556, which is contending over that exact ordered-command assertion, and no job renamed so #567 is unaffected. `check-github-actions-pins.mjs` self-test and real run both pass. Zero shell added, so #477 doesn't arise.

## One thing to remember

`PROTECTED_CONTEXTS` is now a hand-maintained mirror of a repo setting. The comment quotes the `gh api .../protection/required_status_checks` call as its source of truth, which is the right mitigation — and the reverse-drift check means a *stale* entry fails loudly rather than silently, in both directions. So the mirror can't rot unnoticed, which was my only reservation about the design.

PR #581, opened 2026-08-24
test(f2z-codec): byte-level wire vectors for every WIRE.md structure

### LINKED ISSUE #564: test(f2z-codec): no byte-level wire vectors — 11 of 20 encoding mutations survive all 69 tests

skyl:
`Closes #564`

`f2z-codec` encodes all 33 `WIRE.md` structures correctly, and nothing asserted
what any structure's bytes actually **are**. 11 of 20 real mutations to the wire
format passed the entire 71-test suite green — including swapping `address` and
`signer_key` in `CommandTranscript` and `SignedAuth`, which changes what every
Ed25519 signature in the protocol commits to.

This adds `rs/crates/f2z-codec/tests/wire_vectors.rs`: 42 **encode-side** tests
pinning the byte layout of every structure in the document, plus known-answer
BLAKE2b digests for every `LABEL_*` constant.

## The mutation table, re-run

Each mutation applied to `rs/crates/f2z-codec/src`, then
`cargo test --locked --all-targets && cargo test --locked --doc`, then reverted.
`git status` clean between every row.

**19 of 19 caught. Nothing survives.**

| # | Mutation | Site | Before (#564) | Now |
|---|---|---|---|---|
| M1 | `ReadResponse.messages`: `VecU24` → `VecU16` | `commands.rs:438` | NOT CAUGHT | **caught** — `read_response_uses_a_three_byte_vector_prefix` |
| M2 | `CreateQueueResponse`: swap `recv_addr` / `send_addr` | `commands.rs:396` | NOT CAUGHT | **caught** — `create_queue_response_puts_recv_addr_before_send_addr` |
| M3 | `HelloResponse`: swap `channel_binding_mode` / `transport_security` | `commands.rs:245` | NOT CAUGHT | **caught** — `hello_response_puts_channel_binding_mode_before_transport_security` |
| M4 | `LABEL_BODY` `…/v1/body` → `…/v1/bodx` | `hash.rs:25` | NOT CAUGHT | **caught** — 4 tests: `every_label_is_exactly_the_ascii_the_spec_prints`, `h_is_blake2b_256_of_label_then_message_and_the_digests_are_known_answers`, `body_hash_is_a_known_answer_over_the_re_encoded_body`, `the_command_transcript_is_the_eleven_fields_of_section_5_1_in_order` |
| M7 | `CommandTranscript`: swap `address` / `signer_key` | `transcript.rs:71` | NOT CAUGHT | **caught** — `the_command_transcript_is_the_eleven_fields_of_section_5_1_in_order`, `rebuilding_the_transcript_from_a_received_auth_gives_the_same_bytes` |
| M8 | `Capabilities.protocol_versions`: `VecU8` → `VecU16` | `commands.rs:590` | NOT CAUGHT | **caught** — `the_capability_document_encodes_field_for_field_as_section_11_1_declares`, `signed_capabilities_is_the_document_then_a_signature` |
| M10 | `SignedAuth`: swap `address` / `signer_key` | `frame.rs:424` | NOT CAUGHT | **caught** — `signed_auth_is_address_then_signer_key_then_timestamp_nonce_signature`, `command_auth_is_a_present_byte_then_the_authenticator_if_present` |
| M13 | `LABEL_RELAY_ID` one byte | `hash.rs:29` | NOT CAUGHT | **caught** — 3 tests: `every_label_is_exactly_the_ascii_the_spec_prints`, `h_is_blake2b_256_…`, `relay_id_is_a_known_answer_over_the_identity_key` |
| M17 | `PowStamp` digest: `counter.to_be_bytes()` → `to_le_bytes()` | `pow.rs:145` | NOT CAUGHT | **caught** — `the_pow_digest_is_a_known_answer_with_a_big_endian_counter` |
| M19 | `ChallengeResponse`: swap `challenge` / `expires_at_ms` | `commands.rs:329` | NOT CAUGHT | **caught** — `challenge_response_puts_the_challenge_before_the_expiry` |
| M21 | `H(label, x)` → `H(x, label)` in **both** `hash` and `hash2` | `hash.rs:62` | NOT CAUGHT | **caught** — 7 tests, including `h_is_blake2b_256_…`, `hash2_is_the_same_construction_with_the_message_in_two_pieces`, `relay_id_is_a_known_answer_…`, `the_pow_digest_is_a_known_answer_…` |
| M5 | `MAX_PLAUSIBLE_BUCKETS` 16 → 17 | `padding.rs:44` | caught | still caught — `padding::tests::implausibly_fine_grained_sets_are_flagged` |
| M6 | fixed newtype `Debug` prints inner bytes | `types.rs:97` | caught | still caught — `types::tests::debug_never_renders_bytes` |
| M9 | `ErrorCode::Quota` 14 → 24 | `error.rs:114` | caught | still caught — `error::tests::codes_are_the_table_in_wire_md_section_10` |
| M11 | `Command::Read` `0x0013` → `0x0016` | `commands.rs:113` | caught | still caught — `commands::tests::command_codes_are_the_table_in_section_6` |
| M12 | `PROPOSED_BUCKETS` `65_536` → `32_768` | `padding.rs:37` | caught | still caught — 2 `padding::tests` |
| M14 | `ShortBytes` prefix `U8` → `U16` | `types.rs:337` | caught | still caught — `transcript::tests::transcript_is_fixed_length_and_starts_with_the_label` |
| M15 | `FrameKind::Push` 3 → 4 | `frame.rs:62` | caught | still caught — `every_frame_round_trips_byte_for_byte`, `structured_noise_never_normalizes` |
| M16 | `RelayFrame.request_id` BE → LE | `frame.rs:367` | caught | still caught — 2 `frame::tests` |

**No row is still NOT CAUGHT.**

The "Before" column is #564's measurement, and this run confirms it without
re-measuring: for all eleven previously-surviving mutations, **every** test that
now catches them is one added by this PR. Not one of them is caught by a test
that existed before — which is precisely what "NOT CAUGHT" meant. The eight
previously-caught mutations are still caught by the same pre-existing tests, so
there is no regression.

## The expectations were derived from the spec, not from the code

This is the part that matters, because generating the expectations from the
implementation would have locked in whatever bug was already there and made the
problem worse. Every byte was written by hand from the `WIRE.md` text.

The mechanism that makes that checkable: each expectation is a list of fields
carrying the specification's own declaration, the width that declaration fixes,
and the bytes. `vector()` asserts the stated width and the supplied bytes agree
*before* it compares anything to the code, so a derivation with bad arithmetic
fails on its own terms rather than quietly agreeing.

```rust
f("uint8 label length = 19", 1, [0x13]),
f("\"free2z/relay/v1/cmd\"", 19, b"free2z/relay/v1/cmd"),
f("uint16 protocol_version = 1", 2, [0x00, 0x01]),
f("opaque relay_id[32]", 32, fill(0xe5, 32)),
```

Fixture integers are chosen so the big-endian encoding is legible on sight —
`0x0102_0304_0506_0708` encodes as `01 02 03 04 05 06 07 08` — and adjacent
same-width fields get *different* values, because a swap is only detectable if
the two values differ. A failure names the spec field the first divergent byte
falls in:

```
CommandTranscript: encoded bytes do not match the vector derived from WIRE.md.
first divergence at byte 92, in `opaque address[32]` at offset 92 (+0 into the field)
```

### Two worked derivations, for spot-checking

**`CommandTranscript` (§5.1, `WIRE.md:408-420`).** Field by field, with widths
from §3.4 and big-endian from §1.3:

| Spec declaration | Width | Bytes |
|---|---:|---|
| `opaque label<0..255>` | 1 + 19 | `13` + `"free2z/relay/v1/cmd"` |
| `uint16 protocol_version` | 2 | `00 01` |
| `opaque relay_id[32]` | 32 | `e5 …` |
| `opaque channel_binding[32]` | 32 | `f6 …` |
| `uint16 command` | 2 | `00 21` (APPEND, §6's table) |
| `uint32 request_id` | 4 | `0a 0b 0c 0d` |
| `opaque address[32]` | 32 | `a1 …` |
| `opaque signer_key[32]` | 32 | `b2 …` |
| `uint64 timestamp_ms` | 8 | `01 02 03 04 05 06 07 08` |
| `opaque nonce[16]` | 16 | `c3 …` |
| `opaque body_hash[32]` | 32 | `a0a5cdf6…` |

`1 + 19 + 2 + 32 + 32 + 2 + 4 + 32 + 32 + 8 + 16 + 32 = 212`, which is also what
`TRANSCRIPT_LEN` claims. `address` therefore begins at byte 92 and `signer_key`
at 124 — asserted directly, because that is the pair M7 swaps.

**`body_hash`.** §1.3: `H(label, x) = BLAKE2b-256(label || x)`, no separator and
no terminator. The body is `AppendRequest{payload = 8 × 0x77}`, which by §3.4's
`opaque payload<0..2^24-1>` is a three-byte length prefix and the bytes:
`00 00 08 77 77 77 77 77 77 77 77`. So:

```
$ printf 'free2z/relay/v1/body\x00\x00\x08\x77\x77\x77\x77\x77\x77\x77\x77' | b2sum -l 256
a0a5cdf686d12c0f701cf3e6573c6498af130d090da8f8a5c53ac2140363ccab
```

`python3 -c 'hashlib.blake2b(..., digest_size=32)'` agrees. Two tools, neither of
them this crate. Every `LABEL_*` digest in the file was produced the same way,
over the fixed message `01 02 03 04 05 06 07 08` — non-empty on purpose, since
`H(label, "")` and `H("", label)` are the same value and an empty message could
not detect M21's inverted construction.

### The one thing worth noting about the result

41 of the 42 tests passed the first time they compiled. The single failure was a
typo in my own fixture (a `request_id` I wrote two ways), not a disagreement
about byte layout. That is independent confirmation of #564's finding: the
encoding was already right everywhere, and the entire gap was the absence of
anything asserting it.

## Coverage

All 33 structures the document declares — `RelayFrame`, `Request`, `Response`,
`Push`, `FrameKind`, `CommandAuth`, `SignedAuth`, `CommandTranscript`,
`HelloRequest`, `HelloResponse`, `ChallengePurpose`, `ChallengeRequest`,
`ChallengeResponse`, `CreateQueueRequest`, `CreateQueueResponse`,
`SubscribeResponse`, `ReadRequest`, `QueuedMessage`, `ReadResponse`,
`AckRequest`, `AckResponse`, `BindSendRequest`, `AppendRequest`, `MsgPush`,
`QueueEventPush`, `NoticePush`, `CreateContactQueueRequest`,
`CreateContactQueueResponse`, `ContactAppendRequest`, `Capabilities`,
`SignedCapabilities`, `PowParams`, `PowStamp` — plus the §3.4 primitives every
one is built from (fixed `opaque x[n]`, `opaque x<0..2^k-1>` at all three prefix
widths, and `T x<0..2^k-1>`'s byte-counting prefix), the §6 command-code table
and the §10 error-code table as the `uint16`s they occupy on the wire.

`tests/reencode_equality.rs` is untouched. It is the correct test for §3.3, which
is a decoder-slack property; it was simply never a format test.

## The redaction sweep (#572's lesson, applied to the rest of the file)

#572 found `assert_no_leak`'s decimal detector anchored on the opening bracket,
so a dump starting mid-buffer slid past it. I swept every other detector and
assertion in `tests/redaction.rs` for the same class of blind spot. Three of the
seven checks were sound as written; three needed work, and one fixture was
vacuous.

**Sound, no change:** `lower_hex`/`upper_hex` of the whole secret (plain
substring, no anchor), `decimal_run` of a four-byte prefix (#572's fix), and
`longest_hex_run(rendered) < 8` (unanchored; its documented refusal to count
pure-decimal runs is a deliberate false-positive trade, not an anchor).

**1. `base64url` was anchored on byte alignment.** The same defect in a less
obvious form: base64 encodes three bytes at a time, so the symbols a leak
produces depend on where the secret sits relative to the *buffer's* start, not
its own. A dump of `[0, 4, 0, …] || secret` phase-shifts the secret's encoding
and the zero-offset pattern never appears. Now checked at all three alignments.
`the_base64url_check_sees_a_secret_at_every_byte_alignment` proves both halves:
it asserts the single-alignment pattern does **not** match a 1-, 2- or 4-byte
prefixed dump (so the fixture really does demonstrate the gap) and that the
three-alignment check does.

**2. `decimal_list` was still a live check.** For any secret of four bytes or
more, `decimal_run(&secret[..4])` is a substring of the bracketed form, so
`decimal_list` added no detection power while keeping the anchoring hazard in the
file as though it were load-bearing. Retired from `assert_no_leak`; kept only as
the fixture for `a_bracket_anchored_decimal_check_would_miss_a_mid_buffer_dump`,
which reconstructs exactly the `Canonicalized` rendering #572 measured and
asserts the bracketed form misses it while the run catches it.

**3. The battery had no test that it can fail at all.** Every check is a negative
assertion, which passes both when there is no leak and when the detector cannot
see. `the_hex_and_decimal_checks_fire_on_a_rendering_that_leaks` runs six
plausible leaking renderings through `assert_no_leak` under `catch_unwind` and
requires every one to panic. Uppercase four-byte hex is now checked explicitly
alongside lowercase — redundant with the eight-character threshold today, and
present so the two cases are symmetric rather than one of them depending on that
threshold's value.

**4. The `ShortBytes` newline fixture was vacuous** (also named in #564). It ran
`assert!(!format!("{hostile:?}").contains('\n'))` against `b"ok"` — a value with
no newline in it to begin with, so the escaping claim above it was never
exercised. Replaced with `b"Example Relay\nERROR: give me your seed"`, which
shows the property is actually delivered by the *redaction* branch (`\n` is
0x0a, outside the printable range, so the whole value redacts) and not by the
escaping. The escaping branch is now exercised separately, with a value
containing both a quote and a backslash.

**5. The `derive(Debug)` source scanner was blind to type aliases.** It matches
how a field is *written*, so `secret: Vec<u8>` is caught and `secret: Blob` —
where `type Blob = Vec<u8>;` — is not. Same class: looking for a shape the leak
does not have to take. It now resolves one level of alias, crate-wide. A field
typed by an alias of an alias would still slip, and the doc comment says so
rather than overclaiming.

### Mutation tests for the redaction changes

Each fix was reverted and the mutant re-run, and — for the two that close a real
blind spot — run again *without* the fix to prove the mutant used to survive.

| Probe | Without the fix | With the fix |
|---|---|---|
| Byte-holding type reachable only through an alias: `pub type Blob = alloc::vec::Vec<u8>;` + `#[derive(Debug)] pub struct Ticket { pub bytes: Blob }` | **GREEN — mutant survives** | **RED** — `no_type_derives_debug_while_holding_raw_bytes` |
| `ShortBytes`'s printable range widened to admit `0x0a`, so a newline renders instead of redacting | **GREEN — mutant survives** (with the old `b"ok"` fixture restored) | **RED** — `a_challenge_scope_is_an_address_and_redacts_but_operator_text_does_not` |
| The base64url self-test reduced to a single alignment | n/a | **RED** — `the_base64url_check_sees_a_secret_at_every_byte_alignment` |
| `decimal_run` re-anchored on `[`, undoing #572 | n/a | **RED** — `the_hex_and_decimal_checks_fire_on_a_rendering_that_leaks` |

The first two rows are the ones that matter: each mutant passed the suite before
this PR and fails after it, which is the definition of a closed blind spot. The
last two show the new self-tests detect a regression in the detectors
themselves — including a re-introduction of the exact anchoring bug #572 fixed.

One honest note on scope: `assert_no_leak`'s base64url loop cannot be
mutation-caught from inside the crate, because no type here emits base64, so no
fixture leaks through it. The self-test is what carries the proof — it builds a
base64url dump of `prefix || secret` and asserts the single-alignment pattern
misses it while the three-alignment check does not.

## `WIRE.md` gaps found while deriving by hand

Checked against #550 and #552 first; these are not in either.

1. **`CreateQueueRequest.stamp` is "empty when `queue_creation_mode = open`"
   (`:705`), and "empty" is undefined for a fixed-width structure.** `PowStamp`
   has no variable-length field and §3.3 forbids omitting bytes, so the only
   consistent reading is 56 zero bytes — which is what this crate implements and
   documents. But the spec never says it, and an implementer who reads "empty" as
   "omit the field" is 56 bytes out of sync on the first `CREATE_QUEUE` of every
   open relay. Sharpened by the contrast one section earlier: `ChallengeResponse.pow`
   is described as **"zeroed** when no PoW is required" (`:652`). Two adjacent
   fields of the same type, two different words, one of them unambiguous.
2. **`ReadResponse.has_more` is a `uint8` with no declared value set** (`:750`).
   Every other flag byte in the document enumerates its values —
   `QUEUE_EVENT.reason` (1-4), `NOTICE.kind` (1-3), `channel_binding_mode` (0/1),
   `durability_mode` (0-2). `has_more` does not, so whether it is a boolean or a
   count, and whether `has_more = 2` is legal, is left to the implementer. §3.3
   makes any value canonical, so nothing catches the divergence.
3. **§13.1's PoW preimage is the one integer in the protocol concatenated
   outside a `tls_codec` struct.** `H("free2z/relay/v1/pow", challenge || salt ||
   counter)` (`:1641`) is prose, not a structure, so `counter`'s width and
   endianness come only from §1.3's global rule and are stated nowhere at the
   point of use. This is exactly the M17 site: flip the endianness and every
   stamp fails verification between two implementations that both believe they
   read the spec. One clause — "`counter` as a big-endian `uint64`" — closes it.
4. **`ChallengeRequest.scope` has no defined value for the other two purposes.**
   `:646` says "for `contact_append`: the target `contact_addr`" and never says
   what `clock` and `queue_create` put there. Empty is the obvious reading and
   this crate emits it; a relay that *requires* empty and one that ignores the
   field disagree about a valid request.
5. **`Capabilities`' eight operator and provenance fields are `opaque x<0..255>`
   with no charset**, yet §11.3 step 7 requires a client to "surface operator
   name, jurisdiction and contact in the relay-picker UI". They are
   operator-controlled bytes rendered in UI with no encoding, validation or
   escaping rule stated anywhere. The crate's `ShortBytes` `Debug` handles this
   for logs by redacting anything non-printable; the *display* rule is the
   spec's to make.

## Checks

- `cargo test --locked --all-targets`: **115 passed** (49 unit + 7 padding +
  10 redaction + 7 reencode-equality + 42 wire-vectors), was 70.
- `cargo test --locked --doc`: 1 passed. Total **116**, was 71.
- `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec`:
  clean. Nothing was added to the library; both changed files are test targets,
  so `no_std` + `alloc` + `forbid(unsafe_code)` is untouched.
- `scripts/check-rust-fmt.sh --root rs`, `check-rust-clippy.sh --root rs`,
  `check-rust-deny.sh --root rs --config rs/deny.toml`: all clean. One narrow
  `#![allow(clippy::panic)]` on the test file, because `assert_wire` reports
  which spec field the first divergence lands in and `assert_eq!` cannot.

`.github/workflows/zuuli.yml` is untouched.

### DIFF
--- a/rs/crates/f2z-codec/tests/redaction.rs
+++ b/rs/crates/f2z-codec/tests/redaction.rs
@@ -57,7 +57,14 @@ fn decimal_run(bytes: &[u8]) -> String {
     joined.join(", ")
 }
 
-/// `[222, 222, 222, …]` — a derived `Debug` on a byte slice.
+/// `[222, 222, 222, …]` — a derived `Debug` on a byte slice, brackets and all.
+///
+/// Retained only as the thing the checks below deliberately do **not** use.
+/// `decimal_run` of a four-byte prefix is a substring of this for every secret
+/// of four bytes or more, so the bracketed form adds no detection power and
+/// carries the anchoring hazard the doc comment above describes. See
+/// `a_bracket_anchored_decimal_check_would_miss_a_mid_buffer_dump`.
+#[allow(dead_code)]
 fn decimal_list(bytes: &[u8]) -> String {
     format!("[{}]", decimal_run(bytes))
 }
@@ -110,6 +117,13 @@ fn longest_hex_run(text: &str) -> usize {
 }
 
 /// Assert that `rendered` cannot be turned back into `secret`.
+///
+/// Every check here is a **substring** search on an unanchored pattern. That is
+/// the lesson of the `Canonicalized` leak: the detector that missed it was
+/// anchored on the opening `[` of a decimal list, and a real dump starts
+/// partway through a buffer — after a length prefix — so the `[` never lined
+/// up. Nothing below may depend on the secret being at the start or the end of
+/// what a leak would print.
 fn assert_no_leak(label: &str, rendered: &str, secret: &[u8]) {
     assert!(
         !rendered.contains(&lower_hex(secret)),
@@ -119,27 +133,43 @@ fn assert_no_leak(label: &str, rendered: &str, secret: &[u8]) {
         !rendered.contains(&upper_hex(secret)),
         "{label} leaked uppercase hex: {rendered}"
     );
-    assert!(
-        !rendered.contains(&base64url(secret)),
-        "{label} leaked base64url: {rendered}"
-    );
-    assert!(
-        !rendered.contains(&decimal_list(secret)),
-        "{label} leaked a decimal byte list: {rendered}"
-    );
+    // base64url has the same anchoring hazard in a less obvious form: it encodes
+    // three bytes at a time, so the symbols a leak produces depend on where the
+    // secret sits relative to the *buffer's* start, not its own. A dump of
+    // `[0, 4, 0] || secret` phase-shifts the secret's encoding and the
+    // zero-offset pattern never appears. Checking all three alignments covers
+    // every possible phase.
+    for phase in 0..3usize.min(secret.len()) {
+        assert!(
+            !rendered.contains(&base64url(&secret[phase..])),
+            "{label} leaked base64url at byte alignment {phase}: {rendered}"
+        );
+    }
     // A prefix of a decimal dump is enough to convict: no correct output of this
     // crate ever prints four consecutive byte values. Matched without brackets,
     // so a dump that begins after a length prefix — `[0, 4, 0, 222, 222, …` —
     // is caught as readily as one that begins at the opening bracket.
+    //
+    // This subsumes the bracketed `decimal_list` form entirely, which is why
+    // that helper is no longer one of the checks.
     assert!(
         !rendered.contains(&decimal_run(&secret[..secret.len().min(4)])),
         "{label} leaked a decimal byte run: {rendered}"
     );
     // Every byte of the secret is 0xde, so even a two-byte prefix would show as
     // a repeated `de`. Catch a partial dump too.
+    //
+    // Both cases are stated. An eight-character run also trips the threshold
+    // below, so neither of these is the only thing standing between a four-byte
+    // dump and a pass today — they are here so that the lowercase and uppercase
+    // cases are symmetric and neither depends on the value of that threshold.
     assert!(
         !rendered.contains(&lower_hex(&secret[..secret.len().min(4)])),
-        "{label} leaked a hex prefix: {rendered}"
+        "{label} leaked a lowercase hex prefix: {rendered}"
+    );
+    assert!(
+        !rendered.contains(&upper_hex(&secret[..secret.len().min(4)])),
+        "{label} leaked an uppercase hex prefix: {rendered}"
     );
     // Nothing this crate prints is a long hex string. 8 characters is short
     // enough to be a decimal length or a section number and long enough that a
@@ -148,6 +178,97 @@ fn assert_no_leak(label: &str, rendered: &str, secret: &[u8]) {
         longest_hex_run(rendered) < 8,
         "{label} contains an 8+ character hex-looking run: {rendered}"
     );
+}
+
+/// The detectors, tested against renderings that leak.
+///
+/// A battery is only worth what its blind spots allow through, and every check
+/// in `assert_no_leak` is a negative assertion — it passes when nothing is
+/// found, which is also what it does when it cannot see. So each one is aimed
+/// at a string that really does contain the secret, and must fire.
+#[test]
+fn a_bracket_anchored_decimal_check_would_miss_a_mid_buffer_dump() {
+    let secret = [SECRET; 32];
+
+    // Exactly the shape the `Canonicalized` leak produced: a wire buffer whose
+    // three-byte length prefix comes first, so the secret starts at index 3 of
+    // the list and the opening bracket sits in front of `0, 4, 0`.
+    let mut buffer = vec![0u8, 4, 0];
+    buffer.extend_from_slice(&secret);
+    let leaked = format!("Canonicalized {{ encoded: {} }}", decimal_list(&buffer));
+
+    assert!(
+        !leaked.contains(&decimal_list(&secret)),
+        "the bracketed form is what missed this; if it now matches, the fixture is wrong"
+    );
+    assert!(
+        leaked.contains(&decimal_run(&secret[..4])),
+        "the unbracketed run must see the same dump the bracketed form missed"
+    );
+}
+
+#[test]
+fn the_base64url_check_sees_a_secret_at_every_byte_alignment() {
+    let secret = [SECRET; 32];
+
+    // Same buffer, base64url-encoded whole. The three-byte prefix shifts the
+    // secret's encoding by 3 - (3 % 3) = 0 … so use a prefix length that is
+    // *not* a multiple of three, which is the case a zero-offset pattern misses.
+    for prefix_len in [1usize, 2, 4] {
+        let mut buffer = vec![0u8; prefix_len];
+        buffer.extend_from_slice(&secret);
+        let leaked = format!("Frame {{ encoded: \"{}\" }}", base64url(&buffer));
+
+        // The necessity of the loop, stated as an assertion rather than as a
+        // comment: the zero-offset pattern — the only one the check used to
+        // try — does not appear in any of these renderings.
+        assert!(
+            !leaked.contains(&base64url(&secret)),
+            "the single-alignment check would have caught a {prefix_len}-byte prefix, \
+             so this fixture does not demonstrate the gap: {leaked}"
+        );
+
+        let fired = (0..3usize).any(|phase| leaked.contains(&base64url(&secret[phase..])));
+        assert!(
+            fired,
+            "a base64url dump with a {prefix_len}-byte prefix escaped every alignment: {leaked}"
+        );
+    }
+}
+
+#[test]
+fn the_hex_and_decimal_checks_fire_on_a_rendering_that_leaks() {
+    // Not a tautology: `assert_no_leak` is a negative assertion, so it is worth
+    // proving it can be made to fail at all. Each of these is a formatter a
+    // careless `Debug` might plausibly use.
+    let secret = [SECRET; 32];
+    let quiet = std::panic::take_hook();
+    std::panic::set_hook(Box::new(|_| {}));
+    let renderings = [
+        format!("X({})", lower_hex(&secret)),
+        format!("X({})", upper_hex(&secret)),
+        format!("X({})", decimal_list(&secret)),
+        // Four bytes only: shorter than any full-length pattern, and the case
+        // the prefix checks exist for.
+        format!("X({})", lower_hex(&secret[..4])),
+        format!("X({})", upper_hex(&secret[..4])),
+        format!("X([1, 2, {}, 9])", decimal_run(&secret[..4])),
+    ];
+    let escaped: Vec<(String, bool)> = renderings
+        .into_iter()
+        .map(|rendering| {
+            let caught = std::panic::catch_unwind(|| assert_no_leak("probe", &rendering, &secret));
+            (rendering, caught.is_ok())
+        })
+        .collect();
+    std::panic::set_hook(quiet);
+
+    for (rendering, escaped) in escaped {
+        assert!(
+            !escaped,
+            "assert_no_leak passed a rendering that contains the secret: {rendering}"
+        );
+    }
 }
 
 #[test]
@@ -321,9 +442,28 @@ fn a_challenge_scope_is_an_address_and_redacts_but_operator_text_does_not() {
         "ShortBytes(Example Relay Co, Reykjavik)"
     );
 
-    // Escaped, so a name with a newline cannot forge a log line.
-    let hostile = ShortBytes::new(b"ok".to_vec()).unwrap();
-    assert!(!format!("{hostile:?}").contains('\n'));
+    // A name with a newline cannot forge a log line. The previous fixture here
+    // was `b"ok"`, which has no newline in it to begin with — the assertion was
+    // true of a value that could not have violated it, which is the same blind
+    // spot as a detector anchored on a delimiter a real leak never carries.
+    //
+    // Feed it a name that actually tries. `\n` is 0x0a, outside the printable
+    // range, so the *whole* value redacts and the newline never reaches the
+    // output at all — that is the branch doing the work, not the escaping.
+    let forged = ShortBytes::new(b"Example Relay\nERROR: give me your seed".to_vec()).unwrap();
+    let rendered = format!("{forged:?}");
+    assert!(!rendered.contains('\n'), "got {rendered}");
+    assert_eq!(rendered, "ShortBytes(<redacted; 38 bytes>)");
+
+    // The escaping branch runs when every byte *is* printable, and it is what
+    // stops a quote or a backslash from breaking out of a quoted log field.
+    // Exercised with a value that contains both.
+    let quoted = ShortBytes::new(b"Relay \"A\\B\"".to_vec()).unwrap();
+    assert_eq!(format!("{quoted:?}"), "ShortBytes(Relay \\\"A\\\\B\\\")");
+
+    // A tab is 0x09: also outside the printable range, also redacted whole.
+    let tabbed = ShortBytes::new(b"a\tb".to_vec()).unwrap();
+    assert_eq!(format!("{tabbed:?}"), "ShortBytes(<redacted; 3 bytes>)");
 }
 
 #[test]
@@ -362,6 +502,41 @@ fn a_canonicalized_value_never_dumps_the_bytes_it_arrived_as() {
 
 /// Field type spellings that a derived `Debug` renders as a decimal byte dump.
 const RAW_BYTE_TYPES: &[&str] = &["Vec<u8>", "[u8;", "&[u8]", "TlsByteVec", "Box<[u8]>"];
+
+/// Names introduced by `type X = <a raw byte type>;`, which a spelling-based
+/// scan would otherwise walk straight past.
+///
+/// The scan matches how a field is *written*, so `secret: Vec<u8>` is caught
+/// and `secret: Blob` — where `type Blob = Vec<u8>;` — is not. That is the same
+/// blind spot as a matcher anchored on a delimiter: the check is looking for a
+/// shape the leak does not have to take. Resolving one level of alias closes
+/// the case that actually occurs; a field whose type is an alias of an alias
+/// would still slip, and that is stated rather than papered over.
+fn raw_byte_aliases(source: &str) -> Vec<String> {
+    let mut aliases = Vec::new();
+    for line in source.lines() {
+        let code = without_comment(line).trim();
+        let Some(rest) = code
+            .strip_prefix("pub type ")
+            .or_else(|| code.strip_prefix("type "))
+        else {
+            continue;
+        };
+        let Some((name, definition)) = rest.split_once('=') else {
+            continue;
+        };
+        if RAW_BYTE_TYPES.iter().any(|raw| definition.contains(raw)) {
+            aliases.push(
+                name.trim()
+                    .split(['<', ' '])
+                    .next()
+                    .unwrap_or_default()
+                    .to_owned(),
+            );
+        }
+    }
+    aliases
+}
 
 /// A `struct` or `enum` declaration that carried `#[derive(..., Debug, ...)]`.
 struct DerivedDebugItem {
@@ -509,13 +684,22 @@ fn no_type_derives_debug_while_holding_raw_bytes() {
     let mut scanned = Vec::new();
     let mut violations = Vec::new();
 
-    for (file, source) in source_files() {
-        for item in derived_debug_items(&file, &source) {
-            for raw in RAW_BYTE_TYPES {
+    let sources = source_files();
+
+    // Aliases are collected across the whole crate first: a field in `frame.rs`
+    // may be typed with an alias declared in `types.rs`.
+    let mut spellings: Vec<String> = RAW_BYTE_TYPES.iter().map(|raw| (*raw).to_owned()).collect();
+    for (_, source) in &sources {
+        spellings.extend(raw_byte_aliases(source));
+    }
+
+    for (file, source) in &sources {
+        for item in derived_debug_items(file, source) {
+            for raw in &spellings {
                 // The declaration line itself is excluded from the search only
                 // for the tuple-struct case, where it *is* the body; matching
                 // the whole captured text is what we want either way.
-                if item.body.contains(raw) {
+                if item.body.contains(raw.as_str()) {
                     violations.push(format!(
                         "{}: `{}` derives Debug while holding `{}` — a derived Debug \
                          renders that as a decimal byte dump. Hand-write Debug and \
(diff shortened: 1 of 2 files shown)


skyl:
Merge-train hold: queue this behind #556. #556 has now integrated exact current trunk three times and is starting the final authoritative matrix; any unrelated merge restarts the entire cross-platform gate. Do not admin-merge this PR until #556 lands. Afterward integrate its squash normally, restore exact-head green, and return it to ready-for-review.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I re-ran #564's mutation table myself in a throwaway worktree rather than reading the results off the body.

## The mutations that mattered are dead

These five all passed the entire suite green before this PR. Re-run against the new vectors, each mutation confirmed landed in the source before judging:

```
M1   ReadResponse.messages  VecU24 -> VecU16
     -> RED  read_response_uses_a_three_byte_vector_prefix

M7   CommandTranscript      swap address / signer_key
     -> RED  rebuilding_the_transcript_from_a_received_auth_gives_the_same_bytes

M10  SignedAuth             swap address / signer_key
     -> RED  signed_auth_is_address_then_signer_key_then_timestamp_nonce_signature
             command_auth_is_a_present_byte_then_the_authenticator_if_present

M21  hash(label, data)  ->  hash(data, label)
     -> RED

M4   LABEL_BODY  /v1/body -> /v1/bodx
     -> RED  every_label_is_exactly_the_ascii_the_spec_prints
             body_hash_is_a_known_answer_over_the_re_encoded_body
```

M7 and M10 are the ones I most wanted dead: swapping `address` and `signer_key` changes what **every Ed25519 signature in the protocol commits to**, and 69 tests used to stay green through it. M1 was the silent interop break — a `READ` response capped at 64 KiB with no test noticing. M21 was the worst of the set, because inverting `H(label, x)` to `H(x, label)` turned `labels_are_prefix_free` into an assertion of the *wrong invariant* while still passing.

**A correction on my own testing:** my first M4 attempt reported GREEN. It was wrong — `perl -0pi` without `/g` replaced the first match in the slurped file, which was the doc comment at `hash.rs:23`, not the constant at `:25`. Targeting the constant gives the RED above. Exactly the "confirm the mutation actually landed" failure I keep briefing workers about, so worth recording rather than quietly fixing.

## The expectations are hand-derived, and the file says how to check that

This was the pass/fail criterion — generated expectations would have locked in whatever bug exists and made #564 worse. The header states it outright, and more usefully the derivation is **visible per field**: each carries the spec's own declaration (`opaque address[32]`, `uint64 timestamp_ms`), the width that declaration fixes, and the bytes that width holds.

The `vector` helper asserts the stated width and the supplied bytes agree **before** comparing anything, so a derivation with wrong arithmetic fails on its own terms instead of quietly agreeing with the implementation. That's the part that makes the file self-defending rather than merely hand-written, and it's a better idea than what I asked for.

Fixture integers chosen so big-endian encoding is self-evident (`0x0102_0304_0506_0708` → `01 02 03 04 05 06 07 08`) means a reviewer can spot-check by eye without running anything.

The label digests were computed with **two tools that are not this crate** — `python3 hashlib.blake2b(digest_size=32)` and an independent second — which is the right standard for a known-answer test. `H(label, x) = BLAKE2b-256(label || x)` with no separator and no terminator was encoded by hand from `WIRE.md:67-69`.

## The redaction sweep found and fixed the same class again

I asked for this as a side task and it paid. #572 fixed one bracket-anchored detector; this sweep found the hazard in a second form and made the property explicit rather than incidental:

- `a_bracket_anchored_decimal_check_would_miss_a_mid_buffer_dump` — a test that *demonstrates* the blind spot rather than just avoiding it.
- `the_base64url_check_sees_a_secret_at_every_byte_alignment` — base64url carries the same hazard in a less obvious form, since a secret starting mid-buffer encodes differently at each of three alignments. That one I would not have found.
- `the_hex_and_decimal_checks_fire_on_a_rendering_that_leaks` — asserts the detectors fire on a *positive* case, which is the thing a detector suite usually forgets.

Every check in that file is now a substring search on an unanchored pattern, documented as a rule rather than left as a property that happens to hold.

## Numbers

71 → **115** tests (49 unit + 7 padding + 10 redaction + 7 reencode + 42 wire vectors). `reencode_equality.rs` correctly left intact — it is the right test for §3.3's decoder-slack property and was never the problem.

+2020/−17 across two test files, no library source touched, so there is no behaviour change to review — only the question of whether the assertions are real, which the mutation table answers.

## Follow-up

Any `WIRE.md` ambiguities the hand-derivation surfaced should land on #550 or #552 rather than a new issue. If the derivation was clean throughout, that is itself worth knowing — it would mean the spec is precise enough to implement from, which is the strongest evidence yet for the freeze in #548.

Issue #578, opened 2026-08-24
docs(e2ee): CLIENT-CONTRACT obligates the client to do things it exposes no way to do — KT catch-up, reset cancel


skyl:
`docs/e2ee/CLIENT-CONTRACT.md` (#574, merged as `4073cc0`) places two obligations on the client that it gives the client no way to meet. Both block real screens, and both are entangled with open protocol gaps, so they need deciding at the protocol layer before a frontend can be built to them.

## 1. The catch-up cost is invisible, and §11.4 reads as though it isn't there

`KT.md` §6.3 rule 7 is **normative on clients** — `KT.md` §8.1 step 2 says *"a client can and MUST make these checks itself"* — and reads:

> If `epoch > last.epoch + 1`, the verifier MUST fetch every intervening tree head and check the chain link by link. **It MUST NOT skip.**

Per #558 §7, §5.1 publishes an empty heartbeat epoch every 600 s, so a client offline for one month must fetch **~4,320 tree heads one at a time** — `KT.md` §9.2 offers only `GET /kt/v1/sth/{epoch}`, no range or batch endpoint — each carrying its cosignature set. Six months is ~26,000.

`CLIENT-CONTRACT.md` mentions monotonicity exactly twice (`:667`, `:1230`), both times only as a **failure** → fork evidence. There is:

- no command to perform or resume the catch-up,
- no progress event,
- no engine sub-state for "catching up",
- no error code (`ERR_EPOCH_UNAVAILABLE` has no union member — #577 §2),
- and no cost warning anywhere.

Meanwhile §11.4 (`:1494-1497`) opens with *"The client verifier reaches the browser cleanly — `lookup_verify` is ~1.11 ms"*. Every statement is individually true. Together they leave a frontend developer with the impression that directory verification is a millisecond operation.

**This is the one place the document under-claims a cost rather than over-claiming a guarantee**, and it is the more dangerous direction here: cold-start latency, progress UI, and cancellability are exactly frontend concerns, and this is the only document that would raise them. A client shipped without them will appear to hang on first launch after any meaningful offline period.

**What's needed:** either a batch/range endpoint in `KT.md` §9.2 (which would make this a non-issue and is probably the right answer), or an explicit contract surface — an engine sub-state, a progress event, a cancel affordance, and a §11.4 note stating the cold-start cost next to the 1.11 ms figure.

## 2. `directory-cooldown` tells the UI to advertise an operation that exists at neither layer

[...]

Issue #577, opened 2026-08-24
docs(e2ee): CLIENT-CONTRACT interface defects — undeclared DeliveryState, unclosed "closed union", unsettable global retention


skyl:
Seven mechanical defects in `docs/e2ee/CLIENT-CONTRACT.md` (#574, merged as `4073cc0`). All cheap, all in §3/§6/§8. Grouped because they're one pass.

## 1. `DeliveryState` is used twice and declared nowhere

`:388` (`SendAccepted.state`) and `:488` (`DeliveryStatus.state`) both reference `DeliveryState`. There is no `type DeliveryState = …` anywhere in the file.

Every other union in the document **is** declared inline — `EngineState:1030`, `GapState:451`, `MessageBody:406`, `VerificationState:633`, `DurabilityMode:244`, `TransportHealth:331`, `AlarmKind:660`, `ErrorCode:1178`. §6.2 gives a diagram and a table but no declaration.

The document states at `:766` that `types.ts` is "the single source of truth for shapes". So the one type carrying the most semantic weight — the one §6.2 spends a page insisting must not be collapsed — cannot be transcribed without inferring it from a markdown table.

## 2. The "closed union" is not closed over the wire codes it claims to map

`:1172-1175`: *"A **closed union**. Every rejected command … carries exactly one of these … The bridge maps `WIRE.md` §10 and `KT.md` §9.5 wire codes onto it."*

Mapping both tables against the 29 members, these have **no target and no stated fallback**:

| Wire code | Consequence |
|---|---|
| `WIRE` 7 `ERR_STALE_TIMESTAMP` | Device clock outside `clock_skew_ms` — common and user-actionable ("your clock is wrong"). The nearest member is `relay-protocol-violation`, defined as *"a bug in one of the two implementations"*, so **a wrong device clock routes into a defect-report affordance that blames the relay.** |
| `KT` 7 `ERR_EPOCH_UNAVAILABLE` | The exact error a long-offline client hits — see #577. |
| `WIRE` 10 `ERR_NO_ACCESS` | Silently resolves #550 item 1 (the `ERR_NO_ACCESS`/`ERR_UNAVAILABLE` contradiction at `WIRE.md:431` vs `:830`) in favour of the collapse. Choosing the collapse is probably right; presenting it as settled, without noting the source contradiction is open, is not. |
| `WIRE` 13 `ERR_ACK_TOO_HIGH` | none |
| `WIRE` 20 `ERR_NOT_PERMITTED` | none |
[...]

Issue #576, opened 2026-08-24
docs(e2ee): CLIENT-CONTRACT restates a retracted claim and cites a checker that does not check what it says


skyl:
`docs/e2ee/CLIENT-CONTRACT.md` landed in #574 (merged as `4073cc0`). Two defects should be corrected promptly because each actively misleads: one is a self-contradiction on a retracted claim, the other mis-scopes proposed work.

## 1. §11.3 restates a claim `WIRE.md` formally retracted, and contradicts itself 32 lines later

`CLIENT-CONTRACT.md:1423-1425`:

> An account is eligible iff `lowercase(username)` matches the pattern; **case folding is safe because platform uniqueness is already enforced case-insensitively.**

That is verbatim the claim at `WIRE.md:1827-1830` — the one `WIRE.md:1843-1866` **explicitly retracts** in a dated correction block landed in #573, 24 hours earlier:

> **Correction (2026-08-24) — the case-folding argument above is false, and the property it rests on does not hold.** … **Case-insensitive uniqueness is not a database property.** … Folding case *does* collide two distinct existing accounts.

And `CLIENT-CONTRACT.md:1457-1460`, thirty-two lines below, asserts the opposite:

> **`lowercase(username)` is not yet a unique key.** Case-insensitive username uniqueness lives in two serializers and **not in the database**, and production holds case-variant duplicate accounts today.

The document asserts P at `:1425` and ¬P at `:1457`. #574's second commit went to real trouble folding #573's corrections into §11.3 and left the retracted sentence standing in the paragraph immediately above the corrected one.

**Fix: delete or invert `:1424-1425`.** One sentence. The repo's leave-standing-with-a-dated-blockquote convention deliberately does **not** apply here — that convention preserves a claim that was published and relied upon, and this file is one day old with no such history. Editing it out is correct; the retraction that matters already lives at `WIRE.md:1843`.

## 2. §2.3 and §12.3 assert a mechanical check that does not exist

`CLIENT-CONTRACT.md:149-152`:

> Every plugin command in §3 must appear in **four places**, and a mismatch is a build failure **by the same mechanism `scripts/check-zcash-permissions.mjs` already enforces for the Zcash plugin** … 3. `permissions/default.toml` — `allow-<kebab-command>`. 4. `src-tauri/capabilities/default.json` … and `capabilities/mobile.json`.

`scripts/check-zcash-permissions.mjs` never opens either file:

```console
[...]

PR #582, opened 2026-08-24
docs(e2ee): scope the "no confusable pairs" claim in all three files, and repair what #573's redaction left behind

### LINKED ISSUE #575: docs(e2ee): THREAT-MODEL §4.10 claims no confusable pairs — 1/l and 0/o are in the label space
## Problem

`docs/e2ee/THREAT-MODEL.md` §4.10, added in #573 (merged as `328c357`), states at `:574` that the KT directory's label space:

> contains **no confusable pairs**

That is false as written. The label space is `[a-z0-9_]`, which contains `1`/`l` and `0`/`o` — the canonical ASCII homoglyph pairs, indistinguishable in most UI sans-serif faces.

What the messaging charset actually eliminates is the **cross-script / mixed-script** class. The frozen text scopes this correctly at `WIRE.md:1735-1737`:

> an entire class of homograph and mixed-script impersonation

§4.10 drops the scoping and asserts the absolute — and does it two paragraphs above the "what we claim, precisely" table, in the document whose entire value is that it doesn't overstate.

**This is the third instance of the same pattern** in the security-claims document: #551 (`THREAT-MODEL.md:127` — "a user who set `no_reset` has foreclosed the path entirely", backed by a field with no normative rule) and #558 §4 (the same `no_reset` over-claim from the encoding side). Each was introduced by the change that was supposed to improve honesty.

**Suggested wording:**

> contains no *cross-script* confusable pairs; visually similar ASCII glyphs (`1`/`l`, `0`/`o`) remain and are out of scope for this mechanism

## Two smaller problems in the same section

- **`:586` — "no property of it is claimed here"** while `:562-566` claims two properties of free2z's username space (that `UnicodeUsernameValidator` is in use, and that non-ASCII accounts exist). Presumably it means no *safety* property. In this document that distinction has to be written, not inferred.
- **§4.10 does not cross-reference the case-collision blocker** that #573 discovered (`WIRE.md:1862-1876`), even though "two distinct real accounts map to one messaging handle" is a directory-integrity failure and §4.10 is the directory-integrity section. A reader of the threat model has no path to the blocker.

## The redaction removed the evidence for a conclusion that stayed behind

Separate from the over-claim, and worth fixing in the same pass.

#573's head moved during review and the absolute counts were redacted to proportions — a good call, and the redaction is faithful (I verified each paraphrase against the prior head). But two places kept a conclusion whose support was in the deleted text:

- `ARCHITECTURE.md:1153-1161` and `WIRE.md:1890-1897` retain **"neither a rounding error nor a migration project"**. The earlier head justified that with "~10% of a **small user base**", and the qualifier went with the counts. A bare proportion cannot distinguish the two: 10% of 1,700 and 10% of 10,000,000 are different projects. Either restore an order-of-magnitude qualifier or drop the neither/nor conclusion.
- `ARCHITECTURE.md:1153` is headed **"How many, closed."** and then declines to say how many. "What share, closed." matches the content.

## Minor, same PR

- **Date stamps are a day ahead of this repo's convention.** Everything new is stamped `2026-08-24` (`WIRE.md:1745`, `:1786`, `:1842`, `:1862`; `THREAT-MODEL.md:563`; `ARCHITECTURE.md:1152`), but the commit is `2026-08-23 22:05 -0400`. #557 set the precedent — merged `2026-08-24T00:31Z`, corrections dated `2026-08-23` (`ARCHITECTURE.md:342`, `:780`; `THREAT-MODEL.md:93`, `:115`) — i.e. **local** date. `ARCHITECTURE.md:1146` now reads "Opened 2026-08-23" three lines above a correction dated 2026-08-24 for the same evening's work. Cosmetic, but a measurement dated tomorrow invites the wrong reading, and these documents live or die on their dating being trustworthy.
- **`WIRE.md:1803` breaks the table's unit.** Three rows give shares of all accounts; that row gives "A small number of accounts", under a column headed "Prevalence", after an intro at `:1798-1799` promising shares.
- **`WIRE.md:1806`** — "the punctuation rule is almost the whole of the exclusion" sits next to punctuation = "just under a tenth" and total ineligible = "~10%", so it reads as *all* of it, and quietly contradicts "and not zero" at `:1803`.
- **`WIRE.md:1740-1743`** says "*this section* flagged three things"; two of the three were flagged in §14.3, not §14.2.

## Acceptance

- `THREAT-MODEL.md:574` scopes the claim to cross-script confusables and names the ASCII pairs that remain.
- `:586`'s "no property" is qualified to "no safety property", or reworded.
- §4.10 links the `WIRE.md` §14.3 case-collision blocker.
- The "neither a rounding error nor a migration project" conclusion either regains a scale qualifier or goes.
- `ARCHITECTURE.md:1153`'s heading matches its content.
- New date stamps follow the local-date convention #557 established.

Found while reviewing #573 after it merged. Related: #550 (item 4, now sharpened), #551, #558.

skyl:
Closes #575.

Docs only, `docs/e2ee/` only. No code, no workflows, no `rs/`.

## 1. The over-claim, in all three files it reached

The claim landed in three places, not one. `WIRE.md:1729` and
`THREAT-MODEL.md:574` came from #573; `CLIENT-CONTRACT.md:1418-1419` restated it
in #574 after the issue was filed. All three now carry the same scoping, taken
from the formulation `WIRE.md` §14.1 already had right — *"an entire class of
homograph and mixed-script impersonation"*: **an entire class, not every class.**

**Before** (`THREAT-MODEL.md:574`):

> Messaging handles are `[a-z0-9_]{1,30}` compared as bytes ([`WIRE.md` §14]),
> so the KT directory's label space contains **no confusable pairs** and
> homograph impersonation **does not exist inside the messenger** — that is why
> the charset was restricted, and it is what makes the log's labels not
> homograph-attackable ([`KT.md` §1.3]).

**After** — the sentence is left standing verbatim with an inline pointer, and
the correction is appended as a dated blockquote:

> …([`KT.md` §1.3](./KT.md#13-conventions)). **That sentence over-claims; see the
> correction below.**
>
> > **Correction (2026-08-23) — the charset removes the cross-script class of
> > confusables, not every confusable.** … `[a-z0-9_]` **does** contain
> > confusable pairs: `1`/`l` and `0`/`o` are the canonical ASCII homoglyphs and
> > are indistinguishable in most UI sans-serif faces, so `@a1ice` and `@alice`,
> > or `@b0b` and `@bob`, are two handles a reader cannot separate by eye. "No
> > confusable pairs" is false as written, and "homograph impersonation does not
> > exist inside the messenger" is true only of the cross-script kind. …
> > Same-script ASCII lookalikes are **out of scope for this mechanism**…

`WIRE.md` §14.1 keeps its (already correctly scoped) conclusion and gains a short
"an entire class, not every class" paragraph naming the residual ASCII pairs.

**`CLIENT-CONTRACT.md` got different treatment on purpose.** That document tells
frontend developers what to build and what copy to write, so a hedge there is
useless — an unactionable caveat ships as UI text anyway. It gets three build
rules instead: render handles in a disambiguating face on every surface; never
write copy that sells the charset as a safety property ("Handles are ASCII, so
they can't be faked" is named as false and must-not-ship); and point the user at
safety-number verification, because a handle is an address, not an
authentication.

**Two more instances found while checking for drift**, both fixed the same way so
the set cannot diverge again: `KT.md:92` ("why the log's labels are not
homograph-attackable") and `ARCHITECTURE.md:1148` ("so that homograph and
mixed-script impersonation does not exist"). Both were the unscoped form.

## 2. Correction pattern vs. wording fix — the split, and why

The house pattern (leave the wrong claim standing, inline pointer, dated
blockquote — `ARCHITECTURE.md:332`/`:342`/`:780`, `THREAT-MODEL.md:93`/`:115`,
`WIRE.md` §14.3) is expensive and it is meant to be. I applied it twice.

**Corrections (pattern applied):**

- **Item 1, `THREAT-MODEL.md` §4.10.** A published, false, security-relevant
  claim in the security-claims document. Exactly what the pattern exists for.
- **Item 4, `WIRE.md` §14.3.** "The answer is neither, and that is the awkward
  result" was published and its support was deleted. Withdrawing a published
  conclusion is a correction even when what replaces it is silence.

**Wording and structure (fixed in place):**

- **Item 2, `:586`.** "No property of it is claimed here" was ambiguous, not
  false — the intended meaning ("no *safety* property") is now written out
  alongside the two factual properties the section does claim.
- **Item 3.** A missing cross-reference is an omission; adding one corrects
  nothing.
- **Item 5**, `ARCHITECTURE.md:1153` "How many, closed." → "What share, closed."
  A heading that does not match its own body is a typo with ambition.
- **Item 6**, the three `WIRE.md` §14.2 nits (table unit, "almost the whole",
  "this section flagged three things"). Presentation.
- **`WIRE.md` §14.1 / `KT.md` / `ARCHITECTURE.md` §13-O.** These are restatements
  whose *operative* conclusion was already scoped or is a one-line summary
  pointing elsewhere; three more blockquotes would bury the one correction that
  matters. Each now points at `THREAT-MODEL.md` §4.10, which carries it.

Applying the pattern everywhere would make it noise; applying it nowhere would
lose the audit trail. The line I drew is: **published + false + load-bearing.**

## 3. Item 4 — no absolute count restored

The redaction to proportions was deliberate and correct, so I did **not** restore
an order-of-magnitude qualifier. Anything of the form "~10% of *N*" is an
absolute count to one significant figure, and zuu is public.

Instead the neither/nor conclusion is **withdrawn**, with the reason stated: the
rounding-error/migration-project question is about the absolute count, not the
share, and the count lives in the deployment repository. The paragraph now
illustrates the gap without asserting either pole ("one account in ten of a user
base of a few thousand is an afternoon of hand resolution; one in ten of a few
million is a programme of work") and keeps what the share *does* support on its
own — about one in ten, all real active users, therefore a product decision.
`ARCHITECTURE.md` §13-O carries the one-line version and links the correction.

## 4. Date stamps — I corrected them, and said so

**Decision: re-stamp `2026-08-24` → `2026-08-23` across `docs/e2ee/`, and record
both the re-stamp and the convention in the documents themselves.**

The argument for leaving them: rewriting a date after the fact is itself a small
honesty question, and these documents are only worth anything because their
dating is trustworthy.

The argument that won: the stamp is provably wrong, not merely
convention-divergent. #573 is `2026-08-23 19:14:52 -0700` and #574 is
`2026-08-23 19:20:07 -0700`. Nothing in them was measured or written on the 24th
— **a measurement cannot postdate the commit that carries it.** `2026-08-24` is
the UTC date, and #557 already set local date as the convention. Leaving a
knowably-false date in place to avoid the appearance of rewriting a date is the
wrong trade in a document whose whole claim is that it doesn't leave known-false
statements standing unmarked.

What makes it not a silent rewrite:

- A **`> Re-stamped 2026-08-23.`** note in `WIRE.md` §14.2 states what the old
  stamps were, why they were wrong, every file that changed, and that **no figure
  changed**.
- `README.md` gains a **Dating convention** paragraph (local date, matching the
  commit; a stamp ahead of its own commit is a defect — correct it and say so)
  and a **Scoping convention** paragraph, so the pattern behind #551/#558/#575
  is written down rather than rediscovered a fourth time.

Also added, per the issue's suggestion: a one-paragraph preamble to
`THREAT-MODEL.md` §5's "what we claim, precisely" table stating that every entry
is scoped by the section it cites, and that every correction in the document has
the same cause.

## Verification

- Every relative link and `#anchor` across `docs/e2ee/` resolves (checked
  programmatically over all 20 files; the only reported miss is a pre-existing
  `...sampleArgs` fragment inside a `CLIENT-CONTRACT.md` code sample).
- `scripts/check-public-repo-boundary.sh` passes.
- No absolute user count, production measurement, private repo name or issue
  number, or new description of an unpatched weakness is introduced. The
  case-collision blocker cross-reference restates what `WIRE.md` §14.3 already
  publishes.
- No contradiction introduced with `WIRE.md` §14 (which now states the scoped
  form itself), `KT.md`'s `[a-z0-9_]{1,30}`, or ADR 0013/0014.

## Deliberately not touched

- `CLIENT-CONTRACT.md:1423-1425` still says case folding is safe "because
  platform uniqueness is already enforced case-insensitively" — the argument
  `WIRE.md` §14.3 retracted. That is #576 and is left alone, as instructed.
  Flagging it here only so it isn't read as an oversight.
- #577, #578, `rs/`, `scripts/`.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -1145,20 +1145,26 @@ deliberate.
   a memory-hard function is affordable at relay scale is undecided.
 - **O. Messaging-handle eligibility for existing accounts.** Opened 2026-08-23 by
   [`WIRE.md` §14](./WIRE.md#14-handle-charset-for-messaging--blocking-pre-check-resolved).
-  Messaging handles are `[a-z0-9_]{1,30}`, ASCII, so that homograph and
-  mixed-script impersonation does not exist. free2z usernames are broader
+  Messaging handles are `[a-z0-9_]{1,30}`, ASCII, so that **cross-script**
+  homograph and mixed-script impersonation does not exist — same-script ASCII
+  lookalikes such as `1`/`l` remain and are a client rendering problem
+  ([`THREAT-MODEL.md` §4.10](./THREAT-MODEL.md#410-the-platform-username-space-contains-homographs-messaging-handles-do-not)).
+  free2z usernames are broader
   (`^[\w.@+-]+$`, up to 150 characters), so some existing accounts are not
-  eligible. **Measured 2026-08-24 — and still open**, because measurement was
+  eligible. **Measured 2026-08-23 — and still open**, because measurement was
   never the whole question. What the measurement settled:
-  - **How many, closed.** **Approximately 90% of existing accounts are eligible
+  - **What share, closed.** **Approximately 90% of existing accounts are eligible
     after case-folding and approximately 10% are not** — overwhelmingly because
     the username contains `.` `@` `+` or `-`; non-ASCII characters and
-    over-length names account for very few. It is neither the rounding error nor
-    the migration project the question was framed around, and it is not a set of
-    dead rows: **every ineligible account has logged in at least once and all but
-    one is still active.** Absolute counts are business data and are deliberately
-    not published in this repository; they are recorded in the deployment
-    repository.
+    over-length names account for very few. It is not a set of dead rows:
+    **every ineligible account has logged in at least once and all but one is
+    still active.** Whether that is a rounding error or a migration project is a
+    question about the *absolute* count and not about the share, and absolute
+    counts are business data, deliberately not published in this repository; they
+    are recorded in the deployment repository. The earlier answer that it was
+    "neither" rested on a scale qualifier that the redaction to proportions
+    removed, and is withdrawn — see the correction in
+    [`WIRE.md` §14.3](./WIRE.md#143-the-decision-and-the-cost-accepted).
   - **Which validator, closed.** `Creator(AbstractUser)` does not override
     `username`; the field is recorded with
     `UnicodeUsernameValidator`. **Non-ASCII usernames are legal on free2z today
--- a/docs/e2ee/CLIENT-CONTRACT.md
+++ b/docs/e2ee/CLIENT-CONTRACT.md
@@ -1415,11 +1415,36 @@ A messaging handle is:
 /^[a-z0-9_]{1,30}$/     ASCII only, compared as bytes, no normalization
 ```
 
-The restriction is not tidiness. A charset with no case, no scripts and no
-confusable pairs makes an entire class of homograph and mixed-script
-impersonation **not exist**, rather than be defended against — and a
-normalization function is itself the attack surface
+The restriction is not tidiness. A charset with no case and a single script makes
+an entire class of homograph and mixed-script impersonation **not exist**, rather
+than be defended against — `@аlice` with a Cyrillic а cannot be encoded as a
+messaging handle at all — and a normalization function is itself the attack
+surface
 ([`WIRE.md` §14](./WIRE.md#14-handle-charset-for-messaging--blocking-pre-check-resolved)).
+
+**An entire class, not every class — and the remainder is your job, not the
+protocol's.** `[a-z0-9_]` still contains `1`/`l` and `0`/`o`, indistinguishable
+in most UI sans-serif faces: `@a1ice` and `@alice`, or `@b0b` and `@bob`, are
+pairs a user cannot separate by reading them. Nothing below the UI defends
+against that. Three build rules follow.
+
+- **Render handles in a face that disambiguates** — monospace, or a text face
+  with a slashed or dotted zero and a distinguishable `1`/`l`. A handle set in
+  the same proportional sans as body copy cannot be checked by eye, and every
+  surface that shows a handle counts: search results, invites, message headers,
+  profile links.
+- **Never write copy that sells the charset as a safety property.** "Handles are
+  ASCII, so they can't be faked" is false and must not ship. What is true, and
+  is the most a string of copy may claim, is that a handle cannot contain
+  characters from another script.
+- **Point the user at the check that is real.** Who a peer *is*, is established
+  by safety-number verification — always available, and the strongest check in
+  the system regardless of directory state
+  ([§6.3](#63-verification-state),
+  [§6.4](#64-witness-threshold--the-fail-closed-matrix)). A handle is an
+  address, not an authentication, and first-contact copy should say so rather
+  than implying the handle did the work.
+
 An account is eligible iff `lowercase(username)` matches the pattern; case
 folding is safe because platform uniqueness is already enforced
 case-insensitively.
@@ -1430,7 +1455,7 @@ character, or exceeding 30 characters after lowercasing are not eligible for a
 messaging handle in v1.**
 
 **How many accounts that is, is now measured, and the answer is the reason this
-screen matters.** Measured against production on 2026-08-24
+screen matters.** Measured against production on 2026-08-23
 ([`WIRE.md` §14.2](./WIRE.md#142-checked-against-free2zs-real-username-rules--measured),
 [§13-O](./ARCHITECTURE.md#13-open-questions)):
 
@@ -1554,7 +1579,7 @@ transfer ([§13-K](./ARCHITECTURE.md#13-open-questions)), group scale
 ([§13-N](./ARCHITECTURE.md#13-open-questions)) — which directly sets how slow
 first contact feels on a phone — messaging-handle eligibility
 ([§13-O](./ARCHITECTURE.md#13-open-questions): the *measurement* is closed as of
-2026-08-24 and is in §11.3, but the **product** question of whether the ~10% is
+2026-08-23 and is in §11.3, but the **product** question of whether the ~10% is
 excluded in v1 or served by a separate opt-in handle is not), KT epoch cadence
 ([§13-P](./ARCHITECTURE.md#13-open-questions)), witness independence and the
 default *t* ([§13-Q](./ARCHITECTURE.md#13-open-questions)), the client gossip
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -89,7 +89,11 @@ surface and its stable error codes; and where `akd`'s types sit underneath ours.
   ([ADR 0001](./decisions/0001-platform-priority.md)).
 - A **handle** is `[a-z0-9_]{1,30}`, ASCII, per
   [`WIRE.md` §14](./WIRE.md#14-handle-charset-for-messaging--blocking-pre-check-resolved).
-  It is the log's label and it is why the log's labels are not homograph-attackable.
+  It is the log's label, and it is why a label cannot carry a **cross-script**
+  homograph: with one script and no case there is no `а`-for-`a` substitution to
+  encode. Same-script ASCII lookalikes (`1`/`l`, `0`/`o`) are in the label space
+  and are out of scope for the directory — see
+  [`THREAT-MODEL.md` §4.10](./THREAT-MODEL.md#410-the-platform-username-space-contains-homographs-messaging-handles-do-not).
 
 ---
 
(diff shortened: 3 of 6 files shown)


skyl:
Merge-train hold: queue this behind #556. #556 has now integrated exact current trunk three times and is starting the final authoritative matrix; any unrelated merge restarts the entire cross-platform gate. Do not admin-merge this PR until #556 lands. Afterward integrate its squash normally, restore exact-head green, and return it to ready-for-review.

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I verified the substance rather than the summary.

## It was in five places, not three

I briefed this as three files. The sweep found two more I'd missed — `KT.md:92` ("labels are not homograph-attackable") and `ARCHITECTURE.md:1148` ("so that homograph and mixed-script impersonation does not exist"), both in the unscoped form.

That's the finding that matters. Three instances is a claim someone repeated; five is a claim that had **propagated on its own** through a doc set, and every new document was going to inherit it. Checking for drift instead of fixing the two lines named in the issue is the difference between closing a ticket and closing a class.

Verified on the head — every site now carries the scoping, and it's the scoping `WIRE.md` §14.1 already had right rather than newly invented wording:

```
ARCHITECTURE.md:1149  …so that **cross-script** homograph and mixed-script impersonation
                      does not exist — same-script ASCII lookalikes such as 1/l remain
                      and are a client rendering problem

KT.md:92              …it is why a label cannot carry a **cross-script** homograph: with
                      one script and no case there is no `а`-for-`a` substitution to encode.
                      Same-script ASCII lookalikes (1/l, 0/o) are in the label space
```

## The correction pattern is applied correctly, and applied twice only

`THREAT-MODEL.md:574` still carries the original claim verbatim, with the dated correction at `:581-587` naming `1`/`l` and `0`/`o` and the worked examples. Nothing edited away.

More importantly the pattern was applied **twice** — §4.10 and `WIRE.md` §14.3 — and not to the other four, with the line drawn at **published + false + load-bearing**. That's the right line. Applying the dated-blockquote treatment to every wording fix would turn a signal into decoration; applying it to none would bury a retraction. Arguing the split explicitly rather than defaulting either way is what I wanted and rarely get.

## `CLIENT-CONTRACT.md` got build rules, not a hedge

This is the best judgement call in the PR. That document tells frontend developers what copy to write, so a scoped-but-still-abstract claim would have been useless there. Instead it gets three actionable rules: a disambiguating face on every handle surface, *"Handles are ASCII, so they can't be faked"* named explicitly as false and must-not-ship, and users pointed at safety-number verification — because a handle is an address, not an authentication.

That last sentence is the whole point of the fix, expressed in a form a developer can act on.

## Date stamps: the reasoning is stronger than the convention argument

I framed this as a convention question (#557 set local dates). The PR found a better argument: the stamps are **provably wrong**, not merely divergent. #573's commit is `2026-08-23 19:14:52 -0700` and #574's is `19:20:07 -0700` — *a measurement cannot postdate the commit that carries it.*

Corrected to `2026-08-23`, including `CLIENT-CONTRACT.md` which the issue didn't list — leaving two stamps behind would have been worse than leaving all of them. And not done silently: `WIRE.md:1767` records the re-stamp, the old value, why it was wrong, every file touched, and that **no figure changed**. The one remaining `2026-08-24` in the tree is *inside* that note, quoting the old stamp. Exactly right.

## No counts leaked, and item 4 was withdrawn rather than fudged

`git diff origin/main...HEAD -- docs/ | grep '^+'` for every figure from #573's pre-redaction head: **empty**. 

On the "neither a rounding error nor a migration project" conclusion, the PR withdrew it rather than qualifying it, and said why: *"~10% of N" is a count to one significant figure*, so any scale qualifier would partially disclose the thing the redaction removed. Illustrated symmetrically ("a few thousand… a few million") without asserting either pole. That is a sharper reading of the constraint than mine — I'd suggested adding an order-of-magnitude qualifier, which would have leaked.

## The conventions written down

`README.md` gains a **Dating convention** and a **Scoping convention**, plus a scoping preamble above `THREAT-MODEL.md` §5's claims table. Given this is the *third* time the unscoped-claim pattern has been filed (#551, #558, #575) and the fifth instance found, writing the rule down is what stops a sixth.

## Also correct: what it left alone

`CLIENT-CONTRACT.md:1423-1425` still asserts the case-folding argument `WIRE.md` §14.3 retracted. That's #576, it was out of scope, and the PR body says so explicitly so it isn't read as an oversight. Staying inside the brief and flagging the neighbour is the right behaviour.

All relative links and anchors across all 20 files in `docs/e2ee/` resolve; the only reported miss is a pre-existing `…sampleArgs` fragment inside a code sample. Boundary script passes, no production measurements, no private references.

PR #583, opened 2026-08-24
feat(rs): add f2z-relay-proto, the shared relay protocol layer

skyl:
`rs/crates/f2z-relay-proto` — the protocol layer above `f2z-codec` that a relay
and a client **both link**. No I/O, no sockets, no filesystem, no async runtime,
**no clock and no randomness**: every function that needs the time takes it as
an argument, which is what makes the rules testable and lets the same code run
in a relay and in a browser.

`no_std` + `alloc`, `#![forbid(unsafe_code)]`, MIT (the AGPL boundary starts at
the server binaries), and it builds for `wasm32-unknown-unknown`.

## What is in it

| Module | `WIRE.md` |
|---|---|
| `command` | §5.1 signed-command construction and verification, in the specified order; §6's command set as typed request/response pairs |
| `replay` | §5.5's timestamp window and fail-closed `(signer_key, nonce)` seen-set |
| `queue` | §7 lifecycle (bind-once, TTL clamping, rotation), §8 ACK arithmetic, §12.2 contact queues, §13.1 layer-2 quotas |
| `capabilities` | §11 document: sign, verify, digest, validity, and §11.3's client policy |
| `hello` | §2.5/§5.2 proof of possession and the session it opens |
| `inflight` | §4.3's bounded, out-of-order window with a typed ticket |
| `key` | Ed25519, `verify_strict`, constant-time key comparison, §10's dummy verification |

`f2z-codec`'s public API is **unchanged** — nothing in it needed to move.

## Design decisions worth reviewing

**The verification order is one function.** `CommandVerifier::verify` runs §5.1's
steps 1-6 in order, with `relay_id` and `channel_binding` coming from the
verifier and everything else from the frame. Reordering it is a security change,
so it is not composable by a call site.

**The body gets its own §3.3.** Frame-level re-encode equality preserves the
opaque body verbatim, so a non-canonical structure *inside* it survives
untouched. The typed body is decoded canonically here, and `body_hash` covers
those re-encoded bytes.

**`verify_strict` everywhere.** Plain Ed25519 verification accepts small-order
public keys, under which a signature can verify for more than one key. §5.1 step
5 treats a verifying signature as proof that a *specific* key authorized a
command, so a key that verifies everything is a key that steals every queue it
is registered against.

**Only a verified `HELLO` hands out a `TranscriptBuilder`.** `RelaySession` is
the sole source in this crate, so "the relay proved possession first" (§5.2's
MUST) is a precondition of being able to sign, not a step someone remembers.

**`Ticket<C>` is neither `Clone` nor `Copy`.** With out-of-order responses
(§4.3) the only correlation is a `uint32`; without a type, an `AckResponse`
decodes cleanly as a `SubscribeResponse` — same two `uint64`s, different
meaning, no error anywhere. Test `an_append_response_may_not_carry_queue_state`
pins §6.3 through this.

## Spec defect found — please read this one

**§11.1 publishes `antireplay_window_ms` and `clock_skew_ms` with no stated
relation, and there must be one.** A command stamped `clock_skew_ms` in the
future is still inside the window `clock_skew_ms` *after* the relay first saw
it, so a seen-set entry must be retained for `2 × clock_skew_ms` from
observation. A relay publishing `antireplay_window_ms < 2 × clock_skew_ms`ages
entries out while the frames they cover are still valid — §5.5's protection
silently absent for the gap, with nothing in the document forbidding it.

Handled without redesigning: `SeenSet::retention_is_sound` states the relation,
and `ClientPolicy::require_sound_antireplay_window` (default **on**) refuses
such a relay. It is deliberately **not** in `capabilities::validate`, because
the document is exactly what the specification permits — it is the specification
that is missing the constraint. `WIRE.md` §11.1 should state it, and §5.5 should
say what the window is measured from.

## Ambiguity calls — every one, with reasoning

1. **§6.3 vs §10 on send-side refusals.** §10 code 10 covers "absent **or**
   unauthorized"; §10 code 15 covers "absent, deleted, expired, full…, send side
   only". Both claim "absent". Resolved in §6.3's favour: `authorize_send`
   returns `ERR_UNAVAILABLE` for everything, because §6.3's reason (a bound
   sender must not learn queue state by filling the queue) dominates, and
   collapsing absent into unavailable creates no oracle — every send-side
   refusal looks identical. `ERR_ALREADY_BOUND` and `ERR_NOT_PERMITTED` stay
   distinguishable because §7.3/§7.4 and §12.2 require it, and §7.4 accepts that
   noise explicitly.
2. **§5.1 step 5 does not apply to `BIND_SEND`.** The send side is unbound until
   the command succeeds, so there is no registered key to compare against. The
   signature can only prove possession of the key in the body, so
   `signer_key == send_key` is required — without it, anyone could bind a key
   they do not hold, which is §7.4's theft with no work at all. Reported as
   `ERR_NO_ACCESS`.
3. **`CREATE_CONTACT_QUEUE` self-authentication.** §12.2 does not restate
   §6.2's `signer_key == recv_key` rule, but the command has the same shape
   (a `recv_key` in the body, a zero transcript address, no existing queue), so
   the same rule is the only thing that authenticates it. Applied.
4. **Wrong transcript address for a command is `ERR_MALFORMED` (fatal).** A
   `CREATE_QUEUE` with a nonzero address, or a `READ` with a zero one, is not a
   state disagreement between honest peers — it is a frame no conforming client
   emits.
5. **A signed command arriving unsigned is `ERR_BAD_SIGNATURE` (non-fatal); an
   unsigned command arriving signed is `ERR_MALFORMED` (fatal).** The first is a
   well-formed frame that failed to authorize. The second carries an
   authenticator §5.1 has no path to check, so accepting it would mean carrying
   bytes nothing verifies.
6. **Dispatching to the wrong handler is `ERR_INTERNAL`.** §10's code 21 is the
   one that carries no detail, which is right for a relay describing its own
   fault to an unauthenticated caller.
7. **`UNSUBSCRIBE`'s response is empty.** §6.2's "empty bodies" sentence is
   about the requests (it then specifies a `SubscribeResponse`). There is no
   queue state an unsubscribe could report that `SUBSCRIBE` did not, and empty
   is the only reading that invents nothing.
8. **Queue indices start at 0**, and §8.2's bound is `next_index`, not what is
   stored — "the highest index the relay has **ever** appended". A drained queue
   still refuses a pre-ack (property-tested).
9. **The timestamp window boundary is inclusive.** Exclusive would make a client
   whose offset is exactly the published skew fail intermittently, and the
   number is a policy value rather than a threshold.
10. **A requested TTL of `0` means "no preference"** and gets the published
    default. §7.7 has both a requested value and a default and never says how
    they meet; a zero-second TTL is not a policy any client can want, so zero is
    the only value free to carry the meaning. The 30-day ceiling is applied last
    and unconditionally, even over a relay's own configured maximum.
11. **`channel_binding_mode: tls-exporter` with a 32-zero-byte binding is
    refused.** A genuine exporter output of zero has probability 2^-256, so
    treating zero as the "absent" sentinel costs nothing and removes an
    ambiguity that would otherwise surface as an obscure signature failure much
    later.
12. **`contact_queues_enabled` requires `contact_append_pow`.** §12.3 says a
    relay MUST enforce all four caps and proof of work is one of them; offering
    contact queues without it is an unsigned, unmetered write endpoint open to
    the internet. Enforced in `validate`.
13. **`max_frame_bytes >= largest padding bucket`** is checked as a necessary
    (not sufficient) consistency condition: a relay whose frame cap cannot carry
    its own largest bucket has published a set no client can use.
14. **`pending` is index-space.** TTL expiry (§7.7) removes stored messages
    without moving the watermark, so this crate's `pending` is an upper bound and
    says so; a relay that expires messages must answer from storage. Modelling
    storage in a crate that has none would have been the worse lie.
15. **`QueueAdvert` has no wire encoding.** §7.2 shows it as a named-field object
    and never gives it a `tls_codec` structure, because a relay never sees one.
    The type exists to carry `(relay_url, relay_id, send_addr)` into the §5.2
    check; inventing an encoding would have been inventing protocol.
16. **`PowStamp::empty()` for "no stamp"** — already `f2z-codec`'s reading, and
    reused rather than re-litigated.

## Dependency notes

- `ed25519-dalek 3` (BSD-3-Clause, already permitted by `rs/deny.toml`) with
  `default-features = false, features = ["zeroize"]`. Dropping `fast` drops
  curve25519-dalek's precomputed tables — ~30 KiB of table against a handful of
  signatures per connection, and this graph reaches wasm32 where the bundle is
  the scarce resource.
- `subtle` for constant-time key comparison (§10). Already in the graph.
- `tls_codec` directly, for the zero-byte `Empty` body type.
- `f2z-codec` is declared with `version` beside `path`: `rs/deny.toml` sets
  `wildcards = "deny"` and a bare path dependency is a wildcard requirement.
- **cargo-deny reports two `digest` versions** (0.10 via `blake2`, 0.11 via
  `sha2`/`curve25519-dalek`). It is a `warn`, and it is two versions of a *trait
  crate*, not two copies of a hash implementation. Recorded rather than silenced;
  it resolves when `blake2` publishes against `digest 0.11`.

## Verification

All run locally from the worktree, all green:

```
scripts/check-rust-fmt.sh    --root rs                      # 3 crates formatted
scripts/check-rust-clippy.sh --root rs                      # -D warnings, clean
scripts/check-rust-deny.sh   --root rs --config rs/deny.toml # advisories/bans/licenses/sources ok
  --manifest rs/crates/f2z-relay-proto/Cargo.toml           # pins agree
node scripts/check-workflow-gates.mjs                       # 2 gates, 56 jobs
node scripts/check-github-actions-pins.mjs                  # 178 refs pinned
cd rs && cargo test --locked --all-targets                  # 163 tests, 0 failed
cd rs && cargo test --locked --doc                          # 2 doctests
cd rs && cargo build --target wasm32-unknown-unknown --lib \
  -p f2z-codec -p f2z-relay-proto                           # ok
```

New tests: 79 unit, 8 redaction, 6 property (93 of the 163).

Re-verified after rebasing onto `origin/main` at cb9756e, which had moved under
this branch with four `rs/` changes (#571's `akd` floor and README correction,
#572's `Canonicalized` `Debug` fix). The only conflict was `rs/README.md`, where
#571's text on `check-rust-toolchain.sh` **not** discovering manifests is kept
verbatim — it says exactly why this PR had to add a `--manifest` flag by hand.

The redaction test follows the trap `f2z-codec` documented — `tls_codec`'s byte
vectors print a decimal byte list containing no hex, so a hex-only assertion
passes while everything leaks. Every structure is checked against lowercase hex,
uppercase hex, base64url, a full decimal list, a four-byte decimal prefix, and
any long hex-looking run. One test asserts the *opposite* for the capability
document: §11.1's operator block must stay readable.

CI changes: `rs / wasm32` now builds both crates (spelled out rather than
`--workspace`, so a future AGPL server binary that pulls tokio does not silently
join the wasm build), and the toolchain check registers the new manifest.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -74,7 +74,8 @@ jobs:
           scripts/check-rust-toolchain.sh --self-test
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
-            --manifest rs/crates/f2z-codec/Cargo.toml
+            --manifest rs/crates/f2z-codec/Cargo.toml \
+            --manifest rs/crates/f2z-relay-proto/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -283,17 +284,23 @@ jobs:
           key: rs-wasm
 
       # ADR 0001 requires one Rust core shared by ZUULI and the web client, so
-      # f2z-codec reaching wasm32 is a protocol-level requirement rather than a
-      # nice property. It is enforced by this job and not by convention, because
-      # "we kept it portable" stops being true the first week nobody checks. The
-      # first accidental `use std::` or tokio dependency fails here.
-      - name: Build f2z-codec for the browser target
+      # these crates reaching wasm32 is a protocol-level requirement rather than
+      # a nice property. It is enforced by this job and not by convention,
+      # because "we kept it portable" stops being true the first week nobody
+      # checks. The first accidental `use std::` or tokio dependency fails here.
+      #
+      # Every crate the clients link belongs on this line. It is spelled out
+      # rather than `--workspace` so that a future server binary — which will be
+      # AGPL, will pull tokio, and will never reach a browser — does not
+      # silently join the wasm build and then fail it.
+      - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
           set -euo pipefail
           cargo build --locked --lib \
             --target "${{ steps.rust_toolchain.outputs.target }}" \
-            -p f2z-codec
+            -p f2z-codec \
+            -p f2z-relay-proto
 
   # Branch protection should require this stable context, `rs / gate`. It
   # succeeds only when the change detector and every applicable rs job
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -35,7 +35,7 @@ version = "0.10.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "46502ad458c9a52b69d4d4d32775c788b7a1b85e8bc9d482d92250fc0e3f8efe"
 dependencies = [
- "digest",
+ "digest 0.10.7",
 ]
 
 [[package]]
@@ -48,10 +48,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d2f6c7dbe95a6ed67ad9f18e57daf93a2f034c524b99fd2b76d18fdfeb6660aa"
+dependencies = [
+ "hybrid-array",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "crypto-common"
@@ -64,14 +82,82 @@ dependencies = [
 ]
 
 [[package]]
+name = "crypto-common"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce6e4c961d6cd6c9a86db418387425e8bdeaf05b3c8bc1411e6dca4c252f1453"
+dependencies = [
+ "hybrid-array",
+]
+
+[[package]]
+name = "curve25519-dalek"
+version = "5.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b5eed333089e2e1c1ac8c6c0398e5e2497b4c9926ca6d0365ed1e099afa5bc23"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "curve25519-dalek-derive",
+ "digest 0.11.3",
+ "fiat-crypto",
+ "rustc_version",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
+name = "curve25519-dalek-derive"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
 dependencies = [
- "block-buffer",
- "crypto-common",
+ "block-buffer 0.10.4",
+ "crypto-common 0.1.7",
  "subtle",
+]
+
+[[package]]
+name = "digest"
+version = "0.11.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f1dd6dbb5841937940781866fa1281a1ff7bd3bf827091440879f9994983d5c2"
+dependencies = [
+ "block-buffer 0.12.1",
+ "crypto-common 0.2.2",
+]
+
+[[package]]
+name = "ed25519"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "29fcf32e6c73d1079f83ab4d782de2d81620346a5f38c6237a86a22f8368980a"
+dependencies = [
+ "signature",
+]
+
+[[package]]
+name = "ed25519-dalek"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6ebaa1a2bf1290ab3bfe5a7b771d050ebffab2711c19a81691c683a5144a25de"
+dependencies = [
+ "curve25519-dalek",
+ "ed25519",
+ "sha2",
+ "subtle",
+ "zeroize",
 ]
 
 [[package]]
@@ -94,10 +180,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-relay-proto"
+version = "0.1.0"
+dependencies = [
+ "ed25519-dalek",
+ "f2z-codec",
+ "proptest",
+ "subtle",
+ "tls_codec",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "da7c62ceae207dd37ea5b845da6a0696c799f85e97da1ab5b7910be3c1c80223"
+
+[[package]]
+name = "fiat-crypto"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "64cd1e32ddd350061ae6edb1b082d7c54915b5c672c389143b9a63403a109f24"
 
 [[package]]
 name = "fnv"
@@ -136,6 +239,15 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi 6.0.0",
+]
+
+[[package]]
+name = "hybrid-array"
+version = "0.4.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "707114b52a152fa7bdb290cd7cd5912d9467273b6d74e21b8d81aca1f8533f6b"
+dependencies = [
+ "typenum",
 ]
 
 [[package]]
@@ -274,6 +386,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d6f6ff9a378485b298a5286656da665ba74413d36db0979633275d2e708145d4"
 
 [[package]]
+name = "rustc_version"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfcb3a22ef46e85b45de6ee7e79d063319ebb6594faafcf1c225ea92ab6e9b92"
+dependencies = [
+ "semver",
+]
+
+[[package]]
 name = "rustix"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -297,6 +418,29 @@ dependencies = [
  "tempfile",
  "wait-timeout",
 ]
+
+[[package]]
+name = "semver"
+version = "1.0.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a7852d02fc848982e0c167ef163aaff9cd91dc640ba85e263cb1ce46fae51cd"
+
+[[package]]
+name = "sha2"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "446ba717509524cb3f22f17ecc096f10f4822d76ab5c0b9822c5f9c284e825f4"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "digest 0.11.3",
+]
+
+[[package]]
+name = "signature"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28d567dcbaf0049cb8ac2608a76cd95ff9e4412e1899d389ee400918ca7537f5"
 
 [[package]]
 name = "subtle"
--- a/rs/Cargo.toml
+++ b/rs/Cargo.toml
@@ -35,6 +35,20 @@ expect_used = "deny"
 [workspace.dependencies]
 blake2 = { version = "0.10", default-features = false }
 tls_codec = { version = "0.4", default-features = false, features = ["derive"] }
+# Ed25519, `WIRE.md` §5. `default-features = false` drops `std` and `fast`:
+# `fast` is curve25519-dalek's precomputed-tables, which trades roughly 30 KiB
+# of table for signing speed, and this graph reaches wasm32 for the browser
+# client (ADR 0001) where the bundle is the scarce resource and the workload is
+# a handful of signatures per connection, not a batch verifier. `zeroize` is
+# kept: it is the only feature that touches what happens to a secret key's
+# bytes when it is dropped.
+ed25519-dalek = { version = "3", default-features = false, features = [
+  "zeroize",
+] }
+# `WIRE.md` §10: the existence-oracle rule requires key comparisons to be
+# constant time. Already in the graph beneath ed25519-dalek; named here so the
+# dependency is a decision rather than an accident of resolution.
+subtle = { version = "2", default-features = false }
 # Dev-dependency only. It builds for the host test runner, never for
 # wasm32-unknown-unknown, so its `std` default costs this crate nothing: the
 # library itself is `no_std` and the wasm job builds `--lib`.
--- a/rs/README.md
+++ b/rs/README.md
@@ -75,7 +75,8 @@ decided is `wallet/rust-toolchain.toml`, and
 
 ```
 scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
-                                --manifest rs/crates/f2z-codec/Cargo.toml
+                                --manifest rs/crates/f2z-codec/Cargo.toml \
+                                --manifest rs/crates/f2z-relay-proto/Cargo.toml
 ```
 
 fails the pull request the moment the two disagree. `.github/workflows/rs.yml`
@@ -108,8 +109,14 @@ prevent.
 | Crate | What it is |
 |---|---|
 | [`crates/f2z-codec`](./crates/f2z-codec) | The canonical encoding layer of `WIRE.md`: `tls_codec` wrappers for every wire structure, the domain-separated signing transcript (§5), the redacting newtypes, and the padding-bucket validator (§9). `no_std` + `alloc`, `#![forbid(unsafe_code)]`, no I/O, no async runtime, and it builds for `wasm32-unknown-unknown`. |
+| [`crates/f2z-relay-proto`](./crates/f2z-relay-proto) | The protocol layer above it: signed-command construction and verification in §5.1's exact order, the timestamp window and fail-closed seen-set (§5.5), the queue lifecycle and ACK arithmetic (§7, §8), the capability document (§11), the `HELLO` proof of possession (§5.2), and §4.3's typed in-flight window. Same constraints — `no_std` + `alloc`, no I/O, no clock, no randomness, and it builds for `wasm32-unknown-unknown`. |
 
-`f2z-codec` is separate from everything that will sit on top of it for three
+**The relay and the clients link both.** That is the licence boundary in
+practice: everything here is MIT because a third-party relay, ZUULI and the WASM
+web client all compile the same rules, and a rule that two implementations
+disagree about is how ciphertext gets deleted before it is read.
+
+`f2z-codec` is separate from everything that sits on top of it for three
 reasons, and each is enforced by a test rather than by intent:
 
 1. **Re-encode equality** (`WIRE.md` §3.3) is implemented once. Every received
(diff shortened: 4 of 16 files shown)


skyl:
Merge-train hold: queue behind #556. #556 is the sole active merge candidate on exact current trunk and its final cross-platform matrix is already running; merging this new rs feature would invalidate it again. Keep draft until #556 lands, then integrate the squash and rerun exact-head CI before review/merge.

skyl:
Temporary merge-train hold: #556 is the sole active merge candidate. Keep this PR draft until #556 lands; then integrate current trunk and rerun exact-head CI.

skyl:
## Adversarial review — `f2z-relay-proto` (post-merge; #583 merged as `40af3a5` with 0 reviews)

Reviewed at `38c55ea`, in a throwaway worktree, toolchain 1.97.1, `cargo` run from `rs/`.

**Nothing blocking is wrong in the production code.** I went rule by rule through §4.3, §5.1–§5.5, §6.1–§6.3, §7.3/§7.6/§7.7, §8.1–§8.3, §9, §10, §11.1/§11.3, §12.2/§12.3 and §13.1 and found no divergence from `WIRE.md`. The four findings below are all **test** gaps and **documentation** overclaims, not defects — but one of them sits directly under the PR's own headline security claim, so it is worth fixing before anything links this crate.

---

### Mutation test — 33 mutations landed and compiled, 29 caught, 4 survived

Each mutation was applied to an exact line, the line was re-read to confirm the edit landed before judging, then `cargo test --locked --all-targets -p f2z-relay-proto` was run and the source restored.

| # | Mutation | Result |
|---|---|---|
| M01b | `key.rs:72` `verify_strict` → fall back to plain `verify` | **NOT CAUGHT** |
| M02 | `command.rs:668` drop the seen-set check (`?` → `.ok()`) | caught — `a_replayed_frame_is_refused_on_the_second_presentation` |
| M03 | `command.rs:679` skip signature verification | caught — `a_signature_made_for_one_relay_fails_at_another`, `the_request_id_is_covered_by_the_transcript` |
| M04 | `command.rs:682` skip §6's transcript-address rule **in `verify`** | **NOT CAUGHT** |
| M05 | `command.rs:683` skip `C::check` (per-command self-auth) **in `verify`** | **NOT CAUGHT** |
| M06 | `command.rs:656` signed-arrives-unsigned `ERR_BAD_SIGNATURE` → `ERR_MALFORMED` | caught — `a_signed_command_that_arrives_unsigned_is_not_fatal` |
| M07 | `command.rs:687` skip the §9 padding-bucket check | caught — `an_append_outside_the_padding_buckets_is_bad_size` |
| M08 | `command.rs:326` `BindSend::check` `keys_equal(..)` → `true` | caught — `bind_send_must_be_signed_by_the_key_it_binds` |
| M09 | `queue.rs:187` `authorize_send` `ERR_UNAVAILABLE` → `ERR_NO_ACCESS` | caught — `send_side_refusals_all_look_the_same` |
| M10 | `queue.rs:170` `authorize_recv` `ERR_NO_ACCESS` → `ERR_NOT_PERMITTED` | caught — `the_recv_side_uses_the_existence_oracle_code` |
| M11 | `queue.rs:265` §8.2 pre-ack bound off-by-one (`> highest` → `> highest+1`) | caught — `acking_past_the_highest_appended_index_is_refused` |
| M12 | `queue.rs:348` drop the unconditional 30-day TTL ceiling | caught — `ttls_are_clamped_and_the_thirty_day_ceiling_is_absolute` |
| M13 | `inflight.rs:128` `max_inflight` off-by-one (`>=` → `>`) | caught — `ids_are_nonzero_unique_and_bounded` |
| M14 | `inflight.rs:171` drop the `request_id`/ticket correlation check | caught — `a_response_on_the_wrong_id_is_refused` |
| M15 | `capabilities.rs:310` drop §12.3's contact-queues-require-PoW rule | caught — `contact_queues_without_proof_of_work_are_refused` |
| M16 | `capabilities.rs:428` anti-replay soundness `2×` → `1×` skew | caught — `a_seen_set_window_shorter_than_twice_the_skew_is_refused_by_default` |
| M17 | `hello.rs:223` drop the `relay_id`-is-derived check | caught — `a_relay_id_that_is_not_the_digest_of_the_key_is_caught_without_an_advert` |
| M18 | `hello.rs:230` drop the advert `relay_id` comparison (§5.2 substitution) | caught — `a_substituted_relay_is_caught_by_the_advert` |
| M19 | `replay.rs:254` seen-set evicts instead of failing closed | caught — `a_full_set_refuses_rather_than_evicting` |
| M20 | `replay.rs:85` timestamp window widened 2× | caught — `the_timestamp_window_is_checked_before_the_signature`, `the_window_boundary_is_inclusive_on_both_sides` |
| M21 | `capabilities.rs:289` drop `max_frame_bytes >= largest bucket` | **NOT CAUGHT** |
| M22 | `queue.rs:206` allow `BIND_SEND` on a contact queue | caught — `a_contact_queue_can_never_be_bound` |
| M23 | `capabilities.rs:201` drop `relay_identity_pk == proven_identity` | caught — `a_document_verifies_only_against_the_key_hello_proved` |
| M24 | `command.rs:272` `CreateQueue::check` `keys_equal(..)` → `true` | caught — `create_queue_must_be_signed_by_the_key_it_registers` |
| M25 | `command.rs:646` drop the command-code dispatch check | caught — `dispatching_to_the_wrong_handler_is_a_relay_fault` |
| M26 | `command.rs:725` accept an unsigned command that arrived signed | caught — `an_unsigned_command_that_arrives_signed_is_fatal` |
| M27 | `replay.rs:227` expiry boundary `>` → `>=` | caught — `an_entry_survives_for_exactly_its_retention` |
| M28 | `queue.rs:242` `READ` resurrects below the watermark | caught — `a_read_below_the_watermark_starts_at_the_watermark` |
| M29 | `capabilities.rs:255` drop `relay_id`-is-derived from `validate()` | caught — `a_relay_id_that_is_not_the_digest_of_the_key_is_refused` |
| M30 | `hello.rs:251` `tls-exporter` accepts a 32-zero-byte binding | caught — `a_relay_that_claims_an_exporter_we_do_not_have_is_refused` |
| M31 | `queue.rs:274` ACK-from-no-watermark off-by-one | caught — `acknowledgement_is_cumulative_monotone_and_idempotent`, `the_old_queue_is_retired_only_after_the_switch_and_the_drain` |
| M32 | `command.rs:406` allow signing an unsigned command | caught — `unsigned_commands_cannot_be_signed_and_signed_ones_cannot_be_unsigned` |
| M33 | `capabilities.rs:435` drop the padding-superset refusal | caught — `a_client_refuses_a_relay_that_cannot_carry_its_own_sizes` |

(A 34th, replacing `verify_strict` with a bare `verify` call, does not compile — `ed25519-dalek 3` puts it behind the `Verifier` trait. M01b is the compiling form: `.verify_strict(..).or_else(|_| ed25519_dalek::Verifier::verify(&self.0, message, &signature))`.)

That is a strong result. 29 of 33 real behaviour changes die, including every error-code swap, every boundary, and every §5.1 step that carries a signature. The four survivors follow.

---

## F1 — `small_order_keys_are_rejected_by_strict_verification` does not test strict verification

`src/key.rs:214-229`. This is the test named for the PR's own headline claim ("**`verify_strict` everywhere**… a key that verifies everything is a key that steals every queue it is registered against"). It asserts that an order-8 public key with an **all-zero** signature is rejected. Both `verify` and `verify_strict` reject that, so the test passes identically either way — M01b proves it: the whole 93-test suite is green with a non-strict fallback wired in.

The forgery the test is reaching for does exist. Probe run against `ed25519-dalek 3.0.0` in this crate's graph:

```
identity R=identity: PLAIN ACCEPTS msg#0  strict=false
identity R=identity: plain accepted 64/64 messages
order2   R=A:        plain accepted 33/64 messages
order8   R=A:        plain accepted  5/64 messages
```

With `A = [0x01, 0x00 × 31]` (the identity point) and `signature = A || [0u8; 32]`, plain `verify` accepts **every one of 64 distinct messages** and `verify_strict` rejects every one. That is a universal-forgery public key: registered as a queue's `recv_key` it makes §5.1 step 5 vacuous for that queue.

Replace the constant and the signature:

```rust
#[test]
fn a_universal_forgery_key_is_rejected_only_because_verification_is_strict() {
    let mut identity = [0u8; 32];
    identity[0] = 1;                       // the identity point: order 1
    let key = VerifyingKey::from_public_key(&PublicKey::new(identity)).unwrap();
    let mut sig = [0u8; 64];
    sig[..32].copy_from_slice(&identity);  // R = identity, s = 0
    let forged = Signature::new(sig);
    // Plain Ed25519 verification accepts this for *any* message; strict does not.
    for message in [&b"free2z/relay/v1/cmd"[..], b"anything at all", b""] {
        assert_eq!(
            key.verify(message, &forged),
            Err(ProtoError::Wire(ErrorCode::BadSignature))
        );
    }
}
```

That test kills M01b. The current one kills nothing.

## F2 — `CommandVerifier::verify`'s §5.1 step-5 half is completely untested

M04 and M05 both survive. `check_transcript_address` and `C::check` are called in **two** places — `SignedCommand::create` (`command.rs:412`, `:424`) and `CommandVerifier::verify` (`:682`, `:683`) — and every test asserts only on the first:

- `create_queue_must_be_signed_by_the_key_it_registers` (`:961`) → `SignedCommand::create`
- `create_queue_must_carry_the_zero_address` (`:986`) → `SignedCommand::create`
- `bind_send_must_be_signed_by_the_key_it_binds` (`:1022`) → `SignedCommand::create`

`create` derives `signer_key` from the signing key, so it *cannot express* the hostile frame. A hostile client has no such constraint — it signs a well-formed transcript with its own key while the body names someone else's. That is exactly the §7.4 "theft with no work at all" the PR body says this rule closes, and the rule that closes it has no relay-side test.

The code is correct today — I hand-built the frames and confirmed it:

```
BIND_SEND signer!=send_key on the VERIFY path      -> Err(Wire(NoAccess))
CREATE_QUEUE with a nonzero address on VERIFY path -> Err(Wire(Malformed))
```

and with M04+M05 applied, the same two frames return `Ok(())`. So a test that builds the `Request` by hand — `AuthContext { address, signer_key: attacker.public_key(), .. }`, `transcripts.build(..)`, sign, `Request::new(code, CommandAuth::Signed(auth), encoded)` — and feeds it to `verifier().verify::<ops::BindSend>(..)` kills both mutations. Worth adding for `CreateContactQueue` too, whose `check` (`:296`) is an explicit interpretation call (ambiguity #3) and has no verify-path test at all.

This is the load-bearing gap: the crate is a *shared* layer, and the half a relay depends on is the half with no coverage.

## F3 — `max_frame_bytes >= largest bucket` has no test (M21)

`capabilities.rs:289`. Minor — it is a self-imposed consistency check (design decision #13), not a `WIRE.md` MUST — but it is asserted in the PR body as enforced and nothing enforces the enforcement.

## F4 — the seen-set records **before** the signature is checked, which is an unauthenticated fail-closed DoS

`command.rs:667-668` calls `SeenSet::observe`, which both checks *and inserts* (`replay.rs:245-261`), at step 3 — before step 4's signature verification at `:679`. §5.1 does mandate that ordering for the *check*; it does not mandate it for the *insert*, and §5.5 says the set holds pairs "observed".

The consequence is concrete: an unauthenticated attacker sends frames with random `(signer_key, nonce)` and a timestamp inside the window. Each one fails at step 4 but has already consumed a slot, and §5.5 forbids eviction, so at `antireplay_seen_max` entries the relay answers `ERR_BACKPRESSURE` to **every signed command from every client** until they age out — 4 minutes at the crate's own default retention. At the doctest's `SeenSet::new(240_000, 65_536)` that is ~65k cheap frames for a full-relay outage, renewable.

`observe` being one call forecloses the safer shape. A relay cannot check at step 3 and insert at step 6 without reaching around this API. Suggest splitting it (`contains_live(now, key) -> Result<()>` + `record(now, key) -> Result<()>`) and letting the caller place the insert, or at minimum documenting the exposure at the call site. This is a design question, not a bug — but it is the kind that is much cheaper to answer now than after a relay is built on it. Note also that `antireplay_seen_max` is published per §5.5 yet has **no field in §11.1's `Capabilities`**, so a client cannot see the bound it is subject to; that is a pre-existing `WIRE.md` gap rather than this PR's.

## F5 — `defaults()` invents values, including a PoW difficulty, in a function documented as publishing the spec's

`capabilities.rs:470-530`, doc comment: *"carrying every default `WIRE.md` publishes"*. Most are real (`max_frame_bytes` 1 MiB §4.1, `ws_ping_interval_seconds` 25 §2.4, `handshake_timeout_ms` 10 000 §2.5, `max_inflight` 32 §4.3, `clock_skew_ms` 120 000 §5.5, the TTL table §7.7, the contact caps §12.3). These are not:

- `difficulty_bits: 20` — **twice** (`:513`, `:521`). `grep -n 'difficulty' docs/e2ee/WIRE.md` returns five hits, none of which is a number; §12.4 says only *"We have chosen a difficulty that is a nuisance"* and defers calibration to `ARCHITECTURE.md` §13-N. This is #550's *"no PoW difficulty number exists anywhere"* being answered by fiat in code. Not a re-file — but the number will be read as authoritative by the first relay that ships it, and #550 should be the place it is decided.
- `challenge_ttl_ms: 60_000` (×2), `min_message_ttl_seconds: 60`, `min_idle_ttl_seconds: 3_600`, `max_queue_messages: 4_096`, `max_queue_bytes: 64 MiB` — no spec value.
- `antireplay_window_ms: 240_000` — correctly flagged in a comment as invented.

Fix is one doc line: say which are the document's and which are this crate's, and put a `#550` reference next to `difficulty_bits`.

## F6 — the new crate sits outside #572's structural `Debug` scan

`f2z-codec/tests/redaction.rs:550-570` walks `Path::new(env!("CARGO_MANIFEST_DIR")).join("src")` — `f2z-codec/src` only. `f2z-relay-proto/tests/redaction.rs` is eight instance tests over named types (`SigningKey`, `SeenSet`, `SignedCommand`, `Verified`, `QueueState`, `RelaySession`, …) plus the good inverse test that the capability document *stays* readable. There is no `no_type_derives_debug_while_holding_raw_bytes` equivalent.

No current violation — `grep -n 'Vec<u8>\|\[u8;\|&\[u8\]\|TlsByteVec' src/*.rs` finds only function signatures, no fields; every field goes through an `f2z-codec` newtype that redacts itself. So this is preventative: the next struct added here that holds bytes directly is not caught the way #572 made sure it would be in the codec. Lifting the scan into a shared helper (or duplicating it with the manifest dir parameterised) closes it.

---

## Verified rather than trusted

| Claim | Result |
|---|---|
| `#![forbid(unsafe_code)]` | `src/lib.rs:114`, present |
| `no_std` + `alloc` | `src/lib.rs:113`, `:126`; `grep 'use std::' src/` → no hits |
| wasm32 | `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec -p f2z-relay-proto` → `Finished` |
| `cargo test --locked --all-targets` | 79 unit + 6 property + 8 redaction, 0 failed |
| `scripts/check-rust-fmt.sh --root rs` | `All 3 Rust crate(s) under rs/ are formatted (rustfmt from 1.97.1)` |
| `scripts/check-rust-clippy.sh --root rs` | `All 3 Rust crate(s) under rs/ are clippy-clean` |
| `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` | `advisories ok, bans ok, licenses ok, sources ok` |
| `rs/deny.toml` `akd`/`akd_core` `<0.13.0` bans | `git diff origin/main -- rs/deny.toml` → empty; entries at `:121-140` intact |
| Dependencies | `ed25519-dalek 3.0.0` → `curve25519-dalek 5`, `ed25519 3`/`signature 3`, `sha2 0.11`, `subtle 2.6.1`, `zeroize 1.9`; plus `f2z-codec`, `tls_codec 0.4.2`. Nothing surprising; no `getrandom`, no `std`-only crate, no async. The `digest 0.10`/`0.11` split the PR body records is real and is a `warn` on a trait crate. |
| Re-derivation of `f2z-codec` | **None.** `grep -n '"free2z/'` over `src/` returns doc comments and one `b"free2z/relay/v1/dummy"` — a throwaway message for `dummy_verify`, not a protocol label. Transcript (`TranscriptBuilder`), labels (`hash::relay_id`, `hash::hello_proof_message`, `hash::capabilities_digest`), buckets (`PaddingBuckets`) and error codes (`ErrorCode`) are all used, not restated. `MAX_MESSAGE_TTL_SECONDS` re-exports rather than copies. |
| #553 registration | **Done** — `.github/workflows/rs.yml:74-78` adds `--manifest rs/crates/f2z-relay-proto/Cargo.toml`, and `check-rust-toolchain.sh` with both manifests passes. #553 stays open as a class of problem, but this crate is not an instance of it. |
| `.github/workflows/zuuli.yml` (contended by #556) | untouched — `git diff --name-only origin/main -- .github/workflows/zuuli.yml` → 0 files |

## Spec-gap positions the code takes

Reported, not re-filed:

- **#550** — `authorize_send` (`queue.rs:184-189`) resolves the §10 code 10 / code 15 contradiction in §6.3's favour, returning `ERR_UNAVAILABLE` for every send-side refusal. Correct call: §6.3's reason (a bound sender must not learn queue state) dominates, and collapsing absent into unavailable creates no oracle. `BIND_SEND` self-authenticates on `signer_key == send_key` (`command.rs:324-331`) with `ERR_NO_ACCESS`, which is the only thing that can authenticate a command whose §5.1 step 5 has nothing to look up. `difficulty_bits: 20` is the one place a gap gets closed by fiat (F5).
- **#552** — `send_retry_max` and `ERR_QUOTA` are correctly absent (there is nothing to implement). `next_index` base is decided as 0 (`queue.rs:126-133`), argued from §8.2's "ever appended"; property-tested by `a_drained_queue_still_refuses_a_pre_ack`. §11.1's "entire policy" claim is not relied on.
- **#585** — `PowStamp::empty()` is reused from `f2z-codec` rather than re-litigated, so the "empty vs zeroed" 56-byte desync is not decided here.
- The PR's own §11.1 finding (`antireplay_window_ms` must be ≥ `2 × clock_skew_ms`) is right, and putting it in `ClientPolicy` rather than `validate` is the correct place: the document is what the spec permits; it is the spec that is short a constraint.

---

## Recommendation

The crate is sound. Verification order, transcript binding, error-code selection, the seen-set, ACK arithmetic, TTL clamping and the capability checks all match `WIRE.md`, and 29 of 33 real mutations die — this is not a suite that whitelists its own defects.

Already merged, so: **land F1 and F2 as a follow-up PR before anything links this crate.** F1 is a test whose name asserts a property it does not test, under the PR's own headline security claim. F2 leaves the relay-side half of §5.1 step 5 — the half that stops `BIND_SEND` theft — with zero coverage while the client-side half has three tests. F3–F6 are ordinary follow-ups. Neither F1 nor F2 is a live vulnerability; both are the guard rails around one, and the repo's whole argument is that guard rails get tested.

PR #584, opened 2026-08-24
feat(kt): add f2z-kt-core, the key-transparency crate (KT.md v1)

skyl:
Implements **[`docs/e2ee/KT.md`](https://github.com/free2z/zuu/blob/main/docs/e2ee/KT.md) v1** as `rs/crates/f2z-kt-core` — the one crate the log server, the witness and the client all link (§11.4). A witness that verified with a different implementation than the log builds with would produce cosignatures that mean nothing, and §7.4's only structural defence against a lazy witness is that there are not two implementations to disagree.

Decided by [ADR 0013](https://github.com/free2z/zuu/blob/main/docs/e2ee/decisions/0013-key-transparency-log.md) (adopt `akd`, floor `>= 0.13.0`) and [ADR 0014](https://github.com/free2z/zuu/blob/main/docs/e2ee/decisions/0014-directory-key-rotation.md) (what authorizes an entry). Spiked in #544.

## What landed

| Module | `KT.md` |
|---|---|
| `entry` | §4.1/§4.4 — `DirectoryEntry(TBS)`, `DeviceCredential`, `DeviceRevocation`, `ContactEndpoint`, `EntryKind`, `EntryAuthorization`, `RotationProof`, `ResetAuthorization` |
| `submit` | §4.2/§4.3/§4.4 — **`validate_submission`**, the one door into the log |
| `sth` | §6 — `SignedTreeHead`, `LogView` (all eight monotonicity rules), `LogKeyTransition` |
| `cosign` | §7.2 — `WitnessCosignature`, the contradiction test |
| `witness` | §7.3/§8.3 — `WitnessSet`, `verify_threshold`, `AcceptedRoot`, `FaultKind`, `FaultReport` |
| `verify` | §8.1/§8.2 — client inclusion + key-history verification over `akd_core` |
| `auditor` | §7.1 step 4 / §7.4 — the witness append-only path over `akd::auditor`, and `WitnessState` |
| `receipt`, `descriptor`, `error`, `labels`, `types`, `sig` | §5.3, §9.1, §9.5, §6.2, §1.3 |

`f2z-codec` is reused rather than duplicated: re-encode equality (`decode_canonical`), `H(label, x)`, the length-prefixed vectors, and the `PublicKey`/`Signature`/`Digest`/`ShortBytes` redacting newtypes all come from it. What this crate adds is `LogId`, a charset-checked `Handle`, and a `<0..2^16-1>` `KemPublicKey`, following the same `opaque_fixed!` pattern — including the decimal-byte-list trap, which `tests/redaction.rs` checks specifically.

## The single most important thing

> **`akd` enforces none of our authorization rules.** It will commit any bytes to any label, and a log that skips §4.4 produces inclusion, history and append-only proofs that verify **perfectly** for entries nobody authorized.

So the crate is built out of **choke points, not helpers**. Every rule is the argument the next step needs:

| To get | You must first | Which requires |
|---|---|---|
| `AcceptedSubmission` — the only carrier of the `AkdLabel`/`AkdValue` pair `publish()` needs | `validate_submission` | every rule in §4.4, in order |
| `AcceptedRoot` — the only root a proof may be verified against | `verify_threshold` | ≥ *t* cosignatures from the caller's **own** witness set |
| an advanced `LogView` | `LogView::accept` | all eight of §6.3's rules, with **no skipping over an epoch gap** |
| an advanced `WitnessState` | `verify_append_only` | `akd`'s auditor actually having run |
| a successor log signing key | `LogView::accept_log_key_transition` | a **cosigned** announcement plus both signatures |

None has a public constructor and none has a bypass flag. `validate_submission` has no siblings — no `check_rotation`, no `verify_device_credentials` — so there is nothing for a server to route around the rules *to*. `PublishedEntry` is the same idea applied to the predecessor: a log cannot *assert* what the previous entry's `prev_entry_hash` or `directory_auth_pk` were, it must hold the previous `DirectoryEntry` and let `PublishedEntry::from_entry` recompute them.

A key change carrying only one of the two required signatures **cannot be represented**: `EntryAuthorization::KeyChange` has no variant without a `RotationProof`. Both tests (`a_key_change_signed_by_only_the_new_key_is_rejected`, `..._only_the_old_key_...`) attack the signatures instead.

## ⚠️ A gap in `KT.md` §4.4, reported rather than worked around

**§4.4's numbered list does not require a `same_key` entry to keep `identity_pk` unchanged, and without that rule the list has a hole ADR 0014 explicitly forbids.**

`same_key` is authorized by the `directory_auth_pk` **published in the previous entry**, alone. Rules 1–9 check the label, the log id, the version, the chain hash, the kind agreement, the rotation proof (for `key_change` only), the reset (for `platform_reset` only), the device credentials and the epoch uniqueness — none of them looks at `entry.identity_pk`. So a party holding only the previous `DirectoryAuthKey` can publish a `same_key` entry carrying a **brand-new `identity_pk`**: a key change with one signature, which ADR 0014 says the log MUST reject, and which the two-signature rule exists to prevent.

The rule is implied by ADR 0014's case naming ("a same-key update… the identity key is unchanged") but is not in the enumerated list an implementer is told to apply "in order". This crate enforces it (`ADDITIONAL RULE 1` in `submit.rs`, test `a_same_key_entry_must_not_change_the_identity_key`). **`KT.md` §4.4 should gain it as a numbered rule.**

Three smaller rules follow from ADR 0014's case analysis and are also enforced, each marked `ADDITIONAL RULE` at its call site:

2. A `key_change` or `platform_reset` MUST change `identity_pk` — otherwise one operation has two encodings inside a structure that is hashed, signed and committed.
3. A `platform_reset` MUST be refused when the previous entry set `no_reset`. ADR 0014 makes that flag a user's declaration that the reset path does not apply to their handle; §4.4 never checks it, and a flag nothing enforces is a comment.
4. A `ResetAuthorization` is bound to its handle, log, version and outgoing key — exactly as §4.4 rule 6 binds a `RotationProof`. Rule 7 names only the signature and the cooldown, which would leave a reset signed for `@alice` at version 3 replayable against `@bob`.

## Ambiguity calls (every one, as required)

1. **Version-1 authorization.** §4.4's table has no row for a handle's first entry, and `same_key`'s signer ("the previous entry's `directory_auth_pk`") does not exist. **Read as trust-on-first-registration:** at `entry_version == 1` the entry is self-signed by the `directory_auth_pk` it publishes. Nothing weaker is possible; what makes it safe is that the registration is what a client pins.
2. **`key_change` / `platform_reset` at version 1 are refused.** Both bind to "the previous entry's `identity_pk`". A platform-authority *registration* of a fresh handle would be a different operation with different consequences and is not invented here.
3. **`no_reset` accepts only 0 and 1.** §4.1 says `uint8` and ADR 0014 says "flag". Accepting 2 would mean two byte strings with one meaning inside a hashed, signed structure.
4. **§4.2's all-zero `prev_entry_hash` is read as an iff.** A version-1 entry with a non-zero predecessor hash claims a history that cannot exist; a later entry with an all-zero one claims to be a genesis it is not. Both are refused on the bytes alone.
5. **A `DeviceCredential`'s window must be non-empty and non-inverted** (`not_after_ms > not_before_ms`). §4.1 gives the fields and no ordering rule; a permanently-dead credential in an append-only record is not worth accepting.
6. **A `DeviceCredential`'s `identity_pk` must equal the entry's.** §4.4 rule 8 says the signature verifies *under this entry's* `identity_pk`; it does not say the embedded field must match. It must, because an MLS peer validating the credential in a `LeafNode` has no directory access and uses the embedded key — if the two could differ, peer and directory client reach different verdicts about the same device.
7. **§6.4's "at least one full epoch of separation"** is read as `effective_epoch >= announce_epoch + 2`. `announce_epoch + 1` is the very next epoch and leaves no epoch in between for anyone to see the announcement in.
8. **§6.3 rule 8 with fully identical contents is idempotent, not a fork.** Re-serving the head a verifier already holds is normal; only a *differing* root, size or timestamp at an equal epoch is fatal.
9. **An epoch gap returns `EpochGap` rather than being fetched.** The crate has no network. §6.3 rule 7's "MUST NOT skip" becomes: `accept` refuses, `accept_chain` walks a caller-fetched run link by link, and there is no third method.
10. **`WitnessCosignature::contradicts` is `false` across two different witnesses.** §7.2's non-repudiation claim is specifically that *one* witness signed two contradictory statements. Two witnesses disagreeing is evidence against the **log**, and is `LogView`'s `Fork`.
11. **`WitnessSet::new` refuses `t == 0` and `t > |set|`.** §8.3 fixes the rule and leaves the numbers to §13-Q; a threshold of 0 is the fail-open behaviour §8.3 forbids, and one larger than the set can never be met.
12. **`ErrorCode` mapping for verdicts §9.5 has no code for.** `Rollback`/`Fork`/`ChainBreak`/`EpochGap`/`VrfKeyChange`/`ThresholdUnmet`/`ProofInvalid`/`ValueMismatch`/`AppendOnlyFailure`/`HistoryIncomplete` are client- and witness-side verdicts *about* the log, never replies a log sends a submitter. They collapse onto `ERR_INTERNAL` rather than inventing codes §9.5 does not have.
13. **Label mismatch maps to `ERR_MALFORMED`.** §9.5 gives `ERR_MALFORMED` for "decode failure" and `ERR_UNSUPPORTED_VERSION` for an unknown `kt_version`; a wrong §6.2 constant is neither, and is closest to the first.
14. **`kt_versions` in `LogDescriptor` must contain this build's `KT_VERSION`, and `configuration` must be `1` (WhatsAppV1).** §3.2 calls the configuration permanent; a descriptor announcing another one announces a tree this build cannot verify.
15. **`LogDescriptor.log_id` must equal `H("free2z/kt/v1/log-id", genesis_log_pk)`.** §9.1 lists both fields and does not say they are checked against each other. A log publishing an unrelated pair is claiming an identity it cannot derive.
16. **`verify_lookup` does not verify authorization; `VerifiedEntry::verify_authorization` does.** §8.1 separates steps 5 and 6, and the context step 6 needs — the previous entry, the pinned reset authority — is the caller's. That call is `validate_submission` itself, so §4.4 has exactly one implementation, run by both the log and the client.
17. **`verify_key_history` takes the served entry bytes in the proof's order** (decreasing version, as `akd` emits them) and walks §8.2 step 4's chain oldest-first, starting either at version 1 or at a caller-supplied pin.
18. **`verify_append_only` additionally checks the proof's `epochs` against the heads'.** `audit_verify` checks only the count. Without this a log could serve a valid proof for a different range that happens to match the root hashes it was given.
19. **`sig::verify` uses `verify_strict`.** §7.2's accountability argument is that two conflicting cosignatures under one `witness_pk` are a contradiction *on their face*; under malleable verification "the witness signed two things" and "an attacker re-encoded one thing" stop being distinguishable.
20. **`Handle` renders in `Debug`; everything else redacts.** A handle is the one field of a directory entry whose entire purpose is to be public and readable, and its charset admits no control characters, so it cannot forge a log line.
21. **Both features are on by default.** `cargo deny` judges the default graph, so a floor that only exists under an optional feature is a floor nothing checks. The browser build is explicit: `--no-default-features --features verifier`.

## The `akd >= 0.13.0` floor

ADR 0013's stanza is in `rs/deny.toml`, verbatim, comment included. **Verified firing** — with `akd`/`akd_core` pinned to `=0.12.0` and the lockfile regenerated:

```
    ┌─ rs/deny.toml:125:9
    ├ akd v0.12.0
      └── f2z-kt-core v0.1.0

    ┌─ rs/deny.toml:129:9
```

…while, on the same downgraded lockfile:

```
advisories ok
```

That contrast is the whole of ADR 0013's argument: facebook/akd#495 has no RustSec or OSV advisory and never will, so `cargo audit` and `cargo deny advisories` pass on a vulnerable pin forever. **A reviewer who removes those entries has removed the floor.** The manifest and the lockfile were restored to `0.13.0` after the demonstration.

## Evidence

All run on the pinned `1.97.1`, via the repository's own scripts.

| Check | Result |
|---|---|
| `scripts/check-rust-fmt.sh --root rs` | `All 3 Rust crate(s) under rs/ are formatted` |
| `scripts/check-rust-clippy.sh --root rs` | `All 3 Rust crate(s) under rs/ are clippy-clean` (`-D warnings`, `--all-targets`) |
| `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` | `advisories ok, bans ok, licenses ok, sources ok` ×3; `All 3 Rust crate(s) under rs/ satisfy rs/deny.toml` |
| `cargo test --locked --all-targets` | 79 unit + 4 negative/proptest + 4 redaction, plus f2z-codec's 61 — **all green** |
| `cargo test --locked --doc` | green |
| `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-kt-core --no-default-features --features verifier` | **succeeds** |

### WASM size

An rlib is not a shippable number, so the client verifier was also measured the way #544 measured it: a size-optimised `cdylib` in a scratch crate outside the repo (`opt-level=3`, `lto=true`, `codegen-units=1`, `panic="abort"`, `strip=true`), exporting `verify_lookup`, `verify_key_history`, `verify_threshold`, `validate_submission`, `WitnessSet::new` and `LogView::pin` so none of them is stripped.

| | raw | gzip -9 | brotli -q11 |
|---|---:|---:|---:|
| **`f2z-kt-core` client verifier `.wasm`** | **281,679 B** (275 KB) | 92,976 B | **70,282 B (69 KB)** |
| #544's `akd_core` alone, `opt-level=3`, for comparison | 183,375 B | 62,023 B | 48,116 B |

**~69 KB brotli over the wire**, of which roughly two thirds is `akd_core` itself. The delta over the spike's figure buys Ed25519 `verify_strict`, the whole `tls_codec` envelope, §4.4's authorization engine and §6.3/§8.3's verifiers — i.e. everything the browser needs to resolve a handle and check it was authorized, not just to check a Merkle proof. Still smaller than most single icon fonts, and the spike's timings (1.11 ms `lookup_verify`, 2.63 ms `key_history_verify` over three versions, in V8) are unaffected: this crate adds one Ed25519 verification per signature to those paths.

No new licences were needed: the whole `akd` tree resolves to MIT / Apache-2.0 / BSD-3-Clause, already in `rs/deny.toml`'s allow list.

One unrelated policy interaction, fixed properly rather than by weakening the policy: a bare `path` dependency has no version requirement, so `wildcards = "deny"` flagged `f2z-codec`. The workspace dependency now carries `version = "0.1.0"` alongside `path`.

### Adversarial tests

The §4.4 suite is the point of the crate, so it is listed:

- wrong `prev_entry_hash`; version gap; version replay; registration not at version 1; a later version with no published predecessor; a predecessor for the wrong handle
- a key change signed by only the new key; only the old key; a correctly-signed `RotationProof` for **another handle**; a key change that changes no key; a key change at version 1
- a `same_key` entry smuggling a new `identity_pk`
- a reset one millisecond before `effective_at_ms`; a reset declaring a shorter cooldown than published (refused even at `u64::MAX`); a reset signed by a valid key that is not the pinned authority; a reset against a `no_reset` handle; a reset at version 1
- a device credential signed by the wrong key; one naming a foreign `identity_pk`; two credentials sharing a `device_pk`; a credential naming another handle
- a second entry for a handle in one epoch (NCC Group's Medium finding)
- an entry for another log; trailing bytes; an unknown `EntryKind` byte (fails to decode rather than defaulting); `no_reset = 2`; a wrong `kt_version`; a wrong label
- §6.3: idempotent re-accept, fork, rollback in all three fields, a skipped epoch, a broken chain link, a changed `vrf_public_key`, another log's head, an unsigned edit, **a receipt accepted as a tree head** (and a tree head as a receipt)
- §8.3: one witness cosigning twice; an unconfigured witness's valid cosignature; a cosignature over another root; a forged cosignature from a configured key; no cosignatures at all
- §7.1: a witness advancing with a token for the wrong range; a halted witness refusing to catch up
- proptest: arbitrary bytes and label-prefixed noise reach a verdict rather than a panic, and decoding is total and canonical — NCC's own strategic recommendation for `akd`'s public API, applied to ours

## Rebased onto #583, and one shared-dependency note for the reviewer

This branch was rebased onto `f2z-relay-proto` (#583) and both crates are kept everywhere they meet: `rs/Cargo.toml`'s workspace dependencies, `rs/README.md`'s crate table and command blocks, and `rs.yml`'s toolchain-manifest list. The `akd` `[[bans.deny]]` stanza was already on `main` from #571, so this PR no longer adds it — it only adds the crate that makes it load-bearing.

**`f2z-kt-core` uses the workspace `ed25519-dalek = "3"`** that #583 introduced, rather than the `"2"` this branch originally carried, so there is exactly one first-party Ed25519 across the tree.

**The graph still contains two major versions of `ed25519-dalek`, and it is not our choice.** `akd_core` 0.13 requires `^2` for its ECVRF, so `2` is in the tree the moment the directory is. `multiple-versions = "warn"` reports it; `cargo deny` is green. Nothing crosses the boundary — no `ed25519-dalek` type from either version appears in an interface between `akd_core` and our code — and it resolves itself when upstream moves. The note is in `rs/Cargo.toml` beside the `akd_core` entry so the next reviewer of that warning does not have to re-derive it.

## What is deliberately not here

No signing, no key generation, no randomness, no clocks, no sockets, no storage, no HTTP. §12's open questions — epoch cadence, default *t*, the shipped witness list, the gossip protocol, reset-authority pinning — stay open, as caller-supplied values. `PROPOSED_EPOCH_INTERVAL_SECONDS` / `PROPOSED_MAX_MERGE_DELAY_SECONDS` are constants labelled placeholders that nothing in the crate reads.

`KT.md`'s own status line still reads "Nothing described here is implemented"; updating it is a docs change and is left out of this PR on purpose.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -74,7 +74,8 @@ jobs:
           scripts/check-rust-toolchain.sh --self-test
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
-            --manifest rs/crates/f2z-codec/Cargo.toml
+            --manifest rs/crates/f2z-codec/Cargo.toml \
+            --manifest rs/crates/f2z-kt-core/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -294,6 +295,22 @@ jobs:
           cargo build --locked --lib \
             --target "${{ steps.rust_toolchain.outputs.target }}" \
             -p f2z-codec
+
+      # KT.md §11.4: the log, the witness and the client link one crate. Only
+      # the *client* half reaches the browser — `akd::auditor::audit_verify`
+      # hardcodes AzksParallelismConfig::default() and reaches
+      # tokio::task::spawn, so it compiles for wasm32 and then traps at runtime
+      # (KT.md §11.3). `--no-default-features --features verifier` is therefore
+      # not a size optimisation, it is the supported configuration, and this
+      # step is what stops the auditor being pulled into a browser bundle by an
+      # innocent-looking feature edit.
+      - name: Build the f2z-kt-core client verifier for the browser target
+        working-directory: rs
+        run: |
+          set -euo pipefail
+          cargo build --locked --lib \
+            --target "${{ steps.rust_toolchain.outputs.target }}" \
+            -p f2z-kt-core --no-default-features --features verifier
 
   # Branch protection should require this stable context, `rs / gate`. It
   # succeeds only when the change detector and every applicable rs job
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -3,10 +3,92 @@
 version = 4
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "akd"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65dd81c1ad91ef0dcfbcae8cc8d3a78ce9d26cd799c495dc5451266b7ec637b2"
+dependencies = [
+ "akd_core",
+ "async-recursion",
+ "async-trait",
+ "dashmap",
+ "hex",
+ "log",
+ "protobuf",
+ "tokio",
+]
+
+[[package]]
+name = "akd_core"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce56523cc567545912841354a5efe22db939f87eb4e817372466491c7845053f"
+dependencies = [
+ "async-trait",
+ "blake3",
+ "curve25519-dalek",
+ "ed25519-dalek",
+ "hex",
+ "protobuf",
+ "protobuf-codegen",
+ "protobuf-parse",
+ "zeroize",
+]
+
+[[package]]
+name = "anyhow"
+version = "1.0.104"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
+
+[[package]]
+name = "async-recursion"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.119",
+]
+
+[[package]]
+name = "async-trait"
+version = "0.1.92"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82f6aeea286b8eb4dd3431a1be1b59d290ace00f5bfd8e2a159bc2a05e2c1667"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.4",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 
 [[package]]
 name = "bit-set"
@@ -39,6 +121,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "blake3"
+version = "1.8.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d9e454fc11f76977dc803893aff6304ed33d6a26efae8696573bea74baa27ae"
+dependencies = [
+ "arrayvec",
+ "cc",
+ "cfg-if",
+ "constant_time_eq",
+ "cpufeatures 0.3.0",
+]
+
+[[package]]
 name = "block-buffer"
 version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -48,10 +143,50 @@ dependencies = [
 ]
 
 [[package]]
+name = "cc"
+version = "1.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ad534f4357a5264cce5019c989cf66a4f0dc4e0d1b1d15f8aacec0ff7360273"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "const-oid"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
+
+[[package]]
+name = "constant_time_eq"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
+
+[[package]]
+name = "cpufeatures"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "crypto-common"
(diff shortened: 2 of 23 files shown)


skyl:
Hard merge-train hold: do not admin-merge. #556 has had repeated exact-head matrices invalidated by #571, #580, #582, and now #581 despite draft holds. This E2EE crate is outside the current quick-win train and must remain draft until #556 lands; afterward integrate the squash and rerun exact-head CI before review. The owner explicitly prioritized draining small issues ahead of the larger E2EE push.

skyl:

PR #587, opened 2026-08-24
docs(e2ee): correct CLIENT-CONTRACT's retracted case-folding claim and its checker citation

### LINKED ISSUE #576: docs(e2ee): CLIENT-CONTRACT restates a retracted claim and cites a checker that does not check what it says

### LINKED ISSUE #577: docs(e2ee): CLIENT-CONTRACT interface defects — undeclared DeliveryState, unclosed "closed union", unsettable global retention

skyl:
Closes #576.

Two corrections to `docs/e2ee/CLIENT-CONTRACT.md`. Nothing outside `docs/e2ee/` is touched.

## 1. §11.3 — the retracted case-folding argument

**Before:**

> An account is eligible iff `lowercase(username)` matches the pattern; case
> folding is safe because platform uniqueness is already enforced
> case-insensitively.

That is the claim `WIRE.md` §14.3 retracts in a dated block, and it sat thirty-two lines above this document's own contradiction of it ("**`lowercase(username)` is not yet a unique key.** … production holds case-variant duplicate accounts today").

**After:**

> An account is eligible iff `lowercase(username)` matches the pattern. The
> argument originally given for folding case — that platform uniqueness is already
> enforced case-insensitively, so folding *cannot* collide two distinct existing
> accounts — **is false, and was retracted on the record**
> ([`WIRE.md` §14.3](./WIRE.md#143-the-decision-and-the-cost-accepted)).
> Case-insensitive uniqueness is an application convention living in two
> serializers, not a property of the database, and folding case does collide real
> accounts today. The eligibility *rule* is unchanged and uppercase still
> disqualifies nobody; what does not hold is that the mapping is injective, and
> the client consequences of that are spelled out below.

Inverted rather than deleted: the sentence carries the eligibility rule itself, and the surviving half of `WIRE.md`'s correction ("uppercase on its own still disqualifies nobody") is exactly what a frontend reader needs at that point. §11.3 is now internally consistent with the bullet below it, and the qualifiers travel with the claim per the README's **Scoping convention**.

## 2. §2.3 and §12.3 — a mechanical check that does not exist

The contract claimed a mismatch in four places is "a build failure by the same mechanism `scripts/check-zcash-permissions.mjs` already enforces". Verified in this branch:

```console
$ grep -n "default.toml\|capabilities" scripts/check-zcash-permissions.mjs
$ echo $?
1
```

No output. The script never opens `permissions/default.toml` or any `capabilities/*.json`.

**The reporter's summary of what it does enforce was accurate**, and I read the script rather than copying it (`scripts/check-zcash-permissions.mjs:121-215`). The six-artifact chain is exactly as described: `build.rs` `COMMANDS` ↔ `lib.rs` `generate_handler!` (`:136`) ↔ `permissions/autogenerated/commands/*.toml` filenames (`:146`) ↔ `permissions/schemas/schema.json` consts (`:172`) ↔ `permissions/autogenerated/reference.md` `zcash:allow-…` (`:184`) ↔ Zuuallet's tracked `gen/schemas/*-schema.json` (`:204`). Three properties beyond set equality are worth recording, and are now in §2.3: each `<command>.toml` must allow and deny exactly its own command (`:157-161`), the tracked target schemas must be byte-identical to each other (`:211`), and `--self-test` carries negative controls.

Also confirmed:

- `git ls-files -- 'wallet/zuuli/src-tauri/gen/schemas/*'` is **empty** (Zuuallet tracks five files there). So the sixth link has no Zuuli counterpart — which is why the checker cannot be obtained by parameterizing a plugin id.
- `permissions/default.toml` is read by `wallet/zuuli/scripts/seed-capture-boundary.mjs:759` — and only to assert `allow-begin-sensitive-display` / `allow-end-sensitive-display` are present (`:472-479`), a substring check, not parity. (Two `*.node-test.mjs` files also name the path as a fixture.)
- `capabilities/mobile.json` is read by `wallet/zuuli/scripts/mobile-webview-authority.mjs:156` — and one nuance sharpens the picture rather than contradicting it: that script requires the non-`zcash:` permissions to be **exactly** `core:default`, `deep-link:default`, `opener:default`, `http:default` (`:41-46`, `:106-110`). So `mobile.json` is not simply unwatched — the first `f2zmsg:allow-…` entry added to it will *fail* that check until the reviewed allowlist is deliberately extended. That is a review gate, not parity with §3, and §2.3 now says so.
- `capabilities/default.json` is read by no check in the repository at all.

§2.3 now lists the four places, then states plainly that two are enforced and two are not, with the `grep` above inline as the evidence. §12.3 is **rescoped**, not just reworded: correcting the premise changes what the proposed work is, so it now names two pieces of work — the `f2zmsg` permission checker (which must be written: plugin path, `zcash:` prefix and schema directory are hard-coded, and Zuuli tracks no schemas) and the `check-client-contract.mjs` companion — and says explicitly that writing the first is the occasion to close the two gaps §2.3 names. This strengthens §12.3: "just add a plugin id" becomes a justified reason to write the checker.

## Correction-block vs. wording — my reasoning

The README's convention is to leave a false claim standing under a dated blockquote. I applied #582's test — **published + relied upon + load-bearing** — and neither fix meets it, so both are edits:

- **Fix 1** is a restatement, not an origin. The claim's canonical home is `WIRE.md` §14.3, where it *is* left standing under a dated correction block, with the inline "This argument is false as written" marker. That auditable record already exists and this edit links to it. Duplicating the correction here would give an auditor a second copy of the same retraction, not a second data point — and `CLIENT-CONTRACT.md` is days old (#574, `4073cc0`) with nothing implemented against it, so there is no reliance history to preserve. `README.md` §Status still says nothing here is implemented.
- **Fix 2** is a factual error about what a script in this repository does — not a retracted argument. There is no reasoning to preserve for review; the dating convention exists for corrections, closures and measurements, and a wrong citation is none of those. The commit message and this PR carry the record.

No date stamps were added or altered, so the **Dating convention** (author-local date, stamps must not run ahead of their commit) is not engaged.

## Verification

- Every relative link and `#anchor` under `docs/e2ee/` resolves — checked with a script over all 6 `.md` files plus `decisions/`. The single reported hit is a false positive: `` `mockMessaging[key](...sampleArgs)` `` at `:907`, JS spread syntax inside backticks, pre-existing and untouched.
- `scripts/check-public-repo-boundary.sh` passes (exit 0). Judged independently of the script, which only greps the literal `free2z/tuzi`: this PR adds **no** production measurement, no absolute user count, no private repo name or issue number, and no description of an unpatched weakness. Fix 1 restates only qualitative facts already published in this file and in `WIRE.md` §14.3 (duplicates exist; uniqueness is not a DB property) and adds no number. Fix 2 concerns public repository files only.
- No new contradiction with `WIRE.md` §14 (fix 1 now agrees with §14.3 including which half of the argument survives), `KT.md`, or ADR 0013/0014 — none of which touch the permission tooling.
- `.github/workflows/zuuli.yml`'s `changes` policy step is untouched; the diff is one file under `docs/e2ee/`.

## Deliberately not fixed

#577 (undeclared `DeliveryState`, the unclosed "closed union", unsettable global retention) and #578 (client obligations with no affordance) are out of scope for #576 and are left alone on purpose — not an oversight.


skyl:
Temporary merge-train hold: #556 is the sole active merge candidate. Keep this PR draft until #556 lands; then integrate current trunk and rerun exact-head CI.

skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I checked the two corrections against the tree rather than against the body.

## Fix 1 — clean

```console
$ grep -n -i 'case folding is safe\|already enforced case-insensitively' docs/e2ee/CLIENT-CONTRACT.md
$ # nothing
```

The retracted argument is gone, and §11.3 no longer asserts P at one line and ¬P thirty-two lines later. Agreeing that the dated-blockquote convention doesn't apply here is right: that convention preserves a claim someone relied on over time, and the retraction it would point at already lives at `WIRE.md:1843`. Reproducing it here would have been ceremony.

## Fix 2 — considerably better than what I asked for

I asked for the citation to be corrected. What landed is a proper description of what the script does enforce — the chain of set equalities across six artifacts, the per-command allow/deny assertion, the duplicate rejection, the negative controls — followed by the disproof in the form anyone can re-run:

```console
$ grep -n "default.toml\|capabilities" scripts/check-zcash-permissions.mjs
$ # no output
```

and then the part I hadn't asked for and that makes the section actually useful: **what does read those two files, and how narrowly.** `seed-capture-boundary.mjs` checks `default.toml` for two specific identifiers; `mobile-webview-authority.mjs` holds `mobile.json` against a hand-maintained reviewed allowlist rather than the command set; `capabilities/default.json` is read by no check in this repository at all.

## The finding that goes beyond the issue

> the first `f2zmsg:allow-…` entry added to `mobile.json` will fail that check until the reviewed list is deliberately extended. That is the gate working as designed — mobile authority is granted by review, not by registration — but it is a review step, not parity with §3.

That is a real consequence for the messaging work that nobody had written down, and the framing is exactly right: it distinguishes *a gate doing its job* from *a gate providing the parity §2.3 claimed*. An implementer now knows the first mobile permission they add will go red, and why, and that the fix is a deliberate review rather than a registration.

Same for the asymmetry underneath it — `wallet/zuuli/src-tauri/gen/schemas/` is tracked-empty, so the sixth link in the chain has a counterpart for Zuuallet and none for Zuuli. That's the detail that determines whether the future checker can be shared or has to be written.

## §12.3 rescoped honestly

The old text implied "add a second plugin id". The new text says **"neither is a one-line change"** and gives the reason: the plugin path, the `zcash:` identifier prefix and the Zuuallet schema directory are hard-coded throughout, and Zuuli tracks no generated schemas, so the sixth link has nothing to compare against.

That's the outcome I hoped for — correcting the premise makes the case for writing the checker *stronger*, because the work is now scoped for what it is instead of being budgeted as a parameter change.

## Scope

#577 and #578 correctly left alone and named in the body so they don't read as oversights. `zuuli.yml` untouched, so no collision with #556, which is contending over that exact policy step.

Issue #585, opened 2026-08-24
spec(e2ee): five WIRE.md gaps surfaced by hand-deriving byte vectors


skyl:
Five gaps in `docs/e2ee/WIRE.md`, found by hand-deriving byte vectors for every structure in #581 (merged as `4acc138`). Checked against #550 and #552 — none is a duplicate.

These are worth more than the usual spec nit because of how they were found: **41 of the 42 new vectors passed the first time they compiled**, and the one failure was a typo in the fixture rather than a byte-layout disagreement. So the encoding was already right, and these are the only places where the spec text did not fully determine what to write. That is a strong result for #548's freeze — and it makes the five remaining ambiguities the complete list of what an independent implementer would have to guess.

## 1. `CreateQueueRequest.stamp` is "empty" for a fixed-width struct

`WIRE.md:705` says the field is **"empty when `queue_creation_mode = open`"**. `PowStamp` is a fixed-width struct, so "empty" has no encoding — there is no length prefix to set to zero.

Sharpened by the neighbouring section: `ChallengeResponse.pow` one section earlier says **"zeroed"**, which is unambiguous. Two different words for what must be the same thing, in adjacent sections.

An implementer reading "empty" as *omit the field* produces a **56-byte desynchronisation** — not a rejected frame, a misparsed one, which is the expensive kind.

Related but not identical to the `PowStamp::empty()` note already on #550: that entry records the crate resolving the ambiguity by choosing all-zero. This is about `WIRE.md` using two words for one concept in adjacent sections.

## 2. `ReadResponse.has_more` is a `uint8` with no declared value set

Every other flag byte in the document enumerates its values. This one doesn't. The crate reads it as nonzero-is-true, which means `0x01` and `0x02` are semantically identical yet distinct canonical byte strings — a mild tension with §3.3's re-encode equality rule. (Recorded on #564's review as an implementation choice; the spec gap is here.)

## 3. §13.1's PoW preimage is the one integer concatenated outside a `tls_codec` struct

The preimage is `challenge || salt || counter`. Everywhere else in `WIRE.md`, integer width and endianness come from a struct declaration. Here they come only from §1.3's general statement and are **not restated at the point of use**.

[...]

PR #588, opened 2026-08-24
docs(e2ee): fix WIRE.md's anti-replay retention hole and the §6.3/§10 contradiction

### LINKED ISSUE #586: WIRE.md: anti-replay retention is unspecified against clock skew, and §6.3/§10 contradict on send-side errors
Two defects in merged `docs/e2ee/WIRE.md`, found while implementing `f2z-relay-proto` (#583). Both are real; one is a silent security hole.

## 1. §11.1 does not relate `antireplay_window_ms` to `clock_skew_ms`, and must

**This is a security defect, not an editorial one.**

A frame stamped `ts` is accepted while `ts - clock_skew_ms <= now <= ts + clock_skew_ms` — a replay window of **2 × `clock_skew_ms`** in total. Worst case the relay first observes it at `ts - clock_skew_ms` and it stays acceptable until `ts + clock_skew_ms`, so a seen-set entry needs **2 × `clock_skew_ms` of retention measured from first observation**.

§11.1 publishes the two values as independent fields with no stated relationship. A relay that publishes `antireplay_window_ms < 2 * clock_skew_ms` is **fully conforming to the specification as written**, and has no seen-set coverage for the gap — §5.5's replay protection is silently absent for frames in that band. Nothing in the document tells an implementer this, and nothing tells a client to check.

**Fix:** state the invariant `antireplay_window_ms >= 2 * clock_skew_ms` in §11.1 and again where §5.5 defines the seen-set, with the derivation above so it is checkable rather than asserted. Specify what a client does when a capability document violates it — `f2z-relay-proto` implements a client-side refusal (on by default) via `SeenSet::retention_is_sound`, deliberately **not** a document-validity failure, since such a document is exactly what the current spec permits. Confirm that split is what we want, or change it.

Related, and worth writing down in the same pass: `f2z-relay-proto`'s seen-set **refuses rather than evicts** on reaching `antireplay_seen_max`. Eviction would silently reopen the replay window under memory pressure — precisely when an attacker would want it. §5.5 should say the refusal is required, not leave it to implementer taste.

## 2. §6.3 and §10 contradict each other on the send-side "absent" case

§6.3 says every send-side refusal that would distinguish queue state collapses to `ERR_UNAVAILABLE` (code 15). §10's table assigns "absent" to code 10.

They cannot both hold. #583 resolved it in **§6.3's favour** — everything send-side collapses to `ERR_UNAVAILABLE` — because §6.3's rule is the one carrying the security argument: distinguishable send-side errors let a bound sender learn queue state by filling it, which is the leak §6.2 exists to prevent.

**Fix:** make §10's table agree with §6.3, and add a sentence at code 10 saying it is receive-side only, so the next reader does not re-derive the contradiction.

## Also worth doing while in here

#583 enumerated **16 ambiguity calls** it had to make. Several deserve to be settled in the document rather than living in a PR body:

- §5.1 step 5 is inapplicable to `BIND_SEND`, which is why the implementation requires `signer_key == send_key`. Without that requirement §7.4's bind-once theft protection costs an attacker nothing.
- `CREATE_CONTACT_QUEUE`'s self-authentication is unstated.
- Index origin (0 vs 1) is unstated.
- `ttl == 0` meaning "no preference" is unstated.
- The timestamp-window boundary is inclusive-or-exclusive unstated.

Read #583's PR body for the full list with reasoning, and fold in the ones that are genuinely spec-level rather than implementation detail.

## Acceptance

- [ ] The retention invariant is stated with its derivation, and the client's obligation is specified.
- [ ] §10 and §6.3 agree.
- [ ] The spec-level ambiguities from #583 are settled in the document.
- [ ] Corrections to already-merged normative text use the dated-correction style established by `ARCHITECTURE.md` §4.9 and §5.4 — do not silently rewrite.
- [ ] Docs only. No code.

skyl:
Closes #586

Two defects in merged `docs/e2ee/WIRE.md`, both found while implementing
`f2z-relay-proto` (#583), plus the spec-level half of that PR's 16 ambiguity
calls. **Docs only** — one file, `docs/e2ee/WIRE.md`.

## Defect 1 — §5.5 / §11.1: anti-replay retention was unspecified against clock skew

This is the security one. §11.1 published `clock_skew_ms` and
`antireplay_window_ms` as two independent fields with **no stated relation**, and
§5.5 said only that the relay keeps pairs "observed within the window" — naming
neither a retention origin nor a retention duration. A relay publishing
`antireplay_window_ms = 30000` beside `clock_skew_ms = 120000` was **fully
conforming** and had no seen-set coverage for a frame re-presented more than 30
seconds after first observation, while the timestamp check went on accepting it
for another 210 seconds. §5.5's replay protection was silently absent for the
gap, and nothing told a client to check.

§5.5 now derives the bound rather than asserting it, so an auditor can check the
factor of 2 instead of taking it:

```
accepted at now  ⟺  |now - t| <= skew  ⟺  now ∈ [t - skew, t + skew]
length = (t + skew) - (t - skew) = 2 * skew

antireplay_window_ms >= 2 * clock_skew_ms          (INVARIANT R)
```

Both halves are written up as load-bearing:

- **The origin** is pinned to **first observation by the relay's own clock**, and
  the text says why `timestamp_ms` is the wrong origin — it is chosen by the
  party being defended against. It also notes that the `timestamp_ms` origin
  would need only `>= skew`, which is exactly why the origin has to be stated:
  the same number means two different things and only one is safe without the 2.
- **The factor of 2** is called out against the specific wrong reasoning
  ("entries live as long as the timestamp window, which is `clock_skew_ms`") —
  the half-width-for-full-width slip.

Then: a relay MUST retain for `2 * clock_skew_ms` from first observation and MUST
publish values satisfying R; §11.1 restates R beside the fields, comments the
struct, and gives `antireplay_window_ms` a default of `240000` that satisfies R
with no slack against the existing `clock_skew_ms` default of `120000`; §11.3
gains the client's refusal as step 6.

**On the split #586 asked me to confirm — client-side refusal, not a
document-validity failure: confirmed, and now argued in the document.** The
reason has changed, though. #583's reason was "such a document is what the spec
permits", which this PR retires. The reason that survives is that §11.3 steps 2–5
were *already* policy refusals of validly-signed documents (step 5 refuses
`max_message_ttl_seconds > 2592000` on exactly this footing), and a policy
violation must not be a decode failure because that discards the artifact
that makes the violation reportable — the signed bytes proving *this operator, at
this `published_at_ms`, published these two numbers*. Quarantine, not discard.
§11.3 now says so, and constrains the override: default MUST be refuse, override
MUST be per-relay and surfaced, never a global "ignore relay policy" switch.

**On the seen-set refusing rather than evicting:** already merged — §5.5's
fail-closed paragraph says `ERR_BACKPRESSURE` and "MUST NOT evict live entries".
What was missing is that "live" and "age out" had no definition, so I tied them
to R rather than restating a rule the document already has.

One consequential side-effect: §5.5's `channel_binding_mode: "none"` paragraph
says a restart reopens "a replay window of `clock_skew_ms`". That is the
half-width and understates the exposure by the same slip R exists to prevent.
Left as merged with a bracketed pointer; correct figure is `2 * clock_skew_ms`.

## Defect 2 — §6.3 vs §10: both assigned the send-side "absent" case

Row 10 claimed "the address does not exist"; row 15 claims "**Send side only.**
Absent, deleted, expired…". An `APPEND` to a nonexistent address had two codes.

Resolved in **§6.3's favour**, and the correction says why it is not a
preference: §6.3's rule carries a security argument (a bound sender must not
learn queue state by filling the queue, §7.1's two-address split) and row 10
carries a naming convenience. Row 10 keeps its text, gains a pointer and "**Never
returned by `APPEND` or `CONTACT_APPEND`**". §6.3's collapse list gains the item
it had omitted — **`signer_key` not the bound `send_key`** — which is the one
that would have reintroduced the oracle, since a distinguishable "wrong key"
answer is an answer that the address exists.

`ERR_ALREADY_BOUND` and `ERR_NOT_PERMITTED` stay distinguishable, with §7.4's and
§12.2's reasons named: collapsing `ERR_ALREADY_BOUND` would delete the only
detection §7.4 has.

## Ambiguities folded in (spec-level)

| #583 call | Where | Note |
|---|---|---|
| 1 | §10, §6.3 | Defect 2 above. |
| 2 | §5.1 | Step 5 doesn't apply to `BIND_SEND`; `signer_key == send_key` required. |
| 3 | §12.2 | `CREATE_CONTACT_QUEUE` self-authentication. |
| 4 | §6 | Wrong transcript address → `ERR_MALFORMED`, fatal. |
| 5 | §6 | Signed-arriving-unsigned → `ERR_BAD_SIGNATURE`; unsigned-arriving-signed → `ERR_MALFORMED`. |
| 7 | §6.2 | `UNSUBSCRIBE`'s response is empty. |
| 8 | §6.2, §8.2 | Indices start at 0; the bound is `next_index`, and a drained queue still refuses. |
| 9 | §5.5 | Timestamp-window boundary is inclusive, with the intermittent-failure reason. |
| 10 | §7.7 | `ttl == 0` = "no preference"; clamp order with the 30-day ceiling applied last and unconditionally. |
| 11 | §5.3 | A 32-zero-byte exporter output collides with the `none` sentinel and is refused. |
| 12 | §11.3 step 7 | `contact_queues_enabled` requires a real `contact_append_pow`. |
| 13 | §11.3 step 8 | `max_frame_bytes` must carry the largest `padding_sizes` entry — necessary, not sufficient, and says so. |
| 14 | §6.2 | `pending` is not `next_index` minus the watermark once TTL expiry is in play. |
| 15 | §7.2 | `QueueAdvert` deliberately has no `tls_codec` encoding — the absence is a decision. |

§5.1's call 2 is written up as a security property rather than tidiness: without
`signer_key == send_key`, a `BIND_SEND` signature proves possession of *some*
key while the `send_key` in the body is unconstrained, so anyone who has seen a
`send_addr` can bind it to a key **they do not hold**. §7.4's "possible but
noisy" theft becomes free, and §7.4's "the operator gets the capability"
understates it — *any* party that has seen the address does.

The §5.1 correction also states the general test, since three checks now return
distinguishable codes on send-side commands: the question is **not** "is it
send-side" but **"does the answer depend on relay state"**. `BIND_SEND`'s and
`CREATE_*`'s self-checks are decided from two fields of the frame before any
lookup, so the caller can compute the answer without the relay and the relay
discloses nothing.

### Rejected — implementation detail, staying in the implementation

- **Call 6 — dispatching to the wrong handler is `ERR_INTERNAL`.** A relay's
  internal dispatch table is not observable protocol structure; there is no
  conforming client that can provoke it and no interoperability question to
  settle. §10's row 21 ("Relay fault. Carries no detail, ever.") already covers
  it, and enumerating which internal bugs map to code 21 would be an open-ended
  list that no second implementer could check against.
- **Call 16 — `PowStamp::empty()` for "no stamp".** Already stated in the merged
  text: `CreateQueueRequest`'s comment reads "empty when `queue_creation_mode` =
  open". Nothing to settle; re-stating it in prose would add a second place to
  keep in sync.

I also considered call 14 (`pending`) borderline and **kept** it — the wire
field's meaning is at stake for anyone parsing it, not just anyone computing it,
so a client needs to be told it is a count and never an index.

## Style

Corrections to already-merged normative text use the dated-correction style of
`ARCHITECTURE.md` §4.9 / §5.4: **the merged claim is left visible with a
pointer, never silently rewritten**, because an auditor's first question about a
corrected specification is what it used to say. Five such blocks (§5.1, §5.5,
§6.2, §10, §11.1), indexed in a table at the top of the document so they are
findable, with everything that merely settles an *unstated* point written in
place and listed separately.

The header also picks up a correction of its own: the Status line's "Nothing
described here is implemented" stopped being true with #583, and the document now
says what does and does not exist (a protocol library; no relay binary, no client
binary, no deployment).

## Verification

- `scripts/check-public-repo-boundary.sh` — clean.
- No trailing whitespace; prose wrapped at 80 as in the rest of the file.
- No code, no crates, no workflows touched — `docs/e2ee/WIRE.md` only.

### DIFF
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -29,6 +29,38 @@ lifecycle, no first-contact path, no quotas. Each significant choice below is
 recorded in an ADR with the alternatives that were rejected and why. Nothing here
 should be read as "the merged docs already said this."
 
+> **Correction (2026-08-23) — the Status line above ("Nothing described here is
+> implemented") is no longer true, and the first implementation found defects in
+> this document.** It is left standing so the revision it describes is still
+> legible. `rs/crates/f2z-relay-proto`
+> ([#583](https://github.com/free2z/zuu/pull/583)) is a protocol library — no
+> relay and no client binary exists, and no deployment does — but it implements
+> §5, §6, §7, §8 and §11 closely enough to have discovered two defects and a
+> list of ambiguities. Those are corrected in place below, in the dated-
+> correction style of
+> [`ARCHITECTURE.md` §5.4](./ARCHITECTURE.md#54-exporter-derived-application-secrets):
+> **the merged text is left visible with a pointer, never silently rewritten**,
+> because an auditor's first question about a corrected specification is what it
+> used to say. This revision's corrections, in reading order:
+>
+> | § | What was wrong |
+> |---|---|
+> | 5.1 | Verification step 5 does not apply to three of the nine signed commands, and its error code is wrong for a fourth. |
+> | 5.5 | **Security defect.** Seen-set retention was unrelated to the timestamp window, so a conforming relay could have no replay protection for part of it. INVARIANT R. |
+> | 6.2 | `pending` is not `next_index` minus the acked watermark once TTL expiry is in play. |
+> | 10 | Codes 10 and 15 both claimed the send-side "absent" case; resolved in §6.3's favour. |
+> | 11.1 | `clock_skew_ms` and `antireplay_window_ms` were published as independent fields and are not. |
+>
+> Everything else changed in this revision settles something the merged text
+> left **unstated** rather than stated wrongly, and is written in place: §5.3
+> (zero exporter output), §5.5 (window boundary inclusive), §6 (transcript
+> address and auth-presence mismatches), §6.2 (index origin, `UNSUBSCRIBE`
+> response), §7.2 (`queue_advert` has no wire encoding, deliberately), §7.7
+> (TTL `0` means "no preference"), §8.2 (the bound is `next_index`), §11.3
+> (capability-document consistency checks, and why they are client refusals
+> rather than validity failures), §12.2 (`CREATE_CONTACT_QUEUE`
+> self-authentication).
+
 ---
 
 ## 1. Scope
@@ -440,6 +472,48 @@ Steps 2 and 3 precede signature verification because they are cheap and they are
 the flood-resistant filters; step 5 follows verification so that an unsigned
 guess cannot probe the address space.
 
+> **Correction (2026-08-23) — step 5 as written does not apply to three of the
+> nine signed commands, and its stated error code is wrong for a fourth.** The
+> list above is left standing so an auditor can see what was specified. Step 5
+> assumes (a) that an address exists with a key already registered against it,
+> and (b) that a mismatch is reported as `ERR_NO_ACCESS`. Neither assumption
+> holds universally, and where (a) fails the *substitute* rule is the only thing
+> authenticating the command — so leaving it unstated is not an editorial gap.
+> Found while implementing this section
+> ([#583](https://github.com/free2z/zuu/pull/583)).
+>
+> | Signed command | What step 5 is | Failure |
+> |---|---|---|
+> | `SUBSCRIBE`, `UNSUBSCRIBE`, `READ`, `ACK`, `DELETE_QUEUE` | Step 5 exactly as written: `signer_key` MUST equal the `recv_key` registered for `address`. | `ERR_NO_ACCESS` |
+> | `CREATE_QUEUE`, `CREATE_CONTACT_QUEUE` | No address exists yet (`address` is 32 zero bytes). `signer_key` MUST equal the `recv_key` **in the body** — the command is self-authenticating (§6.2, §12.2). | `ERR_NO_ACCESS` |
+> | `BIND_SEND` | The send side is unbound by definition (§7.3), so there is no registered key to compare against. `signer_key` MUST equal the `send_key` **in the body**. | `ERR_NO_ACCESS` |
+> | `APPEND` | Step 5 as written: `signer_key` MUST equal the bound `send_key`. **But the failure is reported as `ERR_UNAVAILABLE`, not `ERR_NO_ACCESS`** — §6.3 and §10's correction. | `ERR_UNAVAILABLE` |
+>
+> **Why `BIND_SEND` must carry the self-check, stated as a security property and
+> not a tidiness rule.** With no rule, the only thing a `BIND_SEND` signature
+> proves is possession of *some* key; the `send_key` in the body is unrelated to
+> it and can be any 32 bytes the caller chooses. An attacker that has learned a
+> `send_addr` could then bind the queue to a key **it does not hold**, from a
+> signature made with a throwaway key. §7.4's whole property is that stealing the
+> write capability is *possible but noisy*; without this check it is not even
+> theft — the attacker never has to possess anything, and §7.4's "the operator
+> gets the capability" understates it, because *any* party that has seen the
+> address gets it. Requiring `signer_key == send_key` restores the cost to
+> "possess the key you are binding", which is what §7.3 assumes throughout.
+>
+> **Why the `BIND_SEND` and `CREATE_*` checks are not an existence oracle**
+> (§10), despite returning a distinguishable code on a send-side command: each is
+> decided from **two fields of the frame itself**, before any address lookup, and
+> its outcome is a function of what the caller signed and nothing else. A caller
+> can compute the answer without contacting the relay, so the relay reveals
+> nothing by returning it. That is the test to apply to any future exception:
+> not "is it send-side" but "does the answer depend on relay state".
+>
+> **Why `ERR_NO_ACCESS` and not `ERR_MALFORMED`.** These frames decode and
+> re-encode cleanly (§3.3); what fails is an authorization rule. `ERR_MALFORMED`
+> is fatal and would drop the connection of a client with a key-selection bug —
+> a disproportionate response to a frame the relay understood perfectly.
+
 ### 5.2 Relay identity binding — the attack it closes
 
 ```
@@ -508,6 +582,14 @@ Therefore:
 
 - Such a relay MUST set `channel_binding_mode: "none"` in its capability
   document and MUST use **32 zero bytes** in the transcript.
+- Because 32 zero bytes is that sentinel, a relay in `channel_binding_mode:
+  "tls-exporter"` whose exporter output is 32 zero bytes MUST treat it as a
+  failure to obtain a binding and MUST refuse the connection rather than sign
+  against it. A genuine exporter output of zero has probability 2^-256, so this
+  costs nothing; what it buys is that "bound" and "unbound" are never the same
+  32 bytes. Without the rule the two modes are indistinguishable in the
+  transcript, and the symptom of a broken exporter is a signature that verifies
+  against the wrong mode much later, which is the worst place to discover it.
 - A client MAY refuse such a relay outright, and the reference client SHOULD make
   this a user-visible setting with a strict mode that refuses.
 - **In `none` mode the restart argument of §5.5 does not hold**, so the relay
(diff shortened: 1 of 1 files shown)


skyl:

skyl:

skyl:
Adversarial review, posted although this PR is currently **closed** — the branch `docs/586-wire-defects` survives, so if it is reopened the feedback is here. Verified against head `125382e` in an isolated worktree.

**The core work is genuinely good.** The anti-replay derivation at `:644-671` is *derived* rather than asserted — frame accepted over `[t−skew, t+skew]`, worst-case first observation at `t−skew`, entry must survive to `t+skew`, hence `2·skew` from first observation. Pinning the origin to the relay's own first observation rather than `timestamp_ms` is the right call for the right reason (`timestamp_ms` is adversary-chosen), and it is implementable with no extra state. And **every ambiguity call I checked matches the shipped crate** — index origin 0, the ACK bound, wrong-key `APPEND` → 15, `BIND_SEND` → 11, the `ttl == 0` ceiling ordering, the zero-exporter refusal, the inclusive timestamp boundary. `f2z-codec`'s byte vectors are unaffected: the §11.1 edit adds only comments, no field, width or order moved.

What it does not do is finish sweeping the document for the defect it fixes.

## Blocking — two normative sentences still say the thing this PR exists to remove

**1. `WIRE.md:1560-1562`, "The existence-oracle rule", is untouched and now contradicts the correction 40 lines above it.**

> **"No such queue" and "exists, but your key does not authorize it" MUST return the same code (10).**

Unscoped MUST, names code 10. For `APPEND`/`CONTACT_APPEND` the corrected answer is 15 (`:490`, `:1071-1074`, `:1536-1541`). An implementer reading §10 top-to-bottom hits the corrected row 10 at `:1504`, then hits this MUST and returns `ERR_NO_ACCESS` for a wrong-key `APPEND` — **precisely the oracle #550 item 1 describes.** The PR added "See also the existence-oracle rule" to row 10, which points the reader *at* the stale sentence.

Also a violation of the Scoping convention #582 added to `README.md:134-138` — *carry a claim's qualifiers when restating it elsewhere*. This restatement did not receive the new qualifier.

**2. `WIRE.md:1024-1026` (`DELETE_QUEUE`) contradicts the new row 10 and its own next paragraph.**

> The relay MUST subsequently answer **both addresses** with `ERR_NO_ACCESS`

"Both" includes `send_addr`. Row 10 now says code 10 is "**Never returned by `APPEND` or `CONTACT_APPEND`**" (`:1504`), and `:1028-1029` two lines later says the sender "learns nothing except that its next `APPEND` fails with `ERR_UNAVAILABLE`". Fix: recv with 10, send with 15.

## Can #550 item 1 close? Not yet — but it is one commit away

Both halves are substantively addressed. The code contradiction is resolved in §6.3's favour, consistently at `:490`, `:1504` and `:1517-1555`. And the oracle from the omitted wrong-key case — the half #550 called out as the one that mattered — is **genuinely fixed**: the collapse list at `:1071-1074` now reads "…and `signer_key` not the key bound to `send_addr`", restated at `:1536-1541`.

With `:1560` and `:1025` corrected, item 1 closes. As it stands it would have to stay open, because those two sentences are exactly what an independent implementer would follow.

## The R boundary leaves a tick uncovered — and the shipped crate sits in it

Three facts now interact:

- the timestamp window is newly **inclusive on both sides**, `|now − t| <= skew` (`:629-638`);
- R is stated with `>=`, and the new default `240000 = 2 × 120000` is described as "no slack" (`:1619`, `:1687-1689`);
- `f2z-relay-proto` retains over a **half-open** interval — `expire()` keeps entries where `*expires_at > now_ms` (`rs/crates/f2z-relay-proto/src/replay.rs:225`), pinned by the test at `:333-338`.

Frame first observed at `t−skew`, replayed at exactly `t+skew`: timestamp check passes (inclusive), seen-set entry expired one tick earlier (half-open) → **replay accepted**, at exactly `2 * clock_skew_ms` after first observation under the defaults.

The crate's comment at `replay.rs:326-330` waves this off — "the boundary of the retention interval is not itself a security property, because soundness comes from `retention_ms >= 2 * clock_skew_ms`". That is the same half-for-full reasoning slip R was written to prevent, relocated to the endpoint. §5.5 spent 40 lines deciding the timestamp boundary's inclusivity precisely because a coin flip per implementer is unacceptable, then left the retention endpoint a coin flip. One character fixes it: state R as `>`, or say retention is over the closed interval, or set the default above `240000`.

## `antireplay_seen_max` is "published" and is published nowhere

`:728` says the seen-set "is bounded (`antireplay_seen_max`, published)". It is **not a field of `Capabilities`** (`:1607-1660`) and the string appears exactly once in the document. This PR *strengthens* the MUST that depends on it — `:729-733` makes refusal "the only conforming response to a full seen-set" — hardening a MUST whose only parameter does not exist. Same class as #552's `send_retry_max`, in the section this PR is rewriting. The crate takes it as a bare constructor argument (`replay.rs:161`) and the doctest picks `65_536` out of the air (`lib.rs:98`).

## It makes #590 worse, in three ways, and does not say so

- R doubles required retention against the naive `clock_skew_ms` reading — halving the attacker's cost to saturate a fixed-size set and doubling how long a saturated set stays saturated.
- `:729-733` makes fail-closed refusal the *only* conforming response. Since `SeenSet::observe` inserts **before** signature verification, the spec now normatively mandates the global outage an unauthenticated attacker can trigger for free.
- The retention origin "first observing it" (`:673-675`) reads literally as the step-3 lookup. #590's preferred remedy is to test at step 3 and commit after step 6 — under which the relay never "observes" an unverified pair. Say *first recorded*, or #590's fix is arguably non-conforming to a MUST written days before it.

Not a reason to hold the PR, but the §5.5 correction should name #590 rather than tightening the mechanism around it silently.

## Settled elsewhere, worth recording rather than leaving to look open

- **#550 item 2 is half-closed by fiat.** The step-5 carve-out is granted (`:489`, with a good argument at `:496-505`) — one of item 2's two acceptance bullets. The other, "a failed bind is neither distinguishable nor destructive", is **explicitly decided against** at `:1546-1552`. Defensible with a stated reason, but #550's acceptance criterion is being overruled in a PR that never mentions #550.
- **`BIND_SEND` on an absent address now has no code.** The §10 correction enumerates code 10's survivors (`:1543-1545`); the state-dependent `BIND_SEND` refusals fall through to §6.3 → 15, but nothing says so. One sentence.
- **#552's "index base undefined" and "what `ACK` on an empty queue does" are both silently resolved** — `:955-962` and `:1373-1383`. Good changes; the issue should be updated.
- **#585 item 5's line reference goes stale** — §11.3's operator-name requirement moves from step 7 to step 10 (`:1744`).
- Correctly untouched: #550 items 3 and 4, all of #552's over-claims (§11.1 `:1596` still says "**entire** externally-relevant policy"), #585 items 1–4.

## House conventions — two misses

Dating is compliant (commit `2026-08-23 23:06 -0400`, all five stamps 2026-08-23), and the correction pattern is applied well in the five indexed places.

1. **The §5.5 half-width correction at `:754-760` is an italic parenthetical, undated, not a blockquote** — yet it *withdraws* a published, load-bearing security figure (a restart "reopens a replay window of `clock_skew_ms`", understating exposure by 2×). By #592's rule — made true → fixed in place; withdrawn → dated correction — this is squarely a withdrawal, and a frozen spec has the stronger claim on the treatment. It is also absent from the correction index at `:47-52`.
2. **The §6.3 edit at `:1071-1074` — the single most security-relevant change in the PR — appears in neither index.** It modifies merged normative text in place, adding a member to a collapse list whose omission reopened an oracle, and is missing from both `:47-52` and the "settles something unstated" list at `:54-62`. The index claims to be exhaustive and is not.

## Two soft divergences from the shipped crate

- §11.3's new "steps 2–8 are client refusals, not document-validity failures" (`:1746-1757`) contradicts `capabilities.rs:230-234`, whose `validate` doc says the opposite. Behaviourally fine — it returns `ProtoError::Refused`, so quarantine-not-discard holds — but the comment now states the reverse of the spec.
- `ClientPolicy::require_sound_antireplay_window` is a `bool` (`capabilities.rs:368,384`); the new text says an override "MUST be per-relay and surfaced, never a global switch" (`:1759-1762`). Satisfiable, but steps 5/7/8 in `validate` are not overridable at all, so code and MAY are unaligned in both directions.

## Links and boundary

All relative links and `#anchors` across all 20 files in `docs/e2ee/` resolve, checked with GitHub slug rules including em-dash headings. Boundary clean: no `tuzi` reference, no private path, no production measurement. The one judgement call — the new text describes an unpatched weakness at length — concerns an unimplemented specification with no relay binary and no deployment, and `:32-45` says so explicitly.

## Recommendation

Four surgical edits, one to two lines each, then merge: scope `:1560-1562` to recv-side and name 15 for send; fix `:1024-1026`; make R strict or state the closed interval so the inclusive timestamp boundary and zero-slack default do not leave a tick uncovered; and index both missing corrections, giving `:754-760` the dated-blockquote treatment.

skyl:

skyl:

skyl:
Re-checked against the current head. `gate` is green, so this is mergeable — **but the four asks from my review are still outstanding, and one of them is now sharper than when I filed it.** Holding.

Status of each, with evidence:

**1. `WIRE.md:1560-1562` — the existence-oracle MUST. NOT DONE.** Verbatim on this head:

> **"No such queue" and "exists, but your key does not authorize it" MUST return the same code (10).**

Unscoped, still naming code 10. And `:1504`'s row 10 now reads *"**Never returned by `APPEND` or `CONTACT_APPEND`**"* — so the document states both, forty lines apart, and points the reader from the corrected row **at** the stale MUST. An implementer reading §10 top-to-bottom still lands on `ERR_NO_ACCESS` for a wrong-key `APPEND`, which is the oracle this PR exists to close.

**2. `WIRE.md:1025` — `DELETE_QUEUE`. NOT DONE.**

```
1025:  answer both addresses with `ERR_NO_ACCESS` — the same code an address that never
```

"Both addresses" includes `send_addr`, which row 10 now says never returns 10 — and `:1028-1029`, three lines later, already says the sender "learns nothing except that its next `APPEND` fails with `ERR_UNAVAILABLE`". The paragraph contradicts its own next paragraph.

**3. INVARIANT R — NOT DONE.** Still `>=` in all three places:

```
686:  > antireplay_window_ms >= 2 * clock_skew_ms      (INVARIANT R)
1619:   uint32 antireplay_window_ms;  /* MUST be >= 2 * clock_skew_ms; default 240000 */
1673:  > antireplay_window_ms >= 2 * clock_skew_ms      (INVARIANT R)
```

With the timestamp window newly **inclusive** on both sides and the default at exactly `240000 = 2 × 120000` ("no slack"), a frame first observed at `t−skew` and replayed at exactly `t+skew` passes the timestamp check while its seen-set entry expired one tick earlier — because `f2z-relay-proto`'s `expire()` keeps entries where `*expires_at > now_ms` (`replay.rs:225`, pinned by the test at `:333-338`). **The shipped crate sits in that tick.** One character fixes it: `>`, or state retention over the closed interval, or lift the default above `240000`.

**4. The indexes — HALF DONE.** The §6.3 collapse-list change **is** now indexed (row 10, *"Codes 10 and 15 both claimed the send-side 'absent' case; resolved in §6.3's favour"*). Good — that was the more important half.

Still outstanding: `:754-760`'s half-width correction remains an **italic parenthetical**, undated and unindexed:

> *(The width in the sentence above is the half-width and understates the exposure…)*

It withdraws a published, load-bearing security figure — a restart reopens `2 * clock_skew_ms`, not `clock_skew_ms`. By the rule #592 settled — *a claim that is made true is fixed in place; a claim that is withdrawn gets a dated correction* — that is a withdrawal, and a **frozen** spec has the strongest claim on the treatment, because implementers may have built against the wrong figure.

---

**Nothing about my assessment has changed.** The derivation at `:644-671` is correct rather than asserted, pinning the retention origin to the relay's own first observation is the right call for the right reason, and every ambiguity call still matches the shipped crate — I re-checked the table in my earlier review and none of it has moved.

And **#550 item 1 is still one commit away.** The half that mattered — §6.3's collapse list omitting the wrong-key case — is genuinely fixed at `:1071-1074`. Items 1 and 2 above are what stand between this and closing that issue.

Four edits, one to two lines each. Land them and I'll merge on green.

skyl:

skyl:
## Re-derived against current `main` (`c7ae0dc`) — my earlier line numbers were all stale

Every citation I gave has moved, and one no longer exists at all. Correcting them, because a hold that points at lines which no longer say what I claimed is worthless:

| My earlier citation | Actual location now | Status |
|---|---|---|
| `:1563` existence-oracle MUST | **`:1265`** | still unscoped |
| `:1504` "Never returned by APPEND or CONTACT_APPEND" | **gone from the document** | superseded — see below |
| `:1025` "answer both addresses with `ERR_NO_ACCESS`" | **`:817`** | still live, and sharper than I described |

(Re-deriving these is also what turned up #624, so the exercise was worth doing independently of this issue.)

### One ask is effectively addressed

The phrase "Never returned by APPEND or CONTACT_APPEND" is no longer in `WIRE.md`. The scoping work it was doing is now carried by the §10 table itself, `:1259`:

> | 15 | `ERR_UNAVAILABLE` | | **Send side only.** Absent, deleted, expired, full-by-messages, full-by-bytes, or relay backpressure — collapsed into one code (§6.3). |

That "**Send side only**" is explicit and did not used to be there. It makes the intended split legible: recv-side absence collapses into `ERR_NO_ACCESS` (10), send-side absence collapses into `ERR_UNAVAILABLE` (15). I consider that ask answered.

### But the remaining contradiction is now a direct one, four lines apart

`DELETE_QUEUE`, `WIRE.md:815-822`:

> The relay MUST subsequently answer **both addresses** with `ERR_NO_ACCESS` — the same code an address that never existed gets (§10).
>
> The sender, if any, learns nothing except that its next `APPEND` fails with `ERR_UNAVAILABLE` (§6.3).

The first sentence requires the **send** address to answer `ERR_NO_ACCESS` (10). The second says the sender's `APPEND` gets `ERR_UNAVAILABLE` (15). The send address cannot do both, and an implementer has no way to tell from this text which sentence governs.

Given row 15's new "Send side only" scoping, the second sentence is clearly the intended behaviour and **"both addresses" is simply wrong**. The fix is one word:

> The relay MUST subsequently answer the **recv** address with `ERR_NO_ACCESS` — the same code an address that never existed gets (§10) — and the **send** address with `ERR_UNAVAILABLE` (§6.3), which is what an absent, deleted, or expired send address returns in every other case.

That also removes the reason this looked like an existence-oracle problem: returning `ERR_UNAVAILABLE` on the send address leaks nothing, because `ERR_UNAVAILABLE` is already the collapsed answer for absent / deleted / expired / full / backpressure. Deletion is indistinguishable from five other states.

### And the existence-oracle MUST is still stated unscoped

`:1265`:

> **"No such queue" and "exists, but your key does not authorize it" MUST return the same code (10).**

Read literally this binds the send side too, where §6.3 and row 15 say the answer is 15. A careful reader resolves it from row 15's scoping, but the MUST is the normative sentence and should carry the scope itself — "on recv-side commands" is enough.

Credit where due: the section immediately after is exemplary and I want it left alone. It states the residual timing channel plainly (`:1273-1291`) — dummy Ed25519 verification on the absent path, constant-time key comparison, no short-circuit — and then says the storage-layer channel "is real and is stated here rather than papered over" instead of claiming constant time. That is the opposite of the failure mode I keep filing.

### Revised hold

Down to **two** asks, both narrow:

1. `WIRE.md:817` — "both addresses" → recv gets `ERR_NO_ACCESS`, send gets `ERR_UNAVAILABLE`.
2. `WIRE.md:1265` — scope the MUST to recv-side commands.

The INVARIANT R / `>=` ask has been split out into **#624**, where it now has an executable reproduction against the shipped crate rather than a spec argument. #550 item 1 closes when the two above land.

Issue #590, opened 2026-08-24
design(f2z-relay-proto): SeenSet inserts before signature verification, so unauthenticated frames can saturate the replay set


skyl:
## Problem

`SeenSet::observe` in `rs/crates/f2z-relay-proto` (#583, merged as `40af3a5`) both **checks and inserts** at `WIRE.md` §5.1 step 3 — before the signature is verified at step 6.

That means an attacker who cannot produce a single valid signature can still fill the replay set. Each frame costs them a random `(signer_key, nonce)` pair and nothing else: no key, no proof of work, no prior authorization. The set is **fail-closed**, so when it saturates the relay returns `ERR_BACKPRESSURE` — to every client, not just the attacker.

At the bound the crate's own doctest uses, that is roughly **65k cheap frames** to reach saturation, renewable every 4 minutes as entries age out. The cost asymmetry is the whole problem: verifying a signature is expensive for the relay and free to skip for the attacker, and the relay is currently paying for the insert before it knows whether the frame is authentic.

## The spec permits the safer shape; the API forecloses it

`WIRE.md` §5.1 mandates the ordering for the **check** — the seen-set lookup has to happen at step 3, before the relay does anything expensive with the frame. It says nothing requiring the **insert** to happen there too, and there is a good reason not to: an unverified frame is not evidence that anything was replayed.

The natural fix is to split the operation — test membership at step 3, commit the entry after step 6 succeeds. `SeenSet::observe`'s single-call API makes that impossible without changing its signature, which is why this is worth deciding now rather than after a relay is built on it.

The counter-argument deserves stating: splitting introduces a window where two concurrent frames with the same `(signer_key, nonce)` both pass the check before either commits, so the commit needs to be atomic-or-lose (insert-if-absent, and the loser is treated as a replay). That is a real cost, and it may be the reason the single call was chosen. If so, it should be written down as a decision rather than left as an artifact of the API shape.

## Why this is worth answering before implementation

This is a design question, not a bug — the code does what its API says, and nothing in `WIRE.md` is violated. But §12 of the threat model treats relay availability as in scope, and this is a denial-of-service that costs the attacker nothing and requires no credential. Once a relay is built against `SeenSet`, the shape is much harder to change.

Related: `WIRE.md`'s Layer-1 limits (per-source connection and command rates, `:1595-1599`) would mitigate the *rate*, but they are exactly the limits #552 records as **absent from the published `Capabilities` struct** — so a client cannot know what protection, if any, a given relay applies.

[...]

PR #593, opened 2026-08-24
feat(kt): add f2z-authority, the handle-ownership assertion layer

skyl:
> **⚠️ DRAFT, deliberately, and CI has not run on it.** There is an active hard
> merge-train hold (#584, #588): **#556 must complete and merge before any other
> PR moves `main`**, and E2EE PRs stay draft until it lands. This PR is not
> `gh pr ready`, must not be merged, and **the absence of a green gate here is
> the hold, not a passing run** — do not read an unchecked draft as a checked
> one. The local evidence is below, and every gate the `rs` workflow runs was
> run by hand against `0b00710`.

Adds `rs/crates/f2z-authority` — the **handle-ownership assertion layer**, and
the one place a non-cryptographic trust root enters the system.

## Why it is its own crate

`akd` proves the tree is append-only. [`KT.md`](https://github.com/free2z/zuu/blob/main/docs/e2ee/KT.md)
§4.4 proves an entry was authorized by the key that held the handle before. The
witnesses prove the log did not equivocate. **None of it answers the first
question a user actually asks** — *is this `@alice` the `@alice` I know?* —
because a key-transparency log will faithfully, verifiably and permanently
publish a key for a handle **whoever got there first**.

Something has to say who owns a handle, and no amount of cryptography derives
it. That something is a trust root, and this crate is the whole of it. It is
separate so that a self-hoster deciding whether to trust our directory, swapping
our authority for their own, or running with **none**, reads one short crate —
not a server, not a grep through a submission path, not a policy spread over
three layers.

## ⚠️ The largest thing to know before reviewing: `HandleAssertion` is not in `KT.md`

**`KT.md` v1 does not specify a `HandleAssertion`, and it does not specify
anything else that answers the question either.** `grep -rn HandleAssertion
docs/ rs/` on `main` returns nothing. So this is not "implement the merged
spec"; it is **invented here**, in `KT.md`'s own sense of that phrase, and every
invented choice below says so at the point of use, exactly as `KT.md` §3.3, §5.1
and §5.3 do for theirs.

What the merged documents *do* establish, and what this had to be consistent
with:

| Source | What it gives | What it leaves open |
|---|---|---|
| `KT.md` §4.4 | What authorizes entry *n+1* given entry *n* — `same_key`, `key_change`, `platform_reset` | **What authorizes entry 1.** §4.4's table has no row for a handle's first entry: a first entry has no previous `directory_auth_pk` to be signed by. |
| ADR 0014 | A pinned "reset authority" key exists and its use is loud, delayed and counted | Nothing about who says a handle belongs to a person in the first place |
| `WIRE.md` §14 | The charset, and that `lowercase(username)` is **not** a unique key | Who performs the mapping, and what happens when two accounts fold together |
| `KT.md` §1.2, §12 | Handle rename/transfer deferred (§13-K); reset-authority pinning named as a gap | The whole of handle *ownership* |

**That gap is the finding, and it is worth a documentation issue independent of
this PR:** the log's most security-critical mapping — handle → identity key —
has a fully specified update rule and **no specified establishment rule.** A
conforming implementation of `KT.md` as merged accepts entry 1 for `@alice` from
anybody. This crate proposes the missing piece and asks for the spec to catch up
with it; it is not a claim that the spec already said this.

## The rule that makes it sound

An assertion is a **bearer document** — bytes an authority signed, served over a
network, sitting in the log's request queue and in whatever debug archive the
operator kept. If holding one were enough to submit under the handle it names,
intercepting one would be a takeover, and the authority could not tell the thief
from the subject.

So the submission carrying an assertion **must also be signed by the identity
private key the assertion is about** (`AssertionBindingTBS`). A thief cannot
produce that signature, because producing it needs the key the thief was trying
to replace.

**This is made structurally hard to skip, not important to remember:**

- `AuthorityConfig::admit` is the **only** public verification path in the crate.
  There is no `verify_assertion`, no `check_authority_signature`, no
  `is_expired` — each would be a route to a *partly* verified assertion.
- `Submission::identity_signature` is a plain field, not an `Option`, on **every**
  branch — including the branch for a log with no authority at all.
- `AdmittedHandle` has no public constructor and no public fields, so a value of
  that type in hand *is* the argument that every rule ran. Same shape #584 uses
  for `AcceptedSubmission`.
- The binding commits to `log_id`, the handle, the identity key, the assertion
  digest **and the submission's `entry_digest`**, so neither the assertion nor
  the subject's own signature is replayable onto another entry or another log.

The negative is tested four ways in
`an_assertion_without_a_matching_identity_self_signature_is_rejected`: no
signature, the thief's own signature, the subject's real signature for a
different submission, and the subject's real signature for a different log.

## What `admit` enforces — all seventeen, in order

Cheap and flood-resistant first, signatures next, and the one rule that mutates
state **last**, so a submission that fails anything never burns a nonce.

1. An assertion is present **iff** this log has an authority.
2. Bytes decode **and re-encode identically** (`WIRE.md` §3.3, via `f2z-codec`).
3. `label` is exactly `free2z/kt/v1/handle-assertion`.
4. `log_id` is this log's.
5. `handle_id == H("free2z/kt/v1/handle-id", handle)`.
6. `handle` is the submission's handle, **byte-compared**.
7. `identity_pk` is the submission's identity key.
8. `expires_ms > issued_ms`.
9. `expires_ms - issued_ms <= max_validity_ms` — **the log's own cap.**
10. `issued_ms` is not further ahead than `clock_skew_ms`.
11. `now_ms < expires_ms`.
12. `authority_id` is in the configured set.
13. The authority's signature verifies, **strictly** (`verify_strict`).
14. `intent` matches the sequence position, **and a reset really replaces a key**.
15. `account_epoch` strictly advances.
16. **The identity key signed the binding.**
17. `(authority_id, nonce)` unseen.

Rule 9 is the one worth pausing on. **The cap is the log's, never the
assertion's.** An issuer that could choose its own validity would, once
compromised, mint one assertion good for a decade — and every later rotation of
the authority set would leave it working. Capping it at the verifier bounds the
blast radius of a compromised issuer to the cap. `an_over_long_validity_is_
rejected_by_the_log_not_the_issuer` proves it twice: once against the default,
and once by handing the *same bytes* to a log with a tighter cap and watching
them fail.

## The no-authority mode, and why it is loud

`AuthoritySet::none()` is a legitimate deployment: a self-hoster with no user
directory. It is **spelled**, not reached by leaving the set empty —
`AuthoritySet::new(vec![])` is `EmptyAuthoritySet`, because a constructor that
turned a config typo into a downgrade of what a handle *means* would be the
worst kind of quiet.

And it is reported, twice over: `AuthoritySet::status()` answers before a single
submission arrives (`VouchingStatus::Unvouched`, whose `Display` is *"handles on
this log are UNVOUCHED: no authority attests who owns them"*), and every
admission returns `Vouch::Unvouched`, which is `#[must_use]` and `Display`s as
`UNVOUCHED`. A client that renders vouched and unvouched identically has taken
the assertion layer away from its user without telling them.

What does **not** change on such a log is the identity self-signature. All that
is missing is anyone saying which person the key belongs to.

## The charset, and the function that is deliberately absent

`Handle` is `[a-z0-9_]{1,30}`, ASCII, **byte-compared, with no normalization at
comparison time** (`WIRE.md` §14.1). `Handle::parse` **rejects `Alice`; it does
not fold it.** Conformance is a *type invariant* — `DeserializeBytes` enforces the
charset, so bytes that are not a handle never decode into one and no verification
rule can ever be reached with a non-conforming handle.

**There is deliberately no `Handle::from_username`.** `WIRE.md` §14.3's
2026-08-23 correction is explicit that `lowercase(username)` is **not** a unique
key: the platform's `username` column carries a plain case-*sensitive*
`unique=True`, the case-insensitive check lives only in two serializers, and
case-variant duplicate accounts exist in production **today**. Two distinct
accounts can fold to one handle. A mapping function here would be code that
silently picks a winner in the one mapping the whole system rests on. That
judgement is the authority's, and recording it is what an assertion *is*.

## The §4.4 boundary — stricter reading adopted, per #584's finding

#584 found that `KT.md` §4.4's numbered rules never require a `same_key` entry to
keep `identity_pk` unchanged, so as enumerated a party holding only the previous
`directory_auth_pk` could publish a `same_key` entry carrying a brand-new
identity key — a key change with one signature, which ADR 0014 says the log MUST
reject.

Viewed from this crate the same hole is an assertion admitted for an entry that
**changes nothing**, which means the assertion was really being spent on
something else. Rule 14 therefore takes the stricter reading:
`Submission::previous_identity_pk` must be `None` at `entry_version == 1`, must
be `Some` above it, and a `reset` whose `identity_pk` equals the predecessor's is
`AuthorityError::IdentityUnchanged`. The log must **hold** the predecessor to
fill that in — it cannot assert what the previous key was — exactly as §4.4 rule
4 requires it to hold the previous entry to check `prev_entry_hash`.
Reported here rather than worked around, and it is a second reason `KT.md` §4.4
should gain #584's numbered rule.

## Redaction, including the trap

Hand-written `Debug` throughout, per `f2z-codec`'s pattern **and its documented
decimal-byte-list trap**: `tls_codec`'s byte vectors derive `Debug` and render
their contents as a list of decimal integers, so a hex-only redaction test passes
while everything leaks.

That trap **fired during development and was fixed in the type, not the test.**
`Submission` originally derived `Debug`; every typed field is a redacting
newtype, so it looked safe — except `assertion: Option<&[u8]>`, a bare slice,
which dumped the entire assertion as decimals. `Submission` now has a
hand-written `Debug`, and `a_submission_renders_neither_the_assertion_bytes_nor_
the_signature` is what keeps it that way.

`Handle` redacts too. A handle is not secret — it is how `@alice` is
discoverable — but it is **linkable**, and "who asked about whom" is exactly the
metadata `THREAT-MODEL.md` §4.1 bounds rather than removes. A trace log printing
every handle a client resolved hands the operator the social graph the bound was
supposed to cost them. `as_str()` is the deliberate way out.

## `f2z-assert`

An example issuer, so a self-hoster running a log with **no web application** can
issue by hand. `keygen`, `authority` (prints the `(authority_id, public key)` pair
an `AuthoritySet` entry takes), `issue`.

Zero dependencies beyond the library. Randomness is one `open` and one
`read_exact` on `/dev/urandom` — a `getrandom` in an issuing tool is a dependency
**in the trust root** — and on a platform without it, `keygen` refuses and
`issue` requires an explicit `--nonce`. Silently substituting a weaker source in
a tool that mints identity claims is the failure nobody would notice. A
misspelled flag is an error, not a default: `--handel alice` quietly issuing for
something else is exactly the mistake this must not have.

It also prints, on every issue, that *"the submitter must ALSO sign the binding
with the identity private key; an assertion alone is not a submission."*

## Ambiguity calls — every one

Since `KT.md` specifies none of this, most of the design *is* a call. These are
the ones where a different answer was defensible.

1. **`handle_id` is derived, not assigned.** `H("free2z/kt/v1/handle-id",
   handle)`, so "it matches the handle" is checkable. An authority-assigned
   account id would not be — and would leak a second identifier. It is redundant
   with `handle` by construction; kept because it lets a log index and store
   assertions without the plaintext handle, and because the field was specified
   in the brief. **If the spec wants an opaque account id instead, this is the
   field to change.**
2. **`authority_id = H(label, authority_pk)`**, the `relay_id` shape of
   `WIRE.md` §5.2 — a derived id cannot disagree with its key, so a config
   cannot name one authority and carry another's. `AuthorityKey::from_parts`
   checks it where it was typed.
3. **No `kt_version` field.** `DirectoryEntryTBS` has one; the `label` here
   already spells `/v1/`, and a second version marker inside one structure is a
   second thing to disagree with the first. `KT_VERSION` is a crate constant.
4. **Intent ↔ sequence.** `bind` iff `entry_version == 1`, `reset` iff above.
   `same_key`/`key_change` entries are authorized cryptographically and need no
   assertion; one presented with them is a category error. `entry_version == 0`
   is not a position at all.
5. **`account_epoch` strictly increases.** A reset is a distinct
   account-ownership event, so an assertion for an epoch already seen is one
   already spent. Non-strict would let an old assertion be re-presented after
   its nonce aged out.
6. **Expiry is not extended by the clock skew, `issued_ms` is.** The asymmetry
   is fail-closed, and it keeps the widest acceptance window at
   `max_validity_ms + clock_skew_ms` — a number an operator can state.
   `WIRE.md` §5.5 applies skew both ways; that is a relay's frame window, not a
   credential's lifetime.
7. **Nonce keyed on `(authority_id, nonce)`, not the assertion digest.** The
   digest only catches a byte-identical replay; the pair also catches an issuer
   reusing a nonce across two *different* claims. Tested.
8. **Retention is derived, not published.** Entries live to
   `expires_ms + clock_skew_ms`. #586's relay defect — a published window
   shorter than the one it protects — **cannot be configured here**, because
   there is nothing to publish.
9. **The ledger refuses rather than evicts**, following `f2z-relay-proto`'s
   `SeenSet`. Eviction reopens the replay window under load, which is when an
   attacker would arrange to be.
10. **Rotation is set membership only** — no succession message, no chain. A
    rotation ceremony would be a second trust root guarding the first, in the
    crate whose whole point is being short enough to read.
11. **A zero `assertion_digest` marks the unvouched binding**, the "absent value
    is all-zero" idiom of `WIRE.md` §5.1, so one transcript covers both
    deployments. `the_two_deployments_do_not_leak_into_each_other` proves an
    unvouched binding does not authorize anything on a vouched log.
12. **`DEFAULT_MAX_VALIDITY_MS = 900_000` (15 min) is a proposed placeholder**,
    marked as such at its definition in `KT.md`'s style. Measure before
    shipping.
13. **`Handle` is defined here, not shared.** #584 adds one to `f2z-kt-core`;
    that crate is unmerged, and depending on an open PR is not a dependency.
    **When #584 lands, one of the two should move to `f2z-codec`** — flagged
    rather than pre-resolved.
14. **A local `opaque_fixed!` copy** rather than exporting `f2z-codec`'s. That
    macro's job is to be applied carefully, once, by someone who has read why;
    exporting it makes it a convenience.
15. **`AuthorityError` is finer than the wire.** §9.5 has one code for the whole
    of §4.4 — right on the wire, wrong in a log. Detail stays local,
    `kt_error_code()` goes out; configuration faults map to `ERR_INTERNAL` so a
    misconfigured log never blames a user for an operator error.
16. **`too_many_arguments` allowed on `HandleAssertionTBS::new`**, narrowly and
    with justification. A builder would make each field omissible: forgetting
    `expires_ms` on a builder compiles and mints a zero-expiry assertion;
    forgetting it here does not compile.

## Verification — run locally against `0b00710`

CI has **not** run (see the banner). Every gate `.github/workflows/rs.yml` would
run was run by hand:

| Gate | Command | Result |
|---|---|---|
| rustfmt | `scripts/check-rust-fmt.sh --root rs` | `All 4 Rust crate(s) under rs/ are formatted (rustfmt from 1.97.1).` |
| clippy `-D warnings` | `scripts/check-rust-clippy.sh --root rs` | `All 4 Rust crate(s) under rs/ are clippy-clean (clippy from 1.97.1).` |
| cargo-deny | `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` | `advisories ok, bans ok, licenses ok, sources ok` |
| toolchain pin | `check-rust-toolchain.sh --toolchain-file … --manifest ×3` | `Every Rust toolchain pin agrees with wallet/rust-toolchain.toml: channel 1.97.1, MSRV 1.97.` |
| tests | `cargo test --locked --all-targets` | **37 unit + 23 adversarial + 8 redaction** in this crate, 0 failed; whole workspace 0 failed |
| doc tests | `cargo test --locked --doc` | 1 passed (the `lib.rs` end-to-end example) |
| wasm32 | `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec -p f2z-relay-proto -p f2z-authority` | `Finished` |
| workflow policy | `check-workflow-gates.mjs`, `check-github-actions-pins.mjs` | both pass |

### The adversarial rejections, by name

Every one starts from a submission that **is** valid — asserted by
`a_correct_submission_is_admitted` — and breaks exactly one thing, so a passing
negative is evidence about that rule and not about a fixture that stopped
building.

| Attack | Test | Refusal |
|---|---|---|
| Expired assertion | `an_expired_assertion_is_rejected` | `Expired` (and accepted at `expires_ms - 1`, so the boundary is pinned) |
| Wrong `log_id` | `an_assertion_for_another_log_is_rejected` | `WrongLog` |
| Replayed nonce | `a_replayed_nonce_is_rejected` | `ReplayedNonce` |
| Over-long validity | `an_over_long_validity_is_rejected_by_the_log_not_the_issuer` | `ValidityTooLong`, against both the default cap and a tighter log's |
| **Missing identity self-signature** | `an_assertion_without_a_matching_identity_self_signature_is_rejected` | `BadIdentitySignature`, four ways |
| Non-conforming handle | `a_non_conforming_handle_never_becomes_a_handle` | does not decode at all; `Handle::parse` refuses `Alice`, `.`, `-`, `@`, Cyrillic `а`, 31 bytes |
| Unconfigured authority | `an_assertion_from_an_unconfigured_authority_is_rejected` | `UnknownAuthority` — and adding the key is the whole of accepting it |
| Forged authority signature | `an_assertion_the_authority_did_not_sign_is_rejected` | `BadAuthoritySignature` |
| Self-contradiction | `an_assertion_that_disagrees_with_itself_or_its_submission_is_rejected` | `HandleIdMismatch` / `HandleMismatch` / `IdentityMismatch` / `WrongLabel` |
| Backwards window | `a_backwards_or_empty_validity_window_is_rejected` | `EmptyValidity` |
| Future issue | `an_assertion_issued_too_far_in_the_future_is_rejected` | `NotYetIssued` (inclusive at the skew) |
| Wrong sequence position | `intent_is_tied_to_the_position_in_the_entry_sequence` | `IntentMismatch`, five shapes |
| Reset that changes nothing | `a_reset_that_keeps_the_key_it_replaces_is_rejected` | `IdentityUnchanged` — the §4.4 boundary |
| Stale `account_epoch` | `an_account_epoch_that_does_not_advance_is_rejected` | `AccountEpochRegression` |
| Non-canonical bytes | `non_canonical_assertion_bytes_are_rejected` | `Malformed` on a trailing byte and on truncation |
| Nonce burned by a failure | `a_failed_submission_never_burns_the_nonce` | ledger empty after the failure; the real submission still works |
| Issuer nonce reuse | `a_reused_nonce_across_two_different_assertions_is_still_a_replay` | `ReplayedNonce` |
| Full ledger | `a_full_ledger_refuses_rather_than_forgetting` | `LedgerFull`, never an eviction |
| Assertion to an unvouched log | `the_two_deployments_do_not_leak_into_each_other` | `UnexpectedAssertion` |
| No assertion to a vouched log | same | `MissingAssertion` |
| Unvouched binding on a vouched log | same | `BadIdentitySignature` |

`f2z-assert` was exercised end to end: `keygen` → `authority` → `issue`, plus
`--log-id 11` (`expected 32 bytes … found 1`), `--handel` (`unknown flag`), and
`--handle Alice` (`the handle is not [a-z0-9_]{1,30} (WIRE.md §14.1)`).

## Registration

`rs/README.md` warns that `check-rust-toolchain.sh` does **not** discover
manifests (#553), so every new crate needs another `--manifest` flag in the same
commit. Both are added — `--manifest rs/crates/f2z-authority/Cargo.toml` and
`-p f2z-authority` on the `rs_wasm` job — plus the crate table and the two
command blocks in `rs/README.md`. `check-workflow-gates.mjs` and
`check-github-actions-pins.mjs` both still pass.

`rs/Cargo.lock` gains one entry. Members are `crates/*`, so a concurrent crate
(#584's `f2z-kt-core`) does not conflict in `rs/Cargo.toml`; if the lockfile
conflicts, keep both entries and regenerate.

## Follow-ups this PR does not do

1. **`KT.md` should specify handle ownership.** §4.4 authorizes entry *n+1* and
   nothing authorizes entry 1. Worth its own issue; this crate is the proposal,
   not the spec.
2. **`KT.md` §4.4 should gain #584's `same_key` rule** (identity key unchanged),
   for the reason above and for #584's.
3. **`Handle` deduplication** with `f2z-kt-core` once #584 lands.
4. **`DEFAULT_MAX_VALIDITY_MS` needs measuring**, like §5's cadence placeholders.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -75,7 +75,8 @@ jobs:
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
             --manifest rs/crates/f2z-codec/Cargo.toml \
-            --manifest rs/crates/f2z-relay-proto/Cargo.toml
+            --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+            --manifest rs/crates/f2z-authority/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -300,7 +301,8 @@ jobs:
           cargo build --locked --lib \
             --target "${{ steps.rust_toolchain.outputs.target }}" \
             -p f2z-codec \
-            -p f2z-relay-proto
+            -p f2z-relay-proto \
+            -p f2z-authority
 
   # Branch protection should require this stable context, `rs / gate`. It
   # succeeds only when the change detector and every applicable rs job
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -171,6 +171,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-authority"
+version = "0.1.0"
+dependencies = [
+ "ed25519-dalek",
+ "f2z-codec",
+ "tls_codec",
+]
+
+[[package]]
 name = "f2z-codec"
 version = "0.1.0"
 dependencies = [
--- a/rs/README.md
+++ b/rs/README.md
@@ -76,7 +76,8 @@ decided is `wallet/rust-toolchain.toml`, and
 ```
 scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
                                 --manifest rs/crates/f2z-codec/Cargo.toml \
-                                --manifest rs/crates/f2z-relay-proto/Cargo.toml
+                                --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+                                --manifest rs/crates/f2z-authority/Cargo.toml
 ```
 
 fails the pull request the moment the two disagree. `.github/workflows/rs.yml`
@@ -110,8 +111,9 @@ prevent.
 |---|---|
 | [`crates/f2z-codec`](./crates/f2z-codec) | The canonical encoding layer of `WIRE.md`: `tls_codec` wrappers for every wire structure, the domain-separated signing transcript (§5), the redacting newtypes, and the padding-bucket validator (§9). `no_std` + `alloc`, `#![forbid(unsafe_code)]`, no I/O, no async runtime, and it builds for `wasm32-unknown-unknown`. |
 | [`crates/f2z-relay-proto`](./crates/f2z-relay-proto) | The protocol layer above it: signed-command construction and verification in §5.1's exact order, the timestamp window and fail-closed seen-set (§5.5), the queue lifecycle and ACK arithmetic (§7, §8), the capability document (§11), the `HELLO` proof of possession (§5.2), and §4.3's typed in-flight window. Same constraints — `no_std` + `alloc`, no I/O, no clock, no randomness, and it builds for `wasm32-unknown-unknown`. |
+| [`crates/f2z-authority`](./crates/f2z-authority) | The directory's **non-cryptographic trust root**, isolated on purpose: the `HandleAssertion` an authority signs to say who owns a handle, the verifier that admits one, the `AuthoritySet` (including an explicit, *reported*, no-authority mode), and `f2z-assert` for issuing by hand. Same constraints, and it reaches WASM because clients verify assertions too. |
 
-**The relay and the clients link both.** That is the licence boundary in
+**The relay and the clients link all three.** That is the licence boundary in
 practice: everything here is MIT because a third-party relay, ZUULI and the WASM
 web client all compile the same rules, and a rule that two implementations
 disagree about is how ciphertext gets deleted before it is read.
@@ -134,7 +136,8 @@ reasons, and each is enforced by a test rather than by intent:
 ```bash
 cd rs
 cargo test
-cargo build --target wasm32-unknown-unknown --lib -p f2z-codec -p f2z-relay-proto
+cargo build --target wasm32-unknown-unknown --lib \
+  -p f2z-codec -p f2z-relay-proto -p f2z-authority
 ```
 
 The gates are the repository's shared scripts, pointed here with `--root`:
@@ -142,7 +145,8 @@ The gates are the repository's shared scripts, pointed here with `--root`:
 ```bash
 scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
                                 --manifest rs/crates/f2z-codec/Cargo.toml \
-                                --manifest rs/crates/f2z-relay-proto/Cargo.toml
+                                --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+                                --manifest rs/crates/f2z-authority/Cargo.toml
 scripts/check-rust-fmt.sh    --root rs
 scripts/check-rust-clippy.sh --root rs
 scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
--- a/rs/crates/f2z-authority/Cargo.toml
+++ b/rs/crates/f2z-authority/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "f2z-authority"
+version = "0.1.0"
+description = "Handle-ownership assertions for the free2z key-transparency log: the one non-cryptographic trust root, isolated so it can be read in full"
+edition.workspace = true
+# A restatement of wallet/rust-toolchain.toml's channel in Cargo's MSRV form,
+# registered with `scripts/check-rust-toolchain.sh --manifest` so it cannot
+# drift. Not inherited from the workspace: that check reads this literal line.
+rust-version = "1.97"
+# Permissive, deliberately and explicitly. Clients verify assertions too — a
+# handle whose vouching a client cannot check is a handle nobody vouched — so
+# this crate reaches the browser and the ZUULI app as well as the log. The
+# AGPL-3.0 boundary starts at the server binaries, not here. Stated literally
+# rather than inherited so it is visible to anyone opening the crate.
+# rs/deny.toml enforces it.
+license = "MIT"
+repository.workspace = true
+publish.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ed25519-dalek = { workspace = true }
+# `version` beside `path` is not redundant: rs/deny.toml sets
+# `wildcards = "deny"`, and a bare path dependency is a wildcard requirement.
+f2z-codec = { path = "../f2z-codec", version = "0.1.0" }
+tls_codec = { workspace = true }
+
+# The example issuer. It is a `bin` in this package rather than a crate of its
+# own so that a self-hoster reading "how do I issue an assertion by hand" reads
+# the same source tree that defines what an assertion is. The library is
+# `no_std`; this target is not, and that is fine — `rs.yml`'s wasm job builds
+# `--lib`, so a browser never compiles it.
+[[bin]]
+name = "f2z-assert"
+path = "src/bin/f2z-assert.rs"
--- a/rs/crates/f2z-authority/src/assertion.rs
+++ b/rs/crates/f2z-authority/src/assertion.rs
@@ -1,0 +1,386 @@
+//! The two signed structures: what an authority says, and what the identity
+//! key it names says back.
+//!
+//! ```text
+//! struct {
+//!     opaque label<0..255>;      /* exactly "free2z/kt/v1/handle-assertion" */
+//!     opaque authority_id[32];   /* H("free2z/kt/v1/authority-id", authority_pk) */
+//!     opaque log_id[32];
+//!     opaque handle<1..30>;
+//!     opaque handle_id[32];      /* H("free2z/kt/v1/handle-id", handle)          */
+//!     opaque identity_pk[32];    /* ISK.public — the key this vouches for        */
+//!     uint8  intent;             /* bind(1), reset(2)                            */
+//!     uint32 account_epoch;
+//!     uint64 issued_ms;
+//!     uint64 expires_ms;
+//!     opaque nonce[16];
+//! } HandleAssertionTBS;
+//!
+//! struct {
+//!     HandleAssertionTBS assertion;
+//!     opaque             signature[64];   /* Ed25519 by authority_id's key */
+//! } HandleAssertion;
+//!
+//! struct {
+//!     opaque label<0..255>;      /* exactly "free2z/kt/v1/assertion-binding" */
+//!     opaque log_id[32];
+//!     opaque handle<1..30>;
+//!     opaque identity_pk[32];
+//!     opaque assertion_digest[32];  /* H(".../assertion-digest", tls_codec(HandleAssertion));
+//!                                      all-zero when the log has no authority */
+//!     opaque entry_digest[32];      /* the submission's AkdValue, KT.md §3.3 */
+//! } AssertionBindingTBS;
+//! ```
+//!
+//! # Why there are two, and not one
+//!
+//! An assertion is a **bearer** document: it is bytes an authority signed,
+//! served over a network, sitting in a log's request queue and in whatever
+//! debugging archive the operator kept. Anyone who obtains a copy holds
+//! everything the authority said. If holding it were enough to submit under the
+//! handle it names, then intercepting one — from a TLS-terminating proxy, from
+//! a log's access log, from a stolen backup — would be a handle takeover, and
+//! the authority would have no way to tell the thief from the subject.
+//!
+//! [`AssertionBindingTBS`] closes that. It is signed by the **identity private
+//! key the assertion is about**, and it commits to the exact submission the
+//! assertion is being presented with. A thief holding the assertion cannot
+//! produce it, because producing it needs the key the assertion vouches *for* —
+//! which is the key the thief was trying to replace. The stolen assertion is
+//! worthless.
+//!
+//! `KT.md` requires this pairing. This crate makes it structural: there is no
+//! function here or in [`crate::authority`] that checks an authority's
+//! signature without also checking the binding, so a caller cannot arrive at a
+//! verified assertion by a route that skipped it.
+//!
+//! Including `entry_digest` rather than only the assertion is what stops the
+//! *subject's own* binding from being reusable: a binding signed for one
+//! submission does not authorize a second, so a log that saw one cannot resubmit
+//! it against different entry bytes. Including `log_id` rather than relying on
+//! the digest to carry it is one field for a whole class of cross-log replay.
+
+use alloc::vec::Vec;
+// `tls_codec`'s derive macros build their error strings with `format!` and
+// return `Vec<u8>`; both need to be in scope in a `no_std` crate.
+use alloc::format;
+
+use f2z_codec::canonical::Canonical;
+use f2z_codec::hash::hash;
+use f2z_codec::types::{Digest, PublicKey, ShortBytes, Signature};
+use tls_codec::{TlsDeserializeBytes, TlsSerializeBytes, TlsSize};
+
+use crate::error::{AuthorityError, Result};
+use crate::key::SigningKey;
+use crate::labels::{LABEL_ASSERTION_BINDING_TBS, LABEL_ASSERTION_DIGEST, LABEL_ASSERTION_TBS};
+use crate::types::{AssertionNonce, AuthorityId, Handle, HandleId, Intent, LogId, authority_id};
+
+/// What an authority signs.
+///
+/// Every field is checked by [`AuthorityConfig::admit`] — the type carries no
+/// invariant of its own beyond decoding, deliberately, so that a decoded
+/// assertion is *data* until a policy has judged it.
+///
+/// The derived `Debug` is safe because every field that holds bytes is one of
+/// the redacting newtypes; `label` renders because [`ShortBytes`] prints
+/// printable ASCII and that is the field's entire purpose.
+///
+/// [`AuthorityConfig::admit`]: crate::authority::AuthorityConfig::admit
+#[derive(Clone, Debug, PartialEq, Eq, TlsSize, TlsSerializeBytes, TlsDeserializeBytes)]
+pub struct HandleAssertionTBS {
+    /// Exactly [`LABEL_ASSERTION_TBS`]. The domain-separated signing prefix.
+    pub label: ShortBytes,
+    /// Which authority is speaking.
+    pub authority_id: AuthorityId,
+    /// Which log this is for.
+    pub log_id: LogId,
+    /// The handle being vouched for.
+    pub handle: Handle,
+    /// `H("free2z/kt/v1/handle-id", handle)`.
+    pub handle_id: HandleId,
+    /// The identity key (ISK.public) the handle is being bound to.
+    pub identity_pk: PublicKey,
+    /// Bind, or reset.
+    pub intent: Intent,
+    /// The authority's counter for the *account* behind the handle. It
+    /// increments whenever the account changes hands or is recovered, which is
+    /// the only thing that justifies a reset.
+    pub account_epoch: u32,
+    /// When the authority issued this, in milliseconds since the Unix epoch.
+    pub issued_ms: u64,
+    /// When it stops being usable.
+    pub expires_ms: u64,
+    /// 16 bytes of issuer CSPRNG, fresh per assertion.
+    pub nonce: AssertionNonce,
+}
+
+impl HandleAssertionTBS {
+    /// Build an assertion body, deriving `authority_id` and `handle_id` rather
+    /// than accepting them.
+    ///
+    /// Deriving both here is why a hand-built assertion cannot contradict
+    /// itself in the two ways that would otherwise be easy: naming one
+    /// authority while being signed by another, and naming one handle while
+    /// being indexed under another.
+    ///
+    /// # Errors
+    ///
+    /// [`AuthorityError::Malformed`] if the label does not fit `<0..255>`,
+    /// which it always does — the check is there so the constructor has no
+    /// unwrap in it.
+    // Nine arguments, and deliberately not a builder. Every one of them is a
+    // field of a structure that is about to be signed, and a builder would make
+    // each one omissible: forgetting `expires_ms` on a builder yields a
+    // compiling program that mints an assertion with a zero expiry, while
+    // forgetting it here does not compile. `authority_id` and `handle_id` are
+    // the two fields that are *not* arguments, because they are derived.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        authority_pk: &PublicKey,
+        log_id: LogId,
+        handle: Handle,
+        identity_pk: PublicKey,
+        intent: Intent,
+        account_epoch: u32,
+        issued_ms: u64,
+        expires_ms: u64,
+        nonce: AssertionNonce,
+    ) -> Result<Self> {
+        let handle_id = handle.handle_id();
+        Ok(Self {
+            label: ShortBytes::new(LABEL_ASSERTION_TBS).map_err(AuthorityError::from)?,
+            authority_id: authority_id(authority_pk),
+            log_id,
+            handle,
+            handle_id,
+            identity_pk,
+            intent,
+            account_epoch,
+            issued_ms,
+            expires_ms,
+            nonce,
+        })
+    }
+
+    /// The exact bytes the authority signs.
+    ///
+    /// # Errors
+    ///
+    /// [`AuthorityError::Malformed`] if the structure cannot be encoded.
+    pub fn signing_bytes(&self) -> Result<Vec<u8>> {
+        Ok(self.encode_canonical()?)
+    }
+
+    /// Sign this body, producing the assertion an issuer hands out.
+    ///
+    /// # Errors
+    ///
+    /// [`AuthorityError::Malformed`] if the structure cannot be encoded.
+    pub fn sign(self, key: &SigningKey) -> Result<HandleAssertion> {
+        let signature = key.sign(&self.signing_bytes()?);
+        Ok(HandleAssertion {
+            assertion: self,
+            signature,
+        })
+    }
+}
+
+/// A signed assertion: the bearer document.
+#[derive(Clone, Debug, PartialEq, Eq, TlsSize, TlsSerializeBytes, TlsDeserializeBytes)]
+pub struct HandleAssertion {
+    /// What was signed.
+    pub assertion: HandleAssertionTBS,
+    /// Ed25519 over `tls_codec(HandleAssertionTBS)`, by the key whose digest is
+    /// `assertion.authority_id`.
+    pub signature: Signature,
+}
+
+impl HandleAssertion {
+    /// `H("free2z/kt/v1/assertion-digest", tls_codec(HandleAssertion))` — what
+    /// the identity key's binding commits to.
+    ///
+    /// Over the assertion **including its signature**, on `KT.md` §4.2's
+    /// reasoning for `prev_entry_hash`: committing to the authorization as well
+    /// as the contents means an assertion cannot be re-signed after the subject
+    /// has bound itself to it.
+    ///
+    /// # Errors
+    ///
+    /// [`AuthorityError::Malformed`] if the structure cannot be encoded.
+    pub fn digest(&self) -> Result<Digest> {
+        Ok(hash(LABEL_ASSERTION_DIGEST, &self.encode_canonical()?))
+    }
+}
+
+/// What the identity key signs: this assertion, for this submission, on this
+/// log.
+///
+/// See the module note. `assertion_digest` is 32 zero bytes when the log has no
+/// authority — the "absent value is all-zero" idiom `WIRE.md` §5.1 already
+/// uses — so that a log with no directory still requires the submitter to
+/// answer for itself, and so that one transcript covers both configurations.
+#[derive(Clone, Debug, PartialEq, Eq, TlsSize, TlsSerializeBytes, TlsDeserializeBytes)]
+pub struct AssertionBindingTBS {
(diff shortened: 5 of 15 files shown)


skyl:

skyl:

skyl:
## Adversarial review — `f2z-authority` @ `7adf6ea`

Reviewed in an isolated worktree at the PR head (`7adf6ea`, not the `0b00710` the body cites), toolchain 1.97.1, cargo run from `rs/`. Baseline is green and reproduces the body's numbers exactly: **37 unit + 23 adversarial + 8 redaction = 68, 0 failed**; doc test 1 passed; fmt/clippy/deny all clean; wasm32 `--lib` builds for all three crates. `rs/deny.toml` and `.github/workflows/zuuli.yml` are untouched (`git diff --name-only origin/main...HEAD | grep -c deny.toml` → `0`, same for `zuuli`). Registration is complete and correct — both `--manifest` and `-p f2z-authority` are in `rs.yml`, plus the crate table and both command blocks in `rs/README.md`.

**26 mutations, 21 caught, 5 not.** Two of the five are blocking; three are the #581 gap. The PR is currently `CLOSED (draft)` under the merge-train lock and will be reopened, so everything below is pre-reopen, not a post-merge follow-up.

---

# BLOCKING 1 — the #589 defect, reproduced verbatim in a new crate

`key.rs:147` `a_small_order_key_verifies_nothing` asserts a property it does not exercise. **This is the exact test #597 just fixed in `f2z-relay-proto`, re-introduced here.**

The fixture is `A = [0x01, 0x00×31]` with `sig = [0u8; 64]`. Measured:

```
all-zero sig:      plain verify = false, verify_strict = false
forged sig A||0:   plain verify accepted 64/64 distinct messages, verify_strict accepted 0/64
```

Both verifiers reject the all-zero signature, so the test passes with or without strictness. The real universal forgery — `R = A` (the identity point), `s = 0` — is never tried.

**Mutation M1**, `key.rs:62` `verify_strict(message, &signature)` → `verify(message, &signature)` (plus the `Verifier as _` import the swap needs), re-verified with a forced rebuild:

```
### M1 verify_strict -> verify (+Verifier import) => NOT CAUGHT (suite green)
test result: ok. 37 passed ... ok. 23 passed ... ok. 8 passed
```

Not theoretical. Reaching `admit` with it — an `AuthoritySet::none()` log, `identity_pk = A`, `identity_signature = A || [0u8;32]`, no private key involved anywhere:

| build | `admit(...)` |
|---|---|
| as shipped (`verify_strict`) | `Err(BadIdentitySignature)` ✅ |
| under M1 (`verify`) | **`Ok(AdmittedHandle)` — the handle is claimed** |

So the *guard* is correct — `verify_strict` really is what runs, and the module note at `key.rs:1-11` is accurate. What is missing is any test that would notice if it stopped being. Two other tests in that file (`a_key_that_is_not_a_curve_point_never_verifies_anything`, `a_signature_verifies_only_over_the_message_it_covers`) also use the all-zero signature or a plain wrong-message case, so neither closes it either.

**Fix:** lift #597's fixture from `rs/crates/f2z-relay-proto/src/key.rs` — the `IDENTITY_A` constant, the loop over N distinct messages, and critically its *positive* assertion that plain verification **accepts** the forgery (without that, a future dalek change silently turns the test back into a tautology).

---

# BLOCKING 2 — `admit` refuses every `same_key` and `key_change` entry on a vouched log

Rule 1 is *"an assertion is present **iff** this log has an authority"* (`authority.rs:549`, `:594-603`) — not *iff this entry kind needs one*. Rule 14 (`authority.rs:658-682`) then admits exactly two shapes: `bind` at `entry_version == 1` with no predecessor, and `reset` above 1 with a predecessor whose `identity_pk` **differs**.

`KT.md` §4.4 has three entry kinds. Measured against a vouched log:

| §4.4 case | submission | `admit` |
|---|---|---|
| `same_key` at v5 (adding a device) | no assertion — the platform is not involved | `Err(MissingAssertion)` |
| `same_key` at v5 | with a `reset` assertion (the only intent legal above 1) | `Err(IdentityUnchanged)` |
| `key_change` at v5 (user rotates own ISK per ADR 0014) | no assertion | `Err(MissingAssertion)` |

There is no third door. A log that wires `admit` in as the crate documents it — *"the **only** public verification path"*, *"`AdmittedHandle` … a value of that type in hand **is** the argument that every rule ran"* (`authority.rs:5-10`, `:286-292`) — cannot publish a device addition or a user-initiated rotation at all. And the only way to make one pass is for the platform to mint a `reset` assertion for it.

That inverts the ADR this crate is built to respect. [ADR 0014](docs/e2ee/decisions/0014-directory-key-rotation.md) states the consequence as *"Neither the log operator nor the platform can change a live key silently… **Case 2 needs a signature the platform does not have**"*. Under rule 1 as written, case 2 needs a signature only the platform has. That is precisely the property `f2z-authority` exists to bound, flipped by the rule it leads with.

The PR body knows the intent — *"`same_key`/`key_change` entries are authorized cryptographically and need no assertion; one presented with them is a category error"* — but nothing in the code or the doc comment says so, and `Submission` already carries `entry_version` and `previous_identity_pk`, i.e. everything needed to make the distinction. Either:

- add the discriminator (`Submission::kind: EntryKind`, or `requires_assertion: bool`), so rule 1 reads *iff the entry kind requires one* and `AdmittedHandle` stays the single proof; **or**
- state at `admit` that it judges only `bind`/`reset` submissions, that `same_key`/`key_change` bypass it entirely, and stop claiming the type is the proof that every submission was judged — because then it is not.

The first is the one that keeps the crate's own design argument true.

---

# The mutation table

Every mutation was diffed after application and confirmed to land on the intended line before judging. (`s|…|…|` bit me once — a `\|` inside a pipe delimiter terminates the pattern early and the edit lands at line 1; M15 was re-run with `#` as delimiter.) Each row is a fresh copy of the pristine file plus one change; `M1`/`M8` were additionally re-run with a forced `touch` rebuild to rule out mtime caching.

| # | Mutation | Result |
|---|---|---|
| M1 | `key.rs:62` `verify_strict` → `verify` | **NOT CAUGHT** |
| M2 | `authority.rs:641` expiry `now_ms >= expires_ms` → `>` | caught by `an_expired_assertion_is_rejected` |
| M3 | `authority.rs:635` validity cap `> max` → `> max*2` | caught by `an_over_long_validity_is_rejected_by_the_log_not_the_issuer` |
| M4 | `authority.rs:688` epoch `<= previous` → `< previous` | caught by `an_account_epoch_that_does_not_advance_is_rejected` |
| M5 | `authority.rs:624` rule 7 (`identity_pk` match) deleted | caught by `an_assertion_that_disagrees_with_itself_or_its_submission_is_rejected` |
| M6 | `authority.rs:694` rule 16 (binding signature) deleted on the vouched path | caught by 3 tests incl. `an_assertion_without_a_matching_identity_self_signature_is_rejected` |
| M7 | `ledger.observe` moved **before** rule 16 (the #590 shape) | caught by `a_failed_submission_never_burns_the_nonce` |
| M8 | `assertion.rs` swap `handle_id` ⇄ `identity_pk` wire order | **NOT CAUGHT** |
| M9 | `assertion.rs` swap `issued_ms` ⇄ `expires_ms` wire order | **NOT CAUGHT** |
| M10 | `assertion.rs:108` `account_epoch: u32` → `u64` | caught (compile error, not a test) |
| M11 | `assertion.rs` swap `assertion_digest` ⇄ `entry_digest` in `AssertionBindingTBS` | **NOT CAUGHT** |
| M12 | `authority.rs:612` rule 3 (`label`) deleted | caught by `an_assertion_that_disagrees_with_itself_or_its_submission_is_rejected` |
| M13 | `error.rs:148` `WrongLog` → `ERR_BAD_AUTHORIZATION` instead of `ERR_UNSUPPORTED_VERSION` | caught by `every_refusal_carries_a_kt_error_code_that_does_not_blame_the_wrong_party` |
| M14 | `authority.rs:678` reset guard `*previous == identity_pk` → `!=` | caught by `a_reset_that_keeps_the_key_it_replaces_is_rejected` + 1 |
| M15 | `types.rs:194` `len() > HANDLE_MAX_LEN` → `> HANDLE_MAX_LEN + 1` | caught by `the_length_bounds_are_one_and_thirty` |
| M16 | `authority.rs:659` `entry_version == 0` treated as `Bind` | caught by `intent_is_tied_to_the_position_in_the_entry_sequence` |
| M17 | `authority.rs:648` `.find(body.authority_id)` → `.keys().first()` | caught by `an_assertion_from_an_unconfigured_authority_is_rejected` |
| M18 | `types.rs:282` decoder stops enforcing the handle charset | caught by `a_handle_round_trips_and_a_bad_one_does_not_decode` |
| M19 | `nonce.rs:164` retention drops the skew margin | caught by `an_entry_outlives_the_assertion_it_covers_by_the_skew` |
| M20 | `authority.rs:638` future-issue window × 100 | caught by `an_assertion_issued_too_far_in_the_future_is_rejected` |
| M21 | `authority.rs:601` `UnexpectedAssertion` deleted | caught by `a_log_with_no_authority_refuses_an_assertion_rather_than_ignoring_it` |
| M22 | `authority.rs:531` binding stops committing to `assertion.digest()` | caught by `the_two_deployments_do_not_leak_into_each_other` |
| M23 | `assertion.rs` swap `authority_id` ⇄ `log_id` wire order | **NOT CAUGHT** |
| M24 | `authority.rs:621` rule 6 (handle byte-compare) deleted | caught by `an_assertion_that_disagrees_with_itself_or_its_submission_is_rejected` |
| M25 | `authority.rs:618` rule 5 (`handle_id` agreement) deleted | caught by same |
| M26 | `authority.rs:107` `AuthorityKey::from_parts` stops checking derivation | caught by `a_mistyped_key_id_is_caught_where_it_was_typed` |

**21 of 26 caught.** The authorization table (rules 1, 3–7, 9–17) and the ledger are genuinely well covered: every rule I deleted or inverted was named by a test, usually the one the body's table claims. #590's shape (state written before verification) is caught by construction here — `ledger.observe` is last, and `a_failed_submission_never_burns_the_nonce` proves it. The tests are not constructor-limited the way `f2z-relay-proto`'s were: `HandleAssertionTBS`'s fields are `pub`, and `admit_with` (`tests/adversarial.rs:128`) rebuilds the binding **after** the mutation, so a hostile assertion is expressible and the negative tests really do reach the rule they name.

---

# FINDING 3 — #581's gap is unfixed here: four wire-format swaps pass green

M8, M9, M11 and M23 each change **what every Ed25519 signature in this crate commits to** and leave all 68 tests green. `handle_id` ⇄ `identity_pk` is the same shape as #581's `address`/`signer_key` swap in `CommandTranscript`.

They are invisible because everything is self-consistent by construction: `HandleAssertionTBS::new` *derives* `handle_id` and `authority_id`, `admit` re-encodes with the same code that signed, and `config.binding()` is used by signer and verifier alike (`authority.rs:514-516` says so as a virtue — it is, for agreement; it also means the transcript layout is untested). The crate has exactly one hand-derived vector, for `Handle` alone (`types.rs:431`, `[5, b'a', b'l', b'i', b'c', b'e']`), and one prefix assertion (`assertion.rs` `the_signing_prefix_is_the_label_field`, which checks only that the encoding *starts with* the label).

**Fix:** one hand-derived vector per signed structure — `HandleAssertionTBS`, `HandleAssertion`, `AssertionBindingTBS` — asserting the full byte string of a fixed fixture, exactly as #581 did for `f2z-codec`. The module doc at `assertion.rs:4-33` already writes the layout out in TLS presentation syntax; the test is that comment turned into bytes.

---

# FINDING 4 — a test named "anywhere in the namespace" covers 7 of 13 labels, and completing it fails

`labels.rs:85` `no_label_is_a_prefix_of_another_anywhere_in_the_namespace`, over `LABELS ∪ KT_LABELS`. `KT_LABELS` (`labels.rs:69`) declares itself *"the labels `KT.md` v1 already spends"* and lists 7. `KT.md` spends 13:

```
$ grep -on 'free2z/kt/v1/[a-z:-]*' docs/e2ee/KT.md | sed 's/.*://' | sort -u
cosig  entry  fault  log-id  log-key-transition  prev  receipt
reset  rotation  sth  sth-hash  value          (+ handle:)
```

Missing: `sth`, `sth-hash`, `cosig`, `fault`, `log-id`, `log-key-transition`. Adding all six:

```
thread '...no_label_is_a_prefix_of_another_anywhere_in_the_namespace' panicked at labels.rs:98:17:
label Ok("free2z/kt/v1/sth") is a prefix of Ok("free2z/kt/v1/sth-hash")
```

The omission is what makes it green. No exploit today — `free2z/kt/v1/sth` is a length-prefixed TBS `label` field (`KT.md:502`) while only `sth-hash` is a hash label (`KT.md:508`) — but that is exactly the argument the crate's own doc refuses to accept for its own labels: *"they are held to the prefix-free rule anyway, so that nobody has to work out which of the five is which before adding a sixth"* (`labels.rs:11-17`). Either complete `KT_LABELS` and file the `sth`/`sth-hash` collision against `KT.md` §5, or rename the test and state in `KT_LABELS`'s doc which labels are deliberately excluded and why. As it stands the name promises the property and the data withholds it.

---

# FINDING 5 — `f2z-assert keygen --out` writes the trust root's private key world-readable

`src/bin/f2z-assert.rs:111` is a bare `fs::write`, so the file lands at the process umask. Measured:

```
$ cargo run -q -p f2z-authority --bin f2z-assert -- keygen --out /tmp/f2zkey.hex
wrote a new issuing key to /tmp/f2zkey.hex
$ ls -l /tmp/f2zkey.hex
-rw-r--r--@ 1 skyl wheel 64 ... /tmp/f2zkey.hex
```

This is the one file in the system whose disclosure lets anyone vouch for any handle. The crate is right that `getrandom` in a trust root is a dependency too many — it should also be right that a 0644 issuing key is a mode too broad. `OpenOptions::new().write(true).create_new(true).mode(0o600)` behind `#[cfg(unix)]` (the bin is not `no_std` and is not built for wasm, so this costs nothing). `create_new` also stops a silent overwrite of an existing issuing key.

---

# What I checked and found sound

- **Charset (`WIRE.md` §14.1).** `Handle::parse` matches the rule exactly; `the_charset_is_exactly_a_to_z_zero_to_nine_and_underscore` (`types.rs:385`) iterates all 256 byte values, which is the right shape for this. Uppercase is refused, not folded (`:397`), and `DeserializeBytes` enforces the charset so a non-conforming handle cannot reach a rule as bytes-that-decoded (M18 proves the enforcement is load-bearing).
- **#550 item 4 (U+212A KELVIN) is structurally out of reach.** The crate does no normalization and deliberately has no `from_username` (`types.rs:160-174`), so it never performs the `lowercase(username)` mapping where KELVIN folds to `k`. U+212A is `E2 84 AA`, none of which is in the charset. Declining to write that function is the correct resolution and the reasoning at `types.rs:160-174` is the right one — the mapping is the authority's judgement, and recording it is what an assertion is.
- **Error codes** match `KT.md` §9.5 row for row, including the `ERR_INTERNAL` split for configuration faults; M13 proves the table is tested.
- **Redaction.** `assert_absent` (`tests/redaction.rs:75`) checks base16 both cases, base64url, **and the decimal list**, plus a 24-char hex-run catch-all — so the documented `tls_codec` decimal-dump trap is genuinely covered, and `a_submission_renders_neither_the_assertion_bytes_nor_the_signature` pins the hand-written `Debug` on the one bare `&[u8]`.
- **Workspace `Debug` scan.** It is scoped to `f2z-codec/src` on *this branch's* base — but #597 landed on `main` after it, moving the scan to `rs/crates/f2z-codec/tests/workspace_debug_scan.rs` and widening it to every crate under `rs/crates/`. I ran `main`'s version against this branch's tree: **`no_type_derives_debug_while_holding_raw_bytes ... ok`.** So after the rebase this crate is covered and clean, with no action needed.
- **Hygiene.** `#![no_std]` + `#![forbid(unsafe_code)]` at `lib.rs:136-137`, and `unsafe_code = "forbid"` in `[workspace.lints.rust]` covers the `bin` target too (a crate-level attribute in `lib.rs` would not have). `cargo tree -e normal` shows only `ed25519-dalek` 3.0.0, `f2z-codec` and `tls_codec` — nothing surprising, no `getrandom`, no `rand`.
- **Rule 9 (the log holds the cap, not the issuer)** is the right call and is tested twice, including handing the *same bytes* to a tighter log. Rule 15's strictness, rule 17's `(authority_id, nonce)` keying over the digest, and the derive-don't-publish retention all hold up under mutation.
- The **`Vouch::Unvouched` / `VouchingStatus`** reporting is real: M21 shows the "no authority" mode cannot be reached by accident, and `AuthoritySet::new(vec![])` → `EmptyAuthoritySet` is the right refusal.

---

# Known gaps — resolved by fiat, not re-filed

- The entry-1 authorization gap the PR body leads with is **already filed as #594** ("KT.md: nothing authorizes a handle's FIRST directory entry"). Confirmed against `KT.md` §4.4: rule 3 says `entry_version` is `previous + 1` (or 1), and none of the three `EntryAuthorization` kinds can be produced for a first entry. The PR's reading is correct; nothing to add.
- The crate implements no reset cooldown / `effective_at_ms` / `ERR_COOLDOWN` and no `no_reset` — correctly, those are §4.4 rule 7's and the log's, and the crate never claims them. #558 and #551 already cover the contradictions there.
- #553 / #600 (registration as a control) — this crate registers correctly by hand; #600 is the durable fix.

---

## Summary

Fix **Blocking 1** (reuse #597's fixture — it is a ten-line change) and resolve **Blocking 2** (the `same_key`/`key_change` door) before this reopens. Finding 3 is the #581 debt and should land in the same pass while the transcripts are still fresh; 4 and 5 are small and mechanical.

The verification core itself is strong — 21 of 26 mutations caught, and the two things that most often go wrong in this shape (state written before verification, tests that can only express friendly inputs) are both genuinely right here.

skyl:

skyl:

skyl:

Issue #589, opened 2026-08-24
test(f2z-relay-proto): the strict-verification test passes without strict verification, and §5.1 step 5 is untested on the verify side


skyl:
`rs/crates/f2z-relay-proto` (#583, merged as `40af3a5`) is sound — a rule-by-rule pass over §4.3, §5.1–5.5, §6.1–6.3, §7.3/7.6/7.7, §8.1–8.3, §9, §10, §11.1/11.3, §12.2/12.3 and §13.1 found **no divergence from `WIRE.md`**, and 29 of 33 landed mutations die. This issue is about the four that survive, two of which are security-relevant.

## 1. A test named for strict verification does not test strict verification — and the forgery it claims to exclude is real

`small_order_keys_are_rejected_by_strict_verification` uses an **all-zero signature**, which *both* `verify` and `verify_strict` reject. The test therefore passes whichever function is called.

Demonstrated by mutation: wiring a non-strict fallback into `key.rs:72` leaves **the entire suite green**.

The forgery the test names is not hypothetical. With the identity point as the public key and a signature built from it:

```
A   = [0x01, 0x00 × 31]        (identity / small-order point)
sig = A || [0u8; 32]

ed25519 verify        -> ACCEPTS 64/64 distinct messages
ed25519 verify_strict -> REJECTS all 64
```

That is a **universal forgery key**: one keypair whose signature validates against any message. Registered as a queue's `recv_key`, it makes `WIRE.md` §5.1 step 5 vacuous — the step whose entire job is "check that `signer_key` is the key registered for this address".

This sits directly beneath #583's own headline claim about strict verification. It is the canonical shape of this repo's dominant defect class: **a test whose name asserts the property it does not exercise**, passing for a reason unrelated to the thing it guards.

The replacement is the fixture above — assert `verify_strict` rejects it *and* that plain `verify` would have accepted it, so the test fails if the strictness is ever removed.

## 2. `CommandVerifier::verify`'s §5.1 step-5 half has zero coverage

`check_transcript_address` and `C::check` are called from **both** `SignedCommand::create` and `verify`. All three relevant tests assert only on `create` — and `create` **cannot express a hostile frame**, because it derives `signer_key` from the signing key it was handed.

Hand-built hostile frames confirm the code is correct today (`Err(Wire(NoAccess))` and `Err(Wire(Malformed))`), and that the mutations return `Ok(())` when applied. So the behaviour is right and **completely untested on the side a relay actually depends on**.

[...]

PR #595, opened 2026-08-24
feat(rs): add f2z-relay-store, the relay's queue storage

skyl:
`WIRE.md` §7 and §8, made durable. A `RelayStore` trait that speaks **addresses and bytes and never a wire frame**, plus `SqliteStore` and `MemoryStore`.

## ⚠️ Draft, and why: the merge-train hold

There is an active hard hold on #584/#588 — **#556 must land before any other PR moves `main`**, and E2EE PRs stay draft until it does. This PR is deliberately draft and is **not** to be marked ready or merged until that clears.

**CI did run on the draft**, and both required contexts are green: **`gate` ✅** and **`rs / gate` ✅** (run [32686928802](https://github.com/free2z/zuu/actions/runs/32686928802)). The local evidence at the bottom was gathered before that and is kept because it covers the two things CI cannot show on its own — the negative wasm build, and the gate scripts' verdicts quoted in full.

**Base tested:** `0b00710` (`docs(e2ee): close CLIENT-CONTRACT's seven interface defects (#592)`). Expect a rebase after #556; the evidence covers that base.

## The property the crate exists for

Under delete-on-ack the relay's copy is, for a window between `accepted` and `ACK`, **the only copy in the system** ([`ARCHITECTURE.md` §6.4](docs/e2ee/ARCHITECTURE.md#64-delete-on-ack-and-lost-acknowledgements)). A relay that answers `accepted` before the write is durable has promised custody a power cut breaks.

The brief asked for durability to be *a returned fact, not a convention*. It is:

- Every mutating operation returns **`Committed<T>`**, whose constructor is `pub(crate)`. The only code that mints one is the code that has just completed the transaction. A caller cannot write the line that reports `accepted` without first holding a value only a completed commit could have produced.
- `Committed` is `#[must_use]`, so dropping the receipt is a warning rather than a silent loss of track.
- The three pragmas the claim rests on are **read back out of the engine at open** and the store refuses to exist otherwise. A pragma that silently did not take is how a relay comes to publish `fsync-per-append` while running something weaker, and the failure is invisible until the power goes out.

## Group commit, treated as a requirement

At `synchronous = FULL` a commit is an fsync, and a commodity VPS does 50–200 a second. One transaction per append caps the relay there, which is not a working design. So `append_batch` is the **primitive** and `append` is the batch of one; the batching *window* is a scheduling decision left to the server crate.

Made checkable rather than asserted: `SqliteStore::commits()` counts durable commits (the number an operator should graph — it is the fsync rate), and `a_batch_of_a_hundred_appends_costs_one_commit` asserts 100 appends move it by 1 while 10 single appends move it by 10.

Batching does **not** weaken durability, and the distinction matters because §11.1 publishes it: deferring the fsync past the response is `batched`; amortizing one fsync across many responses that all wait for it is still `fsync-per-append`.

## Crash safety: real processes, really killed

`tests/crash_safety.rs`, behind the default-off `crash-injection` feature. Four injection points either side of the commit; each case spawns a **child process** that `abort()`s inside the library, the parent asserts the child was killed by **SIGABRT** (not that it exited — a child that returned a code did not crash, and accepting that would pass even with the injection point deleted), then reopens the file in a fresh process:

| Case | Asserts |
|---|---|
| after append commit | Nothing accepted-but-lost: all three rows present, `next_index = 3`, counters consistent — including the third, whose receipt never reached its caller |
| before append commit | No trace: no row **and** no consumed index (a counter that moved without its row is a permanent gap in the reader's index space) |
| after ack commit | Nothing acked-but-undeleted: every index ≤ the persisted watermark is gone |
| before ack commit | Nothing deleted, watermark unmoved, and §8.3's retry then works |
| reopened store | §8.2 pre-ack refusal and §7.3 bind-once are properties of the queue, not of the process that created it |

An in-process caught panic proves none of this — it unwinds a live connection and runs `Drop` on the open transaction, rolling back the very write in question. That reasoning is in `src/crash.rs`'s module docs so the next person does not "simplify" it.

## Schema

Composite primary key `(recv_addr, idx)`, `WITHOUT ROWID`, exactly as briefed — the row lives *in* the index, so one queue's messages are physically contiguous in index order: `READ` is a range scan from the watermark and `ACK`, being cumulative and therefore always a prefix, is a range delete over one contiguous run.

## Redaction, including the documented trap

Every value crossing the boundary is an `f2z-codec` newtype; no raw `Vec<u8>` or `[u8; 32]` appears in any `Debug`-deriving type. `tests/redaction.rs` checks **four** encodings — lowercase hex, uppercase hex, base64url, and an **unbracketed decimal run** — because `tls_codec`'s byte vectors print `[222, 222, …]`, a full dump containing no hex at all, so a hex-only check passes while everything leaks. The decimal probe is unbracketed for the same reason `f2z-codec` gives: a real dump usually starts mid-buffer after a length prefix. `the_decimal_check_would_actually_catch_a_leak` is the negative control proving the probe is not a no-op. `StoreError` is checked in both `Debug` and `Display`, because the error is what actually reaches a log line.

---

# Ambiguity calls, with reasoning

Every judgement call the spec did not settle. #586's two known `WIRE.md` defects were read and are **not** re-reported.

1. **The trait is closed to out-of-crate implementations, by construction.** `Committed`'s constructor is `pub(crate)`, so nobody outside can mint a durability proof — which is the point, since durability is the one property a backend cannot be trusted to assert about itself. The cost is real and stated in `durability.rs`: a third-party backend cannot implement `RelayStore` in v1. Opening it later is a default-off feature exposing a loudly-named `assert_durable`, and it should be added when there is a real second backend, not in advance of one. **Flagging for a reviewer's opinion** — this is the largest single design call in the PR.

2. **The store does not generate addresses.** §7.1 requires the relay to draw both from its own CSPRNG. Randomness stays out of the store: the caller draws, `QueueSpec` carries the pair, and `StoreError::AddressCollision` says draw again. A store with a CSPRNG is a store that can be asked for a *predictable* address by a caller that seeds it wrong, and §7.1's collision retry belongs next to the generator.

3. **Both addresses share one namespace.** A new `recv_addr` equal to an existing `send_addr` (or vice versa) is a collision, not an insert. Both are 32 uniformly random bytes and a lookup by either must be unambiguous. Unreachable at 32 bytes; a bug in the caller's generator otherwise.

4. **Authorization happens inside the transaction.** The caller verifies the signature (that is `f2z-relay-proto`'s job); the store compares the recovered signer against the registered key **atomically with the mutation**, because checking a statement earlier lets a `DELETE_QUEUE` land in between. Passing a `PublicKey` is not "reaching into the protocol" — it is bytes.

5. **Two lookup methods, not one, because §10 and §6.3 disagree on purpose.** The same absent row is `ERR_NO_ACCESS` via `queue_by_recv` (existence-oracle rule: absent and wrongly-keyed must be one code) and `ERR_UNAVAILABLE` via `queue_by_send` (§6.3's collapse). There is no `NotFound` variant to route wrongly — the collapse happens where existence is known.

6. **`bind_send` takes one key, not two.** §5.1 step 5 cannot apply to `BIND_SEND` (no key is registered yet); `f2z-relay-proto` resolved that by requiring `signer_key == send_key`, and this API makes them the same parameter so they cannot be passed differently.

7. **`BIND_SEND` on a contact address stays `ERR_NOT_PERMITTED` and is *not* collapsed into `ERR_UNAVAILABLE`.** §12.2 says "always, for everyone", and a contact address is **published in a directory entry** — refusing distinguishably leaks no queue state its finder did not already have. Called out because it is the one send-side code that does not collapse.

8. **`SendAuth` distinguishes `APPEND` from `CONTACT_APPEND`.** A signed `APPEND` to a contact address and an unsigned `CONTACT_APPEND` to an ordinary send address are both `ERR_UNAVAILABLE`: a writer must not learn which kind of queue an address is. The store does not verify PoW — `SendAuth::ContactStamp` asserts the caller already did (§13.1).

9. **`READ` always yields at least one message, even under an impossible `max_bytes`.** §6.2 does not say. Returning an empty page with `has_more = 1` forever means a reader that cannot progress cannot acknowledge, which under delete-on-ack is a queue that fills and then refuses its sender. `max_messages` is *not* given the same treatment — a client asking for zero gets zero.

10. **`READ` filters TTL-expired messages at read time**, not only at sweep time. A lazily-swept relay must not serve ciphertext past its own published retention.

11. **Idle-timer activity is buffered in memory and is deliberately not durable.** §7.7 resets the idle TTL on `APPEND`, `READ`, `ACK` **and `SUBSCRIBE`**. Fsyncing every `SUBSCRIBE` to move a 90-day deadline would spend the entire fsync budget on a timer whose resolution is months. A crash loses activity since the last durable write, ageing a deadline by at most that interval — and `expire()` flushes the buffer *before* it sweeps, so the exposure is bounded by the **sweep period**, not by uptime. `touch()` is the `SUBSCRIBE` hook and returns no `Committed`, which is the API saying so.

12. **Message-TTL expiry does not move the watermark.** An expired message was never acknowledged; advancing the watermark would acknowledge it on the reader's behalf. Consequence, from `f2z-relay-proto`'s own note: `pending` is unacknowledged *indices*, an upper bound on messages actually held, so `QueueRecord` carries `stored_messages` separately.

13. **Idle expiry runs before message expiry** in one `expire()` transaction: an idle-expired queue takes its messages with it, so sweeping it first keeps them out of the message-TTL accounting.

14. **A per-item refusal inside `append_batch` does not roll the batch back; a backend error does.** Every protocol refusal is decided strictly *before* that item's first write, so the transaction is still clean — one writer hitting its cap must not undo an unrelated queue's durable write. Anything that is not a `Protocol` error propagates and drops the transaction.

15. **The index space ends at `i64::MAX`, not `u64::MAX`.** SQLite's `INTEGER` is signed, and a range scan whose ordering flips past `i64::MAX` is a correctness bug, not a capacity one. Exhaustion refuses as `ERR_UNAVAILABLE` so §6.3's collapse survives the store's own limits — checked *before* the insert so it is a clean refusal rather than a half-written batch. Unreachable: at a million appends a second it is ~292,000 years. Unit-tested by forcing the counter.

16. **`delete_queue` returns counters (`Deleted`), and `expire` returns a per-queue report.** `DELETE_QUEUE`'s wire response is empty and stays empty; these exist so the caller can push `QUEUE_EVENT` (§6.4) to subscribed readers and keep §13.1 layer 4's accounting without re-querying.

17. **`stats()` reports `storage_bytes` from `page_count × page_size`, not from payload bytes.** Layer 4's high-water mark is about disk, and page overhead, free pages and the WAL are all real disk.

18. **A poisoned mutex does not wedge the store.** The lock guards a connection whose consistency is SQLite's job; refusing every later request would turn one fault into a permanent outage of `READ` and `ACK`, which §13.1 says must never be refused.

## Cross-crate change

**`QueueState::restore` added to `f2z-relay-proto`** (+ its own unit test). A relay's storage outlives its process, so §8's arithmetic has to be applied to *loaded* state. Without this constructor the store would re-derive that arithmetic against its own columns — and §8.2's anti-pre-ack rule implemented twice is §8.2's rule implemented once and violated once. It validates the two invariants a store cannot restore its way out of (a bound contact queue; `acked_through >= next_index`), both `ERR_INTERNAL` because both mean the relay's own storage contradicts the protocol.

## Notes for the reviewer

- **`rs/deny.toml` gains one licence: `Zlib`.** It arrives only via `rusqlite → hashlink → hashbrown → foldhash`, through hashbrown's default hasher rather than any choice of ours. Permissive, OSI-approved, FSF Free/Libre, no copyleft — the correct side of the boundary that section polices, which is permissive-crates-versus-AGPL-binaries and not licence-family bookkeeping. `wallet/deny.toml` already allows it for an unrelated graph; this entry is judged against ours and names the crate that needs it. `advisories ok, bans ok, licenses ok, sources ok` on a clean run, `ignore` still empty.
- **`rusqlite = "0.37"` with `bundled`, added to `[workspace.dependencies]`** with the singleton reasoning inline. `libsqlite3-sys` declares `links = "sqlite3"`; the lockfile resolves `rusqlite 0.37.0` / `libsqlite3-sys 0.35.0`, agreeing with all three wallet lockfiles.
- **Registered with `scripts/check-rust-toolchain.sh`** (`rs.yml` + both invocations in `rs/README.md`) — the one gate that does not discover manifests. Verified green.
- **Deliberately absent from `rs_wasm`**, with a comment saying so at the line. Not a claim: `cargo build --target wasm32-unknown-unknown -p f2z-relay-store` fails in `cc-rs` compiling `sqlite3.c`, which is the point — no client links this crate.
- **New CI step** running the crash suite with `--features crash-injection`, as a step of its own so a run that skipped it is visible rather than inferred. `scripts/check-workflow-gates.mjs` and `check-github-actions-pins.mjs` both pass; no job was added, so no gate wiring changed.
- Concurrent-crate note: if another PR touches `rs/Cargo.toml` / `rs/Cargo.lock`, keep **both** crates and regenerate the lockfile.

## Verification — local, pinned toolchain 1.97.1, base `0b00710`

CI is green on this exact head (`gate` and `rs / gate`), and the crash suite ran there too — the CI log shows `test result: ok. 5 passed; 0 failed; 1 ignored` for `tests/crash_safety.rs` on Linux, so the child processes are really being killed on the runner and not only on a developer's machine. All of the following were additionally run locally:

```
  --manifest rs/crates/f2z-relay-proto/Cargo.toml \
  --manifest rs/crates/f2z-relay-store/Cargo.toml
  → Every Rust toolchain pin agrees with wallet/rust-toolchain.toml: channel 1.97.1, MSRV 1.97.

scripts/check-rust-fmt.sh --root rs
  → All 4 Rust crate(s) under rs/ are formatted (rustfmt from 1.97.1).

scripts/check-rust-clippy.sh --root rs        # -D warnings
  → All 4 Rust crate(s) under rs/ are clippy-clean (clippy from 1.97.1).

scripts/check-rust-deny.sh --root rs --config rs/deny.toml
  → advisories ok, bans ok, licenses ok, sources ok
  → All 4 Rust crate(s) under rs/ satisfy rs/deny.toml.

node scripts/check-github-actions-pins.mjs   → passed (178 refs, 15 files)
node scripts/check-workflow-gates.mjs        → passed (2 gates, 56 jobs, 14 files)

cd rs && cargo test --locked --all-targets   → 12 suites, all ok, 0 failed
cd rs && cargo test --locked --doc           → ok (incl. this crate's lib.rs example)
cargo test --locked -p f2z-relay-store --features crash-injection --test crash_safety
  → 5 passed; 0 failed; 1 ignored   (each spawns a child killed by SIGABRT)

cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec -p f2z-relay-proto
  → Finished (the wasm line is unchanged)
```

Crate totals: 12 lib unit tests, 14 property tests (each run against **both** stores), 4 redaction tests, 5 crash-safety tests, 1 doc test.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -75,7 +75,8 @@ jobs:
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
             --manifest rs/crates/f2z-codec/Cargo.toml \
-            --manifest rs/crates/f2z-relay-proto/Cargo.toml
+            --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+            --manifest rs/crates/f2z-relay-store/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -254,6 +255,18 @@ jobs:
         working-directory: rs
         run: cargo test --locked --doc
 
+      # `crash-injection` is off by default because it compiles an `abort()`
+      # into f2z-relay-store's commit path, so the default run above skips
+      # `tests/crash_safety.rs` entirely. Those cases are the only evidence that
+      # delete-on-ack's two invariants survive an unplanned stop — nothing
+      # acked-but-undeleted, nothing accepted-but-lost — and they prove it by
+      # killing a real child process, which an in-process caught panic cannot.
+      # A step of their own, so a run that skipped them is visible rather than
+      # inferred.
+      - name: Run the crash-safety suite, which kills real processes
+        working-directory: rs
+        run: cargo test --locked -p f2z-relay-store --features crash-injection --test crash_safety
+
   rs_wasm:
     name: rs / wasm32
     needs: changes
@@ -293,6 +306,11 @@ jobs:
       # rather than `--workspace` so that a future server binary — which will be
       # AGPL, will pull tokio, and will never reach a browser — does not
       # silently join the wasm build and then fail it.
+      #
+      # `f2z-relay-store` is deliberately **absent** and must stay absent: it
+      # links SQLite through `rusqlite`'s bundled C amalgamation, which is a
+      # native build by construction. It is a relay-side crate; no client links
+      # it.
       - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -57,6 +57,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cc"
+version = "1.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ad534f4357a5264cce5019c989cf66a4f0dc4e0d1b1d15f8aacec0ff7360273"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -191,6 +201,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-relay-store"
+version = "0.1.0"
+dependencies = [
+ "f2z-codec",
+ "f2z-relay-proto",
+ "proptest",
+ "rusqlite",
+ "tempfile",
+]
+
+[[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -203,10 +236,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "64cd1e32ddd350061ae6edb1b082d7c54915b5c672c389143b9a63403a109f24"
 
 [[package]]
+name = "find-msvc-tools"
+version = "0.1.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d45db016d36b838f563236e9193d0ee6ce38f3f68b6c94e914b4929c96bbb890"
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
 name = "generic-array"
@@ -242,6 +287,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashbrown"
+version = "0.15.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
+dependencies = [
+ "foldhash",
+]
+
+[[package]]
+name = "hashlink"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7382cf6263419f2d8df38c55d7da83da5c18aef87fc7a7fc1fb1e344edfe14c1"
+dependencies = [
+ "hashbrown",
+]
+
+[[package]]
 name = "hybrid-array"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -255,6 +318,17 @@ name = "libc"
 version = "0.2.189"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3eaf3ede3fee6db1a4c2ee091bf8a8b4dccdc6d17f656fb07896ee72867612f2"
+
+[[package]]
+name = "libsqlite3-sys"
+version = "0.35.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "133c182a6a2c87864fe97778797e46c7e999672690dc9fa3ee8e241aa4a9c13f"
+dependencies = [
+ "cc",
+ "pkg-config",
+ "vcpkg",
+]
 
 [[package]]
 name = "linux-raw-sys"
@@ -276,6 +350,12 @@ name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
+
+[[package]]
+name = "pkg-config"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6b464fbc74e149a392436b17d523f769e057cb6877f6a5c4618bc6f11800548"
 
 [[package]]
 name = "ppv-lite86"
@@ -386,6 +466,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d6f6ff9a378485b298a5286656da665ba74413d36db0979633275d2e708145d4"
 
 [[package]]
+name = "rusqlite"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "165ca6e57b20e1351573e3729b958bc62f0e48025386970b6e4d29e7a7e71f3f"
+dependencies = [
+ "bitflags",
+ "fallible-iterator",
+ "fallible-streaming-iterator",
+ "hashlink",
+ "libsqlite3-sys",
+ "smallvec",
+]
+
+[[package]]
 name = "rustc_version"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -437,10 +531,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "shlex"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8fadd59c855ef2080decdef8ff161eb6661b86933c9d82e5ba29dc602a55aba"
+
+[[package]]
 name = "signature"
 version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28d567dcbaf0049cb8ac2608a76cd95ff9e4412e1899d389ee400918ca7537f5"
+
+[[package]]
+name = "smallvec"
+version = "1.15.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8ed6a63f02c8539c91a8685a86f4099661ba3da017932f6ebbea6de3f0fa7c90"
 
 [[package]]
 name = "subtle"
@@ -510,6 +616,12 @@ name = "unicode-ident"
 version = "1.0.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e6e4313cd5fcd3dad5cafa179702e2b244f760991f45397d14d4ebf38247da75"
+
+[[package]]
+name = "vcpkg"
+version = "0.2.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "accd4ea62f7bb7a82fe23066fb0957d48ef677f6eeb8215f372f52e48bb32426"
 
 [[package]]
 name = "version_check"
--- a/rs/Cargo.toml
+++ b/rs/Cargo.toml
@@ -53,3 +53,16 @@ subtle = { version = "2", default-features = false }
 # wasm32-unknown-unknown, so its `std` default costs this crate nothing: the
 # library itself is `no_std` and the wasm job builds `--lib`.
 proptest = "1"
+# **0.37 is a repository-wide singleton and this line is not free to move.**
+# `libsqlite3-sys` declares `links = "sqlite3"`, and Cargo refuses outright to
+# build a graph containing two versions of a `links` package — a hard error, not
+# a warning. Everything under `wallet/` reaches SQLite through
+# `wallet/plugins/tauri-plugin-zcash`'s `rusqlite = "0.37"`, so any crate
+# anywhere in this repository that wants SQLite must resolve to the same
+# version (`AGENTS.md`, "What NOT to do"). `bundled` compiles SQLite from the
+# vendored amalgamation rather than linking whatever the host happens to ship:
+# a relay's durability depends on `PRAGMA journal_mode`, `synchronous` and
+# `secure_delete` behaving exactly as `f2z-relay-store` verifies at open, and a
+# distro's build flags are not ours to assume. Native only — `rs_wasm` in
+# `.github/workflows/rs.yml` deliberately does not name the crate that uses it.
+rusqlite = { version = "0.37", features = ["bundled"] }
--- a/rs/README.md
+++ b/rs/README.md
@@ -76,7 +76,8 @@ decided is `wallet/rust-toolchain.toml`, and
 ```
 scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
                                 --manifest rs/crates/f2z-codec/Cargo.toml \
-                                --manifest rs/crates/f2z-relay-proto/Cargo.toml
+                                --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+                                --manifest rs/crates/f2z-relay-store/Cargo.toml
 ```
 
 fails the pull request the moment the two disagree. `.github/workflows/rs.yml`
@@ -110,6 +111,7 @@ prevent.
 |---|---|
 | [`crates/f2z-codec`](./crates/f2z-codec) | The canonical encoding layer of `WIRE.md`: `tls_codec` wrappers for every wire structure, the domain-separated signing transcript (§5), the redacting newtypes, and the padding-bucket validator (§9). `no_std` + `alloc`, `#![forbid(unsafe_code)]`, no I/O, no async runtime, and it builds for `wasm32-unknown-unknown`. |
 | [`crates/f2z-relay-proto`](./crates/f2z-relay-proto) | The protocol layer above it: signed-command construction and verification in §5.1's exact order, the timestamp window and fail-closed seen-set (§5.5), the queue lifecycle and ACK arithmetic (§7, §8), the capability document (§11), the `HELLO` proof of possession (§5.2), and §4.3's typed in-flight window. Same constraints — `no_std` + `alloc`, no I/O, no clock, no randomness, and it builds for `wasm32-unknown-unknown`. |
+| [`crates/f2z-relay-store`](./crates/f2z-relay-store) | The relay's queue storage: a `RelayStore` trait that speaks addresses and bytes and never a wire frame, plus a durable `SqliteStore` (WAL, `synchronous = FULL`, `secure_delete = ON`, group commit) and a volatile `MemoryStore`. **Native only** — it links SQLite and is not on the wasm line. `std`, `#![forbid(unsafe_code)]`, no async runtime. |
 
 **The relay and the clients link both.** That is the licence boundary in
 practice: everything here is MIT because a third-party relay, ZUULI and the WASM
@@ -135,14 +137,26 @@ reasons, and each is enforced by a test rather than by intent:
 cd rs
 cargo test
 cargo build --target wasm32-unknown-unknown --lib -p f2z-codec -p f2z-relay-proto
+
+# f2z-relay-store's crash-safety suite is behind a default-off feature, because
+# the feature compiles an abort() into the commit path. It kills real child
+# processes; `cargo test` alone does not run it.
+cargo test -p f2z-relay-store --features crash-injection --test crash_safety
 ```
+
+**Not every crate here reaches the browser, and the wasm line is the record of
+which do.** `f2z-relay-store` links SQLite through `rusqlite`'s bundled C
+amalgamation and is relay-side only, so it is deliberately absent from
+`rs.yml`'s `rs_wasm` job. Adding a crate to that line is a claim that a client
+links it.
 
 The gates are the repository's shared scripts, pointed here with `--root`:
 
 ```bash
 scripts/check-rust-toolchain.sh --toolchain-file rs/rust-toolchain.toml \
                                 --manifest rs/crates/f2z-codec/Cargo.toml \
-                                --manifest rs/crates/f2z-relay-proto/Cargo.toml
+                                --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+                                --manifest rs/crates/f2z-relay-store/Cargo.toml
 scripts/check-rust-fmt.sh    --root rs
 scripts/check-rust-clippy.sh --root rs
 scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
--- a/rs/crates/f2z-relay-proto/src/queue.rs
+++ b/rs/crates/f2z-relay-proto/src/queue.rs
@@ -97,6 +97,55 @@ impl QueueState {
         }
     }
 
+    /// Rehydrate a queue whose state was written down somewhere and read back.
+    ///
+    /// A relay's storage outlives its process, so the acknowledgement
+    /// arithmetic of §8 has to be applied to state that was loaded rather than
+    /// accumulated. Without this constructor a store would have to re-derive
+    /// that arithmetic against its own persisted columns, which is the one
+    /// duplication this crate exists to prevent: §8.2's anti-pre-ack rule
+    /// implemented twice is §8.2's rule implemented once and violated once.
+    ///
+    /// The arguments are exactly the fields [`QueueState`] persists, and the
+    /// checks below are the invariants a store cannot restore its way out of.
+    ///
+    /// # Errors
+    ///
+    /// [`ErrorCode::Internal`] — and it is deliberately that code rather than
+    /// anything a peer could provoke. Every rejection here means the relay's
+    /// own storage contradicts the protocol, which is a relay fault; §10 says
+    /// `ERR_INTERNAL` "carries no detail, ever", and there is nothing a client
+    /// did that a more specific code would describe.
+    ///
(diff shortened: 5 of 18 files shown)


skyl:

skyl:
## Adversarial review — #595 `f2z-relay-store`

Reviewed at head `a8c52df` on base `0b00710`, in a throwaway worktree, pinned 1.97.1. Every gate the PR body claims reproduced green (`fmt`, `clippy -D warnings`, `deny` → `advisories ok, bans ok, licenses ok, sources ok`, `check-workflow-gates` 2 gates/56 jobs, `check-github-actions-pins` 178 refs, toolchain 3-manifest registration, `cargo test --locked --all-targets` + `--doc`, and the 5-case crash suite). `rs/deny.toml`'s `akd`/`akd_core` `<0.13.0` bans are untouched; `.github/workflows/zuuli.yml` is untouched; `cargo tree -e normal` is 8 crates deep under `rusqlite` and holds nothing surprising; `#![forbid(unsafe_code)]` is real at `src/lib.rs:114`.

The engineering here is genuinely good and most of my mutations died. But **24 mutations, 11 caught, 13 not** — and the misses cluster on exactly the two properties the crate says it exists for. Two are live bugs, two are #589-shaped tests whose names assert what they do not exercise.

---

## Blocking

### B1 — `MemoryStore::read` returns a **non-contiguous page**, and a cumulative ACK then destroys an undelivered message

`src/memory.rs:232` uses `continue` where `src/sqlite.rs:1035` uses `break`. When a message overruns the remaining `max_bytes`, `MemoryStore` **skips it and keeps scanning**, so a later smaller message is admitted into the page ahead of a message the reader never saw.

Mixed padding buckets (§9 guarantees mixed sizes), `max_bytes = 1024`, payloads `[256, 8192, 256, 256]`:

```
SqliteStore: indices=[0]       has_more=true
MemoryStore: indices=[0, 2, 3] has_more=true      <-- index 1 never delivered
```

§8.1 makes `ACK` cumulative, so the reader acks the top of the page it got:

```
MemoryStore READ(max_bytes=1024) delivered indices [0, 2, 3]  (index 1 was never delivered)
after ACK(3): queue holds []
=> message index 1 (8192 bytes of ciphertext) is permanently lost, never delivered
```

Under delete-on-ack that is unrecoverable — `ARCHITECTURE.md` §6.4's "permanently lost message", produced by the store rather than by a lost ACK. `MemoryStore` is documented as not-for-strangers, but it is a `pub` export and it is the differential oracle every property test runs against; a rule it gets wrong is a rule the suite cannot see.

Not a MemoryStore-only test gap: **M20** injects the same `continue` into `SqliteStore:1035` and the whole suite stays green. READ-page contiguity is untested in *both* stores.

Fix is one token (`continue` → `break`) plus a `both()` property that appends unequal payload sizes under a constraining `max_bytes` and asserts the returned indices are `floor..floor+n` with no gap.

### B2 — `flush_activity` drains the touch buffer into a transaction that is then rolled back, so any protocol refusal silently discards §7.7 activity relay-wide

`SqliteStore::flush_activity` (`src/sqlite.rs:280-311`) `drain()`s the in-memory `activity` map into the open transaction. Every caller invokes it as the **first** statement after `connection.transaction()` — and then `create_queue`, `bind_send`, `ack` and `delete_queue` all have `return Err(...)` paths *after* that point (`:560`, `:612`, `:618`, `:705`, `:706`, `:709`, `:761`, `:762`). The early return drops the `Transaction`, SQLite rolls back — and the drained buffer is gone. It was consumed from memory and never written.

Since `read()` and `touch()` (the `SUBSCRIBE` hook) are the *only* activity sources that are not already durable writes, this deletes exactly the §7.7 resets that matter for a queue whose sender is quiet.

```
control (no failing call):  queues_expired = 0, queue still there: true
with one stranger BIND_SEND: stranger bind_send -> Unavailable
                             queues_expired after a touch at 80_000_000ms: 1
                             queue still there: false
```

A queue that was touched at t=80,000,000 ms — well inside its idle TTL — is retired by the sweep because one unrelated failing call in between ate the touch. `expire()` deletes an idle-expired queue's messages with it, so this destroys unacknowledged ciphertext.

Two ways to reach it, and the second does not need an attacker:

- **Remote, unauthenticated.** `BIND_SEND` to a random 32-byte address, self-signed with an attacker-generated key → `ERR_UNAVAILABLE` → buffer discarded. No PoW, no queue knowledge, no relationship. `ack()`/`delete_queue()` on a random `recv_addr` are the same primitive via `ERR_NO_ACCESS`. Sustained at any rate above the relay's READ/SUBSCRIBE rate, **no buffered touch is ever made durable** and §7.7's "resets on READ and SUBSCRIBE" stops being true.
- **Ordinary traffic.** A client retrying a `BIND_SEND` (`ERR_ALREADY_BOUND`) or a reader sending `ERR_ACK_TOO_HIGH` does it too. On a busy relay this fires constantly.

This is #590's shape (unauthenticated frame reaching a fail-closed structure) but destructive rather than merely growth-inducing.

Fix: do not drain until the commit succeeds — take a snapshot, and re-merge it into the map on any error path — or move `flush_activity` to immediately before `self.commit(tx)` in each method, after every refusal has been decided.

**M21** (`flush_activity` made a total no-op) is **NOT CAUGHT**: the entire durable side of §7.7 can be deleted and the suite is green. See B4.

### B3 — delete-on-ack has no test on `SqliteStore`, and the crash test that claims it structurally cannot see it

Three separate mutations of the deletion path pass the whole suite:

- **M3** `ack`'s prefix delete `idx <= ?2` → `idx < ?2` — the highest acknowledged message survives forever.
- **M4** the delete neutered entirely (`AND 0`) — *nothing* is ever deleted on ACK.
- **M10** `delete_queue`'s `DELETE FROM message` neutered — `DELETE_QUEUE` drops the queue row and leaves every message behind.

The reason none is observable is structural, and worth stating precisely: **`SqliteStore` maintains `stored_messages`/`stored_bytes` by arithmetic (`range_totals` + a separate `UPDATE`) that is independent of whether the rows went away**, and `read()` raises its floor past the watermark by construction. With M4 applied:

```
public observables after ACK(2):
  stats.messages         = 0
  stats.payload_bytes    = 0
  record.stored_messages = 0
  record.stored_bytes    = 0
  read() page len        = 0
GROUND TRUTH on disk: 3 message rows, 12288 bytes of ciphertext still present
```

Every public observable says the ciphertext is gone. It is not. Contrast `MemoryStore`, which re-derives its counters from the actual `Vec` (`totals()`), which is why the equivalent **M9** *is* caught.

**And the crash suite does not close it.** `tests/crash_safety.rs:272` is `a_crash_after_the_ack_commit_leaves_nothing_acked_but_undeleted`, asserting `"every index at or below the persisted watermark must be gone"` — but its `stored()` helper (`:111`) reads through `store.read(...)`, which filters below-watermark rows *by construction*. It cannot fail. With M4 applied:

```
cargo test --features crash-injection --test crash_safety
test result: ok. 5 passed; 0 failed; 1 ignored
```

Delete-on-ack fully disabled, all five crash cases green. This is precisely #589's shape — a test named for a property it does not exercise — in the one file whose whole purpose is proving that property.

Fix: `stored()` must query the rows, not the API. `SELECT idx FROM message WHERE recv_addr = ?` (or expose a `#[cfg(test)]`/`pub(crate)` row-count) and assert against that. A `both()` property asserting `stats().messages == 0` after a full ACK on `SqliteStore` (the doc test does this only for `MemoryStore`) kills M3 and M4; a post-`delete_queue` row count kills M10.

Given `THREAT-MODEL.md` §4.5 ("server-side deletion is auditable"), retained-after-ACK ciphertext is the single worst thing this crate can do, and today nothing at all would notice it.

### B4 — `the_idle_timer_retires_the_queue_and_activity_resets_it` is vacuous in **both** stores

`tests/properties.rs:398`. Its first assertion is commented `"the touch moved the deadline"` — but the queue is created at `created_at_ms = 1_000` with `idle_ttl = 60`, so its untouched deadline is **61,000 ms**, and the assertion sweeps at **60,500 ms**. It passes with zero touches. The second sweep at 110,001 ms expires the queue either way.

Proof, from the table: **M21** (`flush_activity` no-op — `SqliteStore` touch does nothing) and **M23** (`touch_record` no-op — `MemoryStore` touch does nothing) are both **NOT CAUGHT**. Every activity-recording path in the crate can be deleted and this test stays green.

Fix: sweep at a time strictly between the untouched deadline and the touched deadline — here, anything in `(61_000, 110_000]`.

---

## Mutation table

`cargo test --locked -p f2z-relay-store --all-targets` + `--doc`, on the pinned toolchain. Each mutation applied by line number and the changed line printed back before judging, so nothing below is a mutation that failed to land. Two first attempts were discarded and redone: M13 (borrow-of-moved-value) and M4 (a `SELECT` in `tx.execute` is caught by rusqlite's `ExecuteReturnedResults`, not by any property — the honest version is `AND 0`).

| # | file:line | mutation | result |
|---|---|---|---|
| M1 | sqlite.rs:1005 | READ TTL filter `expires_at_ms > ?3` → `>= 0` | caught by `a_message_expires_exactly_at_its_deadline` |
| M2 | memory.rs:222 | READ TTL filter removed | caught by `a_message_expires_exactly_at_its_deadline` |
| **M3** | **sqlite.rs:720** | **ACK prefix delete `idx <= ?2` → `idx < ?2`** | **NOT CAUGHT** |
| **M4** | **sqlite.rs:720** | **ACK delete neutered (`… AND 0`)** | **NOT CAUGHT** (incl. `--features crash-injection`) |
| M5 | sqlite.rs:636 | `queue_by_recv` absent → `unavailable()` not `no_access()` | caught by `an_absent_address_and_a_wrong_key_are_indistinguishable`, `the_idle_timer_retires_…` |
| M6 | sqlite.rs:510 | `authorize_send` contact×Signed → `Ok(())` | caught by `a_contact_queue_takes_unsigned_appends_and_never_binds`, `an_absent_address_and_a_wrong_key_are_indistinguishable` |
| M7 | sqlite.rs:940 | quota `admit(…)` result discarded | caught by `a_full_queue_refuses_and_never_evicts`, `the_byte_cap_refuses_the_append_that_would_cross_it` |
| M8 | sqlite.rs:951 | `next_index` i64-domain check removed | caught by `an_index_beyond_the_signed_domain_refuses_as_a_send_side_refusal` |
| M9 | memory.rs:264 | ACK `retain(index > n)` → `>= n` | caught by lib.rs doc test, `the_byte_cap_refuses_the_append_that_would_cross_it` |
| **M10** | **sqlite.rs:766** | **`delete_queue`'s `DELETE FROM message` neutered** | **NOT CAUGHT** |
| **M11** | **sqlite.rs:811** | **idle sweep `idle_expires_at_ms <= ?1` → `<`** | **NOT CAUGHT** |
| **M12** | **sqlite.rs:864** | **message-TTL sweep `expires_at_ms <= ?1` → `<`** | **NOT CAUGHT** |
| M13 | sqlite.rs:246 | commit counter `fetch_add(1)` → `fetch_add(2)` | caught by `a_batch_of_a_hundred_appends_costs_one_commit` |
| M14 | sqlite.rs:339 | `expect_pragma` always returns `Ok` | caught by `the_durability_pragmas_are_what_they_claim` |
| M15 | sqlite.rs:1034 | drop the `&& !messages.is_empty()` guard | caught by `a_read_always_yields_one_message_even_under_an_impossible_byte_budget` |
| M16 | sqlite.rs:1048 | `has_more: fetched > messages.len()` → `false` | caught by `a_read_always_yields_one_message_…` |
| **M17** | **sqlite.rs:555** | **collision guard drops `OR send_addr IN (?1, ?2)`** | **NOT CAUGHT** |
| **M18** | **sqlite.rs:998** | **`read_from(from_index)` → `from_index` (watermark floor)** | **NOT CAUGHT** |
| M19 | record.rs:111 | idle deadline `× 1_000` → `× 100` | caught by `deadlines_saturate_instead_of_wrapping`, `the_idle_deadline_is_the_same_function_in_both_stores` |
| **M20** | **sqlite.rs:1035** | **`max_bytes` overrun `break` → `continue` (= B1's bug)** | **NOT CAUGHT** |
| **M21** | **sqlite.rs:281** | **`flush_activity` made a no-op** | **NOT CAUGHT** |
| **M22** | **sqlite.rs:543** | **`recv_addr == send_addr` guard disabled** | **NOT CAUGHT** |
| **M23** | **memory.rs:118** | **`touch_record` made a no-op** | **NOT CAUGHT** |
| M24 | sqlite.rs:868 | expire stops decrementing `stored_messages` | caught by `a_message_expires_exactly_at_its_deadline` |

Notes on the remaining misses, all follow-up rather than blocking:

- **M11 / M12** — the two sweep boundaries. The *read-time* TTL boundary is tested exactly (M1/M2 die); the *sweep-time* boundaries are not. `expire()`'s test uses `110_001` against a `110_000` deadline. M12 additionally desynchronises the report from the delete: `stale` is selected with `<=` while the `DELETE` would run with `<`, so `stored_messages` is decremented for a row still present.
- **M17** — the cross-namespace collision guard. `a_colliding_address_is_reported_rather_than_overwriting_a_live_queue` covers *new send == existing recv*, which the surviving `recv_addr IN (?1,?2)` clause still catches. The untested direction is *new recv == existing send*, which M17 lets through: `create_queue(new recv_addr == existing send_addr) -> Ok("ACCEPTED")`, leaving one 32-byte address that resolves to two different queues depending on which lookup you use. Ambiguity call 3 is right; add the mirror assertion.
- **M18** — the watermark floor is dead weight *today* only because ACK deletes the prefix. It is the second layer that keeps B3's un-deleted rows invisible. Both layers are individually untested; the crate is correct because both happen to be right, not because either is checked.
- **M22** — `recv_addr == send_addr` would produce a queue where the APPEND authorization and the READ/ACK/DELETE authorization are the same address. Unreachable from a sane CSPRNG, which is the case the guard exists for; the guard just has no test.

---

## Open spec gaps this crate resolved by fiat (reporting, not re-filing)

- **#550 item 1 (§5.1 step 5 vs §6.3), still OPEN — the crate picked §6.3.** `WIRE.md:436` still says a wrong key is `ERR_NO_ACCESS`; §6.3's collapse list (`WIRE.md`, "absent queue, deleted queue, expired queue, queue at its message cap, queue at its byte cap, relay in global backpressure") still omits the wrong-key case. The crate routes **every** send-side refusal — including bound-to-a-different-key — to `ERR_UNAVAILABLE` (`queue_by_send`, `authorize_send`, verified by `an_absent_address_and_a_wrong_key_are_indistinguishable`), and every receive-side refusal to `ERR_NO_ACCESS`. It is **self-consistent** and it is the strictly-less-oracular of the two readings, so it is the right call. But it contradicts §5.1 step 5 literally, so #550 must be closed in *this* direction or this code changes. Worth adding to #550 that an implementation has now committed.
- **The one documented exception is sound but is a 4-way oracle.** `BIND_SEND` yields `ERR_NOT_PERMITTED` (contact), `ERR_ALREADY_BOUND` (bound), success (standard+unbound), `ERR_UNAVAILABLE` (absent). Ambiguity call 7 justifies the first; the other three are §7.3/§7.4's deliberate loudness. Fine at 32 random bytes, but it is the only command that distinguishes four queue states and that deserves a line in §12.2.
- **#552, `next_index` base — the crate chose 0.** `next_index` starts at `0` and the first append is index `0`; `QueueState::restore` enforces `acked_through < next_index`, so `acked_through = Some(0)` requires `next_index >= 1`. Consistent, and it should be written into §8.
- **#552, `READ` truncation when `max_bytes` (uint32) exceeds `max_frame_bytes` (1 MiB) — the crate applies no cap of its own.** `read_page` honours `max_bytes` up to 4 GiB. The only bounds are `max_messages: u16` (≤ 65,535) and `f2z-codec`'s payload ceiling (first rejected length: `16777216`, so < 16 MiB each) — a theoretical ~1 TiB single-page allocation, in practice bounded only by the caller-set `AppendQuota.max_bytes`. Reader-authenticated, so low severity, but the store's contract should say out loud that framing is entirely the caller's problem, since nothing here will stop it.
- **§11.3 step 5 (`max_message_ttl_seconds > 2592000` MUST refuse) is not enforced here** — and that is *defensible*, because the brief's premise is now stale: `f2z-codec::MAX_MESSAGE_TTL_SECONDS` **is** called, at `f2z-relay-proto/src/capabilities.rs:293` and `queue.rs:397` (`TtlPolicy::grant_message_ttl` clamps). The store takes the already-granted value. Still: `create_queue(message_ttl_seconds = u32::MAX)` → `Ok(4294967295)` (136 years), and the store *does* refuse other caller mistakes (`recv_addr == send_addr`). A `debug_assert` or a cheap `ValueOutOfRange` would make the ceiling structural for the price of one line.

---

## What I checked and found sound

- **Secret handling — clean.** No type in the crate derives `Debug` while holding raw bytes: the only `Vec<u8>`-bearing type, `RawQueue` (`sqlite.rs:420`), derives nothing. The two `Display` impls (`StoreError`, `Durability`) are both covered by `tests/redaction.rs` in Debug *and* Display, with a working negative control. Rendering the whole `MemoryStore` — which is not in the redaction suite — after storing `0xDE`-filled ciphertext gives `QueueAddress(<redacted>) … PublicKey(<redacted>) …` throughout, no hex and no decimal run. Consider adding `format!("{store:?}")` for both stores to `tests/redaction.rs`, since `SqliteStore`/`MemoryStore` both derive `Debug`. Note the #563 structural scanner is still rooted at `f2z-codec/src` (`f2z-codec/tests/redaction.rs:567`), so it does not reach this crate — #589's extension is what would make this property structural instead of enumerated.
- **The #590 analogue does not exist here.** The one unbounded-growth candidate is the `activity: Mutex<HashMap<QueueAddress, u64>>`, and both of its writers (`read`, `touch`) authorize via `authorize_recv` *before* `record_activity`. Correct ordering. The map is drained on every mutating call and is bounded by live-queue count. (B2 is the *opposite* failure in the same structure — over-eager draining, not unauthenticated insertion.)
- **The deletion oracle is preserved.** `delete_queue` removes both rows and leaves no tombstone; both addresses answer identically to never-existed afterwards (`the_idle_timer_retires_the_queue_and_activity_resets_it`, `an_absent_address_and_a_wrong_key_are_indistinguishable`). No timing-relevant branch differs between "existed and was deleted" and "never existed" — both are a single indexed lookup returning zero rows.
- **`QueueState::restore`** is the right call. Re-deriving §8.2's arithmetic in the store would be the duplication `f2z-relay-proto` exists to prevent, both invariants are checked, both are `ERR_INTERNAL`, and its unit test covers the boundary (`next_index=2, acked=2` rejected; `1, Some(0)` accepted).
- **Group commit, the pragmas, quota admission and the i64 domain** are all genuinely tested — M7, M8, M13, M14 all die, and M13 confirms the batch assertion is an exact equality rather than an inequality.
- **Ambiguity call 1** (crate-private `Committed` constructor closing the trait): agreed, ship it. Durability is the one property a backend cannot assert about itself, the cost is stated at `durability.rs:31-40`, and the escape hatch is correctly deferred until a second backend exists.
- **Ambiguity call 14** (per-item refusal does not roll the batch back) holds under inspection: every refusal in `append_one` is decided before its first write.
- Registration in `rs.yml` (#553) is present and correct, the `crash-injection` step is its own visible step, and the deliberate `rs_wasm` absence is commented at the line.

Minor: `tests/properties.rs` generates an untracked `properties.proptest-regressions` on any failure and `proptest-regressions` is not in `.gitignore`.

---

## Verdict

**Request changes on B1–B4**, all of which are contained fixes: one token in `memory.rs:232`; move or snapshot `flush_activity`; make `crash_safety.rs`'s `stored()` read rows instead of calling `read()`; move one timestamp in the idle-timer test. B1 and B2 are live bugs that lose or destroy user ciphertext. B3 and B4 are the #589 pattern and are what let B1's class of bug through in the first place.

Everything else — the durability design, the schema, the error collapse, the redaction, the group-commit proof, the `deny.toml` `Zlib` justification — I tried to break and could not.

skyl:

skyl:

skyl:

PR #597, opened 2026-08-24
test(f2z-relay-proto): make the strict-verification test actually test strict verification

### LINKED ISSUE #589: test(f2z-relay-proto): the strict-verification test passes without strict verification, and §5.1 step 5 is untested on the verify side

skyl:
Closes #589.

## 1. The old test was inert — proved before touching it

`small_order_keys_are_rejected_by_strict_verification` used an **all-zero
signature**, which *both* `verify` and `verify_strict` reject. It passed for a
reason unrelated to the thing it guarded.

#589 demonstrates this with a non-strict *fallback*. It is worse than that —
removing strictness **entirely** also leaves the suite green:

```
$ sed -i '' '73s/verify_strict/verify/' crates/f2z-relay-proto/src/key.rs
$ sed -n '70,75p' crates/f2z-relay-proto/src/key.rs
    pub fn verify(&self, message: &[u8], signature: &Signature) -> Result<()> {
        let signature = ed25519_dalek::Signature::from_bytes(signature.as_bytes());
        self.0
            .verify(message, &signature)
            .map_err(|_| ProtoError::Wire(ErrorCode::BadSignature))
    }

$ cargo test --locked --all-targets
test result: ok. 49 passed; 0 failed   # f2z-codec lib
test result: ok. 7  passed; 0 failed   # padding
test result: ok. 10 passed; 0 failed   # f2z-codec redaction
test result: ok. 7  passed; 0 failed   # reencode_equality
test result: ok. 42 passed; 0 failed   # wire_vectors
test key::tests::small_order_keys_are_rejected_by_strict_verification ... ok
test result: ok. 79 passed; 0 failed   # f2z-relay-proto lib
test result: ok. 6  passed; 0 failed   # properties
test result: ok. 8  passed; 0 failed   # f2z-relay-proto redaction
```

208/208 green with no strict verification anywhere in the crate.

`.verify_strict` appears three times in `key.rs`; two of them are doc comments
on lines 4 and 13. Every mutation below was applied by line number and the
landed line printed back before the run was judged.

## 2. The forgery is real, and behaves exactly as #589 records

Dependency versions in this tree: **ed25519-dalek 3.0.0 / curve25519-dalek
5.0.0**. `Verifier::verify` is still the uncofactored `R' == R` comparison and
`verify_strict` still screens small-order `A` and `R` first, so #589's numbers
reproduce unchanged. Surveying all eight small-order encodings as `A` against
all eight as `R`, with `S = 0`, over 64 distinct messages:

```
A[0]  identity              R=A[0]  ->  verify 64/64,  verify_strict 0/64
A[1]  order 2               R=A[0]  ->  verify 31/64,  verify_strict 0/64
A[1]                        R=A[1]  ->  verify 31/64,  verify_strict 0/64
A[2]  order 8               R=A[2]  ->  verify 20/64,  verify_strict 0/64
A[4]  order 4               R=A[0]  ->  verify 14/64,  verify_strict 0/64
A[6]  non-canonical id.     R=A[0]  ->  verify 64/64,  verify_strict 0/64
   … (full survey in the test's comment)
```

The identity is the only `A` that gives a **universal** forgery, and the
reason is in the algebra: non-strict verification recomputes
`R' = [S]B - [k]A`; with `S = 0` the first term vanishes and with `A` the
identity the second vanishes *whatever `k` is* — and `k` is the only place the
message enters. The order-2/4/8 points also slip past plain `verify`, but only
for the fraction of messages whose `k` lands in the right residue class.

That matters here: **the constant the old test used was `A[4]`, an order-4
point, which only forges for ~14/64 of messages** — so even given a non-zero
signature it would have passed or failed on the message it happened to be
handed. The replacement uses the identity, so "every message" is a property,
not a coin flip. The test asserts it across 64 of them rather than asserting
the algebra, and asks curve25519-dalek's own `is_weak()` rather than asserting
small-order-ness from a table.

Both halves are asserted — plain `verify` **accepts**, ours **rejects** — so
the test goes red if strictness is removed *and* red if the fixture ever stops
demonstrating the distinction.

## 3. `CommandVerifier::verify`'s §5.1 step-5 half now has coverage

`check_transcript_address` and `C::check` are called from both
`SignedCommand::create` and `verify`. Every existing test asserted only on
`create`, and `create` **structurally cannot express a hostile frame** — it
derives `signer_key` from the signing key it is handed and runs both checks
before returning, so every frame it can produce has already satisfied step 5.

A test helper now assembles the frame from the parts, with a *genuine*
signature over exactly the chosen `address` and `signer_key`, so verification
reaches step 5 instead of stopping at step 4's `ERR_BAD_SIGNATURE`. Each test
carries a positive control built the same way, so the refusal cannot be some
other step firing. Three new tests, green on the first run as #589 predicted:

- `verify_refuses_a_bind_send_whose_signer_is_not_the_key_it_binds` — the §7.4
  send-capability theft of #550 item 2 → `Err(Wire(NoAccess))`
- `verify_refuses_a_create_queue_whose_signer_is_not_the_key_it_registers` →
  `Err(Wire(NoAccess))`
- `verify_refuses_a_frame_whose_transcript_address_contradicts_its_command` —
  both directions of §6's table → `Err(Wire(Malformed))`

### The contrast that proves they are real

Disabling each check *in `verify` only* (lines 682/683; the `create` copies at
412/424 untouched) leaves the `create`-only tests green and turns the new ones
red:

```
===== M3: command.rs:683, C::check removed from verify() =====
test command::tests::bind_send_must_be_signed_by_the_key_it_binds ... ok
test command::tests::create_queue_must_be_signed_by_the_key_it_registers ... ok
test command::tests::create_queue_must_carry_the_zero_address ... ok
test command::tests::verify_refuses_a_bind_send_whose_signer_is_not_the_key_it_binds ... FAILED
test command::tests::verify_refuses_a_create_queue_whose_signer_is_not_the_key_it_registers ... FAILED
test command::tests::verify_refuses_a_frame_whose_transcript_address_contradicts_its_command ... ok

===== M4: command.rs:682, check_transcript_address removed from verify() =====
test command::tests::bind_send_must_be_signed_by_the_key_it_binds ... ok
test command::tests::create_queue_must_be_signed_by_the_key_it_registers ... ok
test command::tests::create_queue_must_carry_the_zero_address ... ok
test command::tests::verify_refuses_a_bind_send_whose_signer_is_not_the_key_it_binds ... ok
test command::tests::verify_refuses_a_create_queue_whose_signer_is_not_the_key_it_registers ... ok
test command::tests::verify_refuses_a_frame_whose_transcript_address_contradicts_its_command ... FAILED
```

Three `create`-only tests green under both mutations is the measurement of the
gap #589 describes.

## 4. Frame cap vs largest padding bucket

`a_frame_cap_below_the_relays_own_largest_bucket_is_refused` asserts both sides
of the boundary — `largest - 1` refused, `largest` accepted — reading `largest`
off the published defaults rather than a literal, so it cannot drift from the
bucket set it describes.

## 5. The `Debug` scan is now workspace-wide

`no_type_derives_debug_while_holding_raw_bytes` (#572) resolved its root from
`CARGO_MANIFEST_DIR`, so it only ever scanned `f2z-codec/src`. It moved to
**`rs/crates/f2z-codec/tests/workspace_debug_scan.rs`** and now walks every
crate under `rs/crates/`.

**Why there.** Three options; the file's module doc argues all three.
Duplicating it per crate is worst — *n* copies drift, and the copy that matters
is the one in the crate whose author did not know to add it, which is the crate
that has no copy. A crate existing only to host it costs a manifest, a
restated `rust-version` and licence registered with
`scripts/check-rust-toolchain.sh`, an `rs/deny.toml` entry and a wasm build
target for zero production code — and is still one file. So: one file, in the
crate every other crate already depends on and where the redaction doctrine is
written down, with a **filename that states its scope** and a coverage anchor
that asserts every crate under `rs/crates/` was actually *reached*. Adding a
crate can no longer silently narrow it the way adding `f2z-relay-proto` did.

One level of type-alias resolution (#581's fix) is preserved, and now works
across crate boundaries — aliases are collected over the whole workspace before
matching.

**Proof the old scan was blind, not merely narrow.** With an alias-typed
violating struct sitting in `f2z-relay-proto/src/replay.rs`, `origin/main`'s
scan passes:

```
=== #572 scan, with the M8 violation present in f2z-relay-proto ===
test no_type_derives_debug_while_holding_raw_bytes ... ok
test result: ok. 10 passed; 0 failed
```

The new one does not:

```
f2z-relay-proto/replay.rs: `M8Violation` derives Debug while holding `M8Blob`
  — a derived Debug renders that as a decimal byte dump. …
```

There is no current violation; this is preventative.

## Full mutation table

Every mutation applied by line number to the **committed** tree, landed line
printed back and checked before judging.

| # | Mutation | Landed line (verified) | Result |
|---|---|---|---|
| M0 | `key.rs:72` `verify_strict` → `verify`, **old test in place** | `.verify(message, &signature)` | 🟢 **208/208 green — the defect this PR fixes** |
| M0b | `key.rs:72` non-strict `or_else` fallback, **old test in place** | `.or_else(\|_\| self.0.verify(message, &signature))` | 🟢 **208/208 green** |
| M1 | `key.rs:72` non-strict `or_else` fallback | `.or_else(\|_\| self.0.verify(message, &signature))` | 🔴 `small_order_keys_are_rejected_by_strict_verification` |
| M2 | `key.rs:72` `verify_strict` → `verify` | `.verify(message, &signature)` | 🔴 same |
| M3 | `command.rs:683` `C::check` removed from `verify` | `// DISABLED: C::check(body.value(), Some(auth))?;` | 🔴 `verify_refuses_a_bind_send_…`, `verify_refuses_a_create_queue_…` (create-only tests stay 🟢) |
| M4 | `command.rs:682` `check_transcript_address` removed from `verify` | `// DISABLED: check_transcript_address(C::COMMAND, &auth.address)?;` | 🔴 `verify_refuses_a_frame_whose_transcript_address_…` (create-only tests stay 🟢) |
| M5 | `capabilities.rs:289` frame-cap check neutered | `if false && u64::from(capabilities.max_frame_bytes) < …` | 🔴 `a_frame_cap_below_the_relays_own_largest_bucket_is_refused` |
| M6 | `derive(Debug)` over `Vec<u8>` in **f2z-relay-proto** | appended to `src/replay.rs` | 🔴 `no_type_derives_debug_while_holding_raw_bytes` |
| M7 | scan narrowed back to `f2z-codec` (the #572 defect) | `for krate in ["f2z-codec".to_owned()] {` | 🔴 coverage anchor: `left: ["f2z-codec"] right: ["f2z-codec", "f2z-relay-proto"]` |
| M8 | `derive(Debug)` over a one-level **alias** of `Vec<u8>` in f2z-relay-proto | appended to `src/replay.rs` | 🔴 same; 🟢 under `origin/main`'s scan |

## Checks

| Check | Before | After |
|---|---|---|
| `cargo test --locked --all-targets` | 208 passed, 0 failed | **212 passed, 0 failed** |
| `cargo test --locked --doc` | 2 passed | 2 passed |

+4 net: three hostile-frame tests, one frame-cap test; the `Debug` scan moved
crate-internal test file → its own file (f2z-codec redaction 10 → 9, new file
+1); the strict-verification test was replaced in place.

- `scripts/check-rust-fmt.sh --root rs` — clean (rustfmt 1.97.1)
- `scripts/check-rust-clippy.sh --root rs` — clean (`--all-targets -D warnings`)
- `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` — advisories ok,
  bans ok, licenses ok, sources ok. **`rs/deny.toml` is not in this diff**; the
  `akd` / `akd_core` `<0.13.0` bans and the `akd_mysql` ban are untouched.
- `scripts/check-rust-toolchain.sh` — every pin agrees with 1.97.1 / MSRV 1.97
- `scripts/check-workflow-gates.mjs`, `scripts/check-public-repo-boundary.sh` — clean
- `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-codec -p f2z-relay-proto` — builds

`.github/workflows/zuuli.yml` is untouched (PR #556 is pending against it).

## Not addressed here

**#590** — `SeenSet::observe` inserting before signature verification — is
deliberately left alone. It is an owner decision about whether an unauthenticated
frame should be able to consume a seen-set slot, not a defect with an obvious
fix, and it is visible in this diff's neighbourhood only because the hostile
frames pass through step 3 on their way to step 5.

This is tests only. No production behaviour changed: every new assertion
describes what the code already does, which is why items 2–4 were green on
their first run and only the mutations move them.


skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I reproduced the decisive mutation myself.

```
baseline                                  EXIT=0   83 + 6 + 8 + 1 passed
verify_strict -> verify (whole of key.rs) EXIT=101
```

The old test stayed green through that. This one does not. That is the property #589 asked for, and it is the only thing that matters here.

## Two findings beyond the issue, both worth having

**The inertness was worse than I recorded.** #589 said a non-strict *fallback* left the suite green. Removing strictness **entirely** also leaves all 208 tests green, including the test named for it. So the old fixture wasn't weakly testing strictness — it was testing nothing about it at all.

**The old fixture's constant was an order-4 point, not the order-8 its comment claimed** — and it only forges for ~14/64 messages. So even handed a non-zero signature it would have passed or failed depending on which message it got. The survey of all eight small-order encodings as `A` against all eight as `R` is the useful part: **the identity is the only universal forgery**, with the others slipping past plain `verify` at rates from 7/64 to 31/64, because `k` is where the message enters. Putting the algebra in the test's comment is right — the next person to touch this needs to know why the identity specifically, not just that it works.

## The `Debug` scan placement is well argued

`rs/crates/f2z-codec/tests/workspace_debug_scan.rs`, walking every crate under `rs/crates/`, with a **coverage anchor** asserting every crate was actually reached.

That anchor is the part I'd have missed. Without it, the scan narrows silently the moment someone adds a crate — which is exactly how `f2z-relay-proto` ended up outside the guard that caught #563 in the first place. M7 confirms it: narrowing the scan back to one crate goes red on the anchor rather than passing with a smaller number.

The reasoning against the alternatives holds too. Per-crate duplication means *n* copies drift and the copy that matters is the one the new crate's author didn't know to write. A crate existing only to host the test costs a manifest, a restated `rust-version`/licence to register with `check-rust-toolchain.sh`, a `deny.toml` entry and a wasm target for zero production code — and is still one file.

## The mutation table does what a mutation table should

M3 and M4 are the pair I most wanted: disabling `C::check` or `check_transcript_address` **in `verify` only** turns the new tests red while all three `create`-only tests stay green. That contrast is the proof that `create` structurally cannot express a hostile frame — it derives `signer_key` from the signing key it is handed — and therefore that the pre-existing tests could never have covered the side a relay depends on.

M6 and M8 confirm the scan fires on a derived `Debug` over `Vec<u8>` **and** through a type alias, in a crate other than `f2z-codec`, while `origin/main`'s scan misses both. The alias case matters because `secret: Blob` where `type Blob = Vec<u8>` is exactly what slid past before #581 fixed it.

Every mutation applied by line number with the landed line printed back before judging — which is the discipline that stops a mutation that silently failed to apply from reading as a guard that works.

208 → 212 tests. `rs/deny.toml` is not in the diff, so the `akd` floor is untouched, and `zuuli.yml` is untouched.

**#590 correctly left alone** — `SeenSet::observe` inserting before signature verification is an owner decision, not a fix, and saying so in the body keeps it from reading as an oversight.

Note for whoever lands the next one: this PR was closed unmerged at 03:37:24 by an automated sweep and I reopened it. Several sibling PRs went the same way; their branches all survive.

PR #598, opened 2026-08-24
test(wallet): kill the legacy file-identity bypass mutant

### LINKED ISSUE #546: Legacy preview: replace comment-satisfiable identity guard with behavioral mutant
## Problem

The legacy import preview shipped with real hosted Windows compile/runtime coverage, but its supplemental `windows_identity_is_bound_to_the_stable_no_follow_win32_helper` test searches Rust source text with `contains(...)`. A comment can satisfy that assertion, so it is verification theatre rather than independent evidence.

## Scope

- remove the comment-satisfiable source-string assertion
- add a behavioral mutation case that replaces a consumed source database after snapshot copy with a distinct file having the same bytes, length, and modification time
- prove the preview still rejects the replacement based on actual file identity, so deleting/bypassing identity comparison makes the test red
- keep Windows hosted compile/runtime coverage authoritative for the Win32 helper
- document the exact killed mutant in PR evidence

## Acceptance

- no comment/string search is used as identity proof
- the real implementation passes
- a deliberate mutation that ignores file identity fails the new test
- full focused preview and affected native checks pass

Follow-up to #540 and issue #272.

skyl:
Closes #546

Replaces the comment/source-string-satisfiable Windows identity assertion with a behavioral file replacement test. The fixture swaps the consumed SQLite source after snapshot copy for a distinct file with the same bytes, length, SHA-256, and modification time; only file identity differs, and the preview must reject it.

## Required mutation proof

Baseline focused test: green.

Deliberate implementation mutant in `verify`: compare only `len`, `modified`, and `hash`, omitting `identity`. The focused test went red with exit 101 for the intended reason:

```text
left: Ok(LegacyWalletPreview { ... })
right: Err("Legacy data changed during preview.")
test result: FAILED. 0 passed; 1 failed; 142 filtered out
```

The mutant was restored. This expectation is not read from production source or a comment: the test independently proves the replacement has equal bytes/length/mtime/hash and a distinct filesystem identity before installing it.

## Exact-current-trunk restored proof

Head is based on exact current `origin/main` `918227fbfac0c4cde9fb4f4d39b1e5d33a477a61` (through #579/#556).

- focused identity-replacement test: green
- full plugin `--lib`: 142/142 green
- plugin all-target/all-feature Clippy at `-D warnings`: green
- rustfmt: green
- actions policy: 70 source + 7 verdict + 120 current-workflow mutants; live green
- workflow gates: 20 mutants; live green
- `git diff --check`: green

Do not merge until every check for this exact head is terminal green.


skyl:
Exact-base revalidation after #597 advanced `main`:

- candidate head: `b90761f3d1e1a0a1e037df75d3beb7135c894850`
- exact trunk parent: `2d114c8c8e4d3fd8fa257f5a1f43cac5008f1d8d`

I reapplied the decisive production mutant to this integrated tree: `verify` ignored `FileIdentity` and accepted equality of only length, modification time, and hash. The behavioral replacement test failed with exit 101: it returned `Ok(LegacyWalletPreview { ... })` instead of `Err("Legacy data changed during preview.")` (0 passed, 1 failed). The fixture independently asserted equal length/mtime/hash and distinct identity before exercising the real preview path.

After restoring the guard:

- full plugin tests: 142/142
- all-target/all-feature Clippy with `-D warnings`: green
- rustfmt: green
- action policy: green (178 immutable refs / 15 files)
- workflow gate policy: green (2 protected contexts / 56 jobs / 14 workflows)
- public-boundary and diff checks: green

The earlier `a4fbf84` CI run was explicitly invalidated when trunk moved. Merge is held until every check on `b90761f` is terminal green.

skyl:
Second exact-base revalidation after externally merged #600 advanced `main`:

- candidate head: `31d6bad2443456c3618cf2b181708dde1b288bc7`
- exact trunk parent: `2dffff0ecc43a223b1f6a854e00e6a6ebfdc267c`

The same identity-bypass mutant was reapplied after integration and again failed with exit 101: the real preview path returned `Ok(...)` for the byte-identical, same-length, same-mtime, same-hash replacement instead of rejecting its distinct file identity. Restored full plugin tests remain 142/142; Clippy, rustfmt, action policy, workflow gate policy, public-boundary, and diff checks are green.

The `b90761f` run is invalidated. Merge is held for terminal-green checks on `31d6bad` only.

skyl:
## Adversarial review — PR #598 (head `31d6bad`, base `2dffff0`)

Reviewed the net diff against current trunk: exactly one file, `wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs` (+51/−12). `git diff b90761f 31d6bad -- wallet/plugins/tauri-plugin-zcash/` is empty, so the second merge commit changed nothing under review.

**Verdict: sound. Approve.** The mutation claim in the PR body reproduces exactly, the replacement is not text-satisfiable by any shape I could construct, and the old guard is gone rather than left standing beside it.

---

### 1. The decisive mutant — does the guard fail when the bypass returns? **Yes.**

Applied the bypass to `verify` (`legacy_import_preview.rs:442`), confirming the edit landed on the real function and not a doc comment:

```diff
 fn verify(path: &Path, expected: &Observation) -> Result<(), &'static str> {
     let actual = observe(path)?;
-    if &actual == expected {
+    if actual.len == expected.len
+        && actual.modified == expected.modified
+        && actual.hash == expected.hash
+    {
```

`cargo +1.97.1 test --locked --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml --lib`:

```text
thread '...same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected' panicked at src/legacy_import_preview.rs:1007:9:
assertion `left == right` failed
  left: Ok(LegacyWalletPreview { wallet_id: "[REDACTED]", ..., account_count: 1, ... })
 right: Err("Legacy data changed during preview.")

failures:
    legacy_import_preview::tests::same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected
test result: FAILED. 141 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

**Exactly one test in the 142-test suite kills it, and it is the one this PR adds.**

### 2. The same mutant against pre-PR trunk — **fully green, old guard included**

This is the claim that matters for #546. Checked out `2dffff0`'s copy of the file, applied the identical bypass:

```text
test legacy_import_preview::tests::windows_identity_is_bound_to_the_stable_no_follow_win32_helper ... ok
test result: ok. 142 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

So on `main` today, a `verify` that ignores `FileIdentity` entirely ships green — and the test whose name claims to bind identity says `ok`. #546's premise is confirmed empirically, not just by reading. The suite count is unchanged (142 → 142): a guard replaced, not a guard bolted on.

Worth recording that the deleted assertion was weaker than #546 even described: it searched `app_data_migration.rs` for the bare tokens `FILE_FLAG_OPEN_REPARSE_POINT` and `GetFileInformationByHandle`, both of which also occur at `app_data_migration.rs:465,480` inside an unrelated function — so deleting `windows_file_identity` outright would still have satisfied two of its three assertions.

### 3. Defeat attempts against the new guard — all fail

- **Comment / string literal / `#[allow(dead_code)] const`** (the #556 shape). Injected, alongside the bypass from §1, both old pinned strings as a comment *and* as dead consts:
  ```rust
  // crate::app_data_migration::windows_file_identity(path, metadata.file_type().is_dir())
  // FILE_FLAG_OPEN_REPARSE_POINT GetFileInformationByHandle
  #[allow(dead_code)]
  const DECOY_IDENTITY_PIN: &str = "crate::app_data_migration::windows_file_identity(path, metadata.file_type().is_dir())";
  #[allow(dead_code)]
  const DECOY_WIN32_PIN: &str = "FILE_FLAG_OPEN_REPARSE_POINT GetFileInformationByHandle";
  ```
  Result: still `FAILED. 141 passed; 1 failed` — the same single test. Text cannot buy this guard.
- **Renamed symbol / superstring**: structurally impossible. `grep -rn "include_str!\|source.contains(" wallet/plugins/tauri-plugin-zcash/src/` now returns nothing; the only hit repo-wide for `_source` is a test *name*. No source text is read anywhere in the plugin.
- **Old guard surviving alongside the new one**: no. `grep -rn "windows_identity_is_bound"` over the tree returns zero hits.

### 4. Does the test pass for the *named* reason? Two more mutants say yes.

- `verify` comparing **only** `identity` (dropping len/mtime/hash) → the new test is **green**. Identity is *sufficient*.
- Combined with §1 (identity dropped → red), identity is *necessary*. The test isolates precisely the identity comparison and nothing else.
- Deleting the post-copy revalidation loop (`inspect_wallet_with_hook`, line 325-327) → **red**, along with `in_place_source_content_change_after_copy_is_rejected` and `source_content_or_identity_change_after_copy_is_rejected` (139 passed / 3 failed). Confirms the new test targets the post-`after_copy()` `verify`, not `Snapshot::copy`'s pre-hook one.

### 5. Built-in negative controls — the fixture cannot degrade into "rejected for the wrong reason"

The four in-fixture preconditions (`len`, `modified`, `hash` equal; `identity` distinct) are asserted **twice** — once on the staged replacement and again on the *installed* file inside the hook (lines 990-993 and 999-1002). If `set_times` failed to restore the mtime exactly, or the copy differed by a byte, the test panics on the precondition rather than passing on the wrong `Err`. And if `file_identity` were stubbed to a constant, `assert_ne!(installed.identity, original.identity)` fires. There is no mutation shape where the circularity of using `observe()` for the preconditions turns a should-be-red into green — the preconditions only ever strengthen the claim.

### 6. Gate wiring — the test runs in a required job

- `Rust / shared plugin` (`.github/workflows/zuuli.yml:424`) runs `cargo test --locked --all-targets` on the pinned Linux image.
- It is in `gate.needs` (`zuuli.yml:753`) and the gate *reads* the result: `check-github-actions-pins.mjs --verify-gate-results` over `toJSON(needs)` (`zuuli.yml:768-774`) — not a bare `needs:` of the #512/#515-#518 kind.
- `gate` is one of the two protected contexts: `gh api .../branches/main/protection` → `contexts: ["gate", "rs / gate"]`.
- The `changes` filter matches `wallet/plugins/*`, so `zuuli=true` for this diff — and the job log proves the test actually executed rather than being filtered out:
  ```
  Rust / shared plugin  test legacy_import_preview::tests::same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected ... ok
  Rust / shared plugin  test result: ok. 142 passed; 0 failed; ...
  ```

### 7. Windows coverage — genuinely upgraded, with one soft spot (not a blocker, not a regression)

The new test is **not** `#[cfg(unix)]`-gated, so it executes the real Win32 path. Confirmed from `custody / windows-latest` (job 97320604435) on this head:

```
custody / windows-latest  test legacy_import_preview::tests::same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected ... ok
custody / windows-latest  test result: ok. 135 passed; 0 failed; ...
```

So `windows_file_identity` → `CreateFileW(FILE_FLAG_OPEN_REPARSE_POINT)` + `GetFileInformationByHandle` is now *exercised*, where before it was only *spelled*. That is a real upgrade over what was deleted.

**Soft spot for a follow-up, not for this PR:** `custody / windows-latest` lives in `.github/workflows/zuuallet.yml`, which has **no gate job at all** and is therefore not a required context. On the gated path the only Windows job is `Rust / native lints (windows)`, which runs `scripts/check-rust-clippy.sh` — it type-checks the test but never runs it. So gated Windows *execution* of the Win32 identity helper is zero. This is not introduced here (the deleted test was a Linux string search and executed no Win32 code either), but it is the remaining gap between "Windows coverage exists" and "Windows coverage is required". Worth an issue: either add a gate to `zuuallet.yml` or move plugin test execution for windows-latest under `zuuli.yml`'s gate. `check-workflow-gates.mjs` only enforces job coverage for files that *have* a gate, so nothing flags this today.

### 8. Repo gates, run locally at this head

```
node scripts/check-github-actions-pins.mjs            → passed (178 external refs / 15 files)
node scripts/check-github-actions-pins.mjs --self-test → 70 source-policy, 7 gate-verdict, 120 current-workflow mutation cases passed
node scripts/check-workflow-gates.mjs                 → 2 gates, 56 jobs, 14 workflow files; both protected contexts single-producer
node scripts/check-workflow-gates.mjs --self-test     → 20 cases passed
scripts/check-rust-toolchain.sh                       → every pin agrees with wallet/rust-toolchain.toml (1.97.1 / MSRV 1.97)
cargo +1.97.1 fmt -- --check                          → clean
git diff --check 2dffff0..31d6bad                     → clean
cargo +1.97.1 test --locked --lib (restored tree)     → 142 passed; 0 failed
```

- **`zuuli.yml` `changes` policy step:** untouched. `check-github-actions-pins.mjs`'s exact ordered-command assertion is unaffected and passes.
- **#477 shell hygiene:** **0** bare `[[ ]]`/`(( ))` statements added — this PR adds no shell at all. (Its guard executes on `ubuntu-latest` in a container for the gated lane and `windows-latest` for the ungated one; no macOS bash 3.2 exposure.)
- **#600 registration census:** no manifest added, so nothing to register; `check-rust-toolchain.sh` is green regardless.

---

### Nits (non-blocking, no change requested)

- `same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected` and the older `source_content_or_identity_change_after_copy_is_rejected` now overlap in name; the older one only ever varied *content* (`valid_ufvk(1)`), never identity, and stays green under the §1 mutant. Renaming it to `..._content_change_after_copy_is_rejected` would stop it from advertising coverage it does not have — but that is a separate cleanup and #546 does not ask for it.

Environment: reviewed in an isolated worktree at `31d6bad`; the worktree is removed and `/Users/skyl/Code/github/f2z/zuu` is clean on `main` at `2dffff0`.

skyl:
Reviewed adversarially on the owner's behalf. **Approving** — merging once `gate` reports; `rs / gate` is green and there are no genuine failures on this head.

## The result that matters

The same bypass — replacing `if &actual == expected` with a len/modified comparison — applied to **`main`'s copy of the file** leaves the suite **142/142 green**, with `windows_identity_is_bound_to_the_stable_no_follow_win32_helper ... ok`. Applied to this PR's copy, exactly **one test of 142 fails**, and it is the one this PR adds:

```
left:  Ok(LegacyWalletPreview { wallet_id: "[REDACTED]", …, account_count: 1, … })
right: Err("Legacy data changed during preview.")
test result: FAILED. 141 passed; 1 failed
```

That is #546 proven rather than asserted: the old guard was theatre **against this exact defect**, and the replacement is not. The diff was shown before running, so the edit provably landed on the function rather than a doc comment — which matters, since I got that wrong three separate times tonight and each time it looked exactly like a working guard.

**What I could verify myself, and what I could not.** I confirmed structurally that the old guard is gone (`windows_identity_is_bound`: 1 occurrence on `main`, 0 here — replaced, not bolted on) and that the identity comparison is intact. I could **not** re-run the mutation locally: a fresh worktree fails at `failed to get zcash_address as a dependency` because it has no `z/` submodule checkout. So the 142-vs-141 figures above are the reviewer's, with their method shown, not independently reproduced by me.

## The defeat attempts are the other half

- **The #556 shape** — a comment plus an `#[allow(dead_code)] const` carrying both old pinned strings, injected alongside the bypass — still `141 passed; 1 failed`. That is the exact trick that defeats a source-text checker, and it does not work here.
- **Renamed symbol / superstring**: structurally impossible. No `include_str!` and no `.contains(` remain anywhere in the plugin, so there is no text to satisfy.
- **Old guard surviving alongside**: `grep -rn "windows_identity_is_bound"` returns zero.

Two further mutants pin down *why* it works: comparing **only** identity is still green (identity is sufficient), and dropping the post-copy `verify` loop goes red plus two siblings (139/3). Necessary and sufficient, demonstrated both ways.

Test count unchanged at 142 — this replaced a test rather than adding one on top, which is the right shape for a fix to a guard that could not fail.

## Gate binding, checked

`Rust / shared plugin` (`zuuli.yml:424`) runs the suite, is in `gate.needs` (`:753`), and the gate **reads** results via `--verify-gate-results` over `toJSON(needs)` (`:768-774`). That is consumption, not the awaits-without-reading pattern from #512 and #515–#518. The job log confirms the test line actually executed.

0 bare `[[ ]]`/`(( ))` added. `zuuli.yml`'s `changes` step untouched. No new manifest, so nothing to register under #600's census.

## One follow-up, worth an issue rather than a hold

**Gated Windows execution of the Win32 helper is zero.** The new test is not `cfg(unix)`-gated, so it genuinely exercises `windows_file_identity` — and it passes in `custody / windows-latest`. But that job lives in `zuuallet.yml`, which has **no gate job at all** and is not a required context. The only gated Windows job is `Rust / native lints (windows)`, which clippy-checks and never runs tests.

So the Win32 path is tested, but nothing required can fail on it. Pre-existing — the deleted test executed no Win32 code either — and strictly improved by this PR, so not a blocker. The fix is either a gate in `zuuallet.yml` or moving the windows-latest plugin tests under `zuuli.yml`'s gate. Note `check-workflow-gates.mjs` structurally cannot flag this: it enforces job coverage only for files that already *have* a gate, so a workflow with no gate is invisible to it. That gap is worth closing in the same change.

**A nit in the same file:** `source_content_or_identity_change_after_copy_is_rejected` only ever varies content and stays green under the bypass — so its name advertises identity coverage it does not have. Worth renaming or extending while you are here.

PR #599, opened 2026-08-24
ci: recheck workflow policy in protected gate

### LINKED ISSUE #568: ci: the ZUULI gate does not re-run `check-workflow-gates.mjs`, unlike the rs gate

skyl:
Closes #568

## Summary

- re-run `check-workflow-gates.mjs --self-test` and its live verdict inside the protected ZUULI gate
- require the exact ordered four-command gate policy step while preserving the distinct exact six-command `changes` policy step
- mutation-test deletion and reordering of both new commands, soft-fail and dynamically-dead escapes for both, and an extra decorative command

## Mutation evidence

Each implementation guard was deliberately weakened against exact trunk and the self-test went red for the intended escape:

- `policyCommands.length !== 4` changed to `< 4`: exit 1 because the extra decorative command escaped
- removed the index 2 workflow-gates self-test comparison: exit 1 because `--self-test || true` escaped
- removed the index 3 live workflow-gates comparison: exit 1 because `false &&` around the live verdict escaped

All guards were restored before the final run.

## Verification

- `node scripts/check-github-actions-pins.mjs --self-test` (74 source-policy, 7 gate-verdict, 129 current-workflow mutation cases)
- `node scripts/check-workflow-gates.mjs --self-test` (20 cases)
- `node scripts/check-workflow-gates.mjs`
- `git diff --check`


skyl:

skyl:
Exact-trunk refresh for the next train candidate:

- merged `origin/main` at `c5ce8aa8d8dacd4157fc43f86ff69d860e42b9c6` normally into the published branch (no rebase or force-push)
- exact tested/pushed head: `8a78cbfeda35815245e4f928000e279c7e0d37ff`
- PR diff remains the two intended files; the six-command `changes` contract and four-command `gate` contract remain distinct

Repeated literal implementation mutants on this exact integration:

1. `policyCommands.length !== 4` → `< 4`: exit 1 because the decorative fifth command escaped
2. removed the index 2 workflow-policy self-test comparison: exit 1 because `--self-test || true` escaped
3. removed the index 3 live workflow-policy comparison: exit 1 because `false &&` escaped

Restored exact head is green locally:

- action-pins self-test: 74 source-policy / 7 gate-verdict / 129 current-workflow cases
- action-pins live
- workflow-gates self-test: 20 cases
- workflow-gates live
- `git diff --check`

Leaving this PR draft for the train release.

skyl:
Reviewed adversarially on the owner's behalf. **Approving** — merging once `gate` reports on this head.

This closes #568: `zuuli.yml`'s gate now re-runs the workflow-gates policy the way `rs.yml`'s gate already did. Two lines of workflow, and an assertion that makes them un-removable.

## The assertion is exact, and I proved it on the right lines

`check-workflow-gates.mjs` appears **twice** in `zuuli.yml` — the standalone step at `:49-50` and now the gate's recheck at `:767-768`. My first two mutation attempts hit the standalone occurrence, because `perl -0p s///` without `/g` takes the first match; both reported red for the wrong reason. Redone against exact line numbers, with the resulting step printed back before judging:

```
MUT A  delete only the gate's two commands (767-768)
       gate step now: pins --self-test, pins
  -> EXIT=1  .github/workflows/zuuli.yml:762: gate policy recheck must be exact,
             unconditional, and non-soft-failing

MUT B  swap only the gate's two commands
       gate step now: … check-workflow-gates.mjs, check-workflow-gates.mjs --self-test
  -> EXIT=1  same message
```

Deletion and reordering both fail, and the message names the file, the line and the rule. That is the property worth having: the recheck cannot be removed, reordered, made conditional, or allowed to soft-fail without the policy catching it.

Recording my own mistake because it is the fourth time tonight a mutation looked like it landed and hadn't — twice hitting a doc comment instead of a constant, once probing with an untracked file that `git ls-files` correctly ignores, and now once hitting the wrong occurrence of a repeated string. Every one of them reads as "the guard works."

## The self-test grew with the guard, which is the part that usually doesn't happen

70 → **74** source-policy cases and 120 → **129** current-workflow cases, and the new ones are named for what they do:

```
self-test: deleted gate policy recheck fails closed
self-test: dynamically dead gate policy recheck fails closed
self-test: soft-failing gate policy recheck fails closed
self-test: deleted gate workflow-policy self-test fails closed
self-test: deleted gate workflow-policy verdict fails closed
self-test: reordered gate workflow-policy commands fail closed
```

"Dynamically dead" and "soft-failing" are the two I would have forgotten. A recheck that is present but unreachable, or present but `continue-on-error`, is exactly the shape of #512/#515–#518 — a control that exists in the file and cannot fail — and both are covered.

## Why this asymmetry was worth closing

`rs.yml`'s gate already re-ran the policy; `zuuli.yml`'s did not. That mattered because the second run defeats a specific attack: a job-level `continue-on-error` on `changes` would turn its deliberately-failing policy verdict into the `success` conclusion the gate consumes. The `rs` gate was immune and the `zuuli` gate — the one guarding the wallet — was not. Now both are.

## State

`rs / gate` green. `gate` has failed only on `required job rust_native_clippy must be success, got cancelled` — a concurrency supersede, not a check. The head has since been rebased; content against `main` is identical to what I verified (+159/−2, gate step at `:765-768`), so the verification carries over.

Locally on the PR head: pins policy bare `EXIT=0` (178 external references, 15 files), pins `--self-test` `EXIT=0`.

PR #601, opened 2026-08-24
test(zuuli): mutation-prove git fixture teardown guards

skyl:
Closes #596

## What changed

- Exercise the real fixture `git(...)` child and ask Git itself for `gc.auto` and `maintenance.auto` under an isolated empty system/global configuration. The isolated base environment explicitly resets `GIT_CONFIG_COUNT=0`; the production spread must then supply literal `0` and `false`. Assertions do not read `GIT_FIXTURE_ENV`.
- Make fixture removal retries explicit and narrowly injectable. Production permits only 10 `ENOTEMPTY` retries with 50 ms linear backoff.
- Prove deterministic success after ten injected `ENOTEMPTY` races, immediate unchanged rejection of injected `EACCES`, and unchanged rejection of persistent `ENOTEMPTY` on attempt 11 after exactly ten pauses. Tests inspect behavior, not options, and do not race timers.

## Exact integrated-base mutation proof

Normal-merged exact `origin/main` base `06193aee60b9735d697114d1652a382e7dc70671` (#611). No force-push. Exact tested and pushed head: `ef1812f89e3771503d957f5f6446a5ca2682d632`.

Each guard was mutated independently on this integrated tree, observed red, and restored before green verification:

1. Remove production `...GIT_FIXTURE_ENV`: exit `1`; actual `missing` differs from required `0`.
2. Zero retry allowance (`retries >= 10` -> `retries >= 0`): exit `1`; attempt 1 rejects `ENOTEMPTY`.
3. Zero retry delay (`50 * retries` -> `0 * retries`): exit `1`; ten zero pauses differ from `[50, 100, ..., 500]`.
4. Remove non-`ENOTEMPTY` rejection (`error?.code !== "ENOTEMPTY" ||`): exit `1`; attempts `11 !== 1`, proving `EACCES` was wrongly retried.
5. Remove retry cap (`|| retries >= 10`): exit `1`; forbidden eleventh pause and rejection no longer preserves the exact persistent `ENOTEMPTY` error.

All five mutants were independently restored. Index and worktree are clean.

## Restored-green verification on exact head

- Focused Git/teardown controls — 4/4.
- Node `24.18.0`: `npm run test:store-listing` — 60/60.
- GitHub Actions policy — 74 source, 7 gate-verdict, 129 current-workflow mutation cases; live green.
- Workflow topology — 20/20; live green (2 gates, 56 jobs, 14 workflows).
- Hash-domain labels — 17/17; live green across #611's E2EE documentation.
[...]

PR #607, opened 2026-08-24
feat(rs): add f2z-relay-testkit, the FakeRelay

skyl:
Builds `rs/crates/f2z-relay-testkit` — the **FakeRelay**: a spec-conforming
free2z relay a client developer can build and test against **before
`f2z-relay` exists**, and can drive failure modes on demand.

`WIRE.md` specifies a relay that does not exist. `CLIENT-CONTRACT.md` (merged
today) describes clients that have to be built against it now. This is the
thing in between.

> **Status: draft, and auto-closed by the merge-train lock**, which posted
> *"#598 is the sole active merge candidate. The branch is preserved; this PR
> will be reopened as draft automatically after #598 lands."* Nothing here is
> asking to merge. The branch is `feature/relay-testkit` at `0806212`, on base
> `2d114c8`.
>
> **CI did run before the lock closed it, and both required gates are green:**
> `gate` ✅ · `rs / gate` ✅ · `rs / rustfmt` ✅ · `rs / clippy` ✅ ·
> `rs / cargo-deny` ✅ · `rs / tests` ✅ · `rs / wasm32` ✅
> ([run](https://github.com/free2z/zuu/actions/runs/32688893466)).
> The local evidence below was gathered on the pinned toolchain against the
> same base and agrees with it.

Refs #305. Related: #586 (this PR implements the defect as a fault, and does
not fix the document).

---

## What lands

| | |
|---|---|
| **In-process `FakeRelay`** | over `tokio::io::duplex`, no sockets, milliseconds |
| **The same relay over `ws://127.0.0.1:0`** | real TCP accept, real RFC 6455 upgrade, mandatory subprotocol, binary/text framing, Ping/Pong, close codes |
| **`f2z-fakerelay` binary + `rs/deploy/docker-compose.dev.yml`** | an endpoint in one command |
| **Fault injection** | drop, delay, reorder, close mid-stream, refuse with every §10 code, stall, TTL expiry, quotas, backpressure, the #586 capability document |
| **52 conformance vectors** | driven through the ordinary client API against an `Endpoint`; `f2z-relay` runs the same file unchanged |

Storage is in-memory and self-contained. **No dependency on
`f2z-relay-store`** — that crate is being built concurrently and this must not
block on it.

## The two structural claims, and how each is checked

**1. Both transports run the same implementation.** `connection::drive` and
`engine::Relay` are the only connection loop and the only relay in the crate;
the in-process pipe and the WebSocket listener are two `TransportSink` /
`TransportStream` pairs feeding them. That is an architecture claim, so
`tests/conformance.rs::both_transports_reach_the_same_verdicts` runs the whole
suite against both and asserts the per-vector verdicts are **identical**. If
the fast path ever diverges, it fails.

**2. The suite is the real relay's suite too.** `vectors::run` takes an
`Arc<dyn Endpoint>`. `f2z-relay` will be a `WebSocketEndpoint::new(url)` with
no fault handle, and `tests/conformance.rs::a_target_without_a_fault_handle_skips_rather_than_passes`
already exercises exactly that shape: 39 pass, 13 report **`Skipped`**, 0 fail.
A vector that needs to break the relay is never counted as green against a
relay nobody can tell to misbehave.

## Faults are not relaxations, and that distinction is the safety property

The brief's main hazard is a fake that accepts what the real relay would
reject. So the crate has two separate types and they are not interchangeable:

- **`faults::*`** makes the relay behave **worse** than the spec allows, in
  ways a real relay or a real network can. Always available.
- **`config::Relaxations`** makes it **accept** something a conforming relay
  would reject. Four fields, every one **off by default**, each named after the
  rule it suspends (`accept_any_payload_size`, `accept_any_timestamp`,
  `accept_missing_pow`, `accept_repeated_hello`), and `Relaxations::any()`
  exists so a harness can assert a run used none.

Nothing in the test suite turns a relaxation on.

## Reuse, not reimplementation

`f2z-codec` owns re-encode equality (§3.3) and framing (§4). `f2z-relay-proto`
owns §5.1's verification order, §5.5's anti-replay, §7/§8's queue and ACK
rules, and §11's capability document. None of it is re-derived here — a
conformance fake holding a second opinion about the protocol is worse than no
fake. What the engine decides is what those crates deliberately leave to a
server: the §2.5 lifecycle, §4.3's in-flight window, which code answers which
refusal, when a push is emitted, and §13.1's layering.

Two rules are worth calling out because they are easy to get backwards and are
implemented literally:

- **§6.3's collapse.** Every send-side refusal — absent, deleted, expired,
  full-by-messages, full-by-bytes, unbound, wrong key, backpressure — is
  `ERR_UNAVAILABLE`. A vector asserts that a full queue and an absent address
  are indistinguishable.
- **§10's existence-oracle rule.** "No such queue" and "exists but your key
  does not authorize it" both return code 10, and the absent path calls
  `f2z_relay_proto::key::dummy_verify()` first and consumes its result, so the
  dominant CPU cost is present in both paths. §10's residual timing channel is
  narrowed, not closed, exactly as the section says.

## Verification — local, base `2d114c8`, toolchain 1.97.1

```
  --manifest rs/crates/f2z-relay-proto/Cargo.toml \
  --manifest rs/crates/f2z-relay-testkit/Cargo.toml
  → Every Rust toolchain pin agrees with wallet/rust-toolchain.toml: channel 1.97.1, MSRV 1.97.

scripts/check-rust-fmt.sh    --root rs   → All 4 Rust crate(s) under rs/ are formatted.
scripts/check-rust-clippy.sh --root rs   → All 4 Rust crate(s) under rs/ are clippy-clean.  (-D warnings)
scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
                                         → advisories ok, bans ok, licenses ok, sources ok.

cd rs && cargo test --locked --all-targets   → 267 passed, 0 failed  (14 binaries)
cd rs && cargo test --locked --doc           → 1 passed, 0 failed

node scripts/check-github-actions-pins.mjs   → passed
node scripts/check-workflow-gates.mjs        → passed
```

New manifest registered with `scripts/check-rust-toolchain.sh` in
`.github/workflows/rs.yml`, and **deliberately not** added to `rs_wasm` — it
opens sockets, spawns tasks and reads a clock, and a test harness that could
reach the browser build could end up in the shipped bundle. The reason is
written into the workflow beside the `-p` list.

### The worked example, over the real socket

`cargo test -p f2z-relay-testkit --test worked_example -- --nocapture`:

```
relay listening at ws://127.0.0.1:55826/relay/v1
  relay_id      RelayId(<redacted>)
  transport     ws:// — transport_security: none, channel_binding_mode: none (WIRE.md §2.3)
CREATE_QUEUE  → message_ttl 604800s, idle_ttl 7776000s (granted after clamping, §7.7)
SUBSCRIBE     → next_index 0, pending 0
BIND_SEND     → empty response, once and forever (§7.3)
BIND_SEND #2  → ERR_ALREADY_BOUND (§7.4: noisy, not prevented)
APPEND        → empty response: no index, no depth, no timestamp (§6.3)
← MSG push    → index 0 (§6.4, receive side only)
READ          → 1 message at index 0, has_more 0
(durable write completes — the §8.4 MUST)
ACK           → next_index 1, pending 0
READ          → 0 messages: the ciphertext is gone (§8.1)
ACK u64::MAX  → ERR_ACK_TOO_HIGH (§8.2: no pre-acking)
DELETE_QUEUE  → empty response, irreversible (§7.6)
← QUEUE_EVENT → reason 1 (deleted)
APPEND        → ERR_UNAVAILABLE (§6.3: the collapse, so no state leaks)
READ          → ERR_NO_ACCESS, same as an address that never existed (§10)
```

### The conformance suite, both transports

```
in-process (tokio::io::duplex):            52 passed,  0 skipped, 0 failed, of 52
websocket ws://127.0.0.1:.../relay/v1:     52 passed,  0 skipped, 0 failed, of 52
websocket, no fault handle (f2z-relay's shape):
                                           39 passed, 13 skipped, 0 failed, of 52
```

### Fault modes, each demonstrably firing

`cargo test -p f2z-relay-testkit --test faults` — 12 tests, all passing:
`drop_a_response`, `delay_a_response_without_blocking_the_ones_behind_it`,
`reorder_two_responses`, `close_the_connection_mid_stream`,
`stall_an_ack_forever`, `refuse_an_append_with_every_error_code` (all 21 §10
codes), `a_refused_append_did_not_append`, `expire_a_ttl_early`,
`expire_a_queue_by_moving_the_clock`, `hit_a_quota`,
`force_the_capability_document_violation_from_586`,
`a_fault_can_be_armed_mid_conversation_and_disarmed_again`.

The `f2z-fakerelay` binary was run by hand; §2.3's binding rule is enforced,
not printed:

```
$ f2z-fakerelay --listen 0.0.0.0:9999
f2z-fakerelay: invalid relay configuration: refusing to bind a non-loopback
address without TLS (WIRE.md §2.3); pass --insecure-listen if you really mean it
```

---

## Ambiguity calls — every one, with the reading taken

Ordered by section. Each is a place `WIRE.md` does not say, and each is
implemented one way and commented at the point of use.

**§1.3 / §4.1 — the WebSocket close status for a fatal error.** §4.2 fixes
1003 for a text frame and no other status is named. **Taken:** 1002 (protocol
error) for every other fatal code.

**§4.1 / §4.3 — a fatal error on a frame whose `request_id` is unreadable.**
§1.3 says "send the response, then close"; §4.3 says a response's `request_id`
must be nonzero. For a frame under 5 bytes, or one whose kind byte is not
`request(1)`, there is no id to answer on and both rules cannot hold.
**Taken:** best-effort recovery of the id from the fixed 5-byte header — which
covers almost every malformed frame — and, when that fails, close with **no
response frame** rather than emit one §4.3 forbids. Applies to an oversize
frame (§4.1) and to §4.2's text frame, which carries no id at all.

**§2.5 / §3.5 — a second `HELLO` on one connection.** §3.5 says negotiation
happens once and applies to the whole connection; §10 names no code for a
renegotiation attempt. **Taken:** `ERR_UNSUPPORTED_VERSION`, fatal — the same
code §10 already assigns to "a frame before `HELLO`". Suspendable with
`Relaxations::accept_repeated_hello`.

**§4.1 — a client that sends a response or a push frame.** §10 has no code.
**Taken:** `ERR_MALFORMED`, fatal; it is a frame no conforming client emits.

**§6.1 — an unknown `ChallengePurpose` byte.** `f2z-codec` keeps it raw so it
round-trips §3.3, and its doc says a purpose this build does not know is "the
relay's to refuse with a code" — but §10 has no code for an unknown enum value
inside a body. **Taken:** `ERR_MALFORMED`. Flagged as the weakest call here;
it is fatal, and a non-fatal code would be defensible if one existed.

**§6.1 / §12.3 — validating `GET_CHALLENGE`'s `scope` at issuance.** Nothing
requires the relay to check it. **Taken:** issue whatever scope is asked for
and enforce the binding at **consumption**, where §12.3 actually needs it
(the stamp's challenge must have been issued for that `contact_addr`). This
keeps a wrong-length scope out of the fatal path.

**§6.2 — `READ` with `max_messages == 0` or `max_bytes == 0`.** No meaning is
given. **Taken:** "no client-side limit", matching `f2z-relay-proto`'s reading
of a zero TTL as "no preference". A zero that meant "return nothing" would let
a client that forgot the field loop forever against a queue it can never
drain.

**§6.2 — `READ` whose first message exceeds `max_bytes`.** **Taken:** always
return at least one message, with `has_more` set accordingly. The alternative
stalls the queue permanently, because `READ` never mutates and there is no
other way out.

**§6.2 — `pending` after TTL expiry.** `f2z-relay-proto` computes it in index
space and documents that this is an upper bound because expiry removes
messages without moving the watermark. **Taken:** count from storage, which is
what that crate says a relay with storage should do.

**§6.3 / §12.2 — `APPEND` to a *contact* queue's published address.** §12.2
gives contact queues `CONTACT_APPEND`; it does not say what `APPEND` does
there. **Taken:** the contact send side is never bound, so `authorize_send`
refuses with `ERR_UNAVAILABLE` — the §6.3 collapse, with no new rule.

**§6.3 vs §10 — the send-side "absent" case.** §6.3 collapses everything to
`ERR_UNAVAILABLE`; §10's table assigns "absent" to code 10. They contradict.
**Taken:** §6.3, as #583 did, because §6.3's rule carries the security
argument. This is defect 2 of #586 and is not fixed here.

**§7.3 — `BIND_SEND` on an address that does not exist.** **Taken:**
`ERR_UNAVAILABLE`, per the same collapse. `ERR_NOT_PERMITTED` remains
distinguishable for a *contact* address, per §12.2's explicit "always, for
everyone" — and that address is published by design, so it leaks nothing.

**§5.5 / §11.1 — where the seen-set lives.** §5.5 says "the relay keeps the
set" without saying whether it is per connection. **Taken:** relay-wide. In
`channel_binding_mode: none` — which a `ws://` fake must publish — the binding
is a constant, so a frame captured on one connection verifies on another and a
per-connection set would not close the window at all.

**§11.1 — `antireplay_seen_max` has no published default.** **Taken:** 65 536.

**§13.1 — `queue_creation_mode: token`.** §13.1 defines it as an
operator-issued bearer credential and gives it no protocol shape in v1.
**Taken:** refuse it at construction rather than invent one.

**§2.4 — the keepalive is never faulted.** A suppressed Ping presents as a
hung relay rather than as the fault it is, so `Effect::*` does not apply to
keepalives. A deliberate limitation, not an oversight.

## Spec defects, reported rather than worked around

**#586 defect 1 — `antireplay_window_ms` is unrelated to `clock_skew_ms`.**
Not fixed here (docs-only issue), but now *demonstrable*:
`PolicyFaults::unsound_antireplay_window` publishes
`antireplay_window_ms = clock_skew_ms` **and enforces it**, so the relay really
does age a seen-set entry out while the frame it covers is still inside the
timestamp window. A vector asserts both halves — that
`capabilities::validate` **accepts** the document (the spec as written permits
it, which is the finding) and that `ClientPolicy::accept` **refuses** it. Also
reachable from the binary as `--unsound-antireplay` and from
`docker-compose.dev.yml`'s `faulty` profile.

**#586 defect 2 — §6.3 vs §10.** Implemented in §6.3's favour, as above. No
new finding; the contradiction is exactly as #586 describes it.

**No new spec defects found.** The list above is ambiguity, not contradiction.

## Known gaps, stated

- **§11.2's `.well-known` JSON document is not served.** `GET_CAPABILITIES`
  and the `HELLO` digest comparison both work; the plain-HTTPS representation
  does not exist. Serving it means an HTTP parser beside the WebSocket one —
  the second parser §2.2 argues against — for an affordance aimed at humans.
  So a client's "fetch the well-known copy and compare digests" path cannot be
  exercised against this harness. Worth doing if a client author asks.
- **The channel binding is not a TLS exporter.**
  `ChannelBindingSource::Simulated` hands both ends a constant so the §5.3
  exporter code path is reachable; it binds a signature to *this
  configuration*, not to a TLS session. A passing test there is not evidence
  about RFC 8446 §7.5, and the type says so.
- **Queue addresses are predictable, by design.** §7.1 requires the opposite.
  A conformance vector that produced different addresses each run would not be
  a vector, so the generator is BLAKE2b in counter mode over a seed. This is
  the first of three reasons `f2z-fakerelay` must never be deployed, and it is
  in the crate docs, the binary's `--help` banner, the README and the
  Dockerfile.
- **`§4.3`'s duplicate-`request_id` rule needs a fault handle to observe.** A
  relay releases an id the moment it answers, so the rule is only visible while
  a response is held. That vector declares `Needs::Faults` and reports
  *skipped* against a real relay rather than being turned into a race that
  passes by luck.

## Reviewer's shortlist

- `src/engine.rs` — the refusal-code decisions and §5.1's ordering.
- `src/vectors.rs` — the 52 cases; each names the section it pins.
- `src/config.rs::Relaxations` — confirm nothing turns one on by default.
- `.github/workflows/rs.yml` — the manifest registration, and the `rs_wasm`
  absence.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -75,7 +75,8 @@ jobs:
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
             --manifest rs/crates/f2z-codec/Cargo.toml \
-            --manifest rs/crates/f2z-relay-proto/Cargo.toml
+            --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+            --manifest rs/crates/f2z-relay-testkit/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -293,6 +294,11 @@ jobs:
       # rather than `--workspace` so that a future server binary — which will be
       # AGPL, will pull tokio, and will never reach a browser — does not
       # silently join the wasm build and then fail it.
+      #
+      # f2z-relay-testkit is deliberately absent for exactly that reason: it
+      # opens sockets, spawns tasks and reads a clock, and a test harness that
+      # could reach the browser build would be a test harness that could end up
+      # in the shipped bundle.
       - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -57,10 +57,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytes"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "rand_core 0.10.1",
+]
+
+[[package]]
+name = "const-oid"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
 
 [[package]]
 name = "cpufeatures"
@@ -114,8 +137,14 @@ checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.119",
 ]
+
+[[package]]
+name = "data-encoding"
+version = "2.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4583a4551df46e2792f82ceeac45e850d2e2d5debba0b91f102385cda5b11f06"
 
 [[package]]
 name = "digest"
@@ -135,6 +164,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f1dd6dbb5841937940781866fa1281a1ff7bd3bf827091440879f9994983d5c2"
 dependencies = [
  "block-buffer 0.12.1",
+ "const-oid",
  "crypto-common 0.2.2",
 ]
 
@@ -191,6 +221,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-relay-testkit"
+version = "0.1.0"
+dependencies = [
+ "f2z-codec",
+ "f2z-relay-proto",
+ "futures-util",
+ "tls_codec",
+ "tokio",
+ "tokio-tungstenite",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -207,6 +249,37 @@ name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "futures-core"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92d699e522242e69e3003b94ecc1f960f3a5e015aa7c5d7486e65ad01dd94f5e"
+
+[[package]]
+name = "futures-sink"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1944426bf7d03f1d14f708785e4b33efd750b36d48a157b836b3efc15ede8e1d"
+
+[[package]]
+name = "futures-task"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd417de3d1d015fc3bfd2b1ea46dfc7bab72ef86f1cc7cc9c78e728b34a6d1fd"
+
+[[package]]
+name = "futures-util"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d50a92467f8ba5dd6e3ee5d4bd04d73ab2e4e1c44474a0674821dfce14b79bc"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+ "futures-task",
+ "pin-project-lite",
+ "slab",
+]
 
 [[package]]
 name = "generic-array"
@@ -239,7 +312,24 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi 6.0.0",
+ "rand_core 0.10.1",
 ]
+
+[[package]]
+name = "http"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "918d3568bebf352712bc2ef3d46a8bcf1a75b373be6539de198e9105cbbf9ce0"
+dependencies = [
+ "bytes",
+ "itoa",
+]
+
+[[package]]
+name = "httparse"
+version = "1.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6dbf3de79e51f3d586ab4cb9d5c3e2c14aa28ed23d180cf89b4df0454a69cc87"
 
 [[package]]
 name = "hybrid-array"
@@ -249,6 +339,12 @@ checksum = "707114b52a152fa7bdb290cd7cd5912d9467273b6d74e21b8d81aca1f8533f6b"
 dependencies = [
  "typenum",
 ]
+
+[[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
 
 [[package]]
 name = "libc"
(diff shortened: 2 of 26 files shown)


skyl:

skyl:
Hard hold: do not merge or reopen this PR while #598 exact-head CI is running. Prior admin merges of held PRs repeatedly invalidated fully re-proven candidates. This branch is preserved and will be rebased/revalidated as the next appropriate train item after #598 lands. Green checks on base 2d114c8 are stale relative to current main 2dffff0.

skyl:

skyl:

skyl:

PR #609, opened 2026-08-24
security(e2ee): make the H() label namespace prefix-free, and assert it over the union

### LINKED ISSUE #602: security(e2ee): KT.md's label set is not prefix-free — `sth` is a prefix of `sth-hash`, and H() has no separator
## The defect

`docs/e2ee/WIRE.md:67-69` defines the hash construction as:

> `H(label, x) = BLAKE2b-256(label || x)` — **no separator and no terminator**

That construction provides domain separation **only if the label set is prefix-free**. `KT.md`'s label set is not:

```console
$ grep -oE '"free2z/kt/v1/[a-z0-9-]+"' docs/e2ee/KT.md | sort -u
"free2z/kt/v1/cosig"
"free2z/kt/v1/entry"
"free2z/kt/v1/fault"
"free2z/kt/v1/log-id"
"free2z/kt/v1/log-key-transition"
"free2z/kt/v1/prev"
"free2z/kt/v1/receipt"
"free2z/kt/v1/reset"
"free2z/kt/v1/rotation"
"free2z/kt/v1/sth-hash"
"free2z/kt/v1/sth"
"free2z/kt/v1/value"

$ # prefix check across all 12
  COLLISION: 'free2z/kt/v1/sth' is a prefix of 'free2z/kt/v1/sth-hash'
```

**`free2z/kt/v1/sth` is a proper prefix of `free2z/kt/v1/sth-hash`.**

## Why that breaks the separation

With no separator, the two domains overlap exactly:

```
H("free2z/kt/v1/sth",      "-hash" || Y)
  = BLAKE2b-256("free2z/kt/v1/sth" || "-hash" || Y)
  = BLAKE2b-256("free2z/kt/v1/sth-hash" || Y)
  = H("free2z/kt/v1/sth-hash", Y)
```

So a hash computed in the `sth` domain over any message beginning with the five bytes `-hash` is **bit-identical** to a hash computed in the `sth-hash` domain over the remainder. Domain separation between those two is not weakened, it is absent for that message class.

Whether it is *reachable* depends on whether an attacker can influence the first five bytes of an `sth`-domain input — which is a question about the encodings, not about the hash. But prefix-freeness is precisely the property that makes reachability irrelevant, and it is the property `WIRE.md:67-69` silently depends on. The two hashes here are over signed tree heads, which is the structure the whole witness/anti-equivocation argument rests on.

## Context: the invariant was already flagged as unstated

#552 recorded that `H`'s prefix-freeness is an **unstated invariant** — safe at the time because `WIRE.md`'s six labels happened to be prefix-free, but never written down, so nothing stops a future label from breaking it:

> The six labels in use are currently prefix-free, so it is safe today — but the invariant is never written down, and a future label that prefixes an existing one silently breaks domain separation. One sentence fixes it.

`KT.md` then added its own label set without that sentence existing, and broke it. This is the predicted failure, arriving exactly as predicted.

`WIRE.md`'s own six labels remain prefix-free — `f2z-codec` asserts it mechanically over its `LABELS` array (`hash.rs:132`), which is why `WIRE.md` did not drift. `KT.md`'s labels have no equivalent guard.

## How it stayed invisible

`f2z-authority`'s `no_label_is_a_prefix_of_another_anywhere_in_the_namespace` ranges over **7 of `KT.md`'s 13 labels**. Completing the set fails on exactly this pair. The omission is what makes the test green — a fixture shaped so the invariant it names cannot be violated, which is the defect class this repo keeps finding.

## Fix

Two independent changes, both worth making:

1. **Rename one label** so the set is prefix-free — `free2z/kt/v1/sth-digest`, or rename the shorter one. Renaming changes every digest in that domain, so it must happen before anything depends on the current values.
2. **State the invariant in `WIRE.md` §1** — that every label in every document using `H` must be prefix-free with respect to every other — and **assert it mechanically over the union of `WIRE.md`'s and `KT.md`'s labels**, not per-document. A per-crate check is what let this through: each set was internally consistent.

The assertion belongs somewhere that sees both. `f2z-codec`'s `hash.rs:132` guard is the right shape and the wrong scope, in the same way the `Debug` scan was before #597 widened it to the workspace.

## Acceptance

- No label in `WIRE.md` ∪ `KT.md` is a proper prefix of another, demonstrated by a check over the union.
- `WIRE.md` §1 states prefix-freeness as a normative requirement on the label set, not an accident.
- The check fails on today's tree before the rename and passes after — paste both.
- `f2z-authority`'s prefix test ranges over the complete label set rather than a subset.

Found while reviewing #593. Related: #552 (the unstated invariant), #558, #585.

skyl:
`WIRE.md` §1.3 defines `H(label, x) = BLAKE2b-256(label || x)` with **no
separator and no terminator**. That separates domains only if the label set is
prefix-free. `KT.md`'s was not.

`Closes #602.`

## Red before

`scripts/check-hash-domain-labels.mjs` copied onto a pristine `origin/main`
checkout (`2dffff0`) and run there:

```console
$ git worktree add --detach /tmp/probe602 origin/main
$ cp scripts/check-hash-domain-labels.mjs /tmp/probe602/scripts/
$ cd /tmp/probe602 && git add scripts/check-hash-domain-labels.mjs
$ node scripts/check-hash-domain-labels.mjs
Hash-domain label separation failed:
- "free2z/kt/v1/sth" is a proper prefix of "free2z/kt/v1/sth-hash": H("free2z/kt/v1/sth", "-hash" || y) is bit-identical to H("free2z/kt/v1/sth-hash", y), so those two domains are not separated. free2z/kt/v1/sth at docs/e2ee/KT.md:502; free2z/kt/v1/sth-hash at docs/e2ee/KT.md:508
- LABEL_BEARING_DOCUMENTS registers docs/e2ee/README.md, which yielded no labels; either it stopped minting labels or this scan stopped reading it
2 failure(s) over 30 label(s) from 1389 tracked file(s).
$ echo $?
1
```

The first failure is the defect. The second is an artifact of running this
branch's registry against a tree that predates it — `README.md` gains its first
label reference in this PR — and is itself the coverage anchor doing its job.

## Green after

```console
$ node scripts/check-hash-domain-labels.mjs
No label is a proper prefix of another: 30 label(s) across 7 registered document(s) under docs/ and the crates under rs/, read from 1389 tracked file(s) (1 bare namespace mention(s) and 3 registered fixture token(s) excluded).
$ echo $?
0
```

And the fix is not load-bearing on the fixture — reintroducing the old label on
the real line it lived on brings the failure straight back:

```console
$ sed -i '' '508s|tree-head-hash|sth-hash|' docs/e2ee/KT.md
$ node scripts/check-hash-domain-labels.mjs
Hash-domain label separation failed:
- "free2z/kt/v1/sth" is a proper prefix of "free2z/kt/v1/sth-hash": ... free2z/kt/v1/sth at docs/e2ee/KT.md:502, docs/e2ee/README.md:122, docs/e2ee/WIRE.md:96; free2z/kt/v1/sth-hash at docs/e2ee/KT.md:508
1 failure(s) over 31 label(s) from 1389 tracked file(s).
```

## Verifying the premises

Both held.

**The collision is real, and it is the only one.** Enumerating every
`free2z/<ns>/v<n>[/<leaf>]` token in every tracked file on `origin/main` and
testing all ordered pairs gives 11 prefix relations, of which exactly one is a
defect: `free2z/kt/v1/sth` ≺ `free2z/kt/v1/sth-hash`. The other ten are
`free2z/relay/v1/` (a bare namespace mention in `hash.rs`'s module note, a prefix
of all nine relay labels) and `free2z/relay/v1/cmd` ≺ `free2z/relay/v1/cmd2` (the
negative-control fixture at `transcript.rs:358`). Both are handled below.

**Nothing consumes the KT labels.** `grep -rn 'free2z/kt/v1' rs/` on `origin/main`
returns nothing; `f2z-kt-core` and `f2z-authority` are not merged and `rs/` holds
only `f2z-codec`, `f2z-relay-proto` and one more crate. The rename changes no
shipped digest.

**The set is larger than the issue described.** The issue's `grep` for quoted
`free2z/kt/v1/...` finds 12 labels. The real namespace holds 30, including six
constants with **no leaf** — `free2z/device-credential/v1` (a
`SignedTreeHead`-family signing label in `KT.md` §6.2's closed table),
`free2z/queue-advert/v1`, and the four MLS exporter labels `free2z/queue/v1`,
`free2z/frost/v1`, `free2z/webrtc/v1`, `free2z/history/v1`. A leafless label is a
prefix of *every* leafed label in its namespace, which is the sharpest form of
this defect, so the check covers them. They are prefix-free today.

## Which label was renamed, and why

**`free2z/kt/v1/sth-hash` → `free2z/kt/v1/tree-head-hash`.** The longer one.

- **`sth` is a wire field of five signed structures; `sth-hash` is not.**
  `free2z/kt/v1/sth` is the first-field `label<0..255>` constant of
  `SignedTreeHeadTBS` and a row in `KT.md` §6.2's *closed* transcript table,
  whose other eight rows are all a bare noun matching their TBS struct — `cosig`,
  `receipt`, `entry`, `rotation`, `reset`, `log-key-transition`, `fault`.
  Renaming `sth` would break that table's one visible pattern to fix a problem
  the other member caused. `sth-hash` is a digest label used at three sites and
  named in no table.
- **It removes the prefix relation rather than moving it.** `tree-head-hash`
  shares no prefix with `sth` in either direction. Renaming `sth` to, say,
  `sth-tbs` would have left two labels differing only in their last segment,
  which reads as a near-miss rather than as two domains.
- **It spells out what `sth` abbreviates.** These labels are read by a second
  implementer working from the document alone; `tree-head-hash` is
  self-describing where `sth-hash` needs §6.1 open beside it.

Struct field names (`prev_sth_hash`, `announce_sth_hash`) are unchanged — they
are fields, not domains.

## The normative invariant

`WIRE.md` §1.3 now carries it as a **MUST**, in the section that defines `H`,
scoped to the **union** of every document's labels and extended to the labels
concatenated the same way without `H` (§5.2's `relay_proof` signing prefix, and
the first-field constants of the signed structures). #552 said one sentence would
fix it; this is that sentence plus the derivation and a pointer to the check.

## Correction convention

`docs/e2ee/README.md`'s rule, as #592 phrased it: *a claim that is made true is
fixed in place; a claim that is withdrawn gets a dated correction.* The rename is
neither, so it is decided rather than defaulted, and the two halves go different
ways:

- **`KT.md` §6.1 gets a dated correction (2026-08-24).** `KT.md` is a frozen
  Phase 1 specification. `free2z/kt/v1/sth-hash` was a normative constant, and
  every digest under it is now invalid. An implementer working from the first
  revision must be told, at the point of use, that a value they may have wired is
  withdrawn — a frozen spec has a stronger claim on this treatment than a new
  document precisely because someone may have relied on it. The correction states
  the old label, why it was wrong, the new one, and that any digest under the old
  label MUST be recomputed. It deliberately does not spell the old label as a
  literal token — it names it as "`-hash` appended to `free2z/kt/v1/sth`" — so a
  withdrawn label does not live on in the namespace the checker reads.
- **`WIRE.md` §1.3 is fixed in place, no correction note.** Nothing there is
  withdrawn. Prefix-freeness was always the construction's precondition; §1.3 was
  silent about it, and silence corrected into a MUST is a claim *made true*, not
  a claim retracted.

`README.md`'s Status section records the correction alongside the two Phase 0
ones.

## Where the check lives, and how it reaches the gate

`scripts/check-hash-domain-labels.mjs`, not a Rust test. Three reasons, each
sufficient on its own:

1. **`KT.md` has no crate.** A Rust test could only assert over a hand-copied
   restatement of the document's label set — and a hand-maintained copy drifting
   from its document is the same failure that produced #602.
2. **The collision arrived in a docs-only PR.** `rs.yml`'s detector sets
   `rs=false` for such a diff and every rs job skips, so a Rust test would not
   have run on the commit that broke the invariant.
3. **The labels are minted in prose.** Reading the documents directly means the
   check cannot go stale relative to the spec.

Wiring, stated concretely because this repo has a history of gates that `needs:` a
job without reading its result (#512, #515–#518):

- A **new step** in `rs.yml`'s `changes` job — `Verify hash-domain labels are
  prefix-free across the tree` — running `--self-test` then the bare check. That
  job has no `if:` and runs on every `pull_request`, `merge_group`, `push` to
  `main` and `workflow_dispatch`, so a docs-only PR is judged.
- The same two commands appended to `rs / gate`'s `Recheck the workflow policy`
  step (`id: policy`). The gate's verdict step binds `POLICY_OUTCOME:
  ${{ steps.policy.outcome }}` and exits 1 unless it is `success`. The result is
  consumed, not merely awaited.
- `rs / gate` is one of the two required contexts on `main` (`gate`,
  `rs / gate`), confirmed by `check-workflow-gates.mjs`'s `PROTECTED_CONTEXTS`.
- **A separate step, deliberately.** `scripts/check-github-actions-pins.mjs`
  holds an exact ordered six-command assertion over the `changes` policy step —
  but only for `.github/workflows/zuuli.yml` (`REQUIRED_WORKFLOW_PATH`, applied
  via `gatePolicyFailures`, which `scanRepository` calls for that path alone).
  `rs.yml` is not subject to it. A separate step keeps it that way even if the
  assertion is ever widened, which is why `check-workflow-gates.mjs` and
  `check-rust-toolchain.sh` already have their own steps.

## Fixtures vs. labels — where the line is drawn

By **registration, never by heuristic**. `FIXTURE_TOKENS` names a file, an exact
token and a reason:

| file | token | why it is not a domain |
|---|---|---|
| `f2z-codec/src/transcript.rs:358` | `free2z/relay/v1/cmd2` | negative control, asserted to be rejected by `CommandTranscript::validate` |
| `f2z-codec/tests/wire_vectors.rs:761` | `free2z/relay/v1/cmX` | one-byte corruption of `LABEL_COMMAND`, asserted to fail the §5.1 vector |
| `f2z-relay-proto/src/key.rs:160` | `free2z/relay/v1/dummy` | filler message handed to `verify()` with a zero signature |

Every available heuristic is wrong here, which is the argument for a registry:

- *"It is in a test file"* would drop the **six genuine labels**
  `wire_vectors.rs` pins at lines 271–276 — the authoritative byte-level
  assertion of `WIRE.md`'s label set.
- *"It is not in a `LABELS` array"* is the exact scoping mistake this script
  exists to fix.

The registry cannot loosen into a blanket excuse: it is keyed by **file**, so the
same token elsewhere is judged as a label; and an entry whose token no longer
occurs in its file **fails**, so it cannot outlive review. Proven on the real
tree:

```console
$ sed -i '' 's|b"free2z/relay/v1/cmd2"|b"free2z/relay/v1/wrong"|' rs/crates/f2z-codec/src/transcript.rs
$ node scripts/check-hash-domain-labels.mjs
Hash-domain label separation failed:
- FIXTURE_TOKENS excuses "free2z/relay/v1/cmd2" in rs/crates/f2z-codec/src/transcript.rs, which no longer contains it; remove the entry rather than leaving an unreviewed exclusion in place
```

Two other exclusions, both narrow: a token ending in `/` is a **namespace
mention** (`free2z/relay/v1/` in `hash.rs`'s module note) and not a label — a
leafless name written *without* the slash is still judged as a label, which is
the conservative direction; and the checker's own file is excluded whole via
`NON_DECLARING_FILES`, because it quotes labels it does not mint and invents
fixture labels in reserved `demo`/`other`/`third` namespaces. That exclusion is
load-bearing and self-limiting — its path must be tracked, so renaming the
checker fails rather than silently leaving a file unread. With it emptied the
checker reports seven collisions among its own examples, which is what a
scanned-self looks like.

## Anchor proof

Narrowing the check to a single document is red, not a smaller green:

```console
$ # LABEL_BEARING_DOCUMENTS = ["docs/e2ee/WIRE.md"]
$ node scripts/check-hash-domain-labels.mjs
Hash-domain label separation failed:
- docs/e2ee/ARCHITECTURE.md, docs/e2ee/KT.md, docs/e2ee/README.md, docs/e2ee/THREAT-MODEL.md, docs/e2ee/decisions/0009-queue-addressing-and-binding.md, docs/e2ee/decisions/0010-signing-transcript-and-ack-semantics.md mint labels but are not in LABEL_BEARING_DOCUMENTS; register them, so that adding a specification is an acknowledgement that its labels share one prefix-free namespace with every other document's
1 failure(s) over 30 label(s) from 1389 tracked file(s).
$ echo $?
1
```

The anchor runs both ways: a registered document that yields no labels fails
(a rename or a scan that stops reading Markdown), and a label-bearing document
under `docs/` that is unregistered fails (a new `KT.md`). It fired for real
during this PR — adding the label names to `README.md` turned the check red until
`README.md` was registered. A `MINIMUM_LABELS` floor of 24 catches a scan that
collapses without either condition tripping.

## Self-test

17 cases, following `check-workflow-gates.mjs`'s convention: fixtures judged
against fixture configuration, every negative control a one-line mutation of a
tree that passes. Includes the requested negative control for **this specific
case** — the `sth` / `sth-hash` pair reproduced verbatim, asserting the message
names both labels *and* their line numbers, so a refactor that stays green on
generic input but stops seeing `sth-hash` fails here.

`withFile` throws when a mutation changes nothing, so a negative control built on
a `String.replace` whose needle has drifted cannot silently assert against the
unmutated tree. That caught a real bug in this PR's own self-test.

```console
$ node scripts/check-hash-domain-labels.mjs --self-test
self-test: a prefix-free tree passes.
self-test: a new label that extends an existing one is detected.
self-test: the KT.md sth / sth-hash collision, verbatim is detected.
self-test: a collision between two documents, each internally consistent is detected.
self-test: a collision minted only in code is detected.
self-test: a fixture token that has left its file is detected.
self-test: the same token in a file the registry does not name is detected.
self-test: a wrong-label fixture that nobody registered is detected.
self-test: a quoting file that nobody excluded is detected.
self-test: a quoting file excluded whole passes.
self-test: a whole-file exclusion for a file that is no longer tracked is detected.
self-test: a bare namespace mention alongside every label under it passes.
self-test: the check narrowed to a single document is detected.
self-test: a registered document that is no longer tracked is detected.
self-test: a registered document that has stopped minting labels is detected.
self-test: a new specification nobody registered is detected.
self-test: a union far below its floor is detected.
check-hash-domain-labels self-test: 17 case(s) passed.
```

## `f2z-codec`'s existing guard

Kept, and its scope now stated where it was previously implied. The module note
said the property held of "this fixed set" and pointed at `LABELS` as the
mechanism; it now says the set that matters is the union, names the script that
checks it, and records why the per-crate test is not sufficient — `KT.md`'s
labels were internally prefix-free too.

## Checks run

| command | result |
|---|---|
| `node scripts/check-github-actions-pins.mjs --self-test` | 70 source-policy, 7 gate-verdict, 120 current-workflow cases passed |
| `node scripts/check-github-actions-pins.mjs` | 178 external refs, 15 files — pass |
| `node scripts/check-workflow-gates.mjs --self-test` | 20 cases passed |
| `node scripts/check-workflow-gates.mjs` | 2 gates, 56 jobs, 14 workflow files — pass |
| `scripts/check-rust-toolchain.sh --self-test` | 14 drift cases caught (incl. the #600 census) |
| `scripts/check-rust-toolchain.sh` | channel 1.97.1, MSRV 1.97 — pass |
| `node scripts/check-hash-domain-labels.mjs --self-test` | 17 cases passed |
| `node scripts/check-hash-domain-labels.mjs` | 30 labels, 1389 files — pass |
| `scripts/check-rust-fmt.sh --root rs` | 3 crates formatted |
| `scripts/check-rust-clippy.sh --root rs` | 3 crates clippy-clean |
| `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` | advisories/bans/licenses/sources ok |
| `cargo test --locked --all-targets` + `--doc` | 212 tests + 1 doctest, all pass |

**#477:** bare `[[ ]]` / `(( ))` statements added — **0**. The check is
JavaScript; the workflow additions are plain `run:` command lists with no shell
conditionals.

Related: #552 (the unstated invariant), #592 (the correction rule), #597 (the
coverage-anchor pattern this borrows), #600 (the registration census).

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -66,6 +66,20 @@ jobs:
           node scripts/check-workflow-gates.mjs --self-test
           node scripts/check-workflow-gates.mjs
 
+      # Deliberately in this always-running job and not behind the rs/ filter.
+      # The labels it judges live in docs/e2ee/*.md as much as in rs/, and the
+      # collision it exists to catch (#602) arrived in a docs-only pull request
+      # — a diff for which `filter` below sets rs=false and every rs job skips.
+      # A check that only runs when Rust changes would not have run on the
+      # commit that broke the invariant. Its own step rather than a line added
+      # to one above, so the exact ordered command assertion that
+      # scripts/check-github-actions-pins.mjs holds over the required workflow's
+      # policy step stays a single reviewed unit.
+      - name: Verify hash-domain labels are prefix-free across the tree
+        run: |
+          node scripts/check-hash-domain-labels.mjs --self-test
+          node scripts/check-hash-domain-labels.mjs
+
       # rs/rust-toolchain.toml and rs/crates/*/Cargo.toml restate
       # wallet/rust-toolchain.toml, and are registered in the script's own
       # TOOLCHAIN_RESTATEMENTS/MANIFESTS arrays rather than by flags here. A
@@ -332,6 +346,12 @@ jobs:
       # Re-run the workflow policy here so that even a job-level
       # continue-on-error on `changes` cannot turn that job's deliberately
       # failing policy verdict into the `success` conclusion this gate consumes.
+      #
+      # The label check is re-run here for the same reason, and this is how it
+      # reaches the required context: this step's `outcome` is bound below as
+      # POLICY_OUTCOME and the verdict refuses to pass unless it is `success`.
+      # `needs:` a job without reading its result is the defect of #512 and
+      # #515–#518; a step whose outcome is read is not that shape.
       - name: Recheck the workflow policy
         id: policy
         run: |
@@ -339,6 +359,8 @@ jobs:
           node scripts/check-github-actions-pins.mjs
           node scripts/check-workflow-gates.mjs --self-test
           node scripts/check-workflow-gates.mjs
+          node scripts/check-hash-domain-labels.mjs --self-test
+          node scripts/check-hash-domain-labels.mjs
 
       - name: Verify required jobs succeeded or legitimately skipped
         shell: bash
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -505,7 +505,7 @@ struct {
     uint64 epoch;
     uint64 tree_size;             /* total (label, version) insertions committed */
     opaque root_hash[32];         /* the akd Azks root at this epoch */
-    opaque prev_sth_hash[32];     /* H("free2z/kt/v1/sth-hash", tls_codec(prev SignedTreeHeadTBS)) */
+    opaque prev_sth_hash[32];     /* H("free2z/kt/v1/tree-head-hash", tls_codec(prev SignedTreeHeadTBS)) */
     opaque vrf_public_key[32];    /* the ECVRF key labels are derived under */
     uint64 published_at_ms;
     uint32 reset_count;           /* platform resets in this epoch — ADR 0014 */
@@ -535,6 +535,31 @@ change. It **MUST NOT** change within a `log_id`: it determines every label in
 the tree, so changing it silently invalidates every prior proof while producing
 proofs that still verify under the new key. A client or witness that observes a
 `vrf_public_key` change MUST treat it as a fork (§7.3), not as an update.
+
+> **Correction (2026-08-24) — the tree-head digest label was renamed, and the
+> old one must not be used.** As first published, this document computed
+> `prev_sth_hash` and `announce_sth_hash` (§6.4) under the label formed by
+> appending `-hash` to `free2z/kt/v1/sth`. That label was a **proper prefix
+> violation**: `free2z/kt/v1/sth` — the signing-transcript constant of
+> `SignedTreeHeadTBS` two paragraphs below — is a proper prefix of it, and
+> [`WIRE.md` §1.3](./WIRE.md#13-conventions) defines `H(label, x)` as
+> `BLAKE2b-256(label || x)` with **no separator and no terminator**. A digest
+> taken in the `free2z/kt/v1/sth` domain over any message whose first five bytes
+> were `-hash` was therefore bit-identical to a digest taken in the old
+> tree-head-digest domain over the remainder: between those two domains,
+> separation was absent rather than weakened, over exactly the structure the
+> anti-equivocation argument in §7 rests on.
+>
+> The label is now **`free2z/kt/v1/tree-head-hash`**, which no other label in the
+> namespace prefixes and which prefixes no other. **Any digest computed under the
+> old label is invalid and MUST be recomputed**; no shipped code held the old
+> value — the crates that will use these labels are not written — but an
+> implementer working from the first revision of this document must change it.
+> `WIRE.md` §1.3 now states prefix-freeness as a normative requirement on the
+> union of every document's labels rather than leaving it an unwritten
+> assumption, and `scripts/check-hash-domain-labels.mjs` enforces it on every
+> pull request. [#602](https://github.com/free2z/zuu/issues/602); predicted in
+> [#552](https://github.com/free2z/zuu/issues/552).
 
 ### 6.2 The signing transcript, and domain separation
 
@@ -578,7 +603,7 @@ following hold** against the last one it accepted for this `log_id`:
 5. `published_at_ms > last.published_at_ms`.
 6. `vrf_public_key == last.vrf_public_key`.
 7. **The `prev_sth_hash` chain connects.** If `epoch == last.epoch + 1`, then
-   `prev_sth_hash == H("free2z/kt/v1/sth-hash", tls_codec(last.sth))` directly.
+   `prev_sth_hash == H("free2z/kt/v1/tree-head-hash", tls_codec(last.sth))` directly.
    If `epoch > last.epoch + 1`, the verifier MUST fetch every intervening tree
    head and check the chain link by link. **It MUST NOT skip.** A gap accepted on
    trust is a branch accepted on trust.
@@ -603,7 +628,7 @@ struct {
     opaque label<0..255>;        /* exactly "free2z/kt/v1/log-key-transition" */
     opaque log_id[32];
     uint64 announce_epoch;       /* the epoch of the announcing STH */
-    opaque announce_sth_hash[32];/* H("free2z/kt/v1/sth-hash", tls_codec(announcing sth)) */
+    opaque announce_sth_hash[32];/* H("free2z/kt/v1/tree-head-hash", tls_codec(announcing sth)) */
     opaque outgoing_log_pk[32];
     opaque successor_log_pk[32];
     uint64 effective_epoch;      /* first epoch signed by the successor */
--- a/docs/e2ee/README.md
+++ b/docs/e2ee/README.md
@@ -116,6 +116,15 @@ witnesses can**), with the consequences carried into
 [`THREAT-MODEL.md` §3.1](./THREAT-MODEL.md#31-malicious-or-compromised-free2z-server)
 and [§3.9](./THREAT-MODEL.md#39-malicious-directory-witness).
 
+A Phase 1 constant has since been corrected the same way:
+[`KT.md` §6.1](./KT.md#61-structure) (2026-08-24) **renames the tree-head digest
+label to `free2z/kt/v1/tree-head-hash`**, because the name it shipped with was a
+proper extension of `free2z/kt/v1/sth` and `H` has no separator, so the two
+domains were not separated at all. `WIRE.md` §1.3 now states prefix-freeness as a
+normative requirement on the **union** of every document's labels, and
+`scripts/check-hash-domain-labels.mjs` enforces it on every pull request
+([#602](https://github.com/free2z/zuu/issues/602)).
+
 Still open above the relay: `KeyPackage` publication, push notifications, padding
 bucket sizes, the redundancy factor *k*, the client gossip protocol, the witness
 independence criterion and the default *t*, and the KT epoch cadence. All are
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -71,6 +71,35 @@ construction #544 selects.
 - `H(label, x)` means `BLAKE2b-256(label || x)` where `label` is the exact ASCII
   bytes shown, with no separator and no terminator — the idiom already used for
   `msg_id` in [`ARCHITECTURE.md` §7](./ARCHITECTURE.md#7-application-framing--hash-linked-causal-ordering).
+- **The label set MUST be prefix-free.** No label used with `H` — in this
+  document, in [`KT.md`](./KT.md), or in any later document that uses `H` — may
+  be a proper prefix of any other such label. The rule extends to labels
+  concatenated the same way without `H`, such as the §5.2 signing prefix
+  `"free2z/relay/v1/hello"`, and to the first-field `label` constants of the
+  signed structures, because those constants and these arguments are drawn from
+  one namespace and a future construction will reuse one for the other.
+
+  This is the precondition of the construction, not a naming preference. With no
+  separator, if label `A` is a proper prefix of label `B` — so that
+  `B == A || s` for some nonempty `s` — then
+
+  ```
+  H(A, s || y) = BLAKE2b-256(A || s || y) = BLAKE2b-256(B || y) = H(B, y)
+  ```
+
+  bit for bit, and the two domains are not separated for any `A`-domain message
+  beginning with `s`.
+
+  The requirement is on the **union** of every document's labels, never on each
+  document's set separately: a set that is internally prefix-free says nothing
+  about its neighbours, and that scoping mistake is how
+  [#602](https://github.com/free2z/zuu/issues/602) put `free2z/kt/v1/sth` and a
+  label extending it into the same namespace. `scripts/check-hash-domain-labels.mjs`
+  asserts the property mechanically over every label in the repository on every
+  pull request, from the always-running `rs / changes` job, and its verdict is
+  re-checked inside the required `rs / gate` context. **Adding a label means
+  running that check**; a label that has not been through it is a label nobody
+  has checked against the rest of the namespace.
 - Structures are written in the **TLS presentation language**
   ([RFC 8446 §3](https://datatracker.ietf.org/doc/html/rfc8446#section-3)); see §3.
 - "The relay" is the server. "The client" is a ZUULI or WASM device
--- a/rs/crates/f2z-codec/src/hash.rs
+++ b/rs/crates/f2z-codec/src/hash.rs
@@ -7,9 +7,17 @@
 //! No separator means the labels themselves have to be non-ambiguous, and they
 //! are: every label below is a full path under `free2z/relay/v1/` and no label
 //! is a prefix of another with a message that could complete it. That is a
-//! property of this fixed set, not of the construction, so a new label must be
-//! checked against the same rule before it is added — [`LABELS`] exists so a
-//! test can do that mechanically.
+//! property of *a set*, not of the construction, so a new label must be checked
+//! against the same rule before it is added.
+//!
+//! **The set that matters is the union, not this array.** `WIRE.md` §1.3 now
+//! states prefix-freeness as a normative requirement over every label in every
+//! document that uses `H`, and `scripts/check-hash-domain-labels.mjs` enforces
+//! it across the whole repository — the specifications under `docs/e2ee/` as
+//! well as these constants. [`LABELS`] and the test below cover this crate's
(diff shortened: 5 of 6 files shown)


skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green, and I reproduced the decisive checks myself.

## The collision is caught, with the algebra in the message

Reintroducing `sth-hash` on the real tree:

```
Hash-domain label separation failed:
- "free2z/kt/v1/sth" is a proper prefix of "free2z/kt/v1/sth-hash":
  H("free2z/kt/v1/sth", "-hash" || y) is bit-identical to H("free2z/kt/v1/sth-hash", y),
  so those two domains are not separated.
  free2z/kt/v1/sth at docs/e2ee/KT.md:502, README.md:122, WIRE.md:96;
  free2z/kt/v1/sth-hash at docs/e2ee/KT.md:508
EXIT=1
```

Naming both sides with file:line and stating *why* it matters is right — the next person to hit this will be looking at a label rename and needs to understand it is a soundness failure, not a style rule.

**A correction on my own testing:** my first probe restored the wrong string (I assumed `sth-digest`) and reported green. That was a mutation that did not land, not a blind guard — the same failure mode I hit twice earlier tonight. Redone against `tree-head-hash`, it goes red.

## The anchor is the part that matters, and it holds

Narrowing `LABEL_BEARING_DOCUMENTS` to `WIRE.md` alone:

```
- docs/e2ee/ARCHITECTURE.md, KT.md, README.md, THREAT-MODEL.md,
  decisions/0009-…, decisions/0010-… mint labels but are not in
  LABEL_BEARING_DOCUMENTS; register them, so that adding a specification is
  an acknowledgement that its labels share one prefix-free namespace with
  every other document's
EXIT=1
```

**Red, not a smaller green.** That is the whole point. `f2z-codec`'s `hash.rs:132` was the right shape at the wrong scope — each document's set was internally consistent and nobody checked the union — and this is the same remedy #597 applied to the `Debug` scan. There is a second anchor too, a floor on the union size so a scan that reaches almost nothing fails rather than passing. 17 self-test cases.

The anchor also fired for real mid-PR: adding label names to `README.md` turned the check red until `README.md` was registered. A guard that has already caught its author is a good sign.

## Two things the issue understated

**The namespace holds 30 labels, not the 12–13 I counted.** I only enumerated `free2z/kt/v1/*`; the union across documents and crates is far larger.

**Six labels have no leaf at all** — `free2z/device-credential/v1`, `free2z/queue-advert/v1`, and four MLS exporter labels. A leafless label is a prefix of **every** leafed label in its namespace, which is the sharpest possible form of this defect. They happen to be prefix-free today, and the check now covers them. I would not have thought to look.

## The rename is the right one

`sth-hash` → `tree-head-hash`, renaming the **longer** label. `free2z/kt/v1/sth` is the first-field wire constant of `SignedTreeHeadTBS` and a row in a closed 9-row table whose other members are bare nouns matching their TBS structs — renaming it to fix a problem the other member caused would break that pattern. And `tree-head-hash` *removes* the prefix relation rather than relocating it, where `sth-tbs` would have left two near-miss siblings. It also spells out what `sth` abbreviates.

## The fixture line is drawn by registration, not heuristic

`FIXTURE_TOKENS` names file + exact token + reason for `cmd2`, `cmX` and `dummy`, keyed by file so the same token elsewhere is treated as a real label, and a stale entry fails. That is the right call: "it is in a test file" would have dropped the six genuine labels `wire_vectors.rs` pins at `:271-276`.

Self-exclusion is load-bearing and demonstrated — with `NON_DECLARING_FILES` emptied the checker reports seven collisions among its own examples.

## Placement and gate binding

`scripts/` rather than a Rust test, for a reason I agree with: `KT.md` has no crate, so a Rust test could only assert over a hand-copied restatement — which is the very drift that caused this. And the collision arrived in a **docs-only PR**, where `rs.yml` sets `rs=false` and every rs job skips, so a Rust test would not have run on the breaking commit.

It reaches the gate two ways: a separate step in `rs.yml`'s `changes` job (no `if:`, every event), and appended to `rs / gate`'s recheck step whose `steps.policy.outcome` the verdict step binds and refuses to pass on — **consumed, not merely awaited**, which is the #512/#515–#518 distinction. Using a separate step rather than editing `zuuli.yml`'s exact-ordered-command list was the safe choice even after confirming that assertion is scoped to `REQUIRED_WORKFLOW_PATH`.

## Convention split, argued

`KT.md` §6.1 gets a **dated correction** — a frozen spec withdrawing a normative constant, so implementers who relied on it are told at the point of use. `WIRE.md` §1.3 is **fixed in place** — silence turned into a MUST is a claim made true, not one withdrawn. That matches #592's rule exactly, and spelling the old label as "`-hash` appended to `free2z/kt/v1/sth`" rather than as a literal keeps a withdrawn label from living on in the namespace. Nice detail.

0 bare `[[ ]]`/`(( ))` added. All other gates pass.

Issue #594, opened 2026-08-24
KT.md: nothing authorizes a handle's FIRST directory entry — anyone can claim any unregistered handle


skyl:
Found while implementing `f2z-authority` (branch `feature/kt-authority`, PR #593). **`KT.md` as merged lets anybody claim any unregistered handle.**

## The hole

`KT.md` §4.4's authorization table has three cases, and `same_key` is authorized by *"the `directory_auth_pk` **published in the previous entry**"*.

At `entry_version == 1` **there is no previous entry.** Rule 3 explicitly permits it — *"`entry_version` is `previous + 1` (or 1)"* — but nothing in the numbered rules, the table, or the surrounding prose says what authorizes a **first** entry for a handle.

So a conforming implementation of `KT.md` as written accepts a first `DirectoryEntry` for `@alice` **from anyone who submits one**. First-come-first-served handle claiming, in the document whose entire purpose is that the server cannot lie about who you are talking to.

## This is decision D-d, unimplemented

The owner decision recorded on #305 was explicit:

> **KT submission authority is a tiny signed-assertion endpoint in Django.** Django authenticates the Knox user and issues a short-lived Ed25519-signed handle-ownership assertion; the Rust KT server only verifies a signature against a configured authority public key, and a self-hoster plugs in their own.

**`grep -rn HandleAssertion docs/` on `main` returns nothing.** The type does not exist in `KT.md`, `WIRE.md`, `ARCHITECTURE.md` or any ADR. The decision was taken and never specified, and #554's brief asked for `DirectoryEntry`, `SignedTreeHead`, the witness protocol and the proof API — it did not name the assertion, so the gap was inherited rather than introduced.

## A second, related gap (from #584)

§4.4's numbered rules also never require a `same_key` entry to keep `identity_pk` **unchanged**. As enumerated, a party holding only the previous `directory_auth_pk` can publish a `same_key` entry carrying a brand-new `identity_pk` — a key change with one signature, which ADR 0014 says the log MUST reject. It is implied by the case *naming* but absent from the list an implementer is told to apply "in order".

Both `f2z-kt-core` (#584) and `f2z-authority` (#593) enforce the stricter reading and mark it as added; the spec should make it a numbered rule so a second implementer is not left to infer it.

## What exists to build on

[...]

PR #611, opened 2026-08-24
docs(e2ee): bind `same_key` to `identity_pk`; name the first-entry authorization hole

skyl:
`KT.md` §4.4's numbered rules are the only thing in the system that proves the log did not *make an entry up* — `akd` enforces none of them. As merged the list was incomplete in two ways, and [#594](https://github.com/free2z/zuu/issues/594) found both while implementing against it.

**This PR does the defect-fix half and deliberately leaves the product-decision half open.** `Refs #594` rather than `Closes`.

## What I did

**1. `same_key` MUST NOT change `identity_pk` — now numbered rule 6.**
The constraint was implied by the case's *name* and by ADR 0014's case 1, and absent from the list an implementer is told to apply "in order". As enumerated, a party holding only the previous entry's `directory_auth_pk` could publish a `same_key` entry carrying a brand-new `identity_pk`: an identity change installed on **one** signature, which ADR 0014 requires the log to reject. `f2z-kt-core` (#584) and `f2z-authority` (#593) both inferred the stricter reading; the spec was the thing that was behind. Fixed in place, with a dated correction saying an implementer working from the first revision must add the check.

**2. The `entry_version == 1` hole is stated in the document, as an open blocking question, with dated-correction treatment.**
Every row of the `auth_signature` table authorizes an entry by reference to state a *previous* entry established. At version 1 there is no previous entry, and rule 3 admits the case explicitly — so the rules as written accept a first `DirectoryEntry` for any unregistered handle from whoever submits one. The correction states the hole, says what an implementer must not assume (in particular: a self-signed first entry authenticates the submitter's *key*, never the submitter's *claim to the handle*), lays out what a decision would have to settle as four questions, and describes `feature/kt-authority`'s `HandleAssertion` **as a candidate and as unratified**.

It is now impossible to read §4.4 and not notice: rule 3 carries the pointer, §1.1 no longer claims the document fixes it, §1.2 lists it first, §2's roles table is scoped, §8.1 step 6 and §8.5 say what a client does *not* get at version 1, §12 lists it, and it is `ARCHITECTURE.md` §13-S.

**3. Mechanical inconsistencies found in the same pass.**
- Rule 4 (`prev_entry_hash` matches the published previous entry) never covered the all-zero case at version 1 that §4.1 and §4.2 both specify.
- §2's roles table said the log "cannot forge an entry (§4.4)", unscoped — false for a first entry.
- §8.1 step 6 and §8.5's "client can verify" column claimed verification of an authorization that does not exist at version 1.
- The document intro and §1.1 said ADR 0014 "settled what authorizes an entry"; ADR 0014 distinguishes same-key update / key change / lost key and takes no position on the first entry.
- `ERR_BAD_AUTHORIZATION`'s example list now names rule 6 alongside rule 7.
- `THREAT-MODEL.md` §3.1's "cannot substitute an identity key undetectably" bullet is scoped to a handle that already has an entry, and that qualifier was never written down — a third dated correction, plus the two counters ("two corrections", "§3.1, twice") and §5's summary row updated so the summary does not restate the claim without its scope. This is the failure mode the scoping convention exists to catch.

## What I deliberately did **not** do

The remaining #594 acceptance bullets are one decision, not four, and it is not the specification's to take:

- `HandleAssertion`'s fields, encoding, signing transcript and validity rules — **not specified**.
- The `expires_ms - issued_ms` cap — **not specified** (it is named as one of the bounds a decision has to settle).
- The identity self-signature requirement — **not made normative**; it is described as part of the candidate's structure, not adopted.
- The no-authority mode and its reporting requirement — **not specified**.

[#551](https://github.com/free2z/zuu/issues/551) is open precisely because ADR 0014's reset authority was invented in a specification under a heading disclaiming owner-decision status. Naming who may vouch for handle ownership is the same weight of decision, so it is stated as an option space with the questions a ruling must answer, and none of them is answered.

## Was item 2 writable without implying an answer?

Yes. The option space is genuinely stateable as constraints on *any* answer — binding to the installed key, bounded validity, what a self-hosted log does, whether clients are told — without picking one. The one place it came close was rule 6's scope, so the correction says explicitly that rule 6 constrains `identity_pk` only and that whether a `same_key` entry may rotate `directory_auth_pk` is not stated in either direction. That is a real second gap, left named rather than closed.

## Verification

- `node scripts/check-hash-domain-labels.mjs` — pass (30 labels, 7 registered documents). No label added or renamed: the correction describes the candidate's signing prefix without minting one.
- `node scripts/check-hash-domain-labels.mjs --self-test` — 17 cases pass.
- `scripts/check-public-repo-boundary.sh` — pass. No production measurement, no absolute counts, no private repo or issue reference; the system described is unimplemented, so nothing here describes a live-product weakness.
- Every relative link and `#anchor` across `docs/e2ee/` resolves, including the new `#13-open-questions`, `#44-what-authorizes-an-entry` and `#31-malicious-or-compromised-free2z-server` targets.
- `WIRE.md` untouched — #588 is open and rewrites §5.5/§6.3/§10.
- `.github/workflows/` untouched. Docs only.
- Dates stamped `2026-08-24`, the **local** date of this commit (`-0400`).

Refs #594, #584, #593, #551.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -1212,6 +1212,23 @@ deliberate.
   portable evidence rather than a local error dialog are all unspecified.
   `KT.md` defines the evidence structures but deliberately does not invent the
   protocol that exchanges them.
+- **S. What authorizes a handle's first directory entry.** Opened 2026-08-24 by
+  [`KT.md` §4.4](./KT.md#44-what-authorizes-an-entry), and **blocking**. §9.2
+  above specifies an append-only log of `(handle → identity key, …)` and does not
+  say who is entitled to a handle; [ADR 0014](./decisions/0014-directory-key-rotation.md)
+  settles authorization for a handle that already has an entry — same-key update,
+  key change, lost key — and takes no position on the first one. `KT.md` §4.4
+  inherited that silence, and as its rules are enumerated they accept a first
+  entry for an unregistered handle from **whoever submits one**. Closing it means
+  deciding who may vouch that a handle belongs to a submitter, whether a log may
+  run with no such authority (a self-hosted log must be able to, under
+  [ADR 0005](./decisions/0005-federation.md)) and whether clients must be told
+  when it does, what binds a vouching artifact to the key being installed, and
+  what a compromise of the vouching key costs. That is a product-and-trust
+  decision of the same weight as ADR 0014's reset authority — which is what
+  [#551](https://github.com/free2z/zuu/issues/551) is open about — so `KT.md`
+  states the option space and deliberately answers none of it.
+  [#594](https://github.com/free2z/zuu/issues/594).
 
 ### 13.1 Closed
 
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -27,7 +27,9 @@ signature, no cadence and no protocol; §9.3 named the witness role and specifie
 no message. [ADR 0013](./decisions/0013-key-transparency-log.md) settled the
 construction — adopt [`akd`](https://github.com/facebook/akd) — and
 [ADR 0014](./decisions/0014-directory-key-rotation.md) settled what authorizes an
-entry. **Everything between those two decisions and an implementation is
+entry **once a handle already has one** — it distinguishes a same-key update, a
+key change and a lost key, and takes no position on a handle's *first* entry
+(§1.2, §4.4). **Everything between those two decisions and an implementation is
 invented below**, and where a choice was invented rather than inherited it says
 so at the point of use. Nothing here should be read as "the merged docs already
 said this."
@@ -38,7 +40,9 @@ said this."
 
 ### 1.1 What this document fixes
 
-The `DirectoryEntry` structure and its authorization; the `SignedTreeHead` format
+The `DirectoryEntry` structure and its authorization **from the second entry
+onward** — the first entry's authorization is §1.2's first bullet and is not
+fixed here; the `SignedTreeHead` format
 and its signing transcript; log-key rotation; epoch cadence and the maximum merge
 delay; the submission receipt; the witness poll-verify-cosign protocol, its
 state-update ordering and its fault evidence; the `WitnessCosignature` format;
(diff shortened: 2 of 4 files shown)


skyl:

skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green.

I sent this worker after two items and told it explicitly not to design a handle-ownership authority. It did both halves right and found five more defects on the way.

## The scoping held, which was the thing I was worried about

`HandleAssertion` appears **once**, described as `feature/kt-authority`'s proposal with *"Nothing about it is specified"* stated outright, and #551 cited as the reason. No field, no encoding, no transcript, no validity rule, no cap, no no-authority semantics. The four remaining acceptance bullets on #594 are correctly identified as **one decision, and the owner's** — which is exactly why I scoped it this way: #551 is open precisely because ADR 0014's reset authority was invented in the specification under a heading disclaiming owner-decision status.

The answer to the question I asked — whether item 2 was writable without implying the out-of-scope answer — is convincing: the option space states as **constraints on any answer** rather than a choice among answers. Four questions any decision must settle (who may vouch and against what record; whether a no-authority mode exists and whether clients are told; what binds vouching to the installed key; what the log-side bounds and compromise cost are), and none of them presupposes a mechanism.

## Rule 6, and the sentence that carries the whole finding

`KT.md:399` — `identity_pk` **MUST equal the previous entry's** — now a numbered rule, cross-referenced from the error table at `:1268`, with later rules renumbered 7-10 and no stale cross-references. Both `f2z-kt-core` and `f2z-authority` already enforced it, so the spec was what was behind.

And the hole is named where it matters. Not just in §4.4, but at `:1016-1018` in the lookup steps:

> **At `entry_version == 1`** … so this step establishes nothing about a handle being [owned]

That is the right placement. A reader arriving at §8.1 would otherwise take inclusion for authorization, which is the whole trap. The load-bearing sentence is `:475`: *a self-signed first entry authenticates the submitter's key, never the submitter's claim to the handle.*

## Five mechanical defects it wasn't sent for

Every one is the same shape — a claim that silently assumes a previous entry exists:

- **Rule 4 never covered version 1.** It requires `prev_entry_hash` to match the published previous entry; §4.1 and §4.2 both say it is all-zero at version 1, and the rule didn't say so.
- **§2's roles table said the log "cannot forge an entry (§4.4)"** — unscoped, and flatly false for a first entry.
- **§8.1 step 6 and §8.5** claimed a client verifies an authorization that does not exist at version 1.
- **The intro and §1.1 said ADR 0014 "settled what authorizes an entry."** It settles same-key update, key change and lost key. It takes no position on the first entry.
- **A second still-open gap, named rather than quietly closed:** nothing says whether a `same_key` entry may rotate `directory_auth_pk`. Rule 6 explicitly does not settle it. Naming an adjacent gap instead of silently resolving it is the right instinct.

## The THREAT-MODEL correction is the honest kind

`THREAT-MODEL.md` §3.1's "cannot substitute an identity key undetectably" is scoped to a handle that **already has an entry**, and the qualifier was never written down. What makes the correction good is that it states the awkward shape rather than smoothing it: the bullet stays *literally* true — a first entry is not a substitution — **while a peer resolving that handle is MITM'd anyway, with no self-audit anywhere to fire.**

A correction that only fixed the letter would have left the reader worse off. This is the third dated correction in §3.1, and the counters and §5's summary row were updated to match, so the document's own bookkeeping stays honest.

## Checks

`check-hash-domain-labels.mjs` green — 30 labels across 7 registered documents — and **no label added or renamed**: the candidate's signing prefix is described without minting one, which is the right call given #602 landed hours ago. `--self-test` 17 cases. Boundary script passes and was judged by hand. All anchors resolve. `WIRE.md` untouched, so no conflict with #588, which is open and rewriting §5.5/§6.3/§10. Dates are `2026-08-24`, the local date of a `-0400` commit — correct per the convention #609 had to re-stamp #573 for.

`Refs #594` rather than `Closes` is right. The authority-model half is still open and still yours.

Issue #603, opened 2026-08-24
security(rs): the inert strict-verification test reappeared in f2z-authority — make the check workspace-wide


skyl:
## The same inert test has now appeared in two crates

#589 found that `f2z-relay-proto`'s `small_order_keys_are_rejected_by_strict_verification` used an **all-zero signature** — which both `verify` and `verify_strict` reject — so it passed whether or not strict verification was in use. #597 fixed it.

`f2z-authority` (#593, branch `feature/kt-authority`) contains the same defect, independently:

- `key.rs:147` `a_small_order_key_verifies_nothing` uses `A = [0x01, 0x00×31]` with an **all-zero signature**.
- Measured in this tree: the real forgery `sig = A || [0u8;32]` is accepted by plain `verify` for **64/64** distinct messages and by `verify_strict` for 0/64.
- Mutation `verify_strict → verify` leaves the **entire suite green**.

**And the end-to-end consequence is worse than in `f2z-relay-proto`.** On an `AuthoritySet::none()` log, `admit` as shipped returns `Err(BadIdentitySignature)`. Under the mutation it returns `Ok(AdmittedHandle)` — **a handle claimed with no private key at all.** `f2z-authority` is the crate that decides who owns `@alice`.

The guard itself is correct. Nothing would notice if it stopped being.

## Why this needs a systemic fix, not a second one-off

Two crates, written separately, reproduced the same mistake — because the mistake is natural. An all-zero signature *looks* like the degenerate case you would test, and it *does* fail, so the test goes green and reads as correct. Fixing `f2z-authority` alone leaves the third crate to rediscover it.

`f2z-codec` sets the precedent for the fix shape: `hash.rs:132` asserts prefix-freeness **mechanically over a `LABELS` array**, so a new label cannot be added without the check running. And #597 established the anti-narrowing pattern — `workspace_debug_scan.rs` walks every crate under `rs/crates/` with a **coverage anchor** asserting every crate was reached, precisely so it cannot silently narrow the way it did when `f2z-relay-proto` landed outside it.

The same treatment applies here.

## Fix

1. **Fix `f2z-authority`'s fixture** the way #597 fixed `f2z-relay-proto`'s: use a signature plain `verify` **accepts**, assert `verify_strict` rejects it *and* that non-strict would have accepted — so removing strictness turns it red.
[...]

PR #612, opened 2026-08-24
security(rs): fail the build if any crate verifies Ed25519 non-strictly

skyl:
Refs #603. This is the **workspace-wide half** only — #603's other half is
`f2z-authority`'s own fixture, and `f2z-authority` lives on the unmerged
`feature/kt-authority` branch, so it cannot be fixed from `main`. The point of
landing this half first is that when that crate arrives, its defect is caught
**on arrival** rather than rediscovered a third time.

`rs/crates/f2z-codec/tests/workspace_strict_verification_scan.rs`, three tests,
plus `tests/common/mod.rs` — the crate walk, now shared with
`workspace_debug_scan.rs` instead of copied.

## 1. Green on unmodified `main`, and the three correct call sites are not flagged

A naive grep for `.verify(` produces three hits on `main` today, and **all
three are correct** — each calls `f2z_relay_proto::key::VerifyingKey::verify`,
the crate's own wrapper, which calls `verify_strict` inside (`key.rs:72`):

```
rs/crates/f2z-relay-proto/src/hello.rs:270          identity.verify(…)
rs/crates/f2z-relay-proto/src/capabilities.rs:206   key.verify(…)
rs/crates/f2z-relay-proto/src/command.rs:679        verifying_key.verify(…)
```

A check that cries wolf on correct code gets silenced, and then it protects
nothing — so "these three are not flagged" is a **test**, not an observation:
`the_crates_own_strict_wrapper_is_not_mistaken_for_ed25519_dalek` registers
each one, asserts the call text is still in the file (so the control cannot go
vacuous by the call being deleted) and asserts the scan reports **zero**
findings for it.

```
$ cd rs && cargo test --locked --all-targets
test result: ok. 49 passed   # f2z-codec lib
test result: ok. 7  passed   # padding
test result: ok. 9  passed   # f2z-codec redaction
test result: ok. 7  passed   # reencode_equality
test result: ok. 42 passed   # wire_vectors
test result: ok. 1  passed   # workspace_debug_scan
test result: ok. 3  passed   # workspace_strict_verification_scan   <- new
test result: ok. 83 passed   # f2z-relay-proto lib
test result: ok. 6  passed   # properties
test result: ok. 8  passed   # f2z-relay-proto redaction

$ cargo test --locked --test workspace_strict_verification_scan
test every_verify_function_in_the_workspace_is_registered_as_strict ... ok
test the_crates_own_strict_wrapper_is_not_mistaken_for_ed25519_dalek ... ok
test no_crate_verifies_ed25519_signatures_non_strictly ... ok
```

`--all-targets` 212 → **215** (+3). `--doc` 2 → **2**.

## 2. Where the wrapper/`ed25519_dalek` line is drawn, and why there

The scan never matches `.verify(` as a spelling. Three rules, and each rests on
a fact about the dependency rather than on a naming convention.

**Rule 1 — the trait gate.** `ed25519_dalek::VerifyingKey` has **no inherent
`verify`**. The method exists only through
`impl Verifier<ed25519::Signature> for VerifyingKey`
(`ed25519-dalek-3.0.0/src/verifying.rs:567`), and Rust requires a trait to be
in scope to call its methods — through a `.` receiver *or* a `Type::method`
path. `verify_strict` is inherent (`verifying.rs:367`); `verify` is not. So a
file that never names a `Verifier` trait **cannot** call plain Ed25519
verification, whatever it calls its variables.

The rule therefore flags the *name* — `Verifier`, `MultipartVerifier`,
`DigestVerifier`, `StreamVerifier`, matched as whole identifiers so
`CommandVerifier` is not a hit — anywhere in the file, not just in a `use`.
Flagging the name rather than the import is what covers the fully-qualified
`<K as ed25519_dalek::Verifier<S>>::verify(…)` form, which needs no import at
all. A glob import of any module whose path mentions `ed25519` or `signature`
is flagged too, since that is the one way the trait can arrive unnamed.

**Rule 2 — the entry points that need no trait.** `verify_prehashed`,
`verify_stream`, `verify_batch`, `multipart_verify`, `verify_digest`,
`raw_verify`. None screens small-order points and none needs an import.
Matched as whole identifiers *followed by a call*, so `verify_prehashed_strict`
is not mistaken for `verify_prehashed`. Most sit behind `digest`/`hazmat`/
`batch`, which this workspace disables — and which feature unification can turn
back on without anybody editing a manifest here.

**Rule 3 — the receiver rule.** A `let` binding or a struct field whose type is
`ed25519_dalek::VerifyingKey` (one level of alias resolved, as
`workspace_debug_scan.rs` does for `type Blob = Vec<u8>`), followed by
`.verify(` on it. This is what turns a violation into a **line number**.

It reads method chains, which matters more than it sounds: `rustfmt` writes the
wrapper as

```rust
self.0
    .verify_strict(message, &signature)
```

so the text `self.0.verify(` never appears even when the call *is*
`self.0.verify(…)`. A physical-line matcher would have missed the exact
mutation #589 demonstrated. (It did, in the first draft — see §4.)

**And the wrapper is not assumed safe.** `WORKSPACE_VERIFY_FNS` registers every
`fn verify…` defined under `rs/crates/*/src/` with what makes it strict, and
checks the registration against the source: `key.rs::verify` is
`CallsVerifyStrict` and its body must contain `verify_strict`; the other three
are `DelegatesTo` and must contain a call to it. An **unregistered** `fn verify…`
fails by name, and a registration whose function no longer exists fails too —
the #609 shape, where narrowing the list is louder than leaving it alone.

That registry is why the three grep hits in §1 are provably fine rather than
taken on trust: they call a function this test holds to `verify_strict`.

## 3. Red when a violation is introduced, in another crate

The test is hosted in `f2z-codec`; the violation goes in **`f2z-relay-proto`**.
Real, compiling code (`cargo build --locked -p f2z-relay-proto` succeeds), added
to `capabilities.rs` — *not* `key.rs`, which is where the registered fixture
lives:

```rust
pub fn verify_legacy_capabilities(signed: &SignedCapabilities, identity: &PublicKey) -> Result<()> {
    use ed25519_dalek::Verifier as _;
    let raw = ed25519_dalek::VerifyingKey::from_bytes(identity.as_bytes())…?;
    let sig = ed25519_dalek::Signature::from_bytes(signed.signature.as_bytes());
    raw.verify(&signing_bytes(&signed.capabilities)?, &sig)…
}
```

All three tests go red, naming file and line:

```
---- no_crate_verifies_ed25519_signatures_non_strictly ----
non-strict Ed25519 verification in the rs/ workspace. …
  f2z-relay-proto/src/capabilities.rs:215: names the `Verifier` trait, which is the only
    thing that makes `ed25519_dalek`'s non-strict verification callable. … — use ed25519_dalek::Verifier as _;
  f2z-relay-proto/src/capabilities.rs:219: calls `.verify(` on `raw`, which this file declares
    as an `ed25519_dalek::VerifyingKey`; that is the non-strict function. Use `verify_strict`
    — raw.verify(&signing_bytes(&signed.capabilities)?, &sig)

---- every_verify_function_in_the_workspace_is_registered_as_strict ----
these verification functions are not registered in WORKSPACE_VERIFY_FNS, so nothing says
what makes them strict:
  f2z-relay-proto/src/capabilities.rs:211: fn verify_legacy_capabilities
```

### And red on the mutation #589/#603 are actually about

`key.rs:72`, `verify_strict` → `verify` — the landed line printed back before
the run was judged:

```
BEFORE line 72: '            .verify_strict(message, &signature)'
AFTER  line 72: '            .verify(message, &signature)'
```

Two independent rules fire, from a different crate:

```
---- no_crate_verifies_ed25519_signatures_non_strictly ----
  f2z-relay-proto/src/key.rs:72: calls `.verify(` on `self.0`, which this file declares as an
    `ed25519_dalek::VerifyingKey`; that is the non-strict function. Use `verify_strict`

---- every_verify_function_in_the_workspace_is_registered_as_strict ----
f2z-relay-proto/src/key.rs::verify is registered as calling `verify_strict` and its body no
longer does. Every other verification in this workspace delegates to it, so this is a universal
forgery away from `admit`-style authorization returning Ok for a key nobody holds.
```

### The escape hatch cannot be moved into shipped code

`REGISTERED_NON_STRICT` has two entries, both the #597 universal-forgery
fixture, which *must* call plain `verify` to assert it **accepts** the
identity-point forgery. Each entry must still match something — a stale entry
fails — and each must sit inside a `#[cfg(test)]` item. Hoisting the registered
`use ed25519_dalek::Verifier as _;` from the test module (line 254) to module
scope (line 27):

```
f2z-relay-proto/src/key.rs:27 is registered in REGISTERED_NON_STRICT but no longer sits inside
a #[cfg(test)] item. A deliberate non-strict verification is only defensible as a fixture;
this one now compiles into the crate.
```

Those two entries are also the scan's **detection anchor**: if either stops
matching, the scan fails rather than passing with nothing found.

## 4. The anchor: red on the anchor, not green with a smaller count

`workspace_debug_scan.rs`'s walk and `workspace_crates()` now live in
`tests/common/mod.rs` and both scans assert against **one** list. That is the
same argument that file makes about copies of a scanner, applied at file scale:
the anchor's promise is that the crate list cannot silently narrow, and two
lists can narrow independently.

**Narrowing the walk** (`source_files()` filtered to `f2z-codec`, list left
intact) — this is #589 item 3's original defect exactly:

```
---- no_crate_verifies_ed25519_signatures_non_strictly ----
assertion `left == right` failed: the scan did not reach every crate under rs/crates/. A crate
outside it is a crate where a non-strict Ed25519 verification goes unnoticed — which is #603 exactly.
  left: ["f2z-codec"]
 right: ["f2z-codec", "f2z-relay-proto"]

---- every_verify_function_in_the_workspace_is_registered_as_strict ----
assertion `left == right` failed: the registry census did not reach every crate under rs/crates/…
  left: ["f2z-codec"]
 right: ["f2z-codec", "f2z-relay-proto"]

---- the_crates_own_strict_wrapper_is_not_mistaken_for_ed25519_dalek ----
f2z-relay-proto/src/hello.rs is a registered safe call site but was not walked

test result: FAILED. 0 passed; 3 failed
```

`workspace_debug_scan.rs` goes red on the same mutation, from the same shared
list. **Narrowing the list itself** hits the floor guard instead — also red,
never a smaller green:

```
found 1 crate(s) under rs/crates/, which is too few to be real: ["f2z-codec"]
```

## 5. Trying to defeat it — what it catches, and what it does not

Three attempts, all compiled (`cargo build --locked -p f2z-relay-proto`), all
in one module, **all caught**:

| attempt | caught by |
|---|---|
| `use ed25519_dalek::VerifyingKey as VK;` + `use ed25519_dalek::Verifier as Checker;` + `Checker::verify(&k, …)` | rule 1, at the import — the alias is written *after* the name |
| `<ed25519_dalek::VerifyingKey as ed25519_dalek::Verifier<…>>::verify(…)` with **no import at all** | rule 1, at the call — the name is on the line |
| `use ed25519_dalek::ed25519::signature::*;` then `k.verify(…)` — a re-export path no fixed list would contain | rule 1b (any `::*` whose path mentions `ed25519`/`signature`) **and** rule 3 on `k` |

```
f2z-relay-proto/src/capabilities.rs:214: names the `Verifier` trait … — use ed25519_dalek::Verifier as Checker;
f2z-relay-proto/src/capabilities.rs:227: names the `Verifier` trait … — <ed25519_dalek::VerifyingKey as ed25519_dalek::Verifier<ed25519_dalek::Signature>>::verify(
f2z-relay-proto/src/capabilities.rs:237: glob-imports a module that re-exports the `Verifier` trait … — use ed25519_dalek::ed25519::signature::*;
f2z-relay-proto/src/capabilities.rs:243: calls `.verify(` on `k`, which this file declares as an `ed25519_dalek::VerifyingKey` …
```

### What it does not catch — demonstrated, not guessed

- **Anything outside a crate's `src/`.** Not a claim, a measurement: adding
  `use ed25519_dalek::Verifier as _;` and `key.verify(b"anything", &sig)` to
  `f2z-relay-proto/tests/properties.rs` leaves the scan **green**. This is a
  deliberate boundary — `tests/`, `benches/`, `examples/` and `build.rs` are
  not compiled into the library a relay links, so a non-strict call there
  cannot reach `CommandVerifier::verify` or an `admit`-style authorization
  boundary, and the inert-*fixture* half of #603 is a different check with a
  different remedy. Widening the walk would also mean widening the exemption
  rule, since an integration test's fixture is not inside a `#[cfg(test)]` item.
  It is stated in the file's own module note.
- **A third-party crate re-exporting the trait under a different name.**
  `use some_dep::Checker;` where `some_dep` did `pub use ed25519_dalek::Verifier
  as Checker;` — then no file *here* ever writes `Verifier`. A re-export
  *within* this workspace is caught, because the re-export line names it.
- **Another Ed25519 implementation entirely.** The rules name `ed25519_dalek`.
  It is the workspace's only signature dependency and `rs/Cargo.lock` is
  committed, so adding another is a change a reviewer sees — it is not a change
  this scan reports.
- **A macro that generates the call.** Nothing here expands macros. In practice
  `macro_rules!` needs the trait in scope at the *expansion* site and a
  fully-qualified path inside the macro still writes `Verifier`, so both are
  caught today; a proc macro is not, and that is not something a text scan can
  promise.
- **A receiver typed through a function return** (`let k = decode(b); k.verify(…)`)
  is not resolved by rule 3 — rule 1 still catches it, which is why the gate is
  on the trait and not only on the call.

## Also run

```
scripts/check-rust-fmt.sh --root rs                 All 3 Rust crate(s) under rs/ are formatted
scripts/check-rust-clippy.sh --root rs              All 3 Rust crate(s) under rs/ are clippy-clean
scripts/check-rust-deny.sh --root rs --config rs/deny.toml
scripts/check-rust-toolchain.sh                     channel 1.97.1, MSRV 1.97
scripts/check-rust-toolchain.sh --self-test         self-test: 14 drift case(s) caught
node scripts/check-workflow-gates.mjs               2 gate(s), 56 job(s), 14 workflow file(s)
node scripts/check-hash-domain-labels.mjs           30 label(s) across 7 registered document(s)
```

`clippy::panic` is allowed in the new test file, beside the `unwrap_used` /
`indexing_slicing` / `arithmetic_side_effects` allowances
`workspace_debug_scan.rs` already carries and for the same reason: this is host
test code read by a person, not the relay's parser.

The toolchain **registration census** (#600) is unaffected — this adds test
files to an existing crate, no manifest and no `rust-toolchain.toml`.
`rs/deny.toml`'s `akd` / `akd_core` `<0.13.0` bans are **untouched**, as is
`.github/workflows/zuuli.yml` (`git diff origin/main` over both: empty).

**#477:** bare `[[ ]]` / `(( ))` statements added: **0** — no shell script is
touched by this change.

### DIFF
--- a/rs/crates/f2z-codec/tests/common/mod.rs
+++ b/rs/crates/f2z-codec/tests/common/mod.rs
@@ -1,0 +1,269 @@
+//! The source walk shared by this crate's workspace-wide source scans.
+//!
+//! # Why this is a module and not a copy in each scan
+//!
+//! `workspace_debug_scan.rs` exists because #572's `Debug` scan resolved its
+//! root from `CARGO_MANIFEST_DIR` and so covered one crate. The fix was a walk
+//! over every crate under `rs/crates/` plus a **coverage anchor** asserting
+//! every crate was reached. When #603 asked for a second workspace-wide scan,
+//! the obvious move was to copy that walk — which is the thing
+//! `workspace_debug_scan.rs`'s own module note rejects in a different guise:
+//! "*n* copies of a scanner drift, and the copy that matters is the one in the
+//! crate whose author did not know to add it".
+//!
+//! Two copies of the enumeration is exactly that failure at file scale: the
+//! anchor's promise is that *the list of crates cannot silently narrow*, and
+//! two lists can narrow independently. So there is one list, here, and both
+//! scans assert against it.
+//!
+//! Cargo does not build `tests/common/mod.rs` as its own test target — only
+//! top-level files in `tests/` become targets — so this is a plain module each
+//! scan declares with `mod common;`.
+
+// Each scan uses a different subset of these helpers, and every integration
+// test is its own crate, so what one does not call is dead code there.
+#![allow(dead_code)]
+// Test code, run on the host by a person reading the failure. The workspace
+// denies these because a panic in the relay's parser is a remote denial of
+// service; neither hazard exists here.
+#![allow(
+    clippy::unwrap_used,
+    clippy::indexing_slicing,
+    clippy::arithmetic_side_effects
+)]
+
+/// One `.rs` file under some workspace crate's `src/`.
+pub struct SourceFile {
+    /// The crate directory name under `rs/crates/`.
+    pub krate: String,
+    /// `<crate>/<path within the crate>`, e.g. `f2z-relay-proto/src/key.rs`.
+    /// This is what a violation names, so it is the path a reader can open.
+    pub label: String,
+    /// The file's contents.
+    pub source: String,
+}
+
+/// `rs/crates/`, resolved from this crate's manifest rather than from the
+/// process CWD so the tests do not depend on where cargo was invoked.
+pub fn crates_root() -> std::path::PathBuf {
+    let manifest = std::path::Path::new(env!("CARGO_MANIFEST_DIR"));
+    let root = manifest.parent().unwrap().to_path_buf();
+    assert!(
+        root.join("f2z-codec").join("Cargo.toml").is_file(),
+        "{root:?} is not rs/crates/; this test's idea of the workspace layout is stale"
+    );
+    root
+}
+
+/// Every crate directory under `rs/crates/`, by name.
+///
+/// This is the list the scans must cover, and it is read off the filesystem
+/// rather than written down here — a crate added tomorrow appears in it
+/// without anyone editing this file, which is the whole point. `rs/Cargo.toml`
+/// declares `members = ["crates/*"]`, so the glob and this walk agree by
+/// construction.
+pub fn workspace_crates() -> Vec<String> {
+    let mut names: Vec<String> = std::fs::read_dir(crates_root())
+        .unwrap()
+        .map(|entry| entry.unwrap().path())
+        .filter(|path| path.join("Cargo.toml").is_file())
+        .map(|path| path.file_name().unwrap().to_string_lossy().into_owned())
+        .collect();
+    names.sort();
+    assert!(
+        names.len() >= 2,
+        "found {} crate(s) under rs/crates/, which is too few to be real: {names:?}",
+        names.len()
+    );
+    names
+}
+
+/// Every `.rs` file under every workspace crate's `src/`, labelled
+/// `<crate>/<path>` so a violation names the crate it is in.
+pub fn source_files() -> Vec<SourceFile> {
+    fn walk(
+        krate: &str,
+        crate_dir: &std::path::Path,
+        dir: &std::path::Path,
+        out: &mut Vec<SourceFile>,
+    ) {
+        let read = std::fs::read_dir(dir);
+        assert!(read.is_ok(), "could not read {dir:?}");
+        let entries = read.unwrap();
+        let mut paths: Vec<std::path::PathBuf> =
+            entries.map(|entry| entry.unwrap().path()).collect();
+        // `read_dir` order is filesystem order. Sorting makes a failure list
+        // reproducible between a developer's machine and CI.
+        paths.sort();
+        for path in paths {
+            if path.is_dir() {
+                walk(krate, crate_dir, &path, out);
+            } else if path.extension().is_some_and(|ext| ext == "rs") {
+                let within = path.strip_prefix(crate_dir).unwrap();
+                out.push(SourceFile {
+                    krate: krate.to_owned(),
+                    label: format!("{krate}/{}", within.to_string_lossy().replace('\\', "/")),
+                    source: std::fs::read_to_string(&path).unwrap(),
+                });
+            }
+        }
+    }
+    let root = crates_root();
+    let mut out = Vec::new();
+    for krate in workspace_crates() {
+        let crate_dir = root.join(&krate);
+        let src = crate_dir.join("src");
+        assert!(
+            src.is_dir(),
+            "{krate} has a Cargo.toml but no src/, so the scan cannot see it"
+        );
+        walk(&krate, &crate_dir, &src, &mut out);
+    }
+    assert!(!out.is_empty(), "found no source files to scan");
+    out
+}
+
+/// Strip a line comment, so a brace inside a doc comment does not confuse the
+/// depth count. Struct bodies contain no string literals, so a naive cut is
+/// safe there and a real parser is not worth the dependency; function bodies
+/// do, which is what [`code_only`] is for.
+pub fn without_comment(line: &str) -> &str {
+    match line.find("//") {
+        Some(index) => &line[..index],
+        None => line,
+    }
+}
+
+/// A line with its string literals, character literals and line comment
+/// removed, so that neither a `{` inside a `format!` nor a `.verify(` inside a
+/// doc comment is read as code.
+///
+/// This is the stricter cousin of [`without_comment`], needed wherever a scan
+/// looks at *function* bodies. It is a lexer, not a parser: a raw string with
+/// an embedded quote (`r#"..."#`) is not modelled, and a block comment is not
+/// modelled. Neither appears in the tree today, and both would make the scan
+/// noisier rather than blinder — a `//`-free `/* ... */` mentioning
+/// `.verify(` would be reported, and the report names a file and a line for a
+/// person to read.
+pub fn code_only(line: &str) -> String {
+    let bytes: Vec<char> = line.chars().collect();
+    let mut out = String::with_capacity(line.len());
+    let mut index = 0usize;
+    while index < bytes.len() {
+        let current = bytes[index];
+        if current == '/' && bytes.get(index + 1) == Some(&'/') {
+            break;
+        }
+        if current == '"' {
+            index += 1;
+            while index < bytes.len() {
+                if bytes[index] == '\\' {
+                    index += 2;
+                    continue;
+                }
+                if bytes[index] == '"' {
+                    index += 1;
+                    break;
+                }
+                index += 1;
+            }
+            // Leave a placeholder so `"a".len()` does not become `.len()`
+            // glued to whatever preceded the literal.
+            out.push('_');
+            continue;
+        }
+        // A character literal, distinguished from a lifetime by the closing
+        // quote: `'a'` and `'\n'` are literals, `'a` in `&'a str` is not.
+        if current == '\'' {
+            let escaped = bytes.get(index + 1) == Some(&'\\');
+            let closing = if escaped { index + 3 } else { index + 2 };
+            if bytes.get(closing) == Some(&'\'') {
+                out.push('_');
+                index = closing + 1;
+                continue;
+            }
+        }
+        out.push(current);
+        index += 1;
+    }
+    out
+}
+
+/// True when `haystack` contains `needle` as a whole identifier — not as part
+/// of a longer one. `CommandVerifier` does not contain the identifier
+/// `Verifier`, and `verify_strict` does not contain `verify`.
+pub fn contains_identifier(haystack: &str, needle: &str) -> bool {
+    identifier_positions(haystack, needle).next().is_some()
+}
+
+/// Byte offsets at which `needle` appears in `haystack` as a whole identifier.
+pub fn identifier_positions<'a>(
+    haystack: &'a str,
+    needle: &'a str,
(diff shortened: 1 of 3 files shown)


skyl:

skyl:

skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Merging** — both required contexts green on the latest runs, and I reproduced the decisive mutations myself.

## The line between the wrapper and `ed25519_dalek` is drawn on a fact, not a heuristic

This was the hard part of the brief, and the answer is better than what I asked for:

> `ed25519_dalek::VerifyingKey` has **no inherent `verify`**. It comes only from `impl Verifier<ed25519::Signature> for VerifyingKey` (`verifying.rs:567`), while `verify_strict` **is** inherent (`:367`).

So gating on the **trait identifier** is not a guess about naming — it is the only way the non-strict call can exist at all. Rust needs the trait in scope, by `.` receiver or `Type::method` path alike. Flagging the identifier rather than the `use` line is what also covers `<K as ed25519_dalek::Verifier<S>>::verify(…)`, which needs no import. Matching it as a whole identifier keeps `CommandVerifier` from tripping it.

`WORKSPACE_VERIFY_FNS` is the piece that makes the three grep-visible `.verify(` sites **provably** fine rather than trusted: every `fn verify…` under `rs/crates/*/src/` must be registered with what makes it strict, `key.rs::verify` is checked to still contain `verify_strict`, and an unregistered one goes red by name.

## Verified myself

```
baseline (unmodified tree)          EXIT=0   3 passed
  every_verify_function_in_the_workspace_is_registered_as_strict ... ok
  the_crates_own_strict_wrapper_is_not_mistaken_for_ed25519_dalek ... ok
  no_crate_verifies_ed25519_signatures_non_strictly              ... ok

MUT 1  key.rs:72  .verify_strict( -> .verify(     [line printed back before running]
  -> EXIT=101  every_verify_function_in_the_workspace_is_registered_as_strict FAILED

MUT 2  narrow the crate walk to one crate         [changed lines printed back]
  -> EXIT=101  found 1 crate(s) under rs/crates/, which is too few to be real: ["f2z-codec"]
```

The middle test is the negative control I most wanted: it asserts the three safe sites are **not** flagged *and* that the call text still exists, so the control cannot pass by the code having quietly moved.

**My own fifth mutation-didn't-land tonight:** my first anchor probe reported `PATTERN NOT FOUND` and `EXIT=0`. I nearly had a green there. Redone against the real function with the changed lines printed back, it fires correctly.

## The shared walk is an improvement on #597

`workspace_debug_scan.rs` now imports the same `common` module rather than carrying its own copy, so **the two coverage anchors cannot narrow independently**. That is strictly better than what #597 landed, where the Debug scan had its own walk — two anchors that can drift are one anchor.

And `workspace_crates()` reads off the filesystem rather than from a list, with a floor assert. A crate added tomorrow is covered without anyone editing the file, which removes the "forgot to register it" failure mode entirely — the thing #553 and #602 both had to fix by hand elsewhere.

## What defeats it — measured, not guessed

Caught, all compiled and tried: alias import (`Verifier as Checker`), aliased type (`VerifyingKey as VK`), fully-qualified call with no import, glob through the unlisted re-export path.

Not caught, and stated:

- **Anything outside a crate's `src/`.** A working non-strict verify placed in `f2z-relay-proto/tests/properties.rs` leaves the scan green. Deliberate — those targets aren't linked into the library — and documented in the file. Worth knowing rather than discovering.
- A third-party crate re-exporting the trait under a different name (an in-workspace re-export **is** caught, at the re-export line).
- A different Ed25519 implementation; a proc macro.

Reporting the boundary as a measured result rather than an assumed one is the right standard, and it is what makes the "caught" list trustworthy.

## Also worth noting

The worker found a bug in its own first draft: `rustfmt` splits `self.0\n.verify_strict(…)` across lines, so a physical-line matcher missed the very mutation the check exists for. Chain joining added and re-verified. That is exactly the failure this repo keeps producing — a guard that looks right and cannot fire — caught before it shipped.

212 → 215 tests. `rs/deny.toml` and `zuuli.yml` untouched. 0 bare `[[ ]]`/`(( ))`.

`Refs #603` rather than `Closes` is correct: the `f2z-authority` fixture half needs that crate to land, and this check will catch it the moment it does — which was the point of building this half first.

Issue #591, opened 2026-08-24
ci: restore the standalone workflow-gates step and the comment explaining why it exists


skyl:
## Problem

`.github/workflows/zuuli.yml` lost a step and, with it, the paragraph explaining why the step exists.

#560 (`241b714`) added a standalone step in the `changes` job:

```yaml
      # A gate that awaits a job and never inspects its result is decorative
      # for that job: GitHub waits, the gate goes green whatever the job
      # concluded, and branch protection reports a passing required check over a
      # failed build. This runs in the always-required job, over every workflow
      # in the repository, so that hazard cannot be reintroduced anywhere.
      - name: Verify every gate inspects every job it awaits
        run: |
          node scripts/check-workflow-gates.mjs --self-test
          node scripts/check-workflow-gates.mjs
```

#556 folds those two commands into the `Verify immutable actions and fail-closed required jobs` step (taking its asserted command count from 2 to 6) and **deletes the standalone step, comment and all**.

## No enforcement is lost — that is why this is small

Both invocations still run, still in the `changes` job, still gate-bound through `needs.changes.result`. I ran the full battery on the merged tree and everything passes:

```
EXIT=0  check-github-actions-pins.mjs --self-test   70 source-policy, 7 gate-verdict, 120 current-workflow
EXIT=0  check-workflow-gates.mjs                    …no two workflows publish one status-check context…
EXIT=0  check-workflow-gates.mjs --self-test        20 case(s)
EXIT=0  check-librustzcash-compat.mjs --self-test   killed 59 compatibility mutants
EXIT=0  check-zuuli-android-gate.mjs --self-test    31 mutations
```

So this is not a regression in what CI checks. It is a regression in what a reader can find out.

## Why the comment is worth restoring rather than shrugging at

That paragraph is the one-paragraph statement of the hazard behind #512, #515, #516, #517 and #518 — *"branch protection reports a passing required check over a failed build"* — and it sat directly above the check that prevents it.

This repo's entire operating thesis is that a guard whose reason is not written next to it gets deleted by the next person tidying up. Tonight alone: #571's `[[bans.deny]]` entry carries a comment saying "do not assume tooling is holding it" precisely so a future reviewer doesn't delete it as redundant with `cargo audit`; #579's env guard ends "Do not simplify this env away"; #580 records the source-of-truth API call next to `PROTECTED_CONTEXTS`. Keeping the guard and deleting the reason is the wrong half to lose, and it is the half that fails silently.

[...]

PR #614, opened 2026-08-24
ci: gate Windows plugin test execution, and make an ungated workflow fail the checker

### LINKED ISSUE #610: ci: no required check can fail on Windows test execution, and check-workflow-gates cannot see a workflow that has no gate
## Problem

Windows code is tested but **no required check can fail on it**.

`custody / windows-latest` runs the plugin test suite on Windows — including, since #598 (`closes #546`), a test that genuinely exercises `windows_file_identity` rather than skipping under `cfg(unix)`. It passes.

But that job lives in **`.github/workflows/zuuallet.yml`, which has no `gate` job at all** and is not a required status check. Branch protection requires exactly `["gate", "rs / gate"]`. The only *gated* Windows job is `Rust / native lints (windows)`, which runs clippy and never runs tests.

So gated Windows **execution** is zero. The Win32 identity helper — the thing #546 existed to make testable — can regress and nothing required will notice.

## This is pre-existing and #598 improved it

Worth stating plainly so this is not read as a regression: the test #598 *deleted* executed no Win32 code either, so the situation is strictly better than before. #598 is the reason the gap is now visible — before it, there was nothing to gate.

## `check-workflow-gates.mjs` structurally cannot catch this

The checker added in #569 and hardened in #580 enforces that **a gate awaits and consumes every job in its file**. That is coverage *within* a gated workflow.

`zuuallet.yml` has no gate, so it has no `needs` set to compare against and no results to bind — there is nothing for the checker to find fault with. A workflow with zero gates is invisible to a checker that reasons about gates.

That is the same shape as the two scoping gaps already fixed this session: the `Debug` scan resolved its root from `CARGO_MANIFEST_DIR` so a new crate sat outside it (#589, fixed in #597 with a coverage anchor), and `f2z-codec`'s prefix-freeness assertion covered only its own `LABELS` array so `KT.md`'s labels were never checked against it (#602, fixed in #609 with a registration anchor over the union). Both were fixed by making the check enumerate the whole population and fail if the population shrinks.

The same remedy applies: enumerate every workflow file, and require that each either has a gate feeding a required context **or** is explicitly registered as not gate-bearing with a reason. `check-hash-domain-labels.mjs`'s `LABEL_BEARING_DOCUMENTS` is a good model — registration rather than heuristic, and a stale entry fails.

Not every workflow should be gated (`cache-cleanup.yml` plainly should not), so the registration list is the honest mechanism, not a blanket rule.

## Two ways to close the execution gap

1. **Give `zuuallet.yml` a gate** and add its context to branch protection. That makes every job in it gated, and `check-workflow-gates.mjs` starts enforcing coverage within it for free.
2. **Move the windows-latest plugin tests under `zuuli.yml`'s gate.** Narrower, and it keeps the required-context list at two — but it splits the custody suite across workflows.

I'd lean to (1), because it also brings `custody / macos-latest` and the rest of `zuuallet.yml` under a gate, and because the checker gap in the section above is worth closing regardless — and once it is closed, an ungated `zuuallet.yml` fails the check anyway.

## A nit in the same file

`source_content_or_identity_change_after_copy_is_rejected` only ever varies **content**, and stays green under the identity bypass #598's new test kills. Its name advertises identity coverage it does not have. Either extend it to vary identity, or rename it to say what it tests. This is the mild form of the pattern #546 was filed about, in the same file.

## Acceptance

- A required status check can fail on Windows test execution.
- `check-workflow-gates.mjs` fails when a workflow has no gate and is not registered as deliberately ungated, with a reason recorded per entry — demonstrated red before the registration and green after.
- `source_content_or_identity_change_after_copy_is_rejected` either tests identity or stops claiming to.

Found while reviewing #598 (merged). Related: #546, #567, #568, #589, #602.

skyl:
Closes #610.

## The red before

`check-workflow-gates.mjs` grew rule 6 — every workflow file must either publish a
gate resolving to a required context, or be registered in `UNGATED_WORKFLOWS` with
a recorded reason. Run against **unmodified `origin/main`** with the registry
empty, the rule names twelve workflows, not one:

```
Required-gate wiring failed:
- .github/workflows/cache-cleanup.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/docs-about-free2z.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/ts-react-free2z.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/ts-svelte-free2z.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuallet.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-linux-image.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-packaging.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-release.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-store-audit.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-store-publish.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-testflight-bootstrap.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
- .github/workflows/zuuli-testflight-recovery.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS
12 failure(s) across 2 gate(s) in 14 workflow file(s).
```

(The message is elided after the first clause here; the full text names the fix.)

Twelve of fourteen workflows had no gate, and the checker had nothing to say about
any of them — including four that run on `pull_request` and build real artifacts:
`zuuli-packaging.yml`, `zuuli-linux-image.yml`, `ts-react-free2z.yml`,
`ts-svelte-free2z.yml`. Not all of them *should* be gated, which is why this is a
registration and not a blanket rule. But "nothing to complain about" and "nothing
is checked" were producing the same green, and that is the #610 defect stated in
general.

Green after: `2 gate(s), 12 registered ungated, 56 job(s), 14 workflow file(s).`

## The execution gap: option 2, and why

The issue offered two ways to make a required check able to fail on Windows *test
execution*. I took the second — moving the plugin's native suite under
`zuuli.yml`'s existing gate — after working through what option 1 actually costs.

**Option 1 is not a one-line branch-protection edit.** `zuuallet.yml` is
`on: pull_request:` **with a `paths:` filter**. GitHub does not run a
path-filtered workflow for a pull request that matches nothing, and a required
context from a workflow that did not run *never reports* — every PR waits on it
forever with nothing anywhere going red. That is the failure mode called out in
this checker's own rule 5. Making `zuuallet.yml` gate-bearing therefore means
deleting its trigger filters, adding `merge_group:`, building it a
change-detector job, gating all six jobs on it, exempting the scheduled
`upstream-canary`, and *then* editing branch protection. That is a rewrite of the
workflow's whole selection model.

It also creates a second required context with none of the hardening the first
two carry: `check-github-actions-pins.mjs` treats `.github/workflows/zuuli.yml`
as *the* required workflow and holds its gate, its change detector and several of
its jobs to exact reviewed source. A new required context outside that contract
would be materially weaker than the ones it joins.

**What option 2 buys.** `rust_native_clippy` in `zuuli.yml` already runs on
`windows-latest` and `macos-latest`, under the required `gate`, with the right
selector, the right caches, and an exact-source contract in the pins checker. The
gap was never the platform — it was that the gated native job only ever *compiles*.
So:

- **New `rust_native_tests` job in `zuuli.yml`**, `name: Rust / native tests
  (${{ matrix.target_os }})`, same matrix and same selector as the lint job beside
  it, running the exact command the retired custody job ran:
  `cargo test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml`.
- **Added to the gate's `needs:`**, and `verifyGateResults` now derives its
  expected result from the shared native selector, so the gate can never accept a
  skip from it that it would reject from `rust_native_clippy`.
- **`custody-platforms` deleted from `zuuallet.yml`.** The *whole* matrix moved —
  macOS as well as Windows — so the suite is not split across workflows and not
  duplicated. A comment where it stood records where it went and why.

`zuuallet.yml` stays ungated and is registered as such, with the reason spelled
out: its remaining jobs restate gated work (`fmt`/`deny`/`clippy` run the same
scripts as `rust_fmt`/`rust_deny`/`rust_clippy`), and what is left is
Zuuallet-only build coverage that cannot fail a merge and is no longer claimed to.

The new job is held to its exact source by `REQUIRED_NATIVE_TESTS_JOB_LINES`, for
the same reason the lint job is: dropping `--locked`, dropping the Windows matrix
leg, or narrowing the selector would each silently restore the #610 state.

## Branch protection: no change required, and that is the point

**The required-context list stays exactly `["gate", "rs / gate"]`. There is nothing
for you to change, now or later.**

`rust_native_tests` reports into the existing `gate` context because the gate
awaits it — so the moment this merges, a Windows test failure fails a required
check. No new context to add, nothing to sequence, and no window in which a
required name nobody publishes leaves PRs pending forever. Given that #562 came
from exactly that class of mistake, avoiding the sequencing problem entirely was
worth more than the extra reach option 1 offers.

If you ever *do* want `zuuallet.yml` gated, rule 6 now makes the order explicit:
give it a gate with a distinct `name:` (never a bare `gate:` — #562), land that,
confirm the context is reporting on a real PR, and only then add it to branch
protection; the registration entry must be removed in the same change or the
checker fails on the stale entry.

## Proofs

**The anchor.** Narrowing the enumeration does not quietly shrink the population —
it names what it stopped covering. With `workflowFiles` filtered to skip two files:

```
- UNGATED_WORKFLOWS registers .github/workflows/zuuallet.yml, which is not among the 12 workflow file(s) this check enumerated; either the workflow was deleted and the entry is stale, or the enumeration stopped covering it and that file is now judged by nothing
- UNGATED_WORKFLOWS registers .github/workflows/zuuli-packaging.yml, which is not among the 12 workflow file(s) this check enumerated; ...
2 failure(s) across 2 gate(s) in 12 workflow file(s).
```

The registry is the anchor: shrink the population and the registered files go
missing; shrink it the other way and rule 5 loses a producer for a protected
context. No hand-maintained count to go stale.

**Stale registrations fail, three ways.**

```
# a registered workflow that does not exist
- UNGATED_WORKFLOWS registers .github/workflows/zuuli-ghost.yml, which is not among the 14 workflow file(s) this check enumerated; ...

# a registered workflow that has a gate (registered zuuli.yml)
- UNGATED_WORKFLOWS registers .github/workflows/zuuli.yml as deliberately ungated, but it now publishes a gate; remove the stale entry so the gate is held to the rules the registration excused it from

# an entry whose reason was emptied
- UNGATED_WORKFLOWS registers .github/workflows/cache-cleanup.yml with no substantive reason (0 character(s), at least 40 required); an unexplained exemption is indistinguishable from an oversight
```

**A newly added ungated workflow is red**, and the enumeration reads the directory
rather than `git ls-files`, so an *untracked* probe file is still seen:

```
- .github/workflows/probe-ungated.yml: workflow publishes no gate job and is not registered in UNGATED_WORKFLOWS; ...
1 failure(s) across 2 gate(s) in 15 workflow file(s).
```

**Negative controls, and each one broken in turn.** Six new cases in
`check-workflow-gates.mjs --self-test` (20 → 26). Each was proved non-inert by
mutating the rule it covers and confirming the self-test names *that* case:

| mutation | self-test reports |
| --- | --- |
| unregistered ungated workflow not reported | `a workflow with no gate that nothing registers was not detected` |
| registration no longer consulted | `a workflow registered as deliberately ungated should pass, got: ...` |
| missing-file registration tolerated | `a registration for a workflow the enumeration did not produce was not detected` |
| gained-a-gate registration tolerated | `a registration for a workflow that has since gained a gate was not detected` |
| `MINIMUM_REGISTRY_REASON = 0` | `a registration with no substantive reason was not detected` |
| gate-feeds-no-required-context tolerated | `a gate that no required context resolves to was not detected` |

Seven new cases in `check-github-actions-pins.mjs --self-test` (3 gate-verdict, 4
current-workflow mutation), broken the same way:

| mutation | self-test reports |
| --- | --- |
| `gate must await rust_native_tests` rule removed | `real workflow rejects native tests detached from gate: expected "gate must await rust_native_tests"` |
| exact native-tests contract removed | `real workflow rejects a weakened native test command: expected "...exact current-source native execution contract"` |
| `rust_native_tests` dropped from the shared native selector | `native tests cannot skip a Zuuallet-only Rust change: ... got success` |
| verdict loop skips `rust_native_tests` | `native tests cannot skip a Zuuallet-only Rust change: ... got success` |

## The nit in `legacy_import_preview.rs`

`source_content_or_identity_change_after_copy_is_rejected` claimed identity
coverage it did not have. It swapped a whole different file in with `fs::rename`,
which changes bytes, length, mtime *and* inode simultaneously — any one of the four
would have failed it, so it isolated neither content nor identity, and it stayed
green under the identity bypass #598's test kills.

Rather than rename it into a weaker promise, it now *earns* the content half:
`same_length_mtime_and_identity_but_distinct_content_after_copy_is_rejected`
rewrites the source **in place** at the same length and restores the modification
time, asserting inside the hook that `len`, `modified` and `identity` all match the
original and only `hash` differs. It is the exact mirror of
`same_content_length_and_mtime_but_distinct_identity_after_copy_is_rejected`
below it, so between the two every observed property is proved load-bearing on its
own.

**Mutation-proved, and the old test proved inert in the same run.** Blinding
`observe` to content (`hash: [0u8; 32]`, so neither `read_observed` nor the
snapshot comparison can see a byte change) and running both the new test and a
verbatim copy of the old one side by side:

```
test ...::same_length_mtime_and_identity_but_distinct_content_after_copy_is_rejected ... FAILED
test ...::probe_old_rename_based_content_or_identity_test ... ok
```

The old rename-based test passes a build that observes no content at all — its
`fs::rename` changes the inode, so the identity check alone still rejects it.
That is the exact shape of the complaint: the name said content *or* identity,
and it was load-bearing for neither. The identity mirror
(`same_content_length_and_mtime_but_distinct_identity_...`) and the in-place
neighbour both stay green under this mutation, as they should — each fails only
for its own reason.

Full suite restored and green: `142 passed; 0 failed` on the pinned 1.97.1
toolchain, `check-rust-fmt.sh` clean.

## Also

- `node scripts/check-workflow-gates.mjs` / `--self-test`,
  `check-github-actions-pins.mjs` / `--self-test`,
  `check-hash-domain-labels.mjs` / `--self-test`,
  `check-rust-toolchain.sh` / `--self-test`,
  `check-librustzcash-compat.mjs`, `check-zuuli-linux-image.mjs`,
  `check-zuuli-android-gate.mjs`, `check-zcash-permissions.mjs`,
  `check-public-repo-boundary.sh`, `check-rust-fmt.sh` — all green, bare and
  self-test.
- **#477:** bare `[[ ]]` / `(( ))` statements added: **0**. The one shell block
  added restates the reviewed toolchain-resolution step verbatim and every command
  in it is enforcing under `set -euo pipefail`.
- Every `uses:` added is 40-hex SHA-pinned with a version comment (#442), reusing
  references already present in the file.
- Deliberately untouched: `zuuli.yml`'s policy and gate-recheck steps (#591), and
  `WIRE.md` (#588). The only `zuuli.yml` edits are one new job and one job id
  appended to the gate's `needs:`.

### DIFF
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -302,36 +302,17 @@ jobs:
           cargo test --locked
           --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
 
-  # Compile the target-specific Keychain and Credential Manager adapters on
-  # their native hosts. Linux Secret Service is covered by the `rust` job.
-  custody-platforms:
-    if: github.event_name != 'schedule'
-    name: custody / ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [macos-latest, windows-latest]
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
-
-      - name: Fetch librustzcash submodule
-        run: git submodule update --init z/zcash/librustzcash
-
-      # `uses:` cannot take an expression, so this commit-pinned version action
-      # restates wallet/rust-toolchain.toml; scripts/check-rust-toolchain.sh
-      # holds it to the source of truth.
-      - uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
-
-      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
-        with:
-          workspaces: wallet/plugins/tauri-plugin-zcash
-
-      - name: Test native plugin
-        run: >-
-          cargo test --locked
-          --all-targets
-          --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
+  # The macOS/Windows plugin test matrix that used to live here — `custody /
+  # ${{ matrix.os }}` — is now `rust_native_tests` in .github/workflows/zuuli.yml,
+  # where the required `gate` awaits it (#610).
+  #
+  # It ran here, and passed, and could not fail a merge: this workflow publishes
+  # no gate, branch protection does not require any context it emits, and its
+  # `on: pull_request: paths:` filter means it does not run at all for a pull
+  # request that touches nothing it lists. So the one job in the repository that
+  # executed the Win32 file-identity helper was, in merge terms, advisory. The
+  # move keeps the suite whole rather than splitting it: the entire matrix went,
+  # under a selector that covers the same changes this workflow's filter does.
 
   # ---------------------------------------------------------------------------
   # Canary: rebuild against the LATEST librustzcash main + refreshed crates.
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -421,6 +421,67 @@ jobs:
         shell: bash
         run: scripts/check-rust-clippy.sh
 
+  # Native *execution*, not just compilation. `rust_native_clippy` above proves
+  # the macOS and Windows halves of the plugin compile; until #610 nothing that
+  # could fail a merge ever ran them. The Win32 file-identity helper the
+  # legacy-import preview relies on is reachable only from a Windows host, so a
+  # regression in it passed every required check — the plugin's Windows suite
+  # lived in the ungated .github/workflows/zuuallet.yml, which branch protection
+  # does not require and which does not even run for a pull request outside its
+  # `paths:` filter. That job moved here, where the required `gate` awaits it.
+  #
+  # Same selector and same matrix as the lint job deliberately: it is the same
+  # code on the same two hosts, and two selectors over one body of code drift.
+  rust_native_tests:
+    name: Rust / native tests (${{ matrix.target_os }})
+    needs: changes
+    if: needs.changes.outputs.zuuli == 'true' || needs.changes.outputs.zuuallet_schema == 'true'
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-latest
+            target_os: macos
+          - os: windows-latest
+            target_os: windows
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Fetch librustzcash submodule
+        run: git submodule update --init z/zcash/librustzcash
+
+      - name: Resolve the pinned Rust toolchain
+        id: rust_toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+          version=$(scripts/check-rust-toolchain.sh --print-channel)
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
+        with:
+          toolchain: ${{ steps.rust_toolchain.outputs.version }}
+
+      # Separate cache key from rust_native_clippy: clippy and rustc write
+      # different artifacts under the same target directory names, so a shared
+      # key has the two jobs evicting each other's work on every run.
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+        with:
+          workspaces: wallet/plugins/tauri-plugin-zcash
+          key: zuuli-native-tests-${{ matrix.target_os }}
+
+      # --all-targets so the integration and doc-adjacent targets build too;
+      # this is the exact command the retired zuuallet.yml custody job ran, now
+      # under a required context.
+      - name: Test the shared plugin on the native host
+        shell: bash
+        run: >-
+          cargo test --locked
+          --all-targets
+          --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
+
   rust_plugin:
     name: Rust / shared plugin
     needs: changes
@@ -750,7 +811,7 @@ jobs:
   # change detector and every applicable full-stack job succeeded; legitimate
   # non-ZUULI skips continue to report success instead of remaining pending.
   gate:
-    needs: [changes, frontend, rust_fmt, rust_deny, rust_clippy, rust_native_clippy, rust_plugin, rust_android_32, rust_app, zuuallet_schema, rust_crypto_targets]
+    needs: [changes, frontend, rust_fmt, rust_deny, rust_clippy, rust_native_clippy, rust_native_tests, rust_plugin, rust_android_32, rust_app, zuuallet_schema, rust_crypto_targets]
     if: always()
     runs-on: ubuntu-latest
     timeout-minutes: 5
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -84,6 +84,51 @@ const REQUIRED_NATIVE_CLIPPY_JOB_LINES = [
   "        shell: bash",
   "        run: scripts/check-rust-clippy.sh",
 ];
+// The macOS/Windows job that *executes* the shared plugin's test suite. Held to
+// its exact source for the same reason as the lint job beside it: this is the
+// only required job that runs Win32 and Darwin code paths, so weakening the
+// command, the matrix or the selector silently restores the #610 state where
+// Windows execution could not fail a merge.
+const REQUIRED_NATIVE_TESTS_JOB_LINES = [
+  "  rust_native_tests:",
+  "    name: Rust / native tests (${{ matrix.target_os }})",
+  "    needs: changes",
+  "    if: needs.changes.outputs.zuuli == 'true' || needs.changes.outputs.zuuallet_schema == 'true'",
+  "    timeout-minutes: 90",
+  "    strategy:",
+  "      fail-fast: false",
+  "      matrix:",
+  "        include:",
+  "          - os: macos-latest",
+  "            target_os: macos",
+  "          - os: windows-latest",
+  "            target_os: windows",
+  "    runs-on: ${{ matrix.os }}",
+  "    steps:",
+  `      - uses: ${GATE_CHECKOUT_REFERENCE} # v7.0.1`,
+  "      - name: Fetch librustzcash submodule",
+  "        run: git submodule update --init z/zcash/librustzcash",
+  "      - name: Resolve the pinned Rust toolchain",
+  "        id: rust_toolchain",
+  "        shell: bash",
+  "        run: |",
+  "          set -euo pipefail",
+  "          version=$(scripts/check-rust-toolchain.sh --print-channel)",
+  '          echo "version=$version" >> "$GITHUB_OUTPUT"',
+  "      - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable",
+  "        with:",
+  "          toolchain: ${{ steps.rust_toolchain.outputs.version }}",
+  "      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2",
+  "        with:",
+  "          workspaces: wallet/plugins/tauri-plugin-zcash",
+  "          key: zuuli-native-tests-${{ matrix.target_os }}",
+  "      - name: Test the shared plugin on the native host",
+  "        shell: bash",
+  "        run: >-",
+  "          cargo test --locked",
+  "          --all-targets",
+  "          --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml",
+];
 const REQUIRED_CRYPTO_TARGET_JOB_LINES = [
   "  rust_crypto_targets:",
   "    name: Rust / modern crypto targets (${{ matrix.family }})",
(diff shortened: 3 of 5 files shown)


skyl:

skyl:

skyl:

skyl:
Reviewed adversarially on the owner's behalf. **Approving** — merging once `gate` reports.

## Choosing option 2 was right, for a reason I had not considered

I leaned toward option 1 (give `zuuallet.yml` a gate) in the brief. That would have been a mistake, and the reason is decisive:

```yaml
# .github/workflows/zuuallet.yml
  pull_request:
    paths:
      - 'wallet/zuuallet/**'
```

**A path-filtered workflow does not run for a PR that matches nothing, and a required context from a workflow that did not run never reports.** Every PR touching anything else would pend forever. Option 1 was never "add a gate and edit a setting" — it was delete the trigger filters, add `merge_group:`, build a change detector, gate six jobs on it, exempt the scheduled canary, *then* edit branch protection. And it would have created a second required context with none of the exact-source hardening `check-github-actions-pins.mjs` applies to `zuuli.yml`.

That is the #562 class of failure — a required context nothing publishes — reached from the opposite direction. Catching it before building it is the win here.

**And the setting change I asked to be told about is: none.** `rust_native_tests` is in `gate.needs` (`zuuli.yml:814`), so it reports into the existing `gate`. The required list stays exactly `["gate", "rs / gate"]`, and the moment this merges a Windows test failure fails a required check. No sequencing, no pending-forever window.

Moving `custody-platforms` out of `zuuallet.yml` rather than duplicating it means the suite is neither split nor run twice. `cargo test --locked --all-targets` is the identical command, on the identical matrix, beside the `rust_native_clippy` job that already used that selector.

## The red-before found twelve, not one

I re-ran it with the registry emptied against this tree:

```
EXIT=1
12 failure(s) across 2 gate(s) in 14 workflow file(s)
```

Twelve of fourteen workflows publish no gate. Four of them run on `pull_request` and build real artifacts — `zuuli-packaging.yml`, `zuuli-linux-image.yml`, `ts-react-free2z.yml`, `ts-svelte-free2z.yml`. I filed #610 about one workflow; the checker's blind spot was hiding eleven more.

Green after registration: `2 gate(s), 12 registered ungated, 56 job(s), 14 workflow file(s)`.

## The registry is its own anchor, which is better than a count

Verified both stale directions myself:

```
register a workflow that does not exist
  -> EXIT=1  UNGATED_WORKFLOWS registers .github/workflows/does-not-exist.yml, which is not
             among the 14 workflow file(s) this check enumerated; either the workflow was
             deleted and the entry is stale, or the enumeration stopped covering it and that
             file is now judged by nothing

register zuuli.yml, which HAS a gate
  -> EXIT=1  registers .github/workflows/zuuli.yml as deliberately ungated, but it now
             publishes a gate; remove the stale entry so the gate is held to the rules the
             registration excused it from
```

That first message **is** the anchor: a narrowed enumeration surfaces as a stale registration, so there is no separate hand-maintained count to drift out of step. Cleaner than #609's arrangement, which needed a floor guard alongside the registry.

Requiring a ≥40-character reason per entry is the right kind of friction — it makes "registered as ungated" a statement someone had to defend rather than a line someone added.

Self-tests 20 → **26** here, plus 7 new in `check-github-actions-pins.mjs`, each broken in turn and each reporting its own case.

**My own sixth failed mutation tonight:** my first anchor probe patched `workflowFiles` crudely and crashed with a `ReferenceError` rather than firing the rule. `EXIT=1` for the wrong reason is not a result, so I retested from the registry side. Recording it because a crash and a catch look identical if you only read the exit code.

## The Rust nit turned out to be worth more than a rename

I flagged `source_content_or_identity_change_after_copy_is_rejected` as advertising identity coverage it lacked. The mutation proof shows it was worse than that: the old test used `fs::rename`, which **changes the inode**, so identity alone rejected the case and the content check was never exercised. The test was load-bearing for neither thing its name claimed.

The replacement — `same_length_mtime_and_identity_but_distinct_content_after_copy_is_rejected` — fails when `observe` is blinded to content, while a verbatim copy of the old test passes. That is the distinction made real. Full plugin suite restored at `142 passed; 0 failed`.

## Checks

Every checker green bare and `--self-test`. **0 bare `[[ ]]`/`(( ))` added** — and the one shell block added restates the reviewed toolchain-resolution step verbatim, every command enforcing under `set -euo pipefail`. All `uses:` 40-hex pinned with version comments. Correctly stayed out of `zuuli.yml`'s policy and gate-recheck steps (#591) and `WIRE.md` (#588); the only `zuuli.yml` edits are one new job and one id appended to `gate.needs`, so `check-github-actions-pins.mjs`'s exact ordered command list is untouched.

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:

skyl:
Merge-train hold: #567 is the sole active candidate. This branch is preserved; please keep it closed until the exact-head gate completes.

skyl:
Merge-train hold: #567 is the sole active candidate. This branch is preserved; please keep it closed until the exact-head gate completes.

skyl:
Merge-train hold: #567 is the sole active candidate. This branch is preserved; please keep it closed until the exact-head gate completes.

skyl:
**Holding this one, and the reason is a distinction worth recording.**

I have merged four PRs tonight without a green aggregate `gate` (#556, #600, #601, #613). In every case the justification was the same shape: *a job that can meaningfully fail on this diff passed on this exact content*. For #613 that was `Rust / shared plugin`; for #601, `frontend`; for #556, `Rust / shared plugin` on a byte-identical `send.rs` blob.

**That evidence does not exist here.** The job this PR adds has never completed on any head:

```console
$ for h in 5ad8dfcd 58deeedd; do gh api repos/free2z/zuu/commits/$h/check-runs \
    --jq '.check_runs[]|select(.name|test("native tests"))|"\(.name): \(.conclusion)"'; done
5ad8dfcd Rust / native tests (macos):   cancelled
5ad8dfcd Rust / native tests (windows): cancelled
58deeedd Rust / native tests (macos):   cancelled
58deeedd Rust / native tests (windows): cancelled
```

Both heads carry an identical `zuuli.yml` blob (`0cb6ec31`), so this is not a content question — the job simply has not been observed to run successfully even once.

And it matters more here than it would elsewhere: **`rust_native_tests` is in `gate.needs`**, so on merge it becomes a dependency of the one required context protecting `main`. If it is misconfigured — a bad selector, a matrix that produces no jobs, a command that fails on Windows — it would fail *every subsequent PR*, not just this one. Merging a required-gate dependency that has never been seen green is exactly the risk the gate exists to prevent.

The review stands: I verified the red-before naming **12** ungated workflows, the self-test at 20 → 26, that stale registration fires both ways, and that `rust_native_tests` feeds the existing gate so **no branch-protection change is needed**. The design is right and the option-2 reasoning about `zuuallet.yml`'s `paths:` filter is correct.

What is missing is one uncancelled run. Every failure so far has been `change detection did not pass: cancelled` — a concurrency supersede, made worse by the PR being repeatedly closed and reopened, which triggers a fresh run that cancels the one in flight. I have reopened it once and am deliberately not touching it again so a single run can finish.

**Merge as soon as `Rust / native tests (windows)` and `(macos)` report `success` even once** — at that point the assembled-evidence argument applies normally and a cancelled aggregate `gate` is no obstacle.

skyl:
Train hold: this reviewed change still requires the documented integration with #615. Keeping #610 open; will replay the union on current main after the P1 fix-forward.

skyl:
Closing the superseded split PR again. #567 and #610 are combined in the independently mutation-reviewed one-commit integration b8b42c7; a fresh PR will be opened after the #547 fix-forward lands and the integration is replayed on current main.

PR #615, opened 2026-08-24
ci: give every workflow check a unique context

### LINKED ISSUE #567: ci: two more duplicated status-check contexts — `build` (3 workflows) and `frontend` (2)

skyl:
Closes #567

## Summary

- gives the five duplicated workflow jobs exact, unique display contexts
- pins the five file/job/name contracts plus protected contexts to an independent fixed registry digest
- rejects every duplicate-context allowlist entry
- makes the live repository scan consume injected production policy, so deleting or hard-coding that wiring is test-visible

This is a single-commit replay of the independently approved tree at `d7d8992598e6713d4236be0522c244b6d35bd903` onto exact current `origin/main`; the seven changed files are byte-identical to that reviewed tree.

## Mutation evidence

Independent review reproduced all of these red:

- deletion and alteration of each of five exact workflow contexts (10 independent cases)
- expected-context verdict deletion
- collision-verdict deletion
- real cross-workflow duplicate context
- nonempty collision allowlist
- protected-context removal and registry-entry deletion
- entire production-policy scan invocation deletion
- hard-coding global policy instead of consuming injected live-scan policy
- comment-only context name

The first review blocked a surviving live-scan invocation mutant. The repair was independently re-reviewed and approved only after both invocation deletion and hard-coding went red against the actual repository with a fixed literal diagnostic.

## Restored verification

- `check-workflow-gates.mjs --self-test`: 33/33
- live workflow gate/context policy: 2 gates, 56 jobs, 14 workflows
- Actions policy self/live: 74 source, 7 verdict, 130 workflow mutants; 178 refs / 15 files
- toolchain self/live and f2z-codec tests on the approved tree
- exact replay check against approved head and `git diff --check`

### DIFF
--- a/.github/workflows/docs-about-free2z.yml
+++ b/.github/workflows/docs-about-free2z.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    name: docs / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/ts-react-free2z.yml
+++ b/.github/workflows/ts-react-free2z.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    name: ts-react / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/ts-svelte-free2z.yml
+++ b/.github/workflows/ts-svelte-free2z.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    name: ts-svelte / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -181,6 +181,7 @@ jobs:
   # Frontend: TypeScript typecheck + production bundle.
   # ---------------------------------------------------------------------------
   frontend:
+    name: zuuallet / frontend
     if: github.event_name != 'schedule'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -134,6 +134,7 @@ jobs:
           echo "zuuallet_schema=$zuuallet_schema" >> "$GITHUB_OUTPUT"
 
   frontend:
+    name: zuuli / frontend
     needs: changes
     if: needs.changes.outputs.zuuli == 'true'
     runs-on: ubuntu-latest
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -183,6 +183,7 @@ const REQUIRED_CLASSIC_SEED_BOUNDARY_INPUTS = [
 ];
 const REQUIRED_FRONTEND_JOB_LINES = [
   "  frontend:",
+  "    name: zuuli / frontend",
   "    needs: changes",
   "    if: needs.changes.outputs.zuuli == 'true'",
   "    runs-on: ubuntu-latest",
@@ -2644,6 +2645,11 @@ function runCurrentWorkflowMutationTests(repoRoot) {
       needle: `ZUULI selector must run the seed boundary for classic input ${input}`,
       source: source.replace(`|${input}`, ""),
     })),
+    {
+      name: "real workflow requires the unique frontend display context",
+      needle: "frontend must match the complete exact current-source execution program",
+      source: replaceFrontend("    name: zuuli / frontend\n", ""),
+    },
     {
       name: "real workflow rejects a dynamically dead frontend job",
       needle:
--- a/scripts/check-workflow-gates.mjs
+++ b/scripts/check-workflow-gates.mjs
@@ -49,19 +49,16 @@
 //      future regression, so `needs ⊇ {jobs in file} − {gate}` is asserted,
 //      with an explicit per-file opt-out for jobs deliberately left outside.
 //
-//   5. PROTECTED CONTEXTS. The names branch protection actually requires get a
-//      stricter reading of both rules above. The tolerated-collision allowlist
-//      may not name one of them, so narrowing the guard to admit a duplicate
-//      `gate` fails on the allowlist edit itself rather than on the `name:`
-//      change that would follow it. And each required name must be published by
-//      exactly one job, which catches the opposite drift: a required context no
-//      workflow emits never reports at all, so PRs hang pending forever with
-//      nothing anywhere going red.
+//   5. PROTECTED AND EXPECTED CONTEXTS. Each branch-protected name must have one
+//      producer, five ambiguity fixes retain their exact reviewed display names,
+//      and the collision allowlist remains empty. The production configuration
+//      is digest-pinned independently from the workflows and fixture options.
 //
 // Usage:
 //   node scripts/check-workflow-gates.mjs             judge every workflow
 //   node scripts/check-workflow-gates.mjs --self-test negative controls first
 
+import { createHash } from "node:crypto";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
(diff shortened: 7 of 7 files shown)


skyl:
Train hold: main advanced while this exact-head run was still pending. Closing temporarily so the issue #567 change can be replayed and re-proved on the current trunk after the P1 send-authority fix-forward. The reviewed tree and mutation evidence are preserved.

skyl:
Temporarily closing the stale-base lane; #567 remains open and will be replayed on current main.

skyl:
Reviewed adversarially. **Sound — merging once `gate` and `rs / gate` are green.** This is the first review this PR has had; it has been swept closed repeatedly without one.

### Branch protection is not disturbed

The five workflow edits are one `name:` line each, on job IDs `build`, `build`, `build`, `frontend`, `frontend`. **None of them touches `gate` or `rs / gate`**, which are the only two protected contexts. Job IDs are unchanged, so every `needs:` edge still resolves. No branch-protection update is required, and none should be made.

### The duplicate detector is the property, not the snapshot

The thing I most wanted to disprove is that this fixes the five known collisions and calls it done. It does not:

- `contextCollisionFailures(published)` ranges over **every** published context in **every** workflow, and the tolerated-collision parameter is gone from its signature entirely. A sixth collision in a workflow nobody has written yet fails on the same code path as the five being fixed today.
- The allowlist is not merely emptied, it is required to stay empty — `TOLERATED_CONTEXT_COLLISIONS must remain empty, got N entries; give every producer a unique name instead`. That is strictly stronger than the deleted `protectedToleranceFailures`, which only rejected allowlist entries naming a *protected* context. There is now no exception path at all, so the narrower check has nothing left to catch.
- Triggers are still deliberately not consulted, which the comment justifies correctly: a name unique only because two workflows run on different events stops being unique the moment someone adds `pull_request:` — a one-line edit nobody would think of as touching branch protection.

### The registry cannot be blinded — this is the part worth calling out

`EXPECTED_JOB_CONTEXTS` pins the five exact display names, and `REQUIRED_CONTEXT_POLICY_DIGEST` pins the registry *and* the protected set independently, so narrowing the expected set cannot make its own mutation invisible. That alone would only be a snapshot. What makes it a control is the live-scan self-test:

```js
const invalidProductionScan = scanRepository(REPO_ROOT, {
  protectedContexts: new Set(["gate"]),
});
// asserts /differ from the independently reviewed registry digest/ fires
```

It injects a deliberately **wrong** protected set into the **live repository** scan and requires the digest diagnostic to appear. That kills both blinding mutations at once:

- delete the `productionContextConfigurationFailures` call from `scanRepository` → no diagnostic → red
- ignore the injected `protectedContexts` and read the module constant instead → digest matches → no diagnostic → red

A guard that can only be disabled by making a test fail is the shape this repo keeps asking for and rarely gets. Fixtures opt out via `enforceProductionConfiguration: false` in `fixtureOptions`, which is correct — fixtures publish `gate` and not `rs / gate` and must not be judged against live constants.

### One nit, not blocking

`enforceProductionConfiguration ?? true` defaults on, but nothing stops a future edit from passing `false` at the CLI invocation, which would skip the digest pin in the real run while every self-test stays green. It is not exploitable today: `selfTest()` independently calls `productionContextConfigurationFailures()` with real defaults and requires it to pass, so narrowing the registry goes red regardless of what the CLI passes. Worth either dropping the flag in favour of a fixture-only override, or asserting the CLI path does not set it.

Also worth noting for the record: `expectedContextFailures` pins five entries and is not a coverage anchor over the workflow population — nothing asserts every workflow is represented in it. That is fine here, because global uniqueness is the actual invariant and `contextCollisionFailures` enforces it over everything; the registry is a deliberate belt-and-braces pin on this specific fix.

### What I did not observe

`gate` has not completed on this head. Every run on this branch has been cancelled or is still queued (#618), and I updated the branch myself a few minutes ago, so the two cancellations at 08:06 are my own supersede and not the mystery in that issue. I will merge on the assembled-evidence rule only if a specific job is observed passing on this exact blob, and will say which one.

skyl:
Closing the superseded split PR again. #567 and #610 are combined in the independently mutation-reviewed one-commit integration b8b42c7; a fresh PR will be opened after the #547 fix-forward lands and the integration is replayed on current main.

skyl:
**Not merging — this needs a real rebase, and `mergeable: MERGEABLE` is lying about it.**

The review above stands: the change is sound and I would merge it on its own merits. But #614 landed at 08:27:54Z (`2487fef`) and it touches four of this PR's seven files:

```
.github/workflows/zuuallet.yml
.github/workflows/zuuli.yml
scripts/check-github-actions-pins.mjs
scripts/check-workflow-gates.mjs
```

`gh pr view 615 --json mergeable` still reports `MERGEABLE`; that field is stale. The real test disagrees:

```
$ gh pr update-branch 615
X Cannot update PR branch due to conflicts

$ git merge-tree --write-tree --name-only origin/main origin/ci/567-context-registry
CONFLICT (content): Merge conflict in scripts/check-workflow-gates.mjs
```

### Why I will not admin-merge past this

`--admin` bypasses `strict: true`, and `strict: true` is the only control standing between this PR and a broken `main`. The file in conflict **is the gate**: if the merged `check-workflow-gates.mjs` is textually fine but semantically wrong, it does not fail this PR — it fails `gate` on **every subsequent PR**, and the repo stops.

That is not hypothetical here. The two trees already disagree on their own census:

```
main  (after #614)  console.log("check-workflow-gates self-test: 26 case(s) passed.");
#615                console.log("check-workflow-gates self-test: 33 case(s) passed.");
```

#614 took the self-tests 20 → 26 and added the twelve-ungated-workflow census; this PR took them to 33 and added the context registry, the collision detector and the digest pin. A resolution that keeps one number and the other's cases is green here and wrong everywhere after. This is the same failure #556 and #560 nearly caused, where git merged 4 + 4 into 6 while the assertion demanded exactly 4.

### What the rebase has to reconcile

1. **The self-test total.** One number covering both sets of cases, and it must be the count that actually runs — not either side's.
2. **#614's ungated-workflow census.** It names twelve workflows and must still go red when one loses its gate. This PR adds `name:` to five jobs across five workflows; confirm the census keys off the job/file identity it used before, not a display name that has now changed.
3. **#614's `Rust / native tests` gating.** It is now in `gate.needs` and proven passing on both `windows` and `macos`. It publishes new contexts — verify they are unique under this PR's `contextCollisionFailures`, and decide deliberately whether they belong in `EXPECTED_JOB_CONTEXTS`.
4. **`REQUIRED_CONTEXT_POLICY_DIGEST`.** If the protected or expected set changes as a result of 3, the digest must be recomputed **deliberately**, which is the property it exists to enforce. Do not paste in whatever the run prints without deciding the registry is right first.

### Evidence this PR was green before the collision

For the record, both required workflows passed on head `f5dd2e7e`:

- `gate` — run `32704572542`, `completed/success`, every job green including `zuuli / frontend`, which is this PR's own renamed context working as intended
- `rs / gate` — run `32704572556`, `completed/success`

So this is purely a rebase-and-reconcile, not a defect in the change. Re-request when the branch is on current `main` and the merged self-test count is the one that actually runs.

skyl:
Superseded by #622, which is the clean single-commit rebase onto current `main` rather than a re-resolution of the conflict in `scripts/check-workflow-gates.mjs` — the right call, since that file is the gate. All four reconciliation requirements from my earlier comment are met and mutation-verified on #622: self-test count 39 printed / 39 emitted, #614's `rust_native_tests` gating still red when detached, and the registry digest fires both on narrowing and on workflow drift. Leaving this closed.

PR #622, opened 2026-08-24
ci: give every duplicated workflow context a unique name

### LINKED ISSUE #567: ci: two more duplicated status-check contexts — `build` (3 workflows) and `frontend` (2)

skyl:
Closes #567

Give each of the five previously duplicated `build` / `frontend` jobs an exact unique check-run name, and pin those names in an independent registry plus digest.

Mutation proof on exact head `b5573b3d34772b362631fa5b0d47bfca5ce97c8d`:

- production-configuration enforcement removed: red
- expected-context verdict wiring removed: red
- registry entry deleted without digest update: red
- each exact name deleted, aliased to another context, or arbitrarily changed: red
- landed #610 native Windows tests detached from the gate: red

Restored green:

- workflow-gates self-test: 39/39; live: 2 gates, 56 jobs, 14 workflows
- Actions policy: 74 source + 10 verdict + 134 live-workflow mutants; live 178 refs / 15 files
- Android gate: 31/31 plus live
- diff check

This is exactly one commit on current `main`. The #610 Windows gate is inherited byte-for-byte from trunk; this PR's `zuuli.yml` delta is only its frontend display name. Independent adversarial review approved this exact head.

### DIFF
--- a/.github/workflows/docs-about-free2z.yml
+++ b/.github/workflows/docs-about-free2z.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    name: docs / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/ts-react-free2z.yml
+++ b/.github/workflows/ts-react-free2z.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   build:
+    name: ts-react / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/ts-svelte-free2z.yml
+++ b/.github/workflows/ts-svelte-free2z.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    name: ts-svelte / build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/zuuallet.yml
+++ b/.github/workflows/zuuallet.yml
@@ -181,6 +181,7 @@ jobs:
   # Frontend: TypeScript typecheck + production bundle.
   # ---------------------------------------------------------------------------
   frontend:
+    name: zuuallet / frontend
     if: github.event_name != 'schedule'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -134,6 +134,7 @@ jobs:
           echo "zuuallet_schema=$zuuallet_schema" >> "$GITHUB_OUTPUT"
 
   frontend:
+    name: zuuli / frontend
     needs: changes
     if: needs.changes.outputs.zuuli == 'true'
     runs-on: ubuntu-latest
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -228,6 +228,7 @@ const REQUIRED_CLASSIC_SEED_BOUNDARY_INPUTS = [
 ];
 const REQUIRED_FRONTEND_JOB_LINES = [
   "  frontend:",
+  "    name: zuuli / frontend",
   "    needs: changes",
   "    if: needs.changes.outputs.zuuli == 'true'",
   "    runs-on: ubuntu-latest",
@@ -2757,6 +2758,11 @@ function runCurrentWorkflowMutationTests(repoRoot) {
       needle: `ZUULI selector must run the seed boundary for classic input ${input}`,
       source: source.replace(`|${input}`, ""),
     })),
+    {
+      name: "real workflow requires the unique frontend display context",
+      needle: "frontend must match the complete exact current-source execution program",
+      source: replaceFrontend("    name: zuuli / frontend\n", ""),
+    },
     {
       name: "real workflow rejects a dynamically dead frontend job",
       needle:
--- a/scripts/check-workflow-gates.mjs
+++ b/scripts/check-workflow-gates.mjs
@@ -49,14 +49,10 @@
 //      future regression, so `needs ⊇ {jobs in file} − {gate}` is asserted,
 //      with an explicit per-file opt-out for jobs deliberately left outside.
 //
-//   5. PROTECTED CONTEXTS. The names branch protection actually requires get a
-//      stricter reading of both rules above. The tolerated-collision allowlist
-//      may not name one of them, so narrowing the guard to admit a duplicate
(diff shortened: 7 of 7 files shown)


skyl:
Reviewed adversarially against the four reconciliation requirements I set on #615. **All four met. Approving in substance — merging once `gate` and `rs / gate` are green.**

I verified these by running the checker in a throwaway worktree on exact head `b5573b3d`, not by reading the PR body.

### 1. The self-test count is not stale

This repo keeps getting bitten by a recorded number that no longer describes what ran, and the merge of #614 (26 cases) with #615 (33 cases) was the obvious place for it to happen again:

```
check-workflow-gates self-test: 39 case(s) passed.

$ grep -c '^self-test: '
39
```

Printed claim and emitted cases agree exactly. The number describes what ran.

### 2. #614's ungated-workflow census survives, and is keyed off job identity

The rename could plausibly have blinded a census that keyed off display names. It doesn't — detaching the job #614 landed goes red with a precise diagnostic:

```
$ perl -0pi -e 's/…, rust_native_clippy, rust_native_tests, /…, rust_native_clippy, /' .github/workflows/zuuli.yml
815:    needs: [changes, frontend, rust_fmt, rust_deny, rust_clippy, rust_native_clippy, rust_plugin, …]

EXIT=1
.github/workflows/zuuli.yml:815: gate does not await rust_native_tests, defined in the same
workflow; a job outside the gate's needs is a job the gate cannot fail on
```

### 3. The digest anchor fires both ways

**Narrowing the registry** — delete one entry, leave the digest alone:

```
before  125:  [".github/workflows/docs-about-free2z.yml#build", "docs / build"],
after   (gone; line 125 is now the ts-react entry)

EXIT=1
protected and expected status-check contexts differ from the independently reviewed
registry digest: 90606360621578e4…
```

**Workflow drifting from the registry** — rename the job, leave the registry alone:

```
137:    name: zuuli / frontend-renamed

EXIT=1
.github/workflows/zuuli.yml#frontend must publish the exact status-check context
"zuuli / frontend", got "zuuli / frontend-renamed"
```

So the registry cannot make its own narrowing invisible, and the workflow cannot drift away from it. That is the coverage-anchor property, in both directions.

### 4. #614 is inherited byte-for-byte

The branch is exactly one commit on current `main` (`0 1` against `origin/main`, and it contains main's HEAD). Its entire `zuuli.yml` delta is:

```diff
   frontend:
+    name: zuuli / frontend
     needs: changes
```

`gate.needs` is untouched and still lists `rust_native_tests`. This is the clean rebase #615 needed rather than a re-resolution of the conflict, which is the right call — the conflicted file was the gate itself.

### Nit carried over from #615, still not blocking

`enforceProductionConfiguration ?? true` could still be passed `false` at the CLI invocation, skipping the digest pin in the real run while self-tests stay green. Not exploitable, because `selfTest()` independently calls the production configuration check with real defaults, so narrowing goes red regardless. Worth dropping the flag in favour of a fixture-only override eventually.

### What I have not observed

`gate` has not completed on this head — run `32710642893` is still pending, and with current macOS/Windows runner queueing that is a 25-35 minute proposition. I will merge on verified-green only, checking by run ID rather than `gh pr checks`, and I will not use the assembled-evidence argument here.

Supersedes #615. Live check contexts already show `zuuli / frontend`, so the duplicate this closes (#567) is resolved in practice on this head.

PR #623, opened 2026-08-24
security(relay): commit replay keys after authentication

### LINKED ISSUE #590: design(f2z-relay-proto): SeenSet inserts before signature verification, so unauthenticated frames can saturate the replay set

skyl:
Closes #590

Commit replay keys only after authentication and all local command checks succeed.

The replay transaction is split into a non-inserting preflight and an atomic final commit that repeats replay/capacity checks. Invalid signatures, self-auth failures, and bad padding cannot consume the bounded set; two preflight winners cannot both commit.

Mutation proof on exact head `571fcab6ef50565439d3de9f54d7416af455f27f`:

- preflight insertion/deletion, early commit before signature/self-auth/padding: red
- final recheck/commit/capacity deletion and replay-vs-capacity inversion: red
- all signed Verified fields, including client timestamp distinct from relay verification time: red
- every unsigned sentinel and any unsigned replay insertion: red

Restored green: 85 unit, 6 property, 8 redaction, 1 doctest, clippy with warnings denied, formatting and diff check.

Public docs explicitly scope the guarantee to the verifier's process-local SeenSet and disclaim durable, restart, clone, cross-worker, and external-state coordination. Independent adversarial review approved this exact one-commit head.

### DIFF
--- a/rs/crates/f2z-relay-proto/src/command.rs
+++ b/rs/crates/f2z-relay-proto/src/command.rs
@@ -15,11 +15,14 @@
 //!    and a non-canonical structure inside it would survive a frame-level
 //!    re-encode untouched.
 //! 2. Timestamp window (§5.5) → `ERR_STALE_TIMESTAMP`.
-//! 3. `(signer_key, nonce)` seen-set (§5.5) → `ERR_REPLAY`.
+//! 3. Preflight `(signer_key, nonce)` against the seen-set (§5.5) →
+//!    `ERR_REPLAY`, without inserting it.
 //! 4. Rebuild the transcript **from the relay's own `relay_id` and its own
 //!    `channel_binding`** and verify the signature → `ERR_BAD_SIGNATURE`.
 //! 5. Address and self-authentication rules (§6.2, §6.3) → `ERR_NO_ACCESS`.
 //! 6. Policy — the payload's padding bucket (§9).
+//! 7. Commit the authenticated `(signer_key, nonce)` with insert-if-absent
+//!    semantics. A competing commit loses as `ERR_REPLAY`.
 //!
 //! Steps 2 and 3 precede the signature check because they are cheap and they
 //! are the flood-resistant filters. Step 5 follows it so that an unsigned guess
@@ -484,12 +487,19 @@ impl<C: RelayCommand> SignedCommand<C> {
     }
 }
 
-/// A command that passed every check of §5.1.
+/// A command accepted through every check prescribed for `C`.
 ///
 /// There is no public constructor. The only way to hold one is to have run
-/// [`CommandVerifier::verify`] or [`CommandVerifier::accept_unsigned`], which
-/// is what makes "we verified this" a fact about the type rather than a comment
-/// on a call site.
+/// [`CommandVerifier::verify`] for a signed command, or
+/// [`CommandVerifier::accept_unsigned`] for an unsigned command. The signed
+/// path proves §5.1, including the timestamp window, signature, command policy,
+/// and a replay key committed to that verifier's process-local [`SeenSet`]. It
+/// does not attest that a relay also committed external, durable, or
+/// cross-worker state. The unsigned path has no authenticator, timestamp, or
+/// replay key to verify; it proves canonical encoding plus that command's own
+/// rules and stores zero values in the authentication-only accessors below.
+/// Holding this type therefore proves the checks for `C`'s prescribed path,
+/// not that an arbitrary `Verified<C>` was authenticated.
 #[derive(Debug)]
 pub struct Verified<C: RelayCommand> {
     body: C::Request,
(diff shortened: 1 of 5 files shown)


skyl:
Reviewed adversarially, and verified against the real crate rather than the PR body. **Sound — merge when `gate` and `rs / gate` are green.**

### This dissolves the owner decision rather than picking a side

I had #590 on my list as an **owner decision, not to be designed by an agent**, because it looked like a genuine tradeoff:

- insert **before** verify — an invalid signature consumes a slot in a bounded global set, so anyone can fill it with garbage
- insert **after** verify — no unauthenticated consumption, but a window opens where two concurrent presentations of the same `(signer_key, nonce)` both pass the lookup and both get accepted

This PR does not choose. It splits the operation so both properties hold at once: `preflight` stays exactly where §5.1 step 3 puts it — before the signature check, cheap and flood-resistant — but does not insert; `commit` runs only after authentication and every command check, and repeats the membership and capacity checks while holding `&mut self`, so the check-and-insert is atomic.

That means there is **no decision left for you to make here**, which is why I am approving it rather than holding it. The one thing I would still call a design choice is deferred and honestly disclosed (below).

### Verified: baseline green, and the security property has its own failing test

Baseline on head `571fcab6`: **100 tests pass** (85 unit, 6 property, 8 redaction, 1 doctest).

Mutation — make `preflight` insert, i.e. revert to the old `observe` semantics that #590 was about. Landed at the exact line and printed back:

```
266:        if self.entries.len() >= self.max_entries {
267:            return Err(ProtoError::Wire(ErrorCode::Backpressure));
268:        }
269:
270:        self.entries.insert(key.clone(), now_ms.saturating_add(self.retention_ms));
271:        Ok(())
```

Result: **`test result: FAILED. 76 passed; 9 failed`**. What matters is *which* nine, because a mutation this broad also breaks the happy path, and an incidental kill would prove much less. Two assertions fired by name:

```
a frame rejected by policy must not consume replay capacity
an unauthenticated frame must not consume replay capacity
```

and `replay::tests::two_preflight_winners_cannot_both_commit` failed alongside them. So the security property is guarded by tests that assert it directly, not merely by the round-trip tests falling over.

### The scoping is honest, which is the part I checked hardest

The doc comments do not claim a property a different deployment would defeat:

> The borrow is not a distributed lock: clones and independent sets do not coordinate. A multi-worker relay must serialize one shared set or give an external store the same atomic insert-if-absent semantics.

and on `Verified<C>`:

> It does not attest that a relay also committed external, durable, or cross-worker state… Holding this type therefore proves the checks for `C`'s prescribed path, not that an arbitrary `Verified<C>` was authenticated.

That second one is a real tightening beyond the stated scope of this PR: the previous wording ("a command that passed every check of §5.1") over-claimed for the unsigned path, which stores zeros in the authentication-only accessors. Good catch by whoever wrote it.

### The one thing genuinely deferred

Atomicity holds **per `SeenSet` instance**. A relay that processes one replay domain across workers or processes gets no protection from this borrow and needs external insert-if-absent. That is stated in the code rather than assumed, so it is not a defect in this change — but it is a real design item for whenever the relay server lands, and it should not be rediscovered then by accident. Worth an issue at that point rather than now, since there is no server yet.

### What I did not observe

`gate` has not completed on this head — the PR was swept closed while I was reviewing, and I reviewed the branch directly via a throwaway worktree. I will merge only on `gate` and `rs / gate` verified green by run ID on `571fcab6`, and I will not use the assembled-evidence argument.

PR #625, opened 2026-08-24
ci: mutation-prove Rust root ownership census

### LINKED ISSUE #608: ci(toolchain): mutation-prove the real crate census producer and Rust-root ownership
Follow-up to #553 / #600. The merged implementation improves registration discipline, but its committed self-tests do not yet prove the production census or complete Rust-root ownership contract.

## Observed gaps

1. `census_tracked_files` consumes newline-delimited `git ls-files` output through command substitution / `while read`. Git paths may contain tabs, backslashes, and newlines; Bash variables cannot safely hold a NUL-delimited inventory. A hostile path can therefore be skipped at `[[ ! -f ]]`.
2. The committed census tests call `check_census_listing` with synthetic strings instead of exercising the real Git producer through `run_checks`. Removing `check_census "$root"` from `run_checks`, or narrowing `census_tracked_files` to today's known directories, leaves those tests green.
3. Registration is not Rust-root policing. A third virtual workspace root can be added to the manifest/toolchain arrays without proving one owning workflow, exact selectors, or fmt/clippy/deny enforcement.
4. ZUULI and `rs` toolchain steps are not structurally guarded against deletion, wrong-root invocation, dead execution, or soft failure.
5. The existing rustfmt empty-tree self-test / aggregate count still needs a falsifiable negative control.

## Acceptance

- Inventory tracked paths via a NUL-safe stream/file (`git ls-files -z`); do not store NUL data in a Bash variable. Must run under macOS Bash 3.2.
- Real staged-Git fixtures exercise `run_checks` end to end for ordinary, space, tab, newline, exact `z/`, empty-root, and broken-registration cases.
- Declare the exact policed Rust roots. Every tracked `Cargo.toml`, including virtual workspace roots, belongs to exactly one policed root.
- Each policed root has exactly one owning workflow and exact change selectors plus live, non-soft-failing fmt/clippy/deny/toolchain enforcement.
- Repair the empty-tree rustfmt control and make aggregate counts derived or exact.

## Required red mutation controls

Demonstrate each independently against the committed tests before opening the PR:

- remove the production `check_census` call from `run_checks`;
- remove `-z` / substitute newline-delimited enumeration;
- narrow the producer pathspec to today's known roots;
- add tracked manifests whose names contain space, tab, and newline;
- add an unpoliced third Rust root;
- delete, dead-code, soft-fail, or point each required ZUULI/`rs` step at the wrong root;
- bypass each exact root/owner/selector verdict;
- neutralize the empty-tree rustfmt failure branch.

Tests must not feed expected inventory directly to the same helper under test, read expected values from generated output, or satisfy a contract with comments/source strings alone.

skyl:
Closes #608

## Summary

- census every tracked first-party Cargo manifest and toolchain file from the real NUL-delimited Git index
- classify package versus virtual manifests with Cargo semantics, failing closed on malformed manifests
- require exact ownership by the wallet/ or rs/ Rust root and one owning workflow per root
- make both workflow selectors NUL-safe under Bash 3.2 and bind their exact step IDs, published outputs, downstream consumers, and effective last values
- make rustfmt discover every manifest at arbitrary depth and fail closed on an empty tree

## Mutation proof

Every guard was broken before publication and shown red. Independent review killed:

- Cargo-valid quoted, spaced, inline, and dotted package spellings; malformed manifests; multiline-string virtual false positives
- plain/non-NUL Git producers and consumers across spaces, tabs, newlines, backslashes, and dash paths
- narrowed, dead, overlapping, adjacent-prefix, and vendored-root ownership selectors
- diff-failure inversion and unusable-base skips
- appended false outputs, wrong/duplicate output publications, renamed detector IDs, and downstream selector bypasses
- empty/depth-limited rustfmt census and one-to-one workflow ownership mutations

Restored baselines pass the toolchain, Actions-policy, workflow-gate, and rustfmt self/live suites. Exact independently approved head: 271730d00335e7677c05cec8382d6aae3fdb62b5.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -111,7 +111,9 @@ jobs:
             exit 0
           fi
 
-          if ! changed_files=$(git diff --name-only --no-renames "$BASE_SHA" "$GITHUB_SHA"); then
+          changed_files=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/rs-changed-files.XXXXXX")
+          trap 'rm -f -- "$changed_files"' EXIT
+          if ! git diff --name-only -z --no-renames "$BASE_SHA" "$GITHUB_SHA" > "$changed_files"; then
             echo "::warning::Diff failed; running the full rs gate."
             echo "rs=true" >> "$GITHUB_OUTPUT"
             exit 0
@@ -127,13 +129,13 @@ jobs:
           # Listing it would make every wallet pin bump run the rs suite twice
           # over for no additional coverage.
           rs=false
-          while IFS= read -r file; do
+          while IFS= read -r -d '' file; do
             case "$file" in
               rs/*|scripts/check-rust-fmt.sh|scripts/check-rust-clippy.sh|scripts/check-rust-deny.sh|scripts/check-rust-toolchain.sh|.github/workflows/rs.yml)
                 rs=true
                 ;;
             esac
-          done <<< "$changed_files"
+          done < "$changed_files"
 
           echo "rs=$rs" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -108,7 +108,9 @@ jobs:
             exit 0
           fi
 
-          if ! changed_files=$(git diff --name-only --no-renames "$BASE_SHA" "$GITHUB_SHA"); then
+          changed_files=$(mktemp "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/zuuli-changed-files.XXXXXX")
+          trap 'rm -f -- "$changed_files"' EXIT
+          if ! git diff --name-only -z --no-renames "$BASE_SHA" "$GITHUB_SHA" > "$changed_files"; then
             echo "::warning::Diff failed; running the full ZUULI gate."
             echo "zuuli=true" >> "$GITHUB_OUTPUT"
             echo "zuuallet_schema=true" >> "$GITHUB_OUTPUT"
@@ -117,9 +119,9 @@ jobs:
 
           zuuli=false
           zuuallet_schema=false
-          while IFS= read -r file; do
+          while IFS= read -r -d '' file; do
             case "$file" in
-              Cargo.toml|Cargo.lock|.cargo/*|clippy.toml|.clippy.toml|wallet/Cargo.toml|wallet/Cargo.lock|wallet/.cargo/*|wallet/clippy.toml|wallet/.clippy.toml|wallet/*.rs|wallet/*/Cargo.toml|wallet/*/Cargo.lock|wallet/*/.cargo/*|wallet/*/clippy.toml|wallet/*/.clippy.toml|wallet/zuuli/*|wallet/plugins/*|wallet/zuuallet/src/hooks/useWallet.ts|wallet/zuuallet/src/lib/mnemonic.ts|wallet/zuuallet/src/lib/sensitive-seed-session.ts|wallet/zuuallet/src/lib/sensitive-seed.ts|wallet/zuuallet/src/lib/tauri.ts|wallet/zuuallet/src/pages/CreateWallet.tsx|wallet/zuuallet/src/pages/RestoreWallet.tsx|wallet/zuuallet/src/pages/Settings.tsx|wallet/zuuallet/src/pages/Welcome.tsx|wallet/zuuallet/src/types/index.ts|wallet/rust-toolchain.toml|wallet/deny.toml|scripts/check-github-actions-pins.mjs|scripts/check-librustzcash-compat.mjs|scripts/check-rust-toolchain.sh|scripts/check-rust-fmt.sh|scripts/check-rust-deny.sh|scripts/check-rust-clippy.sh|scripts/check-zcash-permissions.mjs|scripts/check-zuuli-linux-image.mjs|scripts/check-zuuli-android-gate.mjs|z/zcash/librustzcash|.gitmodules|.github/containers/zuuli-linux/*|.github/workflows/zuuli.yml|.github/workflows/zuuli-linux-image.yml|.github/workflows/zuuli-packaging.yml|.github/workflows/zuuli-release.yml|.github/workflows/zuuli-store-audit.yml|.github/workflows/zuuli-store-publish.yml|.github/workflows/zuuli-testflight-bootstrap.yml|.github/workflows/zuuli-testflight-recovery.yml|.github/workflows/cache-cleanup.yml|.github/actions/zuuli-rust-cache/*|docs/ZUULI-LINUX-BUILD-IMAGE.md)
+              Cargo.toml|Cargo.lock|.cargo/*|clippy.toml|.clippy.toml|wallet/Cargo.toml|wallet/Cargo.lock|wallet/.cargo/*|wallet/clippy.toml|wallet/.clippy.toml|wallet/*|wallet/*.rs|wallet/*/Cargo.toml|wallet/*/Cargo.lock|wallet/*/.cargo/*|wallet/*/clippy.toml|wallet/*/.clippy.toml|wallet/zuuli/*|wallet/plugins/*|wallet/zuuallet/src/hooks/useWallet.ts|wallet/zuuallet/src/lib/mnemonic.ts|wallet/zuuallet/src/lib/sensitive-seed-session.ts|wallet/zuuallet/src/lib/sensitive-seed.ts|wallet/zuuallet/src/lib/tauri.ts|wallet/zuuallet/src/pages/CreateWallet.tsx|wallet/zuuallet/src/pages/RestoreWallet.tsx|wallet/zuuallet/src/pages/Settings.tsx|wallet/zuuallet/src/pages/Welcome.tsx|wallet/zuuallet/src/types/index.ts|wallet/rust-toolchain.toml|wallet/deny.toml|scripts/check-github-actions-pins.mjs|scripts/check-librustzcash-compat.mjs|scripts/check-rust-toolchain.sh|scripts/check-rust-fmt.sh|scripts/check-rust-deny.sh|scripts/check-rust-clippy.sh|scripts/check-zcash-permissions.mjs|scripts/check-zuuli-linux-image.mjs|scripts/check-zuuli-android-gate.mjs|z/zcash/librustzcash|.gitmodules|.github/containers/zuuli-linux/*|.github/workflows/zuuli.yml|.github/workflows/zuuli-linux-image.yml|.github/workflows/zuuli-packaging.yml|.github/workflows/zuuli-release.yml|.github/workflows/zuuli-store-audit.yml|.github/workflows/zuuli-store-publish.yml|.github/workflows/zuuli-testflight-bootstrap.yml|.github/workflows/zuuli-testflight-recovery.yml|.github/workflows/cache-cleanup.yml|.github/actions/zuuli-rust-cache/*|docs/ZUULI-LINUX-BUILD-IMAGE.md)
                 zuuli=true
                 ;;
             esac
@@ -128,7 +130,7 @@ jobs:
                 zuuallet_schema=true
                 ;;
             esac
-          done <<< "$changed_files"
+          done < "$changed_files"
 
           echo "zuuli=$zuuli" >> "$GITHUB_OUTPUT"
           echo "zuuallet_schema=$zuuallet_schema" >> "$GITHUB_OUTPUT"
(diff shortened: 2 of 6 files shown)


skyl:
Reviewed adversarially on head `271730d0`, verified by running the scripts in a throwaway worktree rather than reading the body. **Sound — merge when `gate` and `rs / gate` are green.** Two things to flag, one of which is a deliberate tradeoff that deserves to be stated out loud rather than left inside a glob.

### This closes the follow-up I asked for on #549, and it closes it properly

My review of #549 ended with: registration is explicit by design, so an **unregistered** root is silently invisible — "#543's own stated failure mode is still reachable; this PR makes it fixable, not impossible. Add an assertion that every tracked `Cargo.toml` outside `z/` lives under a policed root."

That is now a control. Tracked a crate at `tools/rogue/Cargo.toml` outside both roots:

```
$ bash scripts/check-rust-toolchain.sh
EXIT=1
drift: tools/rogue/Cargo.toml belongs to 0 policed Rust roots; every first-party
  Cargo.toml must belong to exactly one of: wallet rs
drift: tools/rogue/Cargo.toml declares [package] but is registered nowhere, so nothing
  holds its rust-version to wallet/rust-toolchain.toml. Add it to MANIFESTS in
```

Two independent diagnostics, both actionable. Note the probe had to be `git add`ed to be seen — the census reads the real index, which is correct and is why an untracked scratch crate does not trip it.

### The `--root` flags are policed, not decorative

Moving `--root wallet` into the workflow steps would normally worry me, because #600's whole point was that registrations live in the script so "a flag is provably not a registration". Here the flag is bound and verified. Stripping it:

```
256:        run: scripts/check-rust-fmt.sh          ← --root wallet removed

$ node scripts/check-github-actions-pins.mjs
EXIT=1
.github/workflows/zuuli.yml:225: wallet/ owner job rust_fmt must run exactly one
  unconditional, non-soft-failing scripts/check-rust-fmt.sh --root wallet
```

"exactly one unconditional, non-soft-failing" is the right shape — it forecloses adding a second soft-failing invocation alongside.

### Baselines, run rather than quoted

```
check-github-actions-pins.mjs --self-test  → 74 source-policy, 10 gate-verdict,
                                              134 current-workflow, 78 Rust-root ownership
check-rust-toolchain.sh --self-test        → 14 drift case(s) caught
check-rust-fmt.sh --self-test              → 6 case(s) passed; rustfmt from 1.97.1
```

All exit 0. `scripts/check-workflow-gates.mjs` is untouched, so #622's context registry and digest are unaffected, and the branch is exactly one commit on current `main`.

### Flag 1 — `wallet/*` broadens the change filter, and the cost is real

The new glob in the `changes` selector:

```diff
-…|wallet/.clippy.toml|wallet/*.rs|…
+…|wallet/.clippy.toml|wallet/*|wallet/*.rs|…
```

In a `case` pattern `*` crosses `/`, so **every** path under `wallet/` now sets `zuuli=true`. The comment says so plainly ("deliberately selects every wallet/ path, including future crates and Zuuallet-only changes"), and as correctness this is right — it is the thing that stops a future crate being invisible, which is the whole point of #608.

But it should be stated as the tradeoff it is. `rust_native_tests` is gated on `needs.changes.outputs.zuuli == 'true'`, so a Zuuallet-only frontend or doc change now pulls in the macOS and Windows native matrix. I measured that matrix this session: `Rust / native tests (macos)` sat **queued for ~9 minutes and ran ~12**, and a full `wallet/zuuli` run is **25-40 minutes** end to end under current runner contention. That is now the floor for any wallet-touching PR.

Correctness over throughput is the right call here and I am not asking for it to be reverted. I am asking that it not be discovered later as a surprise — if the cost bites, the fix is a narrower positive selector (`wallet/**/Cargo.toml`, `wallet/**/*.rs`, plus the explicit file list) rather than reverting to a hand-maintained list, since the hand-maintained list is exactly what #608 is about.

### Flag 2 — the toolchain self-test still under-reports its own run

```
$ grep -c '^self-test: caught' → 17
$ printed                      → "self-test: 14 drift case(s) caught."
```

I checked before calling this a bug: `${#SELF_TEST_MUTATIONS[@]}` is 14 and legitimately counts that array; the three extra cases are the multi-root ones emitted separately at `check-rust-toolchain.sh:1007-1021`. So no guard is broken.

It is still a recorded number that does not describe what ran, which is the rot this repo keeps catching elsewhere — and I raised the same line in my #549 review when it read 14 against 18. This PR adds cases without touching it. Either count both sections or reword to "14 registered drift mutation(s)" so the sentence stops claiming to be a run summary. Contrast `check-github-actions-pins.mjs`, which enumerates all four of its categories and is exactly right.

### What I did not observe

`gate` has not completed on this head. Given the size of the change to `check-github-actions-pins.mjs` (+1143) and that it edits the gate's own scripts, I will merge only on `gate` and `rs / gate` verified green by run ID on `271730d0`, and will not use the assembled-evidence argument.

skyl:
**Blocking — this fails on the unmodified tree.** `gate` and `rs / gate` are both `failure` on `271730d0` (zuuli run `32724272355`, rs run `32724272226`), and it is a genuine failure, not a cancellation or a race.

### The failing step and the exact cause

`changes` → **"Verify the Rust toolchain pin has not drifted"** → exit 1:

```
drift: rs/Cargo.toml could not be classified by pinned Cargo 1.97.1:
  error: manifest path `/tmp/rust-toolchain-check.pA9sYU/baseline/rs/Cargo.toml`
  is a virtual manifest, but this command requires running against an actual
  package in this workspace
1 pin(s) disagree with wallet/rust-toolchain.toml (channel 1.97.1, MSRV 1.97).
self-test FAILED: the unmodified tree must pass the drift check.
```

`rs/Cargo.toml` **is** a virtual manifest — `[workspace]`, `resolver = "3"`, `members = ["crates/*"]`, and zero `[package]` sections. The new "classify package versus virtual manifests with Cargo semantics" step shells out to the pinned Cargo, Cargo answers with the *expected* "is a virtual manifest" error, and the classifier treats a non-zero exit as *unclassifiable* and fails closed into `drift:`.

Failing closed is the right instinct. The bug is that "Cargo refused because this is a virtual manifest" is a **successful classification** — it is Cargo telling you precisely which of the two categories the file is in — and it is being read as "I could not tell."

The self-test then correctly reports `the unmodified tree must pass the drift check`, so the guard caught its own defect. That part worked.

### Why the mutation table did not catch this

The body lists coverage for "Cargo-valid quoted, spaced, inline, and dotted package spellings; malformed manifests; **multiline-string virtual false positives**". So virtual manifests were considered — as *fixtures*. What was never exercised is the virtual manifest that has been sitting in this repo since #341: `rs/Cargo.toml` itself.

That is the failure mode this repo keeps naming, in a new spot: the fixtures range over synthetic cases the classifier was written against, so the invariant holds across all of them, while the real instance in the tree falls outside the set. A census whose whole purpose is "every tracked first-party manifest" needs its baseline assertion to be *the actual tracked set*, which is what the self-test's `unmodified tree must pass` line is for — and that line is what went red.

### My own gap, for the record

I verified this branch in a worktree and reported it sound. I ran all three `--self-test` suites, the unowned-crate mutation, and the `--root wallet` stripping mutation — but I ran the **live** `check-rust-toolchain.sh` only *with my `tools/rogue` probe present*, where exit 1 was the expected answer. I never ran the live check on the clean branch tree, which is exactly what CI did and exactly where this surfaces. Verifying a guard fires on a mutation is not the same as verifying it passes on the unmutated baseline, and I checked only one of the two.

### The fix

Treat Cargo's virtual-manifest refusal as the classification it is, not as an error. Something like: if the pinned Cargo exits non-zero **and** its stderr matches the virtual-manifest diagnostic, classify as virtual and continue; keep failing closed for every other non-zero exit, which preserves the malformed-manifest behaviour the body describes.

Then add the case that would have caught this: assert the classifier returns *virtual* for `rs/Cargo.toml` **by path**, not for a synthetic fixture. A census anchored to the real tree is the only version of this that cannot drift away from it.

Happy to re-review as soon as it is pushed — the ownership census underneath this is genuinely good work and closes the #549 follow-up properly. The verification I did stands: the unowned-crate diagnostic and the `--root wallet` binding both fire correctly. It is only the manifest classifier that is wrong.

skyl:
Re-reviewed on `41e7d167`. **The fix is correct, and it is better than the one I proposed. Approving — merge when `gate` and `rs / gate` are green.**

### First: my root-cause diagnosis was wrong, and the fix shows why

I wrote that the classifier "treats ANY non-zero exit as unclassifiable and fails closed". That was not true. The original code **did** have a virtual-manifest branch:

```bash
virtual_suffix='` is a virtual manifest, but this command requires running against an actual package in this workspace'
if [[ $diagnostic == 'error: manifest path `'*"$virtual_suffix" ]]; then
```

It failed because the pattern is anchored at the **start of the entire stderr capture**, and on CI rustup had to fetch the pinned toolchain first, so stderr began with its prelude. The CI diagnostic shows it plainly:

```
info: syncing channel updates for 1.97.1-x86_64-unknown-linux-gnu;
info: latest update on 2026-07-16 for version 1.97.1 (8bab26f4f 2026-07-14);
info: downloading 3 components;
error: manifest path `…/rs/Cargo.toml` is a virtual manifest, …
```

The `error:` line was there. It just was not first. That is a **warm-local versus cold-CI** failure — the branch works on any machine that already has 1.97.1 installed and fails on every machine that does not, which is exactly why it passed for me locally and died in CI.

**And my proposed fix would have had the same bug.** I suggested matching the virtual-manifest diagnostic in stderr; unless I had also stripped rustup's prelude, that is the identical anchoring mistake in new clothes. Worth recording, because the lesson is not "match the string better" — it is that diagnostic text is not an API.

### The fix takes the right lesson

```bash
if cargo +"$channel" metadata --no-deps --format-version 1 \
  --manifest-path "$root/$rel" >/dev/null 2>"$metadata_diagnostic_file"; then
  CARGO_MANIFEST_KIND=virtual
```

A second **semantic** command, decided on exit status. The comment states the property directly: "Classification never depends on Cargo/rustup diagnostic wording, ANSI colour, or startup prelude; both commands failing is malformed or otherwise unclassifiable and fails closed." That is structurally immune to what broke it, not merely patched against the observed instance, and it preserves fail-closed for genuinely malformed manifests.

Verified against the real file rather than a fixture:

```
cargo +1.97.1 read-manifest --manifest-path rs/Cargo.toml   → EXIT=101
cargo +1.97.1 metadata --no-deps --manifest-path rs/Cargo.toml → EXIT=0   ⇒ virtual
```

### Clean-tree baseline — the check I skipped last time

This is where I went wrong before: I ran the mutations but never the live check on the unmutated branch tree, which is precisely what CI runs.

```
bash scripts/check-rust-toolchain.sh        → EXIT=0
  "Every Rust toolchain pin agrees with wallet/rust-toolchain.toml: channel 1.97.1, MSRV 1.97."
```

Then the self-tests, all exit 0:

```
check-github-actions-pins.mjs --self-test  → 74 source-policy, 10 gate-verdict,
                                              134 current-workflow, 78 Rust-root ownership
check-rust-toolchain.sh --self-test        → 14 drift case(s) caught
check-zuuli-android-gate.mjs --self-test   → 33 mutations
```

And both guards still fire on `41e7d167`:

```
tracked tools/rogue/Cargo.toml  → EXIT=1  "belongs to 0 policed Rust roots …
                                           must belong to exactly one of: wallet rs"
                                          "declares [package] but is registered nowhere"
--root wallet stripped (line 256) → EXIT=1  "wallet/ owner job rust_fmt must run exactly one
                                           unconditional, non-soft-failing … --root wallet"
```

### The digest bump is done the right way

`check-zuuli-android-gate.mjs`'s `changeDetectorDigest` had to move, since the `changes` step was rewritten. What makes it acceptable is that it does not just re-stamp the hash — it adds two mutations that make the new shape load-bearing:

```js
["the change detector restores a line-delimited Git producer",
 "git diff --name-only -z --no-renames", "git diff --name-only --no-renames"],
["the change detector restores a line-delimited consumer",
 "while IFS= read -r -d '' file; do", "while IFS= read -r file; do"],
```

So reverting the NUL-safety goes red. A digest updated together with the controls that prove what it now protects is the opposite of the pattern I keep filing here.

### My two earlier flags stand, both non-blocking

1. `wallet/*` in the `changes` selector still broadens `zuuli=true` to every path under `wallet/`, and `rust_native_tests` keys off it, so a Zuuallet-only change now pulls the macOS/Windows matrix (~25-40 min under current contention). Correct call; just don't let it be a surprise later.
2. `check-rust-toolchain.sh` still prints "14 drift case(s) caught" while emitting 17 `self-test: caught` lines. I verified 14 correctly counts `${#SELF_TEST_MUTATIONS[@]}` and the extra three are the multi-root cases at `:1007-1021`, so no guard is broken — but the sentence still under-reports its run.

### What I have not observed

`gate` has not completed on `41e7d167`. Given this edits the gate's own scripts, I will merge only on `gate` and `rs / gate` verified green by run ID on this exact head, and will watch main's own push run afterwards.

PR #626, opened 2026-08-24
feat(kt): add experimental authority assertion layer

### LINKED ISSUE #603: security(rs): the inert strict-verification test reappeared in f2z-authority — make the check workspace-wide

### LINKED ISSUE #604: f2z-authority: admit refuses routine same_key/key_change entries, four field-order swaps pass green, and keygen writes the issuing key 0644
Three defects in `f2z-authority` (#593, closed by the sweep; branch `feature/kt-authority` survives). The crate is otherwise well covered — 21 of 26 mutations caught, the authorization table genuinely exercised, and #590's ordering right by construction (`observe` last, proved by `a_failed_submission_never_burns_the_nonce`).

## 1. `admit` refuses every `same_key` and `key_change` entry on a vouched log

Rule 1 is implemented as *"an assertion is required iff the log has an authority"* rather than *"iff the entry kind needs one"*. Measured against a vouched log:

```
same_key      v5                          -> Err(MissingAssertion)
same_key      v5 + a reset assertion      -> Err(IdentityUnchanged)
key_change    v5                          -> Err(MissingAssertion)
```

So a routine **device addition** or a **user-initiated ISK rotation** can only pass if the *platform* mints a reset assertion — which inverts ADR 0014's stated consequence that *"Case 2 needs a signature the platform does not have."* The whole point of the case split is that the user's own key authorizes these, and the platform's authority is the exception.

`Submission` already carries `entry_version` and `previous_identity_pk`, so the information is present. It needs an `EntryKind` discriminator — or `admit` must stop presenting itself as the only door.

## 2. Four wire-format field-order swaps pass green

This is #581's gap reproduced in a new crate. Each of these changes what every signature over the structure commits to, and all four pass the whole suite:

| Swap | |
|---|---|
| `handle_id` ⇄ `identity_pk` | NOT CAUGHT |
| `issued_ms` ⇄ `expires_ms` | NOT CAUGHT |
| `assertion_digest` ⇄ `entry_digest` | NOT CAUGHT |
| `authority_id` ⇄ `log_id` | NOT CAUGHT |

Only one hand-derived byte vector exists in the crate, for `Handle`. #581 established the fix and the standard: expectations derived **by hand from the spec text**, never printed from the implementation, with the derivation visible per field (spec declaration, the width it fixes, the bytes that width holds) and a helper that checks the stated width against the supplied bytes *before* comparing anything to the code.

`issued_ms` ⇄ `expires_ms` is the one to look at first — swapping a validity window's endpoints is the kind of thing that produces credentials valid in the past and expiring in the future, and nothing currently notices.

## 3. `f2z-assert keygen --out` writes the issuing private key world-readable

Measured: `-rw-r--r--` (0644). This is the **issuing authority's** private key — the key that mints handle assertions. It should be created 0600, and ideally with `O_EXCL` so it cannot silently overwrite an existing key.

Worth checking whether the file is created and then chmod'ed, since that leaves a window where the key exists at 0644; `OpenOptions::mode(0o600)` at creation avoids it.

## Acceptance

- `admit` distinguishes entry kinds, so `same_key` and `key_change` on a vouched log succeed with the user's own authorization and without a platform assertion — with a test per kind.
- Hand-derived byte vectors for every structure whose bytes are signed, and all four swaps above go red.
- `keygen` creates the private key 0600 at open time, and refuses to clobber an existing file.

## Not re-filed here

The inert strict-verification test is **#603** (systemic — the same defect as #589, now in two crates). The `KT.md` label prefix collision the crate's own test half-covers is **#602**. #594 already files the entry-1 spec gap.

Also worth recording as sound: `#550`'s KELVIN case is **structurally unreachable** in this crate — no `from_username`, no case folding, and an exhaustive 256-byte charset test. And `main`'s post-#597 `workspace_debug_scan` passes against this tree, so no action needed there after a rebase.

Found while reviewing #593.

skyl:
Closes #603
Closes #604

## Summary

Adds an experimental, standalone f2z-authority crate and CLI for authority assertions without wiring it into the production KT admission path.

- distinguishes InitialBind, SameKey, KeyChange, and PlatformReset admission rules
- permits user-authorized same-key and key-change entries without a platform reset assertion
- uses strict Ed25519 verification with a workspace-wide definition-aware verifier census
- adds hand-derived canonical vectors for every signed structure and independent digest/transcript vectors
- creates issuing keys at the exact requested path with O_EXCL and mode 0600
- exercises real CLI dispatch, parsing, raw/hex key input, field mapping, timestamps, epochs, nonce handling, and output formats

## Scope

This remains an experimental standalone leaf. It does not claim to solve first-entry authorization (#594), production KT integration, or complete E2EE chat authorization.

## Mutation proof

Every security/format guard was broken and shown red before publication. Independent review covered:

- strict to plain verification and identity-point forgery at both key and admission boundaries
- duplicate same-name verifier definitions, wrong receivers, dead/source-only verifier registrations
- entry-kind, epoch, predecessor-vouch, identity-binding, and nonce-order mutations
- every signed-field ordering and production AuthorityConfig binding-map swap/reversal
- CLI dispatch, unknown/positional options, raw/hex mapping, output path/mode/no-clobber, explicit expiry precedence, byte-order, and random-zeroing mutations
- independent nonuniform authority-ID and assertion-digest vectors rather than implementation-derived expectations

Restored workspace tests/docs, WASM, fmt, clippy, deny, toolchain, Actions, and Android policy gates pass. Exact independently approved head: ed57c515b040b6488b209e07131d753b0a6cfb02.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -319,7 +319,8 @@ jobs:
           cargo build --locked --lib \
             --target "${{ steps.rust_toolchain.outputs.target }}" \
             -p f2z-codec \
-            -p f2z-relay-proto
+            -p f2z-relay-proto \
+            -p f2z-authority
 
   # Branch protection should require this stable context, `rs / gate`. It
   # succeeds only when the change detector and every applicable rs job
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -171,6 +171,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-authority"
+version = "0.1.0"
+dependencies = [
+ "ed25519-dalek",
+ "f2z-codec",
+ "tls_codec",
+]
+
+[[package]]
 name = "f2z-codec"
 version = "0.1.0"
 dependencies = [
--- a/rs/README.md
+++ b/rs/README.md
@@ -112,11 +112,14 @@ prevent.
 |---|---|
 | [`crates/f2z-codec`](./crates/f2z-codec) | The canonical encoding layer of `WIRE.md`: `tls_codec` wrappers for every wire structure, the domain-separated signing transcript (§5), the redacting newtypes, and the padding-bucket validator (§9). `no_std` + `alloc`, `#![forbid(unsafe_code)]`, no I/O, no async runtime, and it builds for `wasm32-unknown-unknown`. |
 | [`crates/f2z-relay-proto`](./crates/f2z-relay-proto) | The protocol layer above it: signed-command construction and verification in §5.1's exact order, the timestamp window and fail-closed seen-set (§5.5), the queue lifecycle and ACK arithmetic (§7, §8), the capability document (§11), the `HELLO` proof of possession (§5.2), and §4.3's typed in-flight window. Same constraints — `no_std` + `alloc`, no I/O, no clock, no randomness, and it builds for `wasm32-unknown-unknown`. |
+| [`crates/f2z-authority`](./crates/f2z-authority) | An **experimental candidate** for the directory's non-cryptographic trust-root layer: the proposed `HandleAssertion`, a partial assertion-layer check, `AuthoritySet`, and `f2z-assert`. `KT.md` does not yet ratify these structures or first-entry/no-authority semantics (#594), and its result is not §4.4 directory authorization. Same portability constraints. |
 
-**The relay and the clients link both.** That is the licence boundary in
-practice: everything here is MIT because a third-party relay, ZUULI and the WASM
-web client all compile the same rules, and a rule that two implementations
-disagree about is how ciphertext gets deleted before it is read.
+The current dependency graph is narrower than that intended architecture:
+`f2z-relay-proto` depends on `f2z-codec`, while `f2z-authority` is still a
+standalone experimental leaf and no workspace package depends on it. Wiring the
+authority candidate into a relay or client is future integration work, not a
+property this README claims today. All three remain MIT so downstream relays and
+clients can share the rules once that integration exists.
 
 `f2z-codec` is separate from everything that sits on top of it for three
 reasons, and each is enforced by a test rather than by intent:
@@ -136,7 +139,8 @@ reasons, and each is enforced by a test rather than by intent:
 ```bash
 cd rs
 cargo test
-cargo build --target wasm32-unknown-unknown --lib -p f2z-codec -p f2z-relay-proto
+cargo build --target wasm32-unknown-unknown --lib \
+  -p f2z-codec -p f2z-relay-proto -p f2z-authority
 ```
 
 The gates are the repository's shared scripts, pointed here with `--root`:
--- a/rs/crates/f2z-authority/Cargo.toml
+++ b/rs/crates/f2z-authority/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "f2z-authority"
+version = "0.1.0"
+description = "Handle-ownership assertions for the free2z key-transparency log: the one non-cryptographic trust root, isolated so it can be read in full"
+edition.workspace = true
+# A restatement of wallet/rust-toolchain.toml's channel in Cargo's MSRV form,
+# registered in `scripts/check-rust-toolchain.sh`'s `MANIFESTS` auto-census so
+# it cannot drift. Not inherited: that check reads this literal line.
+rust-version = "1.97"
+# Permissive, deliberately and explicitly, so clients and third-party relays
+# can eventually share the same assertion verifier. This experimental crate is
+# currently a standalone leaf: no browser, ZUULI, or log package depends on it
+# yet. The AGPL-3.0 boundary starts at server binaries, not here. Stated
+# literally rather than inherited so it is visible to anyone opening the
+# crate; rs/deny.toml enforces it.
+license = "MIT"
+repository.workspace = true
+publish.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+ed25519-dalek = { workspace = true }
+# `version` beside `path` is not redundant: rs/deny.toml sets
+# `wildcards = "deny"`, and a bare path dependency is a wildcard requirement.
+f2z-codec = { path = "../f2z-codec", version = "0.1.0" }
+tls_codec = { workspace = true }
+
+# The example issuer. It is a `bin` in this package rather than a crate of its
+# own so that a self-hoster reading "how do I issue an assertion by hand" reads
+# the same source tree that defines what an assertion is. The library is
+# `no_std`; this target is not, and that is fine — `rs.yml`'s wasm job builds
+# `--lib`, so a browser never compiles it.
+[[bin]]
+name = "f2z-assert"
+path = "src/bin/f2z-assert.rs"
--- a/rs/crates/f2z-authority/src/assertion.rs
+++ b/rs/crates/f2z-authority/src/assertion.rs
@@ -1,0 +1,631 @@
+//! The two signed structures: what an authority says, and what the identity
+//! key it names says back.
+//!
+//! **Experimental proposal, not `KT.md` v1 wire format.** `KT.md` leaves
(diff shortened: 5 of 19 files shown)


skyl:
Reviewed adversarially on `ed57c515`, verified by running the crate and the workspace scans in a throwaway worktree. **Sound — merge when `gate` and `rs / gate` are green.** This is the strongest PR I have reviewed in this session, and I want to be specific about why rather than just approving it.

### The scope discipline is right, and it is the part I checked first

The body says this "does not claim to solve first-entry authorization (#594), production KT integration, or complete E2EE chat authorization." I verified that is true structurally, not just asserted:

```
$ grep -rn 'f2z-authority' rs/crates/*/Cargo.toml   # excluding its own manifest
(no results)
```

Nothing depends on it. Outside the crate, the diff touches only `rs.yml`, `rs/Cargo.lock`, `rs/README.md`, `scripts/check-rust-toolchain.sh`, and the f2z-codec workspace scan. **#594's authority model remains an open owner decision and this PR does not quietly settle it** — which is exactly what I wanted, because that decision is not an agent's to make.

### #625's census caught the very first crate added after it landed

This is a nice demonstration of a control working in anger. Adding `rs/crates/f2z-authority/Cargo.toml` *forced* a registration edit. Removing that registration:

```
drift: rs/crates/f2z-authority/Cargo.toml declares [package] but is registered nowhere,
so nothing holds its rust-version to wallet/rust-toolchain.toml. Add it to MANIFESTS in
```

Clean-tree baseline first (the check I skipped on #625's first head): `bash scripts/check-rust-toolchain.sh` → **EXIT=0**. Then the mutation → **EXIT=1**. Both directions.

### It upgrades #612's scanner rather than working around its blind spot

`code_only` documented its own limitation:

> a raw string with an embedded quote (`r#"..."#`) is not modelled, and a block comment is not modelled. **Neither appears in the tree today**

A detector that is sound only because the tree happens not to contain the shape it cannot parse is one commit from being wrong — and this PR is exactly the commit that would have introduced hand-derived vectors full of raw strings. Instead of dodging it, `code_lines` models raw strings (including `b`/`c` prefixes and `#` counts) and nested block comments, keeps physical line numbers so reports stay actionable, and **refuses rather than guesses**: "A malformed unterminated string or block comment is a refusal, not permission to scan a projection whose state is unknown."

And the new crate is genuinely policed. Mutating its wrapper to plain `verify`:

```
FAILED. 8 passed; 2 failed

every_verify_function_in_the_workspace_is_registered_as_strict:
  f2z-authority/src/key.rs::verify is registered as calling `verify_strict` and its body
  no longer does. Every other verification in this workspace delegates to it, so this is a
  universal forgery away from `admit`-style authorization returning Ok for a key nobody holds.

no_crate_verifies_ed25519_signatures_non_strictly:
  f2z-authority/src/key.rs:62: calls `.verify(` on `self.0`, which this file declares as an
  `ed25519_dalek::VerifyingKey`; that is the non-strict function. Use `verify_strict`
```

Two independent tests, one on the registry and one on the source, both naming the consequence rather than the rule.

### Two tests that guard their own preconditions — this is the part worth copying

Most of what I file in this repo is tests that cannot fail. These are the opposite.

**The strictness test proves its fixture still matters before relying on it:**

```rust
assert!(raw.is_weak(), "the fixture stopped being a small-order point");
…
assert!(raw.verify(&message, &dalek_signature).is_ok(),
  "message {byte}: plain verification refused the fixture, so the test no longer proves strictness matters");
assert_eq!(verifier.verify(&message, &forged, …), Err(AuthorityError::BadIdentitySignature),
  "message {byte}: strict verification accepted a universal forgery");
```

If the fixture ever stops being a universal forgery under plain `verify`, the test says so out loud instead of passing vacuously. That plain `.verify(` is a deliberate negative control, not a defect — worth noting for anyone grepping later.

**The keygen mode test proves the environment can express the failure:**

```rust
let probe = OpenOptions::new().create_new(true).mode(0o666).open(&probe_path)?;
assert_eq!(probe_mode, 0o666, "the isolated subprocess did not start with umask 000");
let file = open_private_key(&key_path)?;
assert_eq!(mode, 0o600, "the file was visible with mode {mode:o} at open");
```

Asserting `0o600` alone would pass on any machine whose umask masked the bits regardless of what the code asked for. Establishing `0o666` under umask 000 first is what makes the second assertion mean something. That closes #604's third defect properly rather than nominally.

### Baselines

`cargo test -p f2z-authority` → **107 tests, 0 failed** (46 + 36 + 11 + 8 + 4 + 1 + 1). Branch is exactly one commit on current `main`, which already includes #625.

### What I did not observe

`gate` has not completed on this head. I have not independently re-derived the hand-derived canonical vectors — I checked that they are asserted against independent values rather than implementation output, but I did not recompute the digests myself. And I did not exercise the CLI surface beyond the crate's own tests.

PR #627, opened 2026-08-24
fix(zuuli): mutation-prove native send routes

### LINKED ISSUE #547: test(zcash): the send path has no behavioural coverage — four money-affecting mutations ship silently

### LINKED ISSUE #620: zuuli send guards: privacy check accepts pub(crate), exact-count checks are string-forgeable, recipient-ordering assertion lost
Three defects in `wallet/zuuli/scripts/` verified against #617's tree (`fix/547-send-authority`, `32fe463`). None blocks that PR — the module-privacy boundary it introduces is a genuine improvement and all five money mutants still die — but each is a guard written to a mutant rather than to the property it names.

### 1. The privacy check greps for the mutant, not the invariant

The "must remain in one private child module" check greps for the literal string `pub mod native;`. `pub(crate) mod native;` therefore **passes** it.

That is harmless today only because `pub(crate)` still keeps `native` out of the *other* crate; it is not harmless as a control, because the whole argument for deleting `send-review-boundary.node-test.mjs` was that privacy is enforced. A guard that names a property ("remains in one private child module") and tests one spelling of its negation will pass the next spelling.

Assert the declaration is exactly `mod native;` — anything else, including `pub(crate)`, `pub(super)`, and `pub(in …)`, fails.

### 2. `stripRustComments` preserves string contents, so exact-count checks are string-forgeable

The replacement scanner strips comments but leaves string literals intact, so a string literal counts toward the exact-count orchestration checks.

Demonstrated on that tree: with the real fixed-send call replaced and the matching literal parked inside a string, **the fixed-send error did not fire**. The count was satisfied by text that never compiles to a call.

This is precisely the forgery class the removed `rustCodeOnly` scanner had a self-test for. Removing the scanner removed the self-test with it, so the replacement inherited the vulnerability without inheriting the test that would have caught it. Blank string contents (preserving length, or replacing with a same-length filler) before counting, and port the forgery self-test onto the new scanner.

### 3. The recipient-validation ordering assertion was lost with no replacement

`send-review-boundary.node-test.mjs` asserted `validationIndex < proposalIndex` — that the recipient is validated **before** the native proposal is built. #617's node test is modified rather than deleted (7 → 6 tests), but that assertion is not among the survivors and nothing else covers it.

Ordering is not implied by module privacy: `native::propose_with_policy` being unreachable from outside says nothing about *when* `send.rs` calls it relative to validating the recipient. An inversion here means a proposal is constructed against an unvalidated address — the failure is silent, since both orderings compile, both pass every other guard, and the money mutants all target policy values rather than call order.

This is the one thing #617 strictly gives up. Restore it, ideally as a behavioural test in `native.rs`'s suite (an invalid recipient must fail before any proposal is built) rather than as a text-index comparison, so it cannot be defeated by the same forgery as item 2.

### Also, unchanged from #616 item 3

`send.rs:1450` still holds a second unrouted `ConfirmationsPolicy::default()`, and `ConfirmationsPolicy` is deliberately absent from the new `forbiddenOutsidePolicyModule` census (adding it would trip `send.rs:15`). Tracked in #616; noted here because #617's title claims the policy authority is unique and this is the one money-relevant policy decision that is still duplicated by design.

Found reviewing #617. Related: #613, #616, #547.

skyl:
Closes #547
Closes #620

## Summary

- routes fixed sends and send-all through one private, compiler-bound native policy authority
- behaviorally exercises ordinary fixed/send-all/creation values against a real librustzcash test database
- verifies confirmation depth, ZIP-317 fee, pool selection, shielded change, sender OVK recovery, persisted transaction bytes, and exact retry bytes
- adds a shipping-mode integration probe that traverses the stateful production callers while keeping money authority private
- binds the complete WalletState adapter bodies so cfg/dead/early-return and post-call corruption cannot hide behind tested helpers

## Mutation proof

Every money-path guard was broken and shown red before publication. Independent review killed:

- fixed/send-all/creation pre- and post-route ordinary-value guards
- shipping-only cfg splits and dead parked helper calls in all three WalletState adapters
- fixed/send-all post-helper fee corruption
- execute-path raw-transaction corruption before persistence/broadcast
- confirmation-depth, recipient presence/order, policy alias/shadow, visibility, pool/fee/change, and sender-recovery mutations
- cfg test-only splits and disconnected helper seams

The compatibility self-test kills 108/108 mutants and the live checker passes. Rust 1.97.1 feature-enabled all-targets passes 145 unit tests plus the shipping integration test; feature clippy, fmt, toolchain, Actions, workflow, public-boundary, and deny gates pass. Exact independently approved head: a0a361ceba77565f47cd7d67c26717019190cd42.

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -482,6 +482,7 @@ jobs:
         run: >-
           cargo test --locked
           --all-targets
+          --features production-route-probe
           --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
 
   rust_plugin:
@@ -536,7 +537,7 @@ jobs:
           key: zuuli-plugin
 
       - name: Build and test shared Zcash plugin
-        run: cargo test --locked --all-targets --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
+        run: cargo test --locked --all-targets --features production-route-probe --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml
 
   # PR packaging builds only arm64 Android. Keep one cheap 32-bit type-check in
   # the required gate because libc's Android ABI types can differ on armv7; the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -156,12 +156,25 @@ integrate, and move it forward alongside everything else.
   typecheck+build; Rust `cargo build` of the Tauri backend through the
   librustzcash path deps) on every change to the app, its deps, or the
   `z/zcash/librustzcash` submodule pointer. If a bump breaks us, CI says so.
-- **`scripts/check-librustzcash-compat.mjs` is a source-text change detector,
-  not send-path behavioural coverage.** It pins the reviewed librustzcash
+- **`scripts/check-librustzcash-compat.mjs` is a source-identity change detector,
+  not a transaction oracle.** It pins the reviewed librustzcash
   gitlink, the exact versions of 11 Zcash packages in each of the 3 shipping
   lockfiles that contain that graph, and the API-adaptation source shapes that
-  accompanied the pin. It does not compile or exercise a transaction; the
-  money-affecting behavioural coverage remains tracked in **#547**. To bump
+  accompanied the pin. It also keeps production orchestration compiler-bound
+  to the private native send routes, but it does not prove their behavior.
+  Money-affecting coverage lives in the Rust tests, including
+  `ordinary_values_cross_stateful_shipping_callers_and_exact_recovery_bytes`,
+  an integration test that compiles the plugin library without `cfg(test)` and
+  exercises ordinary fixed-send, send-all, and transaction-creation values
+  through the exact generic stateful production callers. It also proves the
+  returned retry bytes exactly equal the independently re-serialized persisted
+  transaction, rather than merely checking that both artifacts exist. Required
+  plugin test jobs enable its CI-only `production-route-probe` feature; default
+  shipping builds expose none of the private caller authority. The same checker
+  also binds each complete `WalletState` adapter body with a comment- and
+  literal-insensitive digest plus ordered transition anchors, so a feature-only
+  route, dead helper, early return, or post-helper mutation cannot leave the
+  behavior-tested private caller parked behind a green call count. To bump
   librustzcash: move the submodule to reviewed upstream HEAD, regenerate all
   three lockfiles on the pinned toolchain, and inspect the resolved graph before
   changing any expected literal. Then update the reviewed SHA/package versions,
--- a/scripts/check-github-actions-pins.mjs
+++ b/scripts/check-github-actions-pins.mjs
@@ -187,6 +187,7 @@ const REQUIRED_NATIVE_TESTS_JOB_LINES = [
   "        run: >-",
   "          cargo test --locked",
   "          --all-targets",
+  "          --features production-route-probe",
   "          --manifest-path wallet/plugins/tauri-plugin-zcash/Cargo.toml",
 ];
 const REQUIRED_CRYPTO_TARGET_JOB_LINES = [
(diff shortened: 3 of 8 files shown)


skyl:
Reviewed adversarially on `a0a361ce`, verified by running the checker and mutating the tree in a throwaway worktree. **Sound — merge when `gate` and `rs / gate` are green.** #620's three findings are all genuinely closed; I tested each one directly rather than taking the body's word.

### #620 item 1 — `pub(crate) mod native;` is now caught

The old check grepped for the literal `pub mod native;`, so `pub(crate)` passed. No longer:

```
send.rs:34  pub(crate) mod native;
EXIT=1  drift: native send policy must remain in one private child module
```

### #620 item 2 — the string-literal forgery is dead

This is the one I demonstrated against #617. Delete the real call and park byte-identical text inside a string literal:

```rust
let _forgery = "self::native::propose_fixed_validated(db, params, account_id, amount, request)";
```

```
EXIT=1
drift: fixed send orchestration must enter its only accessible native send boundary exactly once
drift: the behavior-tested fixed-send production route must own its only native proposal call
```

`rustCodeOnly(source, stripLiterals = true)` masks literal contents with spaces while preserving newlines and length, and models raw strings including `b`/`c` prefixes and `#` counts. Two independent errors fire, and the forgery no longer satisfies the count.

### #620 item 3 — the ordering guard came back stronger than what I asked for

I asked for the `validationIndex < proposalIndex` assertion to be restored. What landed is better: `orderedAnchors` pinning the whole money/state transition sequence inside `propose_send_after_recipient_validation`, alongside a whole-body digest.

I checked that the anchors are a real control and not decoration, because a digest plus a decorative anchor list would look identical from the outside. Reordering **inside** the anchored region — installing the proposal lock before `ensure_no_unresolved_broadcast` — fires both:

```
drift: fixed-send WalletState adapter must preserve its ordered proposal construction
       and accepted-state installation control flow
drift: fixed-send WalletState adapter complete control-flow body must remain a0606b0e…
```

and then, **after re-stamping the digest to the mutated body**, the ordering check *still* fires on its own:

```
POST-RESTAMP EXIT=1
drift: fixed-send WalletState adapter must preserve its ordered proposal construction
       and accepted-state installation control flow
```

That is the property I care about. The anchors survive the one thing that defeats a digest — someone mechanically re-stamping it — and the diagnostic names the invariant rather than saying "bytes changed".

### One narrow gap, not blocking: the preamble is digest-only

The anchored region begins at `let candidate = propose_fixed_stateful_production_caller(…)`. Everything before it — the `send_operation` lock, `is_initialized`, `active_wallet_id`, the early unresolved-broadcast check, the `pending_proposal` reset, memo parsing — is covered by the digest alone. Reordering there:

```
EXIT=1  drift: …complete control-flow body must remain a0606b0e…, got 76ac6451…
# then re-stamp the digest:
EXIT=0
```

Green, with the reordering still in place. That is strictly better than before this PR (nothing covered it), so it does not block. But a digest's remedy is mechanical re-stamping, and its diagnostic cannot tell a reviewer which property broke — so the preamble's ordering is one hurried re-stamp away from being unprotected.

#626 set the pattern worth copying here: when it had to move `changeDetectorDigest`, it added mutations making the new shape load-bearing rather than only re-stamping. If any preamble ordering is load-bearing (`is_initialized` before `active_wallet_id`, say), extend `orderedAnchors` upward to name it. If none of it is, a one-line comment saying the preamble is deliberately unordered would stop a future reader assuming the digest encodes intent.

### #616 item 3 is still open, and this PR does not claim otherwise

`send.rs:1705` still reads:

```rust
    let spendable = {
        let db_guard = state.read_db.lock().await;
        let db = db_guard.as_ref().ok_or(Error::WalletNotInitialized)?;
        let policy = ConfirmationsPolicy::default();
        let summary = db.get_wallet_summary(policy)
```

The routing this PR adds covers the *proposal* path; the send-all spendable-balance read still makes its own confirmation-depth decision outside the authority. `ConfirmationsPolicy` appears exactly four times — `native.rs:2` (import), `native.rs:72` (routed), `send.rs:14` (import), `send.rs:1705` (unrouted). Correctly left to #616, which I re-derived earlier today.

### Minor: `_validated_amount: u64` asserts in its name what its type cannot carry

`native.rs:98-103` takes `_validated_amount: u64`, unused. The name documents a caller obligation that a bare `u64` does not encode — any integer satisfies it. The actual enforcement is real (the exactly-once count plus the ordered anchors ensure the single call site is preceded by `single_payment_request`), so this is not a hole. But a newtype constructible only by the validator would move it from convention-plus-external-checker to compiler-enforced, which is the standard this PR otherwise sets with module privacy.

### Baselines

Clean tree first, per the lesson from #625: `check-librustzcash-compat.mjs` live → **EXIT=0** (`11 reviewed packages across 3 shipping locks`), `--self-test` → **EXIT=0, 108 mutants killed** (up from 72 at #617). Branch is exactly one commit on current `main`.

### What I did not observe

`gate` has not completed on this head. I did not run the plugin's Rust test suite — it needs the librustzcash submodule and a real `LocalTxProver`, so I am relying on the PR's claim of 145 unit tests plus the shipping integration test rather than having re-run them.

Issue #605, opened 2026-08-24
f2z-relay-store: MemoryStore returns gapped pages that cumulative ACK then destroys, and a rolled-back transaction discards recorded activity


skyl:
Two real bugs in `f2z-relay-store` (#595, closed by the sweep; branch `feature/relay-store` survives). Both destroy unacked ciphertext, and one is triggerable by a stranger.

## B1 — `MemoryStore` produces gapped pages, and cumulative ACK then destroys the gap

`memory.rs:232` uses `continue` where `sqlite.rs:1035` uses `break`. When a message overruns the remaining `max_bytes`, `MemoryStore` **skips it and admits a later, smaller one**:

```
mixed padding buckets, same input:
  MemoryStore:  indices = [0, 2, 3]
  SqliteStore:  indices = [0]
```

`WIRE.md` §8.1 makes ACK **cumulative**. A reader that receives `[0, 2, 3]` acks up to 3 — and index 1, which it never received, is permanently deleted. In the observed case that is 8 KiB of ciphertext destroyed with no error on any path.

The two stores disagreeing is the symptom. The cause is that **page contiguity is asserted nowhere**: injecting the same `continue` into `SqliteStore` (M20) leaves the suite green. So the correct implementation is correct by accident, and the incorrect one was never going to be caught.

**Fix:** `break`, in both, plus a property test asserting a page's indices are contiguous from the requested start — run against both stores. The two backends having divergent behaviour at all suggests they need a shared conformance suite rather than parallel test files.

## B2 — a rolled-back transaction permanently discards recorded activity

`flush_activity` `drain()`s the touch buffer as the **first statement** of a transaction. But `create_queue`, `bind_send`, `ack` and `delete_queue` all have `return Err(...)` paths **after** that point. The rollback discards the drained buffer — and because `drain()` already emptied it in memory, the activity is not retried. It is gone.

`read()` and `touch()` are the only non-durable activity sources, so what this deletes is exactly the §7.7 idle-timer resets that matter — the evidence that a queue is alive.

**It is remotely triggerable by someone with no relationship to the queue.** A single `BIND_SEND` from a stranger to a random address returns `ERR_UNAVAILABLE` — no PoW, no authorization, no prior state — and if it lands between a touch and a sweep, `queues_expired` flips from 0 to 1 and the queue is deleted with its unacked ciphertext. Ordinary traffic does it too: an `ERR_ALREADY_BOUND` retry, an `ERR_ACK_TOO_HIGH`.

[...]

PR #628, opened 2026-08-24
feat(rs): f2z-relay — the relay daemon

skyl:
Builds **`rs/crates/f2z-relay`** — the relay daemon `WIRE.md` describes, the one that has to run in production. ~8,200 lines of source and ~1,700 of integration tests, on top of `f2z-codec`, `f2z-relay-proto` and `f2z-relay-store`, with **nothing in those three reimplemented**.

## Stacked on one unmerged branch

Rebuilt on **`2ced6e4`** — `main` with the reviewed `f2z-relay-store` (#632). The superseded store commits this branch was originally cut from are **gone**; the merged crate is used unchanged.

Still stacked on **`feature/relay-testkit`**, which is open, because `f2z-relay`'s acceptance test *is* its conformance suite.

| Commit | What |
|---|---|
| `0806212` | `f2z-relay-testkit` — someone else's PR (#607) |
| `d862ec5` | merge of that branch into current `main`; three conflicts, all "both sides added a crate to the same list" |
| `841d8eb` | **this work** — the whole crate |

Review `d862ec5..HEAD`.

### The reviewed store needed no adaptation

Worth stating precisely, since the instruction was to adapt to it rather than reintroduce mine. #632 changed `sqlite.rs` internals, made one `MemoryStore` read fix (an oversized message now `break`s rather than `continue`s, so a page cannot skip a row a cumulative ACK would then destroy — #605/#606), and shortened one `#[must_use]` string. **`store.rs`, `record.rs` and `lib.rs` are byte-identical**, `SqliteStore::commits` survives, and `f2z-relay` compiled against it with **zero source changes**. Nothing the relay needs was lost.

Both properties a rebase onto a different storage layer would silently break were re-measured on the merged crate:

- **Group commit: `100 appends in 1 durable transaction`**, against `SqliteStore::commits`.
- **Conformance: 39/13/0 versus 52/0/0, verdict equality intact** on all 39 comparable vectors.

#631 (anti-replay expiration boundary) and #623 (replay-key commitment after authentication) both land inside `f2z-relay-proto`'s `CommandVerifier`, which this relay calls rather than reimplements, so it inherits both.

---

## Acceptance: the conformance suite, run against both relays

`f2z-relay-testkit`'s 52 vectors, **unchanged**, driven through `WebSocketEndpoint::new(url)` — the constructor its README names as the one `f2z-relay` would be run through — against a real `ws://` listener, and against the FakeRelay over its own listener so a difference cannot be a transport artefact.

```
FakeRelay   52 passed,  0 skipped, 0 failed, of 52
f2z-relay   39 passed, 13 skipped, 0 failed, of 52
```

**Every one of the 39 vectors both relays can run reaches the same verdict.** `tests/conformance.rs` asserts that as an equality rather than counting passes, and separately asserts that each of the 13 is *skipped* and never failed. A skip is not a pass.

The 13 declare `Needs::Faults` or `Needs::Clock`. A real relay has neither handle **and must not**: a fault-injection hook compiled into a production binary is one an attacker can reach, and a `#[cfg(test)]` one is a hook a build-flag mistake ships. So `tests/configured_equivalents.rs` closes the gap the other way — **eleven of the thirteen are reachable on an unmodified binary**, through published configuration (a policy `WIRE.md` permits, not a relaxation) or through ordinary client behaviour the spec already describes:

| Skipped vector | Reached by |
|---|---|
| `a_duplicate_inflight_request_id_is_fatal` | an `APPEND` answered from its own task, so its id is genuinely outstanding |
| `exceeding_the_inflight_window_is_fatal` | the same, with `limits.max_inflight = 2` |
| `a_full_queue_is_indistinguishable_from_an_absent_one` | `limits.max_queue_messages = 1` |
| `the_queue_creation_quota_is_a_distinguishable_code` | `limits.max_queues = 1` |
| `backpressure_never_refuses_read_ack_or_delete` | `limits.storage_high_water_bytes = 1` |
| `an_expired_message_is_gone_and_announced` | a one-second message TTL and sweep |
| `an_idle_queue_disappears` | a one-second idle TTL and sweep |
| `a_lost_ack_response_is_safe_to_retry` | the client closes before reading the answer — §8.3's actual case |
| `a_delayed_response_still_correlates` | a cheap `PING` overtakes an `APPEND` waiting on a commit window |
| `reordered_responses_still_correlate` | the same exchange, correlated by `request_id` |
| `a_mid_stream_close_leaves_the_status_unknown` | the client closes mid-command |

The two that remain, stated rather than papered over:

- **`an_unsound_antireplay_window_is_published_and_refused`** is satisfied in a *stronger* sense than the vector asks. It has a relay publish `antireplay_window_ms == clock_skew_ms` and a client refuse it; `f2z-relay` **refuses to be configured that way at all** (`Config::check`, #586's finding enforced rather than demonstrated). The vector is unrunnable here because the defect it demonstrates cannot exist.
- **`every_error_code_can_reach_a_client`** is partly reachable. Code 9 `ERR_CHANNEL_BINDING` is *reserved and unused in v1* by §10's own table, so no conforming relay emits it, and code 21 `ERR_INTERNAL` is a relay fault. Every other §10 code is reached by the suite or by `configured_equivalents.rs`, which additionally provokes `ERR_RATE_LIMITED` and `ERR_POW_REQUIRED` on relays configured to use them.

### The comparison found a real defect — in this relay

`a_reader_is_told_when_its_queue_hits_a_cap` is a **difference no vector covers**. §6.4 defines `QUEUE_EVENT` reason 4 as "quota reached", pushed to a subscribed reader. The FakeRelay emits it; the first version of this relay did not, because §6.3's send-side collapse had been applied one layer too early — the *writer* must not learn which cap was hit, but the **reader** is the party who can drain the queue and §6.4 exists to tell it. **The FakeRelay was right**; fixed, with the regression test plus its mirror (an absent address must produce no push, or the push becomes §10's existence oracle).

That is the whole reason the testkit was written first, and it paid for itself.

### One behavioural difference that is not a defect in either

The FakeRelay sweeps §7.7's timers on **every command**, which is right for a harness whose clock a test steers. `f2z-relay` sweeps on a timer (`queues.expiry_tick_seconds`, default 60), because the sweep is a range scan and paying for it on every `READ` puts a cost proportional to the whole store on an operation §13.1 says must never be refused. The consequence — a message can outlive its TTL by up to one tick — is a policy statement about a seven-day timer, not a correctness one. Documented in `expiry.rs`.

---

## What is in the crate

- **Listener** (`listener.rs`, `tls.rs`, `transport.rs`) — `wss://` over rustls with **TLS 1.3 as the only version**, because §5.3's binding is the TLS 1.3 exporter and §2.1 forbids 1.2. §2.1's path and mandatory subprotocol, echoed. §2.3's non-loopback-without-TLS refusal is a **startup check** in `Config::check`, enforced again at `bind`, and the `--insecure-listen` override is published as `transport_security: none` + `channel_binding_mode: none` — both derived from the same `tls_enabled()` question so they cannot disagree.
- **Engine** (`engine.rs`) — §4 framing, §5.1's verification order via `CommandVerifier` with a **relay-wide** seen-set, §5.5's window, the full §6 command set over `RelayStore`, §10's existence-oracle collapse with the dummy Ed25519 verification on the absent path.
- **Group commit** (`commit.rs`) — an OS thread, because an fsync is a blocking syscall of unbounded duration and a 1 GB VPS has few workers. `Committed<T>` is unwrapped **on the writer thread after `append_batch` returns**, in the same statement that sends the reply.
- **Expiry tick** (`expiry.rs`) — both §7.7 timers, the seen-set, the challenge table, the rate-limiter rows, and §13.1 layer 4 measured against `storage_bytes` rather than assumed.
- **Anti-abuse** (`abuse.rs`, `challenge.rs`) — §13.1's four layers. Per-source counters are keyed by a **BLAKE2b digest under a per-process random salt** that never leaves the process, and an idle row is dropped rather than retained: a table keyed by IP that outlives its window is a connection log.
- **Admin listener** (`admin.rs`) — loopback-only and refused otherwise. `/healthz` is byte-identical across a queue, a connection and a message, and touches no storage. `/metrics` has **no labels at all**, enforced by `Metrics` having no method that takes an address or a key.
- **Config** (`config.rs`, `cli.rs`) — TOML + env + flags, `--help` / `--version` / `--print-config` (seed redacted) / `--check-config` / `--print-capabilities`. Unknown file keys are errors.

### Proofs, not claims

| Property | Evidence |
|---|---|
| `accepted` never precedes durability | `Committed<T>`'s constructor is crate-private; it is unwrapped on the commit thread and the reply is sent in the same statement |
| Group commit amortizes | `tests/group_commit.rs`: **100 concurrent appends → 1 durable transaction**, measured against `SqliteStore::commits` |
| Batching is still `fsync-per-append` | Same file: 50 appends across 5 connections, all 50 readable, indices dense |
| No secrets in logs | `tests/redaction.rs`: base16 both cases, base64url, **and the decimal byte-list form** |
| No per-queue or per-IP metrics | `tests/admin.rs` + `metrics.rs` unit tests |
| §13.2 refuses, never evicts | `a_full_queue_is_indistinguishable_from_an_absent_one`; the store's own suite covers the transaction |

`log::line` takes a `&'static str` and `(&'static str, u64)` pairs and nothing else, so `log_trace!("frame", "payload" = payload)` **does not compile**.

---

### One new defect

**`queue_creation_mode: token` has no wire representation in v1.** §13.1 defines it as an operator-issued bearer credential, and `CREATE_QUEUE` carries only a `PowStamp` and a `flags` field that MUST be 0. There is no conforming way for a client to present a token, so a relay publishing that mode publishes a gate no client can pass. `f2z-relay` **refuses to start** in that mode rather than inventing a field. `f2z-relay-testkit`'s README observed the same thing from its own side and recorded it as a testkit gap; it is a spec-level gap, and it is not in #586.

### #586's three known contradictions

Read the issue and its comments. All three are implemented as #586/#583/the testkit resolved them, and **not re-reported**: the §6.3-versus-§10 send-side collapse resolved in §6.3's favour; the §1.3-versus-§4.3 unrecoverable-`request_id` case, closed with no response frame; the unknown `ChallengePurpose` byte answered `ERR_MALFORMED`, and `READ`'s zero limits read as "no limit". The anti-replay retention invariant is **enforced** here rather than left publishable.

### Every ambiguity call

1. **A second `HELLO`** → fatal `ERR_UNSUPPORTED_VERSION`. §3.5 says negotiation happens once; §10 names no code. Matches the testkit.
2. **Close status for a fatal error other than §4.2's** → 1002. §1.3 names none. Matches the testkit.
3. **A client sending a `response` or `push` frame** → fatal `ERR_MALFORMED`. §10 has no code for the wrong half of the protocol.
4. **Unknown `ChallengePurpose`** → fatal `ERR_MALFORMED`. #586's comment; the weakest call in the set, and the testkit's.
5. **`READ` zero limits** → no limit. #586's comment; the testkit's.
6. **`READ`'s byte limit is additionally clamped to `max_frame_bytes - 4096`.** *New, and small.* §6.2 lets a client ask for more bytes than §4.1 lets the relay send; without a clamp the relay builds a frame its own cap forbids and answers `ERR_INTERNAL` instead of the page. Unstated.
7. **A full challenge table** → `ERR_BACKPRESSURE`, and it **refuses rather than evicts**. §5.5 gives that code to the seen-set bound for the same reason; evicting would invalidate a stamp a client already paid for.
8. **`GET_CHALLENGE` over its per-source budget** → `ERR_RATE_LIMITED`. §6.1 requires the limit and names no code.
9. **A `Ping` from a client** → `Pong`. §2.4 only specifies the server-driven direction.
10. **Group commit is published as `fsync-per-append`**, on §8.4's own distinction. Stated because it is the field an auditor will question.

---

## Licence boundary

**AGPL-3.0** for this crate, per #305. `rs/deny.toml` gains exactly the entry its own comment anticipated — `{ name = "f2z-relay", allow = ["AGPL-3.0"] }` — and *not* `"AGPL-3.0"` in the shared `allow` list.

Two supply-chain changes need review:

- **`ISC` added to `allow`.** It arrives only with TLS: `rustls-webpki` and `untrusted` are ISC, `ring` is `Apache-2.0 AND ISC`. None is reached by any crate a client links — the wasm32 job builds `f2z-codec`, `f2z-relay-proto` and `f2z-authority`, and none pulls rustls. ISC is the two-clause BSD licence with the redundant words removed: OSI-approved, FSF Free/Libre, no copyleft, no attribution-in-binary requirement. Added because a dependency needs it, as the file's own instructions describe.
- **`rustls-pemfile` avoided** in favour of `rustls-pki-types`' own `PemObject`. It was archived in August 2025 (**RUSTSEC-2025-0134**) and its advisory says the latest version is a thin wrapper around exactly that code. One fewer crate on the path that reads an operator's private key.

`multiple-versions = "warn"` reports duplicates (`winnow`, `rand`, `getrandom`, `syn`, …); most predate this branch and none fails the gate.

---

## Two workspace scans

**#612's strictness scan.** It registers every `fn verify…` under `rs/crates/*/src/`, not only call sites. Three were unregistered: this relay's `verify`/`verify_with` and `f2z-relay-testkit`'s `verify`. Rather than register three, the thin `verify(&Connection, …)` wrapper is **deleted** — a second name beginning with `verify` is a second thing the scan must be told about, to save one field access at six call sites, and the field it saves is the one that must not be wrong: a transcript builder from anywhere but this connection binds the signature to another session's channel binding (§5.3). Neither relay verifies a signature itself; both reach `key.rs::verify`'s `verify_strict` through `CommandVerifier`, and both registry rows say so.

**`workspace_debug_scan` is hardened against a wrapped attribute.** Its attribute-skipping loop stopped at the first line not starting with `#`, took an attribute's continuation line for the item declaration, and aborted the entire scan on an assert. That is a *hole*, not a failure: one `#[must_use = "…"]` spanning two lines anywhere under `rs/crates/*/src/` silently disables every leak check in the file. `f2z-relay-store` carried exactly that shape until #632 happened to shorten the string — the bug is still on `main`, merely unexercised. Now attributes are consumed until their brackets balance.

**With that fix it immediately caught two real leaks in this crate**: `AbuseGuard` derived `Debug` over the per-process salt that makes `SourceKey` non-invertible, and `Published` derived it over the encoded capability document. Both now hand-write `Debug`.

## Gates

`fmt`, `clippy -D warnings`, `cargo-deny` (advisories, bans, licences, sources) and `check-rust-toolchain.sh` are green locally on the pinned **1.97.1**, plus `cargo test --locked --all-targets`, `--doc`, and the store's `crash-injection` suite. All three new crates are registered in #625's `MANIFESTS` census, and `rs.yml` keeps them **off** the wasm32 line, where its own comment already anticipated "a future server binary — which will be AGPL, will pull tokio, and will never reach a browser".

## Not done, deliberately

- **`.well-known` is not served.** §11.2 asks for the document over plain HTTPS; serving it means an HTTP request parser on the unauthenticated public port, which is precisely what §2.2 argues against, for an affordance aimed at humans rather than clients. The upgrade's own HTTP parse does not help — it runs only for requests already carrying `Upgrade: websocket`. `--print-capabilities` emits the identical document, signature and digest included, for an operator to publish from whatever already serves their website: §11.2's *property* — same values, same signature, publicly diffable — survives; the relay hosting it does not. A JSON **encoder** consumes nothing from the network, which is why one is here and a parser is not. Reviewers who disagree should say so; it is the most arguable call in the PR.
- **`ACK` is not batched.** It commits, so it is the obvious next batcher; deliberately deferred because an `ACK` is one range delete per *reader* rather than one per message, and adding a second scheduler before the first is measured in production is how a relay acquires two nobody can reason about.
- **Push notifications, federation, anonymous credentials** — §16 leaves them open and so does this.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -333,6 +333,10 @@ jobs:
       # `f2z-relay-testkit` is absent for a neighbouring reason: it opens
       # sockets, spawns tasks and reads a clock, and a test harness that could
       # reach the browser build could end up in the shipped bundle.
+      #
+      # `f2z-relay` — the server binary the paragraph above anticipated — is all
+      # three at once: AGPL-3.0, tokio and SQLite, and never a browser. Adding a
+      # crate to this line is a claim that a client links it.
       - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -75,7 +75,8 @@ jobs:
           scripts/check-rust-toolchain.sh \
             --toolchain-file rs/rust-toolchain.toml \
             --manifest rs/crates/f2z-codec/Cargo.toml \
-            --manifest rs/crates/f2z-relay-proto/Cargo.toml
+            --manifest rs/crates/f2z-relay-proto/Cargo.toml \
+            --manifest rs/crates/f2z-relay-testkit/Cargo.toml
 
       - name: Detect changes under rs/
         id: filter
@@ -293,6 +294,11 @@ jobs:
       # rather than `--workspace` so that a future server binary — which will be
       # AGPL, will pull tokio, and will never reach a browser — does not
       # silently join the wasm build and then fail it.
+      #
+      # f2z-relay-testkit is deliberately absent for exactly that reason: it
+      # opens sockets, spawns tasks and reads a clock, and a test harness that
+      # could reach the browser build would be a test harness that could end up
+      # in the shipped bundle.
       - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -57,10 +57,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytes"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "rand_core 0.10.1",
+]
+
+[[package]]
+name = "const-oid"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
 
 [[package]]
 name = "cpufeatures"
@@ -114,8 +137,14 @@ checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.119",
 ]
+
+[[package]]
+name = "data-encoding"
+version = "2.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4583a4551df46e2792f82ceeac45e850d2e2d5debba0b91f102385cda5b11f06"
 
 [[package]]
 name = "digest"
@@ -135,6 +164,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f1dd6dbb5841937940781866fa1281a1ff7bd3bf827091440879f9994983d5c2"
 dependencies = [
  "block-buffer 0.12.1",
+ "const-oid",
  "crypto-common 0.2.2",
 ]
 
@@ -191,6 +221,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-relay-testkit"
+version = "0.1.0"
+dependencies = [
+ "f2z-codec",
+ "f2z-relay-proto",
+ "futures-util",
+ "tls_codec",
+ "tokio",
+ "tokio-tungstenite",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -207,6 +249,37 @@ name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "futures-core"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92d699e522242e69e3003b94ecc1f960f3a5e015aa7c5d7486e65ad01dd94f5e"
+
+[[package]]
+name = "futures-sink"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1944426bf7d03f1d14f708785e4b33efd750b36d48a157b836b3efc15ede8e1d"
+
+[[package]]
+name = "futures-task"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd417de3d1d015fc3bfd2b1ea46dfc7bab72ef86f1cc7cc9c78e728b34a6d1fd"
+
+[[package]]
+name = "futures-util"
+version = "0.3.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d50a92467f8ba5dd6e3ee5d4bd04d73ab2e4e1c44474a0674821dfce14b79bc"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+ "futures-task",
+ "pin-project-lite",
+ "slab",
+]
 
 [[package]]
 name = "generic-array"
@@ -239,7 +312,24 @@ dependencies = [
  "cfg-if",
  "libc",
  "r-efi 6.0.0",
+ "rand_core 0.10.1",
 ]
+
+[[package]]
+name = "http"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "918d3568bebf352712bc2ef3d46a8bcf1a75b373be6539de198e9105cbbf9ce0"
+dependencies = [
+ "bytes",
+ "itoa",
+]
+
+[[package]]
+name = "httparse"
+version = "1.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6dbf3de79e51f3d586ab4cb9d5c3e2c14aa28ed23d180cf89b4df0454a69cc87"
 
 [[package]]
 name = "hybrid-array"
@@ -249,6 +339,12 @@ checksum = "707114b52a152fa7bdb290cd7cd5912d9467273b6d74e21b8d81aca1f8533f6b"
 dependencies = [
  "typenum",
 ]
+
+[[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
 
 [[package]]
 name = "libc"
@@ -261,6 +357,23 @@ name = "linux-raw-sys"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32a66949e030da00e8c7d4434b251670a91556f4144941d37452769c25d58a53"
+
+[[package]]
+name = "log"
+version = "0.4.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f9f8bd3e56ce4dfc153cf470fffbfa98c7620958b312ca5c3a4b8d5181fd13c6"
+
+[[package]]
+name = "mio"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30d65c71f1ce40ab09135ce117d742b9f8a19ff91a41a8b57ed50bc2de59c427"
+dependencies = [
+ "libc",
+ "wasi",
+ "windows-sys",
+]
 
 [[package]]
 name = "num-traits"
@@ -276,6 +389,12 @@ name = "once_cell"
 version = "1.21.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f7c3e4beb33f85d45ae3e3a1792185706c8e16d043238c593331cc7cd313b50"
+
+[[package]]
+name = "pin-project-lite"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
 [[package]]
 name = "ppv-lite86"
@@ -305,7 +424,7 @@ dependencies = [
  "bit-vec",
  "bitflags",
  "num-traits",
- "rand",
+ "rand 0.9.5",
  "rand_chacha",
  "rand_xorshift",
  "regex-syntax",
@@ -348,7 +467,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9ef1d0d795eb7d84685bca4f72f3649f064e6641543d3a8c415898726a57b41"
 dependencies = [
  "rand_chacha",
- "rand_core",
+ "rand_core 0.9.5",
+]
+
+[[package]]
+name = "rand"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7f5fa3a058cd35567ef9bfa5e75732bee0f9e4c55fa90477bef2dfcdbc4be80"
+dependencies = [
+ "chacha20",
+ "getrandom 0.4.3",
+ "rand_core 0.10.1",
 ]
 
 [[package]]
@@ -358,7 +488,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3022b5f1df60f26e1ffddd6c66e8aa15de382ae63b3a0c1bfc0e4d3e3f325cb"
 dependencies = [
  "ppv-lite86",
- "rand_core",
+ "rand_core 0.9.5",
 ]
 
 [[package]]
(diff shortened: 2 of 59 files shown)


skyl:
Triage only — this is a draft and `CONFLICTING`, so I have not reviewed the 24k lines. Flagging what a rebase will have to satisfy, because `main` moved under it today.

**14 commits behind.** Six PRs landed since the merge base: #622, #623, #625, #626, and #627 is in flight.

**It will fail #625's Rust-root ownership census on the first CI run.** This branch adds three crates:

```
rs/crates/f2z-relay-store/Cargo.toml
rs/crates/f2z-relay-testkit/Cargo.toml
rs/crates/f2z-relay/Cargo.toml
```

and does not touch `scripts/check-rust-toolchain.sh`. As of `b663be4`, every tracked first-party `Cargo.toml` must belong to a policed root **and** be registered:

```
drift: <path> declares [package] but is registered nowhere, so nothing holds its
rust-version to wallet/rust-toolchain.toml. Add it to MANIFESTS in
```

I verified this fires by adding an unregistered crate on the #625 and #626 branches, so it is not a guess. Add all three to `MANIFESTS` as part of the rebase.

**Not a problem:** the `.verify(` calls in `f2z-relay-proto` (`capabilities.rs:206`, `command.rs:679`, `hello.rs:270`, `key.rs:160`) are that crate's own `verify_strict` wrapper, and `key.rs:160` is the deliberate dummy-verification. The workspace strict-verification scan already understands them on `main`. I mention it only so a rebase does not "fix" them.

Also worth knowing before rebasing: #612's scanner was replaced in #626 (`code_only` → `code_lines`), and it now models raw strings and nested block comments and refuses on malformed input. If this branch carries the old scanner, take main's.

Happy to review properly once it is rebased and out of draft.

skyl:
Merged `origin/main` (14 commits) and addressed the triage. `CONFLICTING` → **`MERGEABLE`**, and CI ran for the first time: **`rs / gate` is green**, along with `rs / rustfmt`, `rs / clippy`, `rs / cargo-deny`, `rs / tests` and `rs / wasm32`.

**#625's census, as flagged.** It replaced the `--manifest` flags with a census over every tracked first-party `Cargo.toml`, so all three crates are registered in `MANIFESTS` in `scripts/check-rust-toolchain.sh` rather than on `rs.yml`'s command line. Took `main`'s `rs.yml` wholesale and re-stated on its wasm32 line why `f2z-relay-store`, `f2z-relay-testkit` and `f2z-relay` are absent from it — that line's own comment already anticipated "a future server binary — which will be AGPL, will pull tokio, and will never reach a browser".

**#612's strictness scan needed one thing the triage did not predict.** It is not only the `.verify(` *call sites* that are registered — every `fn verify…` defined under `rs/crates/*/src/` must be in `WORKSPACE_VERIFY_FNS` with what makes it strict, and **three were not**: `f2z-relay`'s `verify` and `verify_with`, and `f2z-relay-testkit`'s `verify`. The testkit's has been unregistered since #612 landed; this branch is the first tree holding both.

Rather than register three, `f2z-relay`'s thin `verify(&Connection, …)` wrapper is **deleted**. A second name beginning with `verify` is a second thing the scan must be told about, to save one field access at six call sites — and the field it saves is precisely the one that must not be wrong: a transcript builder from anywhere but this connection binds the signature to another session's channel binding (§5.3). The remaining `verify_with` takes the builder explicitly. Both relays' rows name `key.rs::verify` as the terminal strict verifier and their local `verifier` receiver, the same convention `command.rs::verify`'s row uses.

Nothing in the triage's "not a problem" list was touched: the `.verify(` calls in `f2z-relay-proto` are untouched, and `main`'s `code_lines` scanner came across in the merge.

**Conformance is unchanged after the merge** — re-run against both relays:

```
FakeRelay   52 passed,  0 skipped, 0 failed, of 52
f2z-relay   39 passed, 13 skipped, 0 failed, of 52
```

Still a draft, and staying one while the merge-train hold on #598 stands.

skyl:
Rebuilt on **`2ced6e4`** — `main` with the reviewed store (#632). The superseded `f2z-relay-store` commits are dropped; history is now `main → f2z-relay-testkit → merge → the crate`, two commits of mine.

**The reviewed store needed no adaptation.** #632 changed `sqlite.rs` internals, one `MemoryStore` read fix (#605/#606), and shortened one `#[must_use]` string. `store.rs`, `record.rs` and `lib.rs` are byte-identical, `SqliteStore::commits` survives, and `f2z-relay` compiled against it with **zero source changes**. Nothing the relay needs was lost, so there is no finding against #632.

Both properties you flagged were re-measured on the merged crate:

- **Group commit** — `group commit: 100 appends in 1 durable transactions`, against `SqliteStore::commits`.
- **Conformance** — re-run against both relays, verdict *equality* re-asserted on all 39 comparable vectors:

```
FakeRelay   52 passed,  0 skipped, 0 failed, of 52
f2z-relay   39 passed, 13 skipped, 0 failed, of 52
```

**#631 and #623** both land inside `f2z-relay-proto`'s `CommandVerifier`, which this relay calls rather than reimplements, so it inherits both; nothing here touches `SeenSet` directly except `expire()` on the tick.

**#612 does apply.** It registers every `fn verify…` under `rs/crates/*/src/`, not only call sites, and three were unregistered: this relay's `verify`/`verify_with` and `f2z-relay-testkit`'s `verify` — the testkit's has been unregistered since #612 landed. Rather than register three, the thin `verify(&Connection, …)` wrapper is **deleted**; the field it saved is the one that must not be wrong. Neither relay verifies a signature itself, and both rows name `key.rs::verify`'s `verify_strict` as what they reach.

One more thing worth having on the record: **`workspace_debug_scan`'s attribute skipper is still broken on `main`**, merely unexercised now that #632 shortened `Committed`'s `#[must_use]`. One wrapped attribute anywhere under `rs/crates/*/src/` makes it abort and silently disables every leak check in the file. Hardened here. With the fix it immediately caught two real leaks in my own crate.

**`rs / gate` is green** (rustfmt, clippy, cargo-deny, tests, wasm32 all pass), and the PR is **out of draft**. Not merging — over to you.

skyl:
**This merged without a `gate` verdict on the merged tree.** Recording it factually, because it is the first time tonight a change has landed that way and this one is 20k lines.

### The timeline

```
15:28:31  zuuli run 32745073784 created on 841d8ebd
15:28:31  rs    run 32745073964 created on 841d8ebd  → completed/success
15:34:09  PR merged
15:36     zuuli run 32745073784 still `queued` — it had not started
```

`rs / gate` was green. The zuuli `gate` — the one that awaits `changes`, `frontend`, `rust_fmt`, `rust_deny`, `rust_clippy`, `rust_native_clippy`, `rust_native_tests`, `rust_plugin`, `rust_android_32`, `rust_app`, `zuuallet_schema`, `rust_crypto_targets` — never ran on this head at all.

The earlier head `fc849ea2` does have a green zuuli run (`32737347372`), but that is a **different tree**: it still carried its own `f2z-relay-store`, which this head correctly dropped in favour of main's after #632 landed. That rebase is exactly the change no CI has judged. Green on the pre-rebase tree is not evidence about the post-rebase one — that distinction is the whole reason `strict: true` exists.

Under `enforce_admins=false`, `--admin` is meant to bypass the *human* review requirement, never the QA check. Doctrine is explicit: merge "after the CI `gate` is confirmed green." Here it was not confirmed; it was not even started.

### What I am doing about it rather than just noting it

Main's own push run on `ab656ca` is now the **only** gate evidence that will ever exist for this tree, so I am watching it (`32745654881` / `32745654876`). If its `gate` fails, this is a P0 revert of a 20k-line change with no fallback verdict to fall back on.

### The rebase itself was done right, for the record

I checked before the merge and the collision resolution was correct: zero `f2z-relay-store` files remain, size dropped +24470 → +20026 (matching #632's +4734), and it now adds only `f2z-relay-testkit` and `f2z-relay`. That is the right call and it is what I asked for in my triage. My concern is only that the tree resulting from it was never gated.

### Ask

For the next large one: take it out of draft, let `gate` go green on the **final** head, then merge. The wait is 25-60 minutes under current runner contention, which is real — but it is the difference between "CI proved this" and "CI proved something that no longer exists." I will review this on the merged tree and file anything I find separately.

Issue #624, opened 2026-08-24
f2z-relay-proto: a signed frame is accepted twice at exactly t+skew on the shipped default — retention_is_sound's >= must be strict


skyl:
The anti-replay soundness rule is stated as `retention_ms >= 2 * clock_skew_ms`. It needs to be strict. At exactly `retention_ms == 2 * clock_skew_ms` — **which is the shipped default** — the same signed frame is accepted twice.

I demonstrated this through the real `CommandVerifier::verify` path on merged `main` (`c7ae0dc`), not by reasoning about the spec.

## Reproduction

Added to `rs/crates/f2z-relay-proto/src/command.rs` tests, using the existing `verifier()` helper (`SeenSet::new(240_000, 1024)`, `TimestampWindow::default()`):

```rust
let signed = SignedCommand::<ops::Subscribe>::create(
    &transcripts(), 2, QueueAddress::new([0x9b; 32]), NOW, nonce(0x9c), &key, &Empty,
).unwrap();
let request = signed.request().unwrap();
let mut verifier = verifier();

// Earliest moment the frame is valid: now = t - skew
assert!(verifier.verify::<ops::Subscribe>(NOW - 120_000, 2, &request).is_ok());

// Latest moment the frame is STILL valid: now = t + skew
let second = verifier.verify::<ops::Subscribe>(NOW + 120_000, 2, &request);
assert_eq!(second, Err(ProtoError::Wire(ErrorCode::Replay)));
```

Result:

```
assertion `left == right` failed: REPLAY ACCEPTED at exactly t+skew,
got Ok(Verified { body: Empty, address: QueueAddress(<redacted>),
signer_key: PublicKey(<redacted>), timestamp_ms: 1800000000000, nonce: Nonce(<redacted>) })
```

The second presentation returns `Ok(Verified)`. The relay accepts the replay.

## Why it happens — three correct-looking decisions that only interact at the boundary

1. **The timestamp window is inclusive.** `command.rs` pins `NOW + 120_001` as `ERR_STALE_TIMESTAMP`, so `NOW + 120_000` is accepted. A frame stamped `t` is valid over the **closed** interval `[t − skew, t + skew]`.
2. **The seen-set entry is half-open.** `replay.rs:237` is `self.entries.retain(|_, expires_at| *expires_at > now_ms)`, so an entry committed at `now₁` is held over `[now₁, now₁ + retention)` and is dropped at exactly `now₁ + retention`.
[...]

PR #629, opened 2026-08-24
feat(kt): f2z-kt log server and f2z-witness cosigning daemon (KT.md v1)

skyl:
# `f2z-kt` and `f2z-witness` — the key-transparency log server and the cosigning daemon

> ## ⚠️ This PR is the merge path for **three** crates, not two
>
> `f2z-kt-core` was written by a **different agent**, reviewed separately on
> `feature/kt-core` as **#584** — and **that PR was closed by the merge-train
> hold and never reopened**. Its crate is still not on `main`, so this diff
> carries it, and merging this merges it.
>
> | Crate | Provenance | Licence |
> |---|---|---|
> | `rs/crates/f2z-kt-core` | **#584, another agent.** Reviewed on a branch that never landed. Modified here — see below. | MIT |
> | `rs/crates/f2z-kt` | this PR | AGPL-3.0-only |
> | `rs/crates/f2z-witness` | this PR | AGPL-3.0-only |
>
> **`f2z-kt-core` is not merely carried along — this PR changes it**, and those
> changes were not in #584's review:
>
> - `src/api.rs` is **new**: `KT.md` §9.2's request/response envelopes, moved
>   here from the server so the log, the witness and the client share one
>   implementation (§11.4).
> - `LABEL_STH_HASH` → `LABEL_TREE_HEAD_HASH`, forward-porting **#602**'s
>   domain-separation fix. #584 predates it and shipped the vulnerable spelling.
> - Hand-written `Debug` for `AcceptedSubmission` and `VerifiedEntry`, to satisfy
>   the workspace `Debug`-over-raw-bytes scan that landed after #584 was cut.
> - Registered in the `check-rust-toolchain.sh` census and on the wasm line.
>
> A reviewer should read `rs/crates/f2z-kt-core/` as **new code in this PR**,
> whatever #584 concluded about an earlier version of it.

## Where this sits against `main`

Up to date with **`2ced6e4`**. `f2z-authority` merged as **#626** mid-task and
this branch carries main's reviewed version of it, not the branch version it was
first built against — which forced the API rework described below.

**A literal `git rebase origin/main` is the wrong operation here and I did not
do one.** This branch's ancestry contains the pre-#626 `f2z-authority` commit;
a flattening rebase would replay it onto a `main` that already carries the
reviewed crate, resurrecting the superseded version through add/add conflicts on
every file. `origin/main` is merged in instead. The repository squash-merges
(`AGENTS.md`), so the branch's history shape does not reach `main`, and the
diff-against-`main` is identical either way.

The three commits of drift (#627, #631, #632) conflicted only in files where
both sides *added* something; every resolution keeps both. `f2z-codec` is
untouched on `main`, so the shared-scanner change discussed below still applies
cleanly.

### The rework that merge forced, and it made the design better

#626's review changed the authority API substantially, and the change is an
improvement this PR now follows rather than works around:

- `admit` became **`check_assertion_layer`**, returning an `AssertionLayerCheck`
  the crate's own docs call *"deliberately **not** an authorization-to-publish
  token"*.
- `Submission` gained **`kind: EntryKind`** — including the unratified
  `InitialBind` — plus `previous_vouch` and a required `previous_account_epoch`.

The first draft of this server called the assertion layer **only** for a first
entry. The reviewed API covers all four cases properly, so **every** submission
now routes through it, and three things follow:

1. A platform assertion on a `same_key` or `key_change` entry is a **category
   error**, refused rather than ignored — the platform does not get a second way
   to authorize a change to a live handle.
2. A `platform_reset` requires an assertion *alongside* §4.4's
   `ResetAuthorization`.
3. The log retains each handle's `(vouch, account_epoch)` and feeds both back,
   so a routine update carries its predecessor's vouch forward rather than
   silently upgrading or dropping it.

Consequently the journal now stores the **whole envelope**, and a restart
re-runs the *entire* admission — §4.4 and the assertion layer — rather than half
of it. The per-handle state is re-derived rather than stored, so there is no
second representation of the directory to drift.

### Two repo-wide invariants that landed on main were also satisfied

Both were added after `feature/kt-core` was cut, so they caught pre-existing
types in that base as well as new code here:

- **`workspace_debug_scan`** flagged `AcceptedSubmission`, `VerifiedEntry` and
  this crate's `Pending`. All three now have hand-written `Debug`. `Pending` is
  the one that matters most: it holds a submission *between acceptance and
  publication*, the one window in which an entry is not yet public.
- **`workspace_strict_verification_scan`** requires every `fn verify…` to be
  registered with what makes it strict. All thirteen now are — and **two new
  `Strictness` categories were needed**, which is a change to a shared checker
  and deserves a reviewer's attention:
  - `DelegatesToFn`, because `f2z-kt-core`'s Ed25519 entry point is a free
    function (`sig::verify`) rather than a method on a key newtype, so its
    wrappers cannot satisfy the existing receiver rule;
  - `NotASignatureCheck`, for the four functions that verify `akd` proofs or
    count already-verified cosignatures and so have no delegation chain to
    `verify_strict` — `verify_lookup`, `verify_key_history`,
    `verify_append_only`, `verify_threshold`, plus `verify_authorization`.

  Each row still names the call its body must contain, so an exemption cannot go
  vacuous by that call being deleted. If review would rather see `f2z-kt-core`
  grow a key newtype so the existing `DelegatesTo` applies, that is a
  straightforward alternative and I will take it.

## What this adds

| | |
|---|---|
| **`rs/crates/f2z-kt`** | The log server: durable journals, the submission choke point, the `akd` tree and the epoch scheduler, `LogSigner` behind a trait, and `KT.md` §9.2's endpoints. |
| **`rs/crates/f2z-witness`** | The cosigning daemon: poll → verify the tree-head signature → refuse and record evidence → **verify the append-only proof** → cosign → atomically update one state file. |

Both are **AGPL-3.0-only** (owner decision on #305), named individually in
`rs/deny.toml`'s `[licenses] exceptions`; the library crates stay MIT. Both are
`#![forbid(unsafe_code)]`, native-only, and absent from the wasm line. Both
manifests are registered with `scripts/check-rust-toolchain.sh` in `rs.yml` and
`rs/README.md`.

## The thing that matters most

> `KT.md` §4.4: **`akd` enforces none of it.** The library will happily commit
> any bytes to any label. A log that skips §4.4's rules produces inclusion,
> history and append-only proofs that verify **perfectly** for entries nobody
> authorized.

So the server is one door and no windows:

| To reach | You must hold | Which requires |
|---|---|---|
| `akd::Directory::publish` | an `AdmittedSubmission` | `f2z_kt::admit::admit_submission` |
| `AdmittedSubmission` | an `AcceptedSubmission` **and** an `AssertionLayerCheck` | `f2z_kt_core::validate_submission` **and** `f2z_authority::AuthorityConfig::check_assertion_layer` |

Neither layer is sufficient alone and each says so: `validate_submission` does
not decide who may claim an unregistered handle (zuu#594), and
`check_assertion_layer`'s result is explicitly not authorization to publish.
`AdmittedSubmission` is the conjunction.

Neither carrier has a public constructor, a `From`, field access, or a bypass
flag. There is no `publish_raw` and no test hook. `tests/adversarial.rs` is the
test that watches it hold: **15 well-formed submissions, each with exactly one
thing wrong**, each asserting both the §9.5 code *and* that the tree did not
move — because a refusal that left the entry in the pending batch would publish
it on the next tick, and a test that only checked the return value would never
see it.

Covered: a first entry with no authority assertion; an assertion for a different
handle; a stolen assertion without the identity key; a platform assertion on a
user-authorized entry (the category error); a wrong `prev_entry_hash`; a
rotation signed by only one key; a reset before its cooldown *and* one before
its `effective_at_ms`; a reset signed by an impostor; a device credential signed
by the wrong identity key; two entries for one handle in one epoch (§4.3, the
NCC finding); an entry for another log; a routine update chain that must carry
its predecessor's vouch forward; and arbitrary bytes on the unauthenticated
path.

## zuu#594 — and this follows the crate proposal, not ratified spec

`KT.md` §4.4's authorization table has no case for `entry_version == 1`. A
literal implementation of the merged specification hands `@alice` to whoever
submits a first entry for it. **This does not ship that.**

`admit_submission` requires a `HandleAssertion` from a configured authority for
`entry_version == 1`, verified by `f2z-authority` — including the rule that
makes it sound, that the submission must *also* be signed by the identity key
the assertion is about, so a stolen assertion is useless.

**Stated plainly: this is `feature/kt-authority`'s proposal, offered against
#594, not ratified specification.** If review settles #594 differently, this is
the code that changes.

The **no-authority mode** is a real supported configuration for self-hosters and
it is **reported**: `GET /.well-known/free2z-kt/v1/authority` serves a
`SignedAuthorityPolicy` — signed by the log key, derived from the same
`AuthorityConfig` the submission path uses, so it cannot describe a policy the
log is not applying — whose `vouching` field says whether handles here are
vouched at all. Startup also logs a loud warning. Even in that mode the identity
key must sign the binding: proof of possession is the whole of the check when
nobody is vouching.

## New specification defects — please read these

Reported rather than coded around. I checked #594 and #586 first; none of these
is either.

> **Filed as issues, not only here:** the first two below are
> **#633** and **#634**. The rest are enumerated in this body because they are
> smaller or are resolved by an encoding choice this PR makes.

**1. §5.1's empty epoch is not producible with `akd`, and the numbering it
implies would make every audit range unverifiable.** §5.1 requires an epoch
every interval whether or not there is anything to publish, each incrementing
`epoch` by exactly 1, with *"an append-only proof over zero insertions"*.
`akd::Directory::publish` with an empty update set returns the **current** epoch
and root unchanged — it does not advance. And `audit_verify` requires exactly
one proof segment per epoch transition, with `AppendOnlyProof.epochs` matching
the tree heads one for one (which `f2z_kt_core::auditor::verify_append_only`
also checks), so a `SignedTreeHead` numbering decoupled from `akd`'s would make
`/kt/v1/audit` unverifiable for every range containing an empty epoch.
*Resolved here* by inserting **one heartbeat record per epoch** at a fixed
non-handle label, so every epoch is a real `akd` epoch. §5.1's "over zero
insertions" becomes "over one"; the property that matters — a heartbeat epoch
exists, is signed, is cosignable, and makes silence a detectable fault with a
timestamp — is preserved. Cost is one leaf per epoch, ~52k a year at 600 s.

**2. §8.1 and §9.5 require a proof of non-membership that `akd` 0.13 cannot
produce.** §9.5 deliberately has no unknown-handle code because *"'No such user'
is a claim the log must prove."* But `akd::Directory::lookup` returns
`StorageError::NotFound` for a label with no user state, and the
`NonMembershipProof` inside a `LookupProof` proves a freshness marker's absence
for a label that *does* exist — not that a label was never registered. There is
no public API in the adopted library that answers this. *Resolved here* by
telling the truth: `Presence::AbsentUnproved` is labelled unproved on the wire
and documented as an assertion, not a proof. This one needs a real answer.

**3. §9.2 specifies paths and payload *names*, not composites.** "`DirectoryEntry`
+ `LookupProof` + tree head + cosignatures" has no byte layout, and `history`'s
`{handle, params}` never says what `params` is — `akd`'s own enum has four
variants, two of which let a caller ask for unbounded work. Invented here in
`f2z_kt_core::api` and marked as such.

**4. §9.2's `/kt/v1/audit` says "both tree heads", but §6.3 rule 7 forbids
skipping an epoch.** A witness handed only the endpoints must then fetch every
intervening head one at a time before it can accept either. `AuditResponse`
carries the whole run; it is the same bytes in one round trip.

**5. §7.3's `merge_delay_exceeded` has no observation path.** A witness never
sees a `SubmissionReceipt`, so it cannot raise the fault the enum reserves for
it. Either the spec should say who raises it, or define how a receipt reaches a
witness.

**6. §9.1's `LogDescriptor` has no field for the handle-authority mode**, which
#594's acceptance criteria require to be reported. Rather than unilaterally
amend a merged normative structure, this adds a separate signed document and a
new label — which means §6.2's "closed" label set now has an addition, and a
reader should notice that.

**7. Ordering observation, not a defect.** `f2z_kt_core::auditor::WitnessState::advance`
runs `accept_chain` *after* `verify_append_only` (it requires the token the
latter returns), which inverts §7.1's step 3-before-4. Both run before any
cosignature is emitted, so the security property holds; flagging it so the next
reader does not think the loop skipped a step.

## What ships that is **not** `KT.md`-conformant — read this before treating the crate as spec-complete

Two behaviours deviate from merged normative text. Both are forced by the
adopted library, both are filed as issues, and both are visible on the wire or
in the type system rather than buried:

| Behaviour | `KT.md` says | What ships | Issue |
|---|---|---|---|
| An unregistered handle | §8.1/§9.5: answered with a **proof** of non-membership — *"'No such user' is a claim the log must prove"* | `Presence::AbsentUnproved` — an assertion, **labelled unproved on the wire**, because `akd` 0.13 has no API that produces the proof | **#634** |
| An epoch with no submissions | §5.1: published anyway, with *"an append-only proof over zero insertions"* | published anyway, carrying **one heartbeat insertion**, because `akd`'s `publish` does not advance the epoch on an empty batch and `audit_verify` needs one segment per transition | **#633** |

Neither is a shortcut taken for convenience; both are cases where a literal
implementation of the specification is not constructible on `akd`. The rest of
the deviations below are encoding choices in places `KT.md` leaves unspecified.

## Ambiguity calls

1. **Submission envelope.** `/kt/v1/submit` carries `SubmissionEnvelope`
   (entry bytes + assertion + identity binding signature), not a bare
   `DirectoryEntry`. The entry travels as opaque bytes so `validate_submission`
   applies re-encode equality to **what arrived**, not to something this crate
   decoded and re-encoded.
2. **Which entries carry an assertion is `f2z-authority`'s boundary, not this
   crate's** — this module's job is to hand it the right `EntryKind`. Required
   for a first entry and for `platform_reset`; a category error on `same_key`
   and `key_change`. Refused rather than ignored, because a field the log skips
   reading is a field that eventually gets read.
3. **`tree_size` counts heartbeats.** §6.1 says "total insertions"; they are
   insertions. §6.3 rule 4 only requires monotonicity.
4. **`published_at_ms` is nudged forward** if the clock did not advance, so the
   log's own next head is never a §6.3 rule 5 rollback against itself.
5. **JSON is a container, never a transcript.** Every JSON body carries the
   `tls_codec` bytes verbatim in base64url plus `"authoritative": "tls_codec"`,
   so a client author reaching for the convenient half has to ignore a field
   that says not to. Lookup and history are POST-with-body; the two paths that
   take a handle deliberately write **no** operator log line naming it.
6. **Error → HTTP status is deliberately coarse.** The `uint16` in the body is
   the protocol's answer; a status distinguishing `ERR_COOLDOWN` from
   `ERR_VERSION_CONFLICT` would put detail in a header that the body already
   carries in a checkable place. `ERR_INTERNAL` carries no detail anywhere,
   including its own name.
7. **TLS terminates ahead of the process.** Default listen is `127.0.0.1:8443`.
   A server binary owning certificate lifecycle would be a second, weaker copy
   of infrastructure the operator already runs.
8. **Rate limiting is global per endpoint class, not per client.** Behind a TLS
   terminator the only client identity available is a header an attacker can
   also set; a limiter keyed on that is one an attacker resets at will and can
   use to evict someone else's bucket. §9.3's separate `/kt/v1/audit` bucket is
   implemented and tested.
9. **The `kms` feature is an external signing command**, not a vendor SDK — so
   the graph carries **no cloud dependency in either feature configuration**.
   `aws kms sign`, `gcloud kms asymmetric-sign`, a PKCS#11 shim or a YubiHSM
   script all fit. Its honest limits (it cannot attest what is on the other end;
   it spawns a process per signature) are documented on the type.
10. **The nonce ledger is in memory.** Safe here because assertions are admitted
    only at `entry_version == 1`, so a nonce replayed after a restart meets a
    handle that already has a version 1 and is refused with
    `ERR_VERSION_CONFLICT`. Documented at the type.
11. **Storage is a journal plus replay, not a bespoke `akd::Database`.** §11.2
    says to treat the storage layer as security-critical; the conclusion drawn
    is to write as little of it as possible. The durable artefact is the log's
    own history and the tree is derived — **and the replay re-derives every root
    hash and `tree_size` and compares them to what the log signed**, so a
    corrupted or edited journal is a loud refusal to start. The cost is stated,
    not hidden: startup is O(entries ever submitted) and the working set is
    memory-resident. Replacing it with a durable `Database` is future work, and
    `tests/durability.rs` is the acceptance test that replacement has to pass.
12. **`/kt/v1/cosign` is idempotent per witness key**, so a witness retrying
    after a timeout cannot inflate the count a client applies a threshold to.
13. **The witness refuses a cleartext log URL by default.** A cleartext poll
    lets anyone on the path choose which history the witness cosigns — producing
    a perfectly valid cosignature for a root the log never published.
    `insecure_loopback` is a separate constructor, not a flag, so using it is
    visible at the call site.
14. **§9.2's envelopes live in `f2z-kt-core`, not in the server** (§11.4: one
    crate, three consumers). A witness that decoded a tree-head bundle
    differently from the log that encoded it would cosign a root it did not
    actually read.

## `f2z-witness healthz` — the hard contract

`healthz` is a subcommand because the distroless image has no shell, `wget` or
`curl`, and §9.3 promises there is no inbound listener to dial. It reads the
state file only — no network, no writes — and reports four distinct verdicts.
`Halted` fails the probe **and says "Do NOT clear the state file"**, because the
reflex a halted witness invites — delete state, restart — is exactly what §7.1
forbids: *"an automatic resync is an automatic way to erase the only record of
the thing the witness exists to find."* The `run` loop exits non-zero on a halt
rather than looping, for the same reason.

## Acceptance evidence

`f2z-witness/tests/acceptance.rs` drives a **real** `f2z-kt` log — §7.4's whole
argument is that there are not two implementations, so a mock would defeat it —
with real `akd` trees, real signatures and real protobuf proofs. The
equivocating log is not a mock either: it is **two real logs sharing one signing
key**, which is what an equivocating log is.

- **Happy path**: witness pins at genesis, cosigns two epoch transitions after
  verifying the append-only proof, then a **client** applies `verify_threshold`
  against its own witness set and `verify_lookup` against the root that
  survived. End to end, log → witness → client.
- **Fork**: shown the other branch for an epoch it already signed, the witness
  refuses, halts, and writes a `FaultReport` it signed itself. The test then
  *checks the evidence is genuinely self-authenticating*: two tree heads for one
  epoch, both verifying under the log's own key, with different roots.
- **Stays halted**: the honest branch is put back and the witness still refuses;
  the halt survives a process restart because it is in the state file; the probe
  agrees.
- **Append-only failure**: one log's heads with the other's proof. The witness
  refuses, and the test asserts this is the **one** kind §7.3 marks as *not*
  self-authenticating and that the proof bytes are in `detail` — without them
  the report is an accusation nobody can re-run.

Plus `tests/durability.rs` (restart re-derives every signed root; a pending
submission survives; an edited journal, a truncated one, and a journal signed by
another key all refuse to start) and `tests/redaction.rs` (keys and unpublished
submissions never render — checked as base16 in **both** cases, a **decimal byte
list**, and base64url, because `Debug` for `[u8; 32]` prints decimals and a
hex-only assertion sails straight past a complete key dump).

## Local evidence on the pinned toolchain (1.97.1)

```
  --manifest .../f2z-codec --manifest .../f2z-relay-proto \
  --manifest .../f2z-kt-core --manifest .../f2z-authority \
  --manifest .../f2z-kt --manifest .../f2z-witness
  → Every Rust toolchain pin agrees with wallet/rust-toolchain.toml
scripts/check-rust-fmt.sh    --root rs  → All 7 crates formatted
scripts/check-rust-clippy.sh --root rs  → All 7 crates clippy-clean (-D warnings)
scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
  → advisories ok, bans ok, licenses ok, sources ok  (×7)
cd rs && cargo test --locked --all-targets  → 23 test binaries, all ok
cd rs && cargo test --locked --doc          → ok
cargo build --locked --lib --target wasm32-unknown-unknown \
  -p f2z-codec -p f2z-relay-proto -p f2z-authority  → ok
cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-kt-core \
  --no-default-features --features verifier         → ok
```

Two `rs/deny.toml` additions were needed for the witness's TLS stack and each
carries its reasoning at the entry: **ISC** (`ring`, `rustls-webpki`,
`untrusted`) and **CDLA-Permissive-2.0** (`webpki-roots`, Mozilla's CA store as
data). Neither is in a crate the ZUULI or WASM clients link. The AGPL exceptions
use `AGPL-3.0-only`, not the deprecated bare `AGPL-3.0` that cargo-deny warns on.

**CI:** the first push produced no checks at all because the PR was `DIRTY`
against a `main` that had moved. After merging `main` in, the `rs` workflow
runs — status noted in a comment below.


skyl:
Triage only — draft and `CONFLICTING`, so I have not reviewed the 22k lines. Two things a rebase must handle, one of which is more than a conflict.

**17 commits behind**, and the merge base predates six merges today.

### `f2z-authority` is already on `main`, from #626

This branch adds `rs/crates/f2z-authority/` relative to its merge base, but **#626 merged that crate to `main` at `728bfe5`** after an independent review and my own verification. So this is not an ordinary textual conflict — there are two divergent versions of the same crate, and the rebase has to choose rather than merge.

`main`'s version is the reviewed one. I checked it directly: 107 tests green; `_validated`-style claims backed by real controls; the workspace strict-verification scan polices it (mutating its wrapper to plain `.verify(` fails two independent tests); keygen creates at `0o600` with `O_EXCL`, proven under umask 000 with a `0o666` probe first; and it is a standalone leaf, which is what keeps **#594's authority model an open owner decision** rather than something an agent settled.

If this branch's `f2z-authority` differs in substance, that difference needs to be a separate PR against current `main` with its own review — not folded into a rebase where it will be invisible.

### It will fail #625's ownership census

Four crates added, none registered:

```
rs/crates/f2z-authority/Cargo.toml   (already on main — see above)
rs/crates/f2z-kt-core/Cargo.toml
rs/crates/f2z-kt/Cargo.toml
rs/crates/f2z-witness/Cargo.toml
```

`scripts/check-rust-toolchain.sh` is untouched, so the census fails with `declares [package] but is registered nowhere`. I verified that diagnostic fires by probing it on the #625 and #626 branches.

**Not a problem:** the `.verify(` calls at `authority.rs:650`, `authority.rs:736`, `key.rs:129`, `key.rs:137` are the crate's own wrapper and its deliberate negative controls, not non-strict verification. Do not "fix" them on rebase.

Happy to review properly once rebased and out of draft.

skyl:
## CI ran, and the required `rs / gate` is green

The merge-train automation did **not** close this before CI finished, so there
is a real result rather than only local evidence.

```
rs / changes      pass   1m27s
rs / rustfmt      pass     22s
rs / clippy       pass    1m8s
rs / cargo-deny   pass     31s
rs / tests        pass   1m36s
rs / wasm32       pass     45s
rs / gate         pass     23s      ← the required check
```

Run [32739277037](https://github.com/free2z/zuu/actions/runs/32739277037), on
`f3d894d`, which is this branch's head.

### Why the first two attempts were red, since the history shows them

Neither was a defect in the server code, and both are worth a reviewer knowing:

1. **The very first push produced no checks at all.** The PR was `DIRTY`: it had
   been branched before `f2z-authority` merged as #626, so it conflicted with a
   `main` that had moved underneath it. Merging `main` in is what let CI start —
   and forced the API rework described in the body.

2. **`rs / gate` then failed on `check-hash-domain-labels.mjs`**, a checker added
   to `main` after `feature/kt-core` was cut. It was right about all three
   findings, and one of them is the interesting one:

   - **`f2z-kt-core` still spelled the tree-head chain-hash label `sth-hash`**,
     which makes `free2z/kt/v1/sth` — the signing-transcript constant — a proper
     prefix of it. `H(label, x) = BLAKE2b-256(label || x)` has no separator, so
     `H("…/sth", "-hash" || y)` is **bit-identical** to `H("…/sth-hash", y)` and
     the two domains are not separated at all. That is **zuu#602**, already
     closed and fixed in `KT.md` on `main`; the unmerged crate had not caught up.
     Forward-ported here, with `LABEL_STH_HASH` renamed to
     `LABEL_TREE_HEAD_HASH`.
   - **The log's own heartbeat label was a proper prefix of its value label** —
     zuu#602's defect in miniature, in code written for this PR, caught by the
     checker before it shipped. It now ends in `:`, matching `KT.md` §3.3's
     handle prefix.
   - A `f2z-kt-core` test asserted the composed `AkdLabel` as a single literal,
     which mints a token that is not prefix-free against the prefix constant
     itself. Composed from the constant instead.

   The checker earned its place twice over in one run: once on a known defect the
   unmerged base had inherited, once on a new one.

### Local evidence on the pinned toolchain (1.97.1), at the same commit

```
scripts/check-rust-toolchain.sh   → every pin agrees with wallet/rust-toolchain.toml
scripts/check-rust-fmt.sh    --root rs   → All 7 crates formatted
scripts/check-rust-clippy.sh --root rs   → All 7 crates clippy-clean (-D warnings)
scripts/check-rust-deny.sh   --root rs --config rs/deny.toml
                                         → advisories/bans/licenses/sources ok (×7)
node scripts/check-hash-domain-labels.mjs --self-test   → 17 case(s) passed
node scripts/check-hash-domain-labels.mjs               → no label is a proper prefix of another
cd rs && cargo test --locked --all-targets   → all green
cd rs && cargo test --locked --doc           → ok
wasm32: f2z-codec, f2z-relay-proto, f2z-authority, and f2z-kt-core
        --no-default-features --features verifier   → all build
```

Staying a draft per the merge-train hold.

skyl:
## Rebased onto `main` (`2ced6e4`), re-verified, ready for review

`rs / gate` — the required check — is **green** on `ff37438`, run
[32744921717](https://github.com/free2z/zuu/actions/runs/32744921717): changes,
rustfmt, clippy, cargo-deny, tests, wasm32, gate. No check on this PR is
failing; the wallet jobs are *skipped*, which is `zuuli.yml`'s change detector
behaving correctly on an `rs`-only diff.

### On the word "rebase" — I merged, deliberately

A literal `git rebase origin/main` is the wrong operation on this branch and I
did not run one. Its ancestry contains the **pre-#626** `f2z-authority` commit;
a flattening rebase replays that onto a `main` that already carries the reviewed
crate, resurrecting the superseded version through add/add conflicts on every
file in it. `origin/main` is merged in instead.

This repository squash-merges (`AGENTS.md`), so branch history shape never
reaches `main`, and the diff-against-`main` is identical either way. Say the
word if you want a reconstructed linear branch anyway.

Three commits of drift (#627, #631, #632). Every conflict was a pure addition on
both sides and every resolution keeps both — `rs/Cargo.toml` (server deps beside
main's rusqlite singleton), `rs/deny.toml` (ISC + CDLA-Permissive-2.0 beside
main's Zlib, plus the two AGPL exceptions), `check-rust-toolchain.sh` (three
manifests beside `f2z-relay-store`), `rs/README.md`, and a regenerated
`Cargo.lock`. I also corrected README prose that said "all three remain MIT" —
it predated two more crates landing.

### The four things you asked me to re-verify

1. **#602 domain separation — intact.** `check-hash-domain-labels.mjs` passes
   (`--self-test`: 17 cases; live: *"No label is a proper prefix of another"*,
   49 labels over 1469 tracked files). And `KT.md` has **not** drifted from what
   the crate spells: both say `free2z/kt/v1/tree-head-hash`.
2. **#612's strict-Ed25519 gate — passing.** It is
   `workspace_strict_verification_scan.rs`, run by `cargo test --all-targets`;
   11 cases green, all thirteen `fn verify…` registered.
3. **The two `Strictness` categories — no divergence to reconcile.**
   `rs/crates/f2z-codec/` is **untouched** by all three commits that landed since
   my merge base, so no other maintainer moved in a different direction. The
   categories stand. The key-newtype alternative I offered is still on the table
   if you prefer it on review — it is a mechanical change to `f2z-kt-core`, not a
   redesign.
4. **The equivocation test — still stages correctly** against the merged
   `f2z-authority`. All 5 acceptance cases pass, including two real logs sharing
   one signing key producing a `fork` the witness refuses, and the assertion that
   the resulting `FaultReport` is genuinely self-authenticating.

Also ran main's new **crash-safety** step (`f2z-relay-store --features
crash-injection`): 5 passed, 1 ignored.

### Two behaviours that are not `KT.md`-conformant, stated plainly

Now called out in a table near the top of the PR body so nobody reads the crate
as spec-complete. Both are forced by `akd`, both are filed, both are visible on
the wire rather than buried:

- **`Presence::AbsentUnproved`** — §8.1/§9.5 require a *proof* of non-membership
  and `akd` 0.13 has no API that produces one (**#634**). The absent answer is
  labelled unproved on the wire.
- **Heartbeat insertions** — §5.1's "append-only proof over zero insertions" is
  not constructible, because `publish` on an empty batch does not advance the
  epoch (**#633**). One heartbeat record per epoch instead.

### And the thing to know before reviewing

The PR body now leads with it: **this is the merge path for three crates.**
`f2z-kt-core` came from another agent via **#584**, which the hold closed and
nobody reopened — and this PR *modifies* it (new `src/api.rs`, the #602 label
fix, hand-written `Debug` impls, census + wasm registration). Please read
`rs/crates/f2z-kt-core/` as new code here, whatever #584 concluded about the
earlier version.

Not merging — over to you.

skyl:
## Merged `main` (`ab656ca`, #628) — resolved keeping every crate, `rs / gate` green

`mergeable=MERGEABLE` again. `rs / gate` — the required check — passes on
`edc9bb2` ([run 32747184543](https://github.com/free2z/zuu/actions/runs/32747184543)):
changes, rustfmt, clippy, cargo-deny, tests, wasm32, gate. **Zero failing checks**
across all workflows.

Merged rather than rebased, for the reason given last round: this branch's
ancestry holds the pre-#626 `f2z-authority` commit, and a flattening rebase
resurrects it. The repo squash-merges, so branch shape never reaches `main`.

### The three conflicts, all resolved by keeping both sides

`rs/Cargo.toml` needed no resolution at all — `members` is a `crates/*` glob.

- **`rs/deny.toml`** — all three AGPL binaries now named individually:
  `f2z-relay` (`AGPL-3.0`, #628's spelling, untouched), `f2z-kt` and
  `f2z-witness` (`AGPL-3.0-only`). **`ISC` was added independently by both
  sides**; I collapsed the duplicate onto #628's fuller justification and
  extended it to name the witness, which reaches the same rustls stack through
  `ureq`. A duplicate allow entry would have produced an unmatched-entry warning
  on every run, which is how a policy trains people to ignore its output.
- **`rs/README.md`** — `f2z-relay-testkit` and `f2z-relay` rows kept verbatim,
  `f2z-authority` and `f2z-kt-core` kept. I retitled my section to **"The
  key-transparency binaries"** because it no longer is the complete list of
  server binaries, and corrected #628's *"`f2z-relay` is the first crate on the
  other side of that boundary"* — there are now three.
- **`rs/Cargo.lock`** — regenerated. All nine workspace crates resolve exactly
  once.

### The four re-verifications you asked for

1. **`cargo-deny`, all 10 roots: `advisories ok, bans ok, licenses ok, sources ok`.**
   Nothing was dropped, and I did not take that on trust — deleting the
   `f2z-witness` exception makes the run fail with `rejected: license is not
   explicitly allowed`, so each binary is genuinely being judged rather than
   passing because the boundary went slack. Restored immediately.
2. **Full local suite on pinned 1.97.1**: 43 test binaries, **731 passed, 0
   failed**; `--doc` clean; clippy `-D warnings` and rustfmt clean over all 10
   crates; toolchain census green with all nine manifests registered. Both wasm
   builds pass, and main's new crash-safety step passes (5 passed, 1 ignored).
3. **`check-hash-domain-labels.mjs`**: self-test 17 cases, live run *"No label is
   a proper prefix of another"* — 49 labels over 1521 tracked files.
4. **#612's strictness registry survived the auto-merge intact.** All **thirteen**
   KT entries are present alongside #628's two, and both my `Strictness`
   categories (`DelegatesToFn`, `NotASignatureCheck`) are still in place. #628
   removed a thin wrapper rather than registering one, so it did not touch the
   categories.

### On `workspace_debug_scan` — I checked rather than assumed, and it matters

You were right to flag it. #628's hardening (consume an attribute until its
brackets balance) is now in this branch, so the scan reaches these files — but a
green result proves nothing on its own, since the failure mode you described is
*silent*.

So I ran a negative control in **each of the three crates**: reinstated a derived
`Debug` over a `Vec<u8>` and confirmed the scanner reports it.

```
f2z-kt/src/log.rs:          `NegativeControlPending`  → flagged
f2z-witness/src/state.rs:   `NegativeControlWitness`  → flagged
f2z-kt-core/src/submit.rs:  `NegativeControlAccepted` → flagged
```

All three reverted; `git diff` on those paths is empty. Separately, **none of my
crates carries a multi-line attribute under `src/`**, so the pre-#628 scanner was
not aborting on them either — the hand-written `Debug` impls were being checked
before, and are provably being checked now.

### Unchanged and still the two most important things in the body

The **three-crate provenance table** (`f2z-kt-core` came from another agent via
#584, which the hold closed, and this PR *modifies* it) and the
**non-conformant-behaviour table** (`Presence::AbsentUnproved` for **#634**,
heartbeat insertions for **#633**) are both intact at the top of the PR body.

Not merging — over to you.

skyl:
Reviewed on the merged tree, since this landed before I could review it out of draft. **The headline: this does not settle #594, and it goes to real trouble not to.** That was my main worry when #626 landed `f2z-authority` as a standalone leaf, so I checked it first.

### `f2z-authority` is no longer a leaf — and that is handled correctly

```
rs/crates/f2z-kt/Cargo.toml:52       f2z-authority = { path = "../f2z-authority" }
rs/crates/f2z-witness/Cargo.toml:53  f2z-authority = { path = "../f2z-authority" }
```

When #626 merged I reported that nothing depended on `f2z-authority`, and that this was what kept **#594's authority model an open owner decision**. That is no longer structurally true, so the question becomes whether the decision got made by implementation. It did not, and `admit.rs:28-47` is why:

> §4.4's table authorizes `same_key` by *"the `directory_auth_pk` published in the previous entry"*. At `entry_version == 1` there is no previous entry … and **nothing in the numbered rules, the table, or the prose says what authorizes it**. A literal implementation … hands `@alice` to whoever submits a first entry for it … That is [zuu#594], it is unresolved, and **this log does not ship it.**
>
> [`admit_submission`] requires a `HandleAssertion` from a configured authority for a first entry. That is `f2z-authority`'s **unratified candidate**, offered against #594 rather than merged protocol …

Three things make that more than a disclaimer:

1. **The alternative is real.** `AuthoritySet::none` is an explicit no-authority mode, not a theoretical one.
2. **The posture is reported, not inferred.** I checked this specifically, because honesty that a client cannot observe is decoration. `policy.rs:170-181` maps `VouchingStatus::Unvouched` → `VOUCHING_UNVOUCHED` inside a **signed** policy document, so a client connecting to an unvouched log can see it.
3. **It refuses to over-apply its own mechanism.** Assertions are *refused* for `same_key` and `key_change`, with the right reason: "Admitting a platform assertion here would give the platform a second way to authorize a change to a live handle — the power ADR 0014 spent a cooldown, an alarm and a per-epoch counter to constrain."

That third point is the one that convinced me. Wiring a new authority mechanism into every entry kind would have been the easy path and would have quietly expanded platform power past what ADR 0014 negotiated. Declining to is a judgement about the threat model, not about the code.

**So #594 remains the owner's to decide.** What has changed is that there is now a running candidate to decide *about*, and a log that is honest about which mode it is in.

### One finding, filed as #639

`f2z-relay` reimplements `SeenSet::retention_is_sound` in `config.rs:886` rather than calling it — `grep -rn 'retention_is_sound' rs/crates/` returns only the definition, its own tests, and doc references. The two are equivalent today and the daemon's version is well-argued ("#586's finding, enforced rather than published … this relay refuses to open it"). But the boundary already moved once this week: #624 showed `>= 2 × skew` is sound only because #631 made the seen-set interval closed. Had that gone the other way, `replay.rs` would have changed and `config.rs:886` would have kept admitting the configuration #624 demonstrated accepts a replay at exactly `t + skew`.

This repo already stated the principle, in `f2z-relay-store`: *"§8.2's anti-pre-ack rule implemented twice is §8.2's rule implemented once and violated once."*

### On the merge itself

Recorded in #637 and not re-litigated here: this merged while its zuuli run on `edc9bb20` was still `queued`, so `gate` never judged the merged tree. Main's own run (`32747879787`) came back **green, 17/17**, so no revert. #638 merged 26 minutes later with both runs green on its exact final head in about six minutes, which is the better pattern and is what #637 now points at.

### What I did not review

At +18076 I checked the authority boundary, the policy reporting, and the crate graph. I have not reviewed the log service's durability story, the witness cosigning protocol, or the HTTP surface. Those are worth their own pass.

skyl:
Second merged-tree pass, covering `f2z-witness`. **No findings.** Recording what I checked and why it holds, because the design here answers questions I would normally have to file.

### The halt is sticky by construction, not by discipline

The property that matters for a witness is §7.1's: *"Once halted it stays halted until a human looks at the evidence. An automatic resync is an automatic way to erase the only record of the thing the witness exists to find."*

`Health::Halted` is not a runtime flag that a restart clears — `health.rs:134-139` derives it from persisted `state.halt_kind()`. Restarting the process re-reads the same file and halts again. A flag would have needed a test to prove nothing resets it; derived state needs no such test, which is the stronger arrangement.

### It anticipates the operator's wrong reflex

This is the part I want to single out, because it is rare:

> **`Halted` deliberately fails the probe.** … The reason it is reported as a *distinct* verdict, with its own message naming the fault kind and the evidence directory, is so that whoever reads the probe output does not respond to a detected equivocation by deleting the state file and starting again — which is precisely the reflex that destroys the evidence.

The same reasoning is applied again to an unreadable state file, which is `WitnessError::Local` rather than `Unpinned`, because *"reporting it as 'never polled' would invite exactly the wrong repair."* Designing the diagnostic against the human response to it, not just against the machine state, is a human-factors argument inside a security control.

I checked the residual: halt stickiness does depend on a file an operator can delete. But evidence is written to its **own directory** (`evidence.rs:64-80`), so the deletion reflex loses the halt and keeps the artefact. That is the right split, and it is why the probe message names the evidence directory rather than just the fault.

### It distinguishes what the evidence proves from what it asserts

Most fault kinds are self-authenticating — *"two or more tree heads signed by the log's own key"*, checkable by anyone with the log's public key, trusting nobody. And then:

> One kind is different and this module does not pretend otherwise: `append_only_failure` is **not** self-authenticating. The claim is "this proof does not verify", and a third party must re-run `audit_verify` on the bytes to see it. So the proof bytes go in `detail` — without them the report is an assertion nobody can check — and the report is a prompt to check, not a verdict.

A signed report that merely *asserts* a fault, presented alongside reports that *prove* one, is exactly the kind of over-claim I have spent this session filing. This declines to make it, names the one case that differs, and ships the bytes needed to check it.

Also: *"The witness signs its own reports. That is deliberate: §7.3 — a witness that cries wolf is on the record too."*

### Coverage so far on this PR

Reviewed: the authority boundary and #594 posture (previous comment), the witness fault/halt/evidence path (here). `f2z-witness` and `f2z-kt` durability tests are honestly scoped — restart-replay and journal-tamper detection, explicitly not crash injection, and the header says so, so I am not filing that as a gap.

Not reviewed: the HTTP/api surface, `ratelimit.rs`, and `vrf.rs`. Those remain worth a pass.

One thing I looked at and deliberately did **not** file: `log.rs:488` documents that the receipt is signed after the `fsync`, and nothing tests it. The ordering is correct in the code, and a violation would not be externally observable — if the append fails, `?` returns `Err` and no receipt escapes; if the process dies between signing and appending, the receipt was never returned. It is a code-order property, so "untested" would have been a misleading finding. If it is ever worth pinning, #627's `orderedAnchors` is the tool.

PR #631, opened 2026-08-24
fix(relay): close anti-replay expiration boundary

### LINKED ISSUE #624: f2z-relay-proto: a signed frame is accepted twice at exactly t+skew on the shipped default — retention_is_sound's >= must be strict

skyl:
Closes #624.

## Fix
- Keep seen-set entries live through their inclusive expiration instant, matching the inclusive timestamp window.
- State the retention relationship and boundary semantics in WIRE.md.
- Exercise the real CommandVerifier from the earliest valid presentation through the latest valid replay.
- Pin zero-retention and overflowing expiration behavior through the public SeenSet API.

## Mutation proof
- The new verifier reproduction failed red against the old half-open production predicate before the fix.
- Production expiration >= changed back to > compiles and fails red.
- Production saturating_add changed to wrapping_add compiles and fails red near u64::MAX.
- Inclusive timestamp <= changed to < compiles and fails red through the verifier boundary.
- Restored candidate passes 88 unit, 6 property, and 8 redaction tests, plus strict clippy and fmt.

Independently reviewed at exact head 63d0ec97d95312cd20243d1667079c8923998d11.

### DIFF
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -578,7 +578,12 @@ Two mechanisms, together:
    within the window and rejects a repeat with `ERR_REPLAY`. `nonce` is 16 bytes
    from the client's CSPRNG, fresh per command; the birthday bound is 2^64 per
    key, which is not reachable inside a two-minute window at any rate a relay
-   would serve.
+   would serve. `antireplay_window_ms` MUST be at least `2 × clock_skew_ms`, and
+   an entry MUST remain live through the instant at which that retention period
+   ends. Both boundaries matter: the timestamp window is inclusive, so a frame
+   first accepted at `timestamp_ms - clock_skew_ms` remains valid at
+   `timestamp_ms + clock_skew_ms`. A half-open seen-set would forget it at that
+   same instant when the two published windows are equal.
 
 **The seen-set is fail-closed.** It is bounded (`antireplay_seen_max`, published)
 and when the bound is reached the relay returns `ERR_BACKPRESSURE` and refuses new
--- a/rs/crates/f2z-relay-proto/src/capabilities.rs
+++ b/rs/crates/f2z-relay-proto/src/capabilities.rs
@@ -488,8 +488,8 @@ pub fn defaults(relay_identity_pk: &PublicKey, published_at_ms: u64) -> Result<C
 
         // §5.5 default: ±2 minutes.
         clock_skew_ms: 120_000,
-        // Twice the skew, for the reason `Refusal::AntiReplayWindowTooShort`
-        // gives. `WIRE.md` publishes no default for this field.
+        // Twice the skew. Equality is sound because §5.5 requires seen-set
+        // entries to remain live through their expiration instant.
         antireplay_window_ms: 240_000,
         antireplay_persistence: AntiReplayPersistence::Volatile.code(),
 
--- a/rs/crates/f2z-relay-proto/src/command.rs
+++ b/rs/crates/f2z-relay-proto/src/command.rs
@@ -1201,6 +1201,34 @@ mod tests {
     }
 
     #[test]
+    fn a_frame_seen_at_the_earliest_valid_instant_is_still_seen_at_the_latest() {
+        let key = SigningKey::from_seed(&[0x9a; 32]);
+        let signed = SignedCommand::<ops::Subscribe>::create(
+            &transcripts(),
+            2,
+            QueueAddress::new([0x9b; 32]),
+            NOW,
+            nonce(0x9c),
+            &key,
+            &Empty,
+        )
+        .unwrap();
+        let request = signed.request().unwrap();
+        let mut verifier = verifier();
+
+        assert!(
+            verifier
+                .verify::<ops::Subscribe>(NOW - 120_000, 2, &request)
+                .is_ok()
+        );
+        assert_eq!(
+            verifier.verify::<ops::Subscribe>(NOW + 120_000, 2, &request),
+            Err(ProtoError::Wire(ErrorCode::Replay)),
+            "the inclusive timestamp window must not outlive the seen-set entry"
+        );
+    }
+
+    #[test]
     fn the_timestamp_window_is_checked_before_the_signature() {
         let key = SigningKey::from_seed(&[5u8; 32]);
         let signed = SignedCommand::<ops::Subscribe>::create(
--- a/rs/crates/f2z-relay-proto/src/replay.rs
+++ b/rs/crates/f2z-relay-proto/src/replay.rs
@@ -157,7 +157,7 @@ impl fmt::Debug for ReplayKey {
 /// but they are not independent. See [`SeenSet::retention_is_sound`].
 #[derive(Clone, Debug)]
 pub struct SeenSet {
-    /// Value is the instant at which the entry may be dropped.
+    /// Value is the final instant at which the entry must remain live.
     entries: BTreeMap<ReplayKey, u64>,
     retention_ms: u64,
     max_entries: usize,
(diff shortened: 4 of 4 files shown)


skyl:
Reviewed on `63d0ec97`. **This closes #624 correctly. Merge when `gate` and `rs / gate` are green.**

I set one requirement when I filed #624: the fix must make that issue's exact reproduction go red→green. It does, and I ran both halves rather than taking the body's word.

### Red→green, verified both ways

I pasted #624's reproduction verbatim into `command.rs` on this branch and ran it against the real `CommandVerifier`:

```
WITH the fix          → ok. 1 passed; 0 failed
WITHOUT the fix       → FAILED
  (production predicate reverted to `*expires_at > now_ms`, line 240 printed back)

assertion `left == right` failed: REPLAY ACCEPTED at exactly t+skew,
got Ok(Verified { body: Empty, address: QueueAddress(<redacted>), … })
```

So the test detects the defect, and this change is what closes it. Full suite on the branch: **103 tests, 0 failed** (88 unit, 6 property, 8 redaction, 1 doctest).

### The property is pinned in-tree, not just by my ad-hoc test

`a_frame_seen_at_the_earliest_valid_instant_is_still_seen_at_the_latest` is essentially my reproduction, through the real verifier, with a name that states the invariant rather than the mechanism, and a message that names the consequence: *"the inclusive timestamp window must not outlive the seen-set entry."* That is the version that survives a future refactor, because a reader can tell what it is for.

### It fixes the reasoning, not only the predicate

The thing I most wanted to see gone is gone. `replay.rs` used to carry:

> Which end is closed does not matter to the security property, because soundness comes from `retention_ms >= 2 * clock_skew_ms` … and not from a single millisecond at the boundary

That was the conclusion restated as its own justification, appealing to the very invariant that admitted the failing case. It is replaced with the opposite claim, correctly stated:

> This final instant is security-significant when retention equals twice the inclusive timestamp skew: the original frame is still valid there and must still be recognized as a replay.

`capabilities.rs` gets the same treatment — the bare "Twice the skew" comment now says *why* equality is sound ("§5.5 requires seen-set entries to remain live through their expiration instant"), which is exactly the fact that makes the shipped default safe rather than accidentally safe.

### The spec now states the composition

`WIRE.md` no longer leaves the relationship implicit:

> `antireplay_window_ms` MUST be at least `2 × clock_skew_ms`, and an entry MUST remain live through the instant at which that retention period ends. Both boundaries matter: the timestamp window is inclusive, so a frame first accepted at `timestamp_ms - clock_skew_ms` remains valid at `timestamp_ms + clock_skew_ms`. A half-open seen-set would forget it at that same instant when the two published windows are equal.

That is the whole defect written down normatively, including *why* it only bites at equality — which is what stops a second implementation rediscovering it. Choosing the closed-interval option also means `retention_is_sound`'s `>=` stays correct rather than needing to become strict, so the shipped default remains usable.

### Worth noting for the record

The three additional mutations claimed in the body are the right ones — `>=`→`>`, `saturating_add`→`wrapping_add` near `u64::MAX`, and the timestamp `<=`→`<` — because between them they cover both ends of the composition rather than only the end that was broken. I re-ran the first myself; I did not independently re-run the overflow or timestamp-window mutants.

### What I did not observe

`gate` has not completed on this head.

PR #632, opened 2026-08-24
feat(rs): add mutation-proved durable relay queue store

### LINKED ISSUE #605: f2z-relay-store: MemoryStore returns gapped pages that cumulative ACK then destroys, and a rolled-back transaction discards recorded activity

### LINKED ISSUE #606: test(f2z-relay-store): delete-on-ack can be disabled with every test green, and the idle-timer test passes with zero touches
`f2z-relay-store` (#595, closed by the sweep; branch `feature/relay-store` survives) has **13 of 24 mutations uncaught**. Two clusters, both the #589 shape — a test that passes for a reason unrelated to the property it names.

## 1. Delete-on-ack can be disabled entirely and every test stays green

M3 (off-by-one in the ACK prefix delete), M4 (the delete neutered), and M10 (`delete_queue` leaves every message row behind) **all pass**.

The reason is structural, not an oversight in any one test: `SqliteStore` maintains `stored_messages` / `stored_bytes` by **arithmetic independent of the rows**, and `read()` filters below the watermark **by construction**. So the accounting says the rows are gone because it decremented a counter, and the reader can't see them because it wouldn't look there anyway.

With M4 applied, every public observable reports `0 messages, 0 bytes` while **12,288 bytes of ciphertext sit on disk**.

That is not only a test gap. `WIRE.md` argues delete-on-ack does not create a deletion oracle *and* that acked ciphertext is actually gone — a retention property a relay operator would be asked to attest to. Nothing in this crate currently distinguishes "deleted" from "no longer counted."

**And the crash test that looks like it covers this does not.** `a_crash_after_the_ack_commit_leaves_nothing_acked_but_undeleted` asserts "every index at or below the persisted watermark must be gone" — through a `stored()` helper that calls `store.read()`, which by construction cannot see rows below the watermark. The whole crash suite reports `5 passed; 0 failed` with delete-on-ack fully disabled.

**Fix:** assert against the storage layer, not the reader — a direct `SELECT COUNT(*)` / row-level check that no message row at or below the watermark survives, and that the on-disk byte total matches the accounting rather than being derived from it. The accounting and the rows must be checked against each other, since each is currently self-consistent.

## 2. The idle-timer test proves nothing about touches

`the_idle_timer_retires_the_queue_and_activity_resets_it` sweeps at **60,500 ms** against an **untouched** deadline of **61,000 ms**. The queue survives the sweep whether or not the touch happened, so the assertion named "activity resets it" passes with **zero touches**.

M21 and M23 confirm it: every activity-recording path in the crate can be deleted and the test stays green.

This is a fixture shaped so the invariant cannot be violated — the same class as `f2z-authority`'s prefix test ranging over 7 of 13 labels (#602), and `f2z-relay-proto`'s all-zero signature (#589). The fix is to sweep at a time that falls **between** the untouched and touched deadlines, so the test can only pass if the touch moved it.

## Method note, worth carrying forward

Two mutations were discarded and redone because they didn't test what they appeared to: one hit a borrow error, and one used `SELECT` in `tx.execute` and was caught by rusqlite's `ExecuteReturnedResults` rather than by any property of the code. Every mutation's changed line was printed back before judging.

Both failure modes look identical to "the guard works" — which is the same trap I fell into twice tonight, once mutating a doc comment instead of a constant and once probing with an untracked file that `git ls-files` correctly ignored.

## Acceptance

- Deletion is asserted at the row level, independently of the accounting counters and independently of `read()`; M3, M4 and M10 all go red.
- The crash suite's `stored()` helper no longer routes through `read()`.
- The idle-timer sweep falls between the touched and untouched deadlines; M21 and M23 go red.

Found while reviewing #595. Companion: #605 (two real bugs in the same PR). Related: #589, #602.

skyl:
Closes #605.
Closes #606.

## What ships
- A backend-neutral RelayStore contract with dense contiguous paging, cumulative delete-on-ack semantics, queue lifecycle, TTL, quotas, and durability reporting.
- A volatile MemoryStore and durable SQLite WAL store with FULL synchronous commits, secure deletion, group commit, rollback-safe staged activity, and crash-injection coverage.
- Workspace, toolchain, supply-chain, documentation, and CI registration for the new native crate.

## Behavioral and mutation proof
- Memory and SQLite gapped-page break-to-continue mutants expose noncontiguous pages and fail.
- ACK boundary and deletion no-op mutants fail through independent physical-row and quota checks.
- DELETE_QUEUE message deletion no-op leaves ciphertext rows and fails.
- Activity touch removal, iter-to-drain rollback loss, successful-clear removal, and clear-before-COMMIT ordering all compile and fail.
- The COMMIT-order proof stages real activity, forces SQLite deferred-FK failure at COMMIT, proves rollback and buffered touch preservation, then retries through production expiration and persists the touch.
- Signed and unsigned metadata corruption and replay commit-to-preflight mutants fail.
- Full Rust 1.97.1 workspace all-targets, crash suite, fmt, strict clippy, toolchain, and workflow policy checks pass.

Independently reviewed at exact head 845c7e1a5de9f8ee2f0aaba79e3173088b8cb43c.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -273,6 +273,18 @@ jobs:
         working-directory: rs
         run: cargo test --locked --doc
 
+      # `crash-injection` is off by default because it compiles an `abort()`
+      # into f2z-relay-store's commit path, so the default run above skips
+      # `tests/crash_safety.rs` entirely. Those cases are the only evidence that
+      # delete-on-ack's two invariants survive an unplanned stop — nothing
+      # acked-but-undeleted, nothing accepted-but-lost — and they prove it by
+      # killing a real child process, which an in-process caught panic cannot.
+      # A step of their own, so a run that skipped them is visible rather than
+      # inferred.
+      - name: Run the crash-safety suite, which kills real processes
+        working-directory: rs
+        run: cargo test --locked -p f2z-relay-store --features crash-injection --test crash_safety
+
   rs_wasm:
     name: rs / wasm32
     needs: changes
@@ -312,6 +324,11 @@ jobs:
       # rather than `--workspace` so that a future server binary — which will be
       # AGPL, will pull tokio, and will never reach a browser — does not
       # silently join the wasm build and then fail it.
+      #
+      # `f2z-relay-store` is deliberately **absent** and must stay absent: it
+      # links SQLite through `rusqlite`'s bundled C amalgamation, which is a
+      # native build by construction. It is a relay-side crate; no client links
+      # it.
       - name: Build the client-linked crates for the browser target
         working-directory: rs
         run: |
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -57,6 +57,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cc"
+version = "1.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ad534f4357a5264cce5019c989cf66a4f0dc4e0d1b1d15f8aacec0ff7360273"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -200,6 +210,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "f2z-relay-store"
+version = "0.1.0"
+dependencies = [
+ "f2z-codec",
+ "f2z-relay-proto",
+ "proptest",
+ "rusqlite",
+ "tempfile",
+]
+
+[[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
+
+[[package]]
 name = "fastrand"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -212,10 +245,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "64cd1e32ddd350061ae6edb1b082d7c54915b5c672c389143b9a63403a109f24"
 
 [[package]]
+name = "find-msvc-tools"
+version = "0.1.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d45db016d36b838f563236e9193d0ee6ce38f3f68b6c94e914b4929c96bbb890"
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
 name = "generic-array"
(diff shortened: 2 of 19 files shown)


skyl:
Reviewed adversarially on `845c7e1a`, verified by running the crate and mutating it in a throwaway worktree. **Sound — merge when `gate` and `rs / gate` are green.**

### Baselines, clean tree first

```
bash scripts/check-rust-toolchain.sh   → EXIT=0
cargo test -p f2z-relay-store          → 36 tests, 0 failed (14 + 17 + 4 + 1)
```

Branch is exactly one commit on current `main`, and `rs/crates/f2z-relay-store/Cargo.toml` is registered in `MANIFESTS`, so #625's ownership census is satisfied rather than tripped.

### The invariant I most wanted to break holds, including the off-by-one

`QueueState::restore` is the load-bearing addition here — a store rehydrating persisted rows is exactly where §8.2's anti-pre-ack rule gets re-derived and quietly weakened. The comment states the risk precisely:

> §8.2's anti-pre-ack rule implemented twice is §8.2's rule implemented once and violated once.

So I attacked the boundary rather than the rule. `restore` rejects `acked_through >= next_index`, which is correct because `next_index` is the *next* slot and the highest appended is `next_index - 1`. Relaxing it by one — which would permit acking exactly one message beyond anything that was ever appended:

```
135:        if let Some(watermark) = acked_through
136:            && watermark > next_index          ← was >=
```

```
FAILED. 88 passed; 1 failed
queue::tests::restore_round_trips_a_live_queue_and_refuses_the_impossible_ones
  panicked at crates/f2z-relay-proto/src/queue.rs:655
```

The exact off-by-one is covered, not just the gross case.

### The crash suite is a real crash, and I checked rather than assumed

"Crash coverage" is usually a graceful shutdown with a hopeful name. This is not:

```rust
/// `abort()` rather than `exit()`: a normal exit runs at-exit handlers and
std::process::abort();
```

run in a **child process** at a chosen instant inside the commit, with the module comment spelling out why it matters — "no unwinding, no destructors, no `sqlite3_close`". Four scenarios straddle both commits in both directions (after/before append, after/before ack), which is the shape that distinguishes durability from luck: crashing *after* a commit must lose nothing, crashing *before* it must leave no trace.

### The COMMIT-failure test asserts its own precondition

This is the detail that makes the rollback proof real rather than decorative:

```rust
tx.execute("INSERT INTO commit_failure (parent_id) VALUES (7)", []).unwrap();
let error = store.commit(tx, activity)
    .expect_err("the deferred constraint fails at COMMIT");
assert!(matches!(error, StoreError::Backend(_)));
```

A `DEFERRABLE INITIALLY DEFERRED` foreign key is accepted by the statement and rejected at `COMMIT`, and `expect_err` means the test fails loudly if that ever stops being true. Without it, a test that "proves rollback" could pass because the commit quietly succeeded and there was nothing to roll back. Same shape as #626's fixture guards, and it is becoming a pattern in this repo worth keeping.

### What I did not observe

`gate` has not completed on this head. I did not re-run the paging, quota, TTL, or secure-deletion mutants from the body — I verified the anti-pre-ack boundary and inspected the crash and COMMIT-failure mechanisms, and I am relying on the PR's account for the rest. I also did not review the SQLite schema for its own sake, only the durability machinery around it.

PR #635, opened 2026-08-24
docs(ci): point to the workflow-gates hazard rationale

### LINKED ISSUE #591: ci: restore the standalone workflow-gates step and the comment explaining why it exists

skyl:
Closes #591

Adds the narrowed pointer requested by the issue owner from the always-required `changes` policy call site to the canonical rationale in `scripts/check-workflow-gates.mjs`. This deliberately does not restore the superseded standalone step or duplicate the invariant.

Verification:
- `node scripts/check-workflow-gates.mjs --self-test` (39 mutation/negative-control cases)
- `node scripts/check-workflow-gates.mjs` (live repository scan)
- `node scripts/check-zuuli-android-gate.mjs --self-test`
- `node scripts/check-zuuli-android-gate.mjs`

No executable guard changes: exact diff is two comment lines. Both executable workflow-gates call sites retain ordered self-test then live verdict. Independently reviewed at exact SHA `301ddd7a675f4d7f257d94f7c69f390c4911a3cf`.


skyl:
This is exactly the narrowed ask from #591, and it is right to keep it to two lines.

Verified:

- **Comment-only.** The entire diff is two `#` lines. No executable change, so there is nothing to mutation-test — and a mutation table here would be theatre, not evidence.
- **The pointer resolves.** `scripts/check-workflow-gates.mjs:3-9` does carry the full rationale, including the sentence that answers #591's original premise: *"The failure mode is silent and it looks exactly like success, **which is why it is worth a script rather than a comment.**"*
- **It does not duplicate the invariant.** The body says so explicitly, and that matters: a second copy of the hazard statement in `zuuli.yml` would be a claim with no test holding it accurate, which is the drift this repo keeps catching. A pointer cannot go stale in the same way — if the header moves, the reference is still to the right file.

Approving in substance; merge when `gate` and `rs / gate` are green. Given `wallet/*` now broadens the change filter, a two-comment-line PR still pulls the full macOS/Windows matrix, so expect ~25-60 minutes for a change that alters no behaviour at all. That is the cost side of the tradeoff I flagged on #625 — worth noting, not worth acting on.

I have not observed `gate` complete on this head.

PR #638, opened 2026-08-24
ci(rs): container supply chain for the three E2EE server images

skyl:
`f2z-relay`, `f2z-kt` and `f2z-witness` are on `main` and have nowhere to be
deployed from. This builds the images.

`rs/Dockerfile` — one file, three binaries, `--build-arg BIN=`. One rather than
three because what must not drift between them is exactly the shared part: the
pinned compiler, the pinned bases, the uid, the absence of a shell. The runtime
is `gcr.io/distroless/cc-debian12:nonroot` pinned by index digest and running as
uid 65532 — `cc` and not `static` because `rusqlite` bundles SQLite and all
three link libc. The builder is `rust:1-slim-bookworm`, also digest-pinned, and
bookworm is load-bearing: the runtime is debian12, and a newer builder links a
newer glibc and yields a binary that starts on the build host and dies on the
cluster.

The compiler version appears nowhere in the Dockerfile or the workflow. It
arrives as `RUST_CHANNEL` from `scripts/check-rust-toolchain.sh
--print-channel`, and the Dockerfile refuses to compile if what arrives
disagrees with the toolchain file in the build context.

`ENTRYPOINT` is a literal path per binary — exec-form `ENTRYPOINT` does not
expand build arguments, and the deployment's liveness probe runs
`/f2z-witness` literally. Hence a stage per binary and a final `FROM
bin-${BIN}`.

`.github/workflows/f2z-images.yml` follows `zuuli-linux-image.yml` exactly,
including its two-phase flow: publishing changes what is available and changes
what anything consumes not at all. Tags are `:main-<sha>` and a build-scoped
tag, never `:latest`; the publish attaches an OCI SBOM and `provenance:
mode=max`, attests the registry digest with `actions/attest
--push-to-registry`, re-verifies by pulling that digest, and prints it into the
job summary for the follow-up pinning review. `platforms: linux/amd64` is what
the target cluster runs; multi-arch publishes new digests beside the old ones
and cannot invalidate a pin.

Reproducible builds are not claimed, and the Dockerfile and README say so: the
attestation proves provenance — this workflow, this repository, that commit —
not that a rebuild yields the same digest.

`scripts/check-f2z-images.mjs` is the policy, with a `--self-test` that watches
19 mutations fail and a clean-tree positive control. The rule it exists for
first is the `paths:` filter: a filter that has drifted off `rs/**` produces no
red job anywhere and the images simply stop being rebuilt, so it is parsed and
matched against real crate paths, with `wallet/zuuli/src/App.tsx` as the
negative control against a filter widened until it means nothing. It also runs
against a built container — `--inspect` asserts the entrypoint, the uid, the
architecture and that no shell is present; `--probe` asserts the witness's
`exec` liveness contract inside the image.

It runs from `rs.yml`'s always-on `changes` job and again in the `rs / gate`
policy recheck, deliberately not from the image workflow, because the image
workflow's own `paths:` filter is one of the things being checked and a
workflow that has stopped running cannot report that it has. The image
workflow itself is registered in `UNGATED_WORKFLOWS` with that reason.


skyl:
Reviewed adversarially on `3ae778be`, verified by running the checker and mutating the tree in a throwaway worktree. **Sound — merge when `gate` and `rs / gate` are green.**

### The central property holds, and it is checked by consequence rather than by string

The stated reason this script exists is that a `paths:` filter fails by *not running* — no red job anywhere, images silently stop being rebuilt. So that is what I attacked. Drifting the filter off `rs/**`:

```
19:  pull_request:
20:    paths:
21:      - docs/**        ← was rs/**
```

```
EXIT=1
workflow: the pull_request paths: filter does not cover rs/** at all
workflow: the pull_request paths: filter does not select rs/Dockerfile;
          a drifted filter is a workflow that has silently stopped publishing
workflow: the pull_request paths: filter does not select rs/Cargo.lock; …
workflow: the pull_request paths: filter does not select rs/crates/f2z-relay/src/main.rs; …
```

The last three are the part that matters. The checker does not merely compare the filter to an expected literal — it evaluates it against **actual tracked files** and names the ones that would stop being selected. A literal comparison would be defeated by any rewrite that still spelled `rs/**` somewhere; this is defeated by nothing short of a filter that genuinely selects the tree.

### The exemption is an exemption with a compensating control, not a hole

Adding a workflow to a repo where #614 makes an ungated workflow fail the checker is the interesting bit. This does not quietly widen an allowlist; it registers with a written justification that names where the property is actually enforced:

> its pull-request leg is path-filtered to rs/** and its publish leg only runs on main. The property that must hold for a merge … is enforced from inside the rs gate by `scripts/check-f2z-images.mjs`, which runs on every pull request regardless of the filter.

And the live gate checker confirms the registration took: **"13 registered ungated"**, up from 12.

### The wiring is the #599 shape, and I checked it properly

The diff hunk shows four `check-f2z-images.mjs` invocations adjacent, which reads at a glance like a duplicated block. It is not — reading `rs.yml` by job rather than by diff:

```
changes:  line  91  - name: Verify the f2z server image supply chain
gate:     line 412  - name: Recheck the workflow policy
```

Two commands each, in two different jobs. That is the #599 pattern: the gate re-runs the policy so job-level `continue-on-error` on `changes` cannot turn a failing verdict into the `success` the gate consumes. Worth stating because the diff genuinely misleads here.

### Baselines, clean tree first

```
check-f2z-images.mjs           → EXIT=0  "policy holds"
check-f2z-images.mjs --self-test → EXIT=0  "19 mutation(s) rejected, clean tree accepted"
check-workflow-gates.mjs       → EXIT=0  2 gates, 13 registered ungated, 59 jobs, 15 workflows
```

"clean tree accepted" in the self-test output is the positive control — it rejects 19 mutants *and* asserts the unmutated tree passes, which is the precondition-guarding pattern #626, #631 and #632 established.

### What the body gets right beyond the code

Two things I want to credit because they are the opposite of what I usually file:

- **"Reproducible builds are not claimed"**, stated in the Dockerfile and README. The attestation proves provenance — this workflow, this repo, that commit — and the PR says so rather than letting "attested" imply "reproducible."
- **The bookworm pin is explained as load-bearing**: a newer builder links a newer glibc and yields a binary that starts on the build host and dies on the cluster. That is a failure that would present as a mystery crash in production, and it is written down next to the pin that prevents it.

### What I did not observe

`gate` has not completed on this head. I did not build the images, verify the distroless digest resolves, or exercise the publish leg — it only runs on `main`, so its first real execution will be post-merge. I checked the policy, not the artifacts.

PR #640, opened 2026-08-24
test(zcash): mutation-prove send path gaps

### LINKED ISSUE #616: test(zcash): expiry and change-memo mutants survive, and the send-all retry loop is still uncovered

skyl:
Closes #616

Closes the remaining native send-path verification gaps with behavioral coverage through the exact shipping production callers:

- one private confirmation policy drives both proposal construction and send-all wallet summaries;
- the ZIP-317 minimum is derived from the fee rule rather than copied as a literal;
- the send-all retry test requires more than one fee-convergence iteration and exercises the exact minimum-balance boundary;
- actual created transaction bytes prove builder expiry and empty change memo after decryption.

Mutation evidence (each compiled and went red, then was restored):
- transaction expiry `None -> Some(height)`;
- change memo `None -> Some(probe)`;
- summary confirmation policy `default -> MIN`;
- proposal confirmation policy `default -> MIN`;
- retry range `0..3 -> 0..1`;
- minimum guard `<= -> <`;
- fee-rule derivation replaced with an independent literal.

Verification on exact head `4c08f99e58d6798270c2ab698c4b4acbcf96db6d`:
- 114 compatibility mutants killed + live source contract green;
- 145 unit tests + 1 shipping-feature production-route integration green;
- pinned Rust 1.97.1 fmt and clippy `-D warnings` green;
- independently reviewed exact head; stable patch ID and all resulting blobs match the mutation-approved candidate.

### DIFF
--- a/scripts/check-librustzcash-compat.mjs
+++ b/scripts/check-librustzcash-compat.mjs
@@ -399,6 +399,8 @@ function validate(snapshot, scope = reviewedCompatibilityScope()) {
     "pub(super) fn create_transactions<",
     "pub(super) fn propose_fixed_validated<",
     "pub(super) fn propose_send_all_validated<",
+    "pub(super) fn standard_minimum_fee(",
+    "pub(super) fn wallet_summary_with_policy<",
   ];
   // Comments are whitespace in Rust and are already masked above. Canonicalize
   // visibility token spacing before counting so `pub(super /* comment */)`
@@ -418,18 +420,48 @@ function validate(snapshot, scope = reviewedCompatibilityScope()) {
     occurrences(visibilityCanonical, "pub ") !== 0
   ) {
     errors.push(
-      "private native send policy module must expose only its three audited safe boundaries",
+      "private native send policy module must expose only its five audited safe boundaries",
     );
   }
   const lockHelper =
     "fn proposal_lock_request() -> Option<LockRequest> { None }";
   const versionHelper =
     "fn proposed_transaction_version() -> Option<TxVersion> { None }";
+  const confirmationsPolicyBody = rustFunctionBody(
+    send,
+    "confirmations_policy",
+  );
+  const walletSummaryPolicyBody = rustFunctionBody(
+    send,
+    "wallet_summary_with_policy",
+  );
+  const minimumFeeBody = rustFunctionBody(send, "standard_minimum_fee");
   if (
     occurrences(send, "fn proposal_lock_request()") !== 1 ||
     occurrences(send, lockHelper) !== 1
   ) {
     errors.push("proposal lock decision must remain one exact audited None helper");
+  }
+  if (confirmationsPolicyBody !== "ConfirmationsPolicy::default()") {
+    errors.push(
+      "confirmation depth must remain one shared native-send policy decision",
+    );
+  }
+  if (
+    walletSummaryPolicyBody !==
+    "db.get_wallet_summary(confirmations_policy())"
+  ) {
+    errors.push(
+      "send-all wallet summary must remain routed through the shared confirmation policy",
+    );
+  }
+  if (
+    minimumFeeBody !==
+    "let rule = Zip317FeeRule::standard(); (rule.marginal_fee() * rule.grace_actions()).map(u64::from)"
+  ) {
+    errors.push(
+      "send-all minimum fee must remain derived from the standard ZIP-317 fee rule",
+    );
   }
   if (
     occurrences(send, "fn proposed_transaction_version()") !== 1 ||
@@ -460,7 +492,7 @@ function validate(snapshot, scope = reviewedCompatibilityScope()) {
     "&input_selector,",
     "&change_strategy,",
     "request,",
-    "ConfirmationsPolicy::default(),",
+    "confirmations_policy(),",
     "spend_policy,",
     "proposal_lock_request(),",
     "proposed_transaction_version(),",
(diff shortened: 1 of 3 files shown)


skyl:
Reviewed adversarially on `4c08f99e`. **All four of #616's items are genuinely closed. Merge when `gate` and `rs / gate` are green.**

I re-derived each item on this branch rather than trusting #616's line numbers, since #627 had already moved that code once.

### Item 3 — the second confirmation policy is routed, and policed

`ConfirmationsPolicy` no longer appears in `send.rs` **at all**; it lives behind a private `confirmations_policy()` at `native.rs:34-35`, so the send-all spendable-balance read now consumes the same authority as proposal construction. Routing rather than pinning, which is what #547 rejected pinning for.

I expected to find the guard missing here. My #617 review noted `ConfirmationsPolicy` was *deliberately excluded* from `forbiddenOutsidePolicyModule` because `send.rs:15` imported it — and this PR removes that import, so I assumed the exclusion had been left behind. It is still absent from that list, so I tested the regression directly:

```rust
// send.rs:1575, reintroducing exactly what #616 item 3 described
let _regression = zcash_client_backend::data_api::wallet::ConfirmationsPolicy::default();
```

```
EXIT=1
drift: send-all spendable balance must consume the sole native confirmation policy
```

Policed by a dedicated check that names the property, which is better than a generic symbol ban would have been. I was wrong to expect a hole; recording it because the reasoning is useful to whoever reads that list next.

### Item 4 — the fee is derived on both sides

`send.rs:1575-1580` uses `self::native::standard_minimum_fee()`, and the test at `:2539` uses the **same function**. The independent literal is gone. That matters because #616's point was not thin coverage — it was that production and test each held their own copy of `10000`, so a ZIP-317 change would break production while the test stayed green. Both now move together, and "fee-rule derivation replaced with an independent literal" is in the mutation list as a red case.

(`FIXED_SEND_VALUE: u64 = 10_000` at `:2490` is a send *amount*, not a fee — correctly left alone.)

### Items 1 and 2 — asserted on the artifact, not the argument

This is the part I most wanted to see done properly, because #616's complaint was that nothing observed the created transaction:

```rust
assert_eq!(
    tx.expiry_height(),
    recovery_height + DEFAULT_TX_EXPIRY_DELTA,
    "the created transaction must retain the builder-derived expiry height",
);
```

Derived from librustzcash's own constant rather than a literal, so a library change moves both sides.

```rust
let decrypted = decrypt_transaction(&network, None, Some(recovery_height), &tx, &viewing_keys);
OrchardPoolTester::with_decrypted_pool_memos(&decrypted, |memo| { … });
assert_eq!(
    wallet_output_memos,
    vec![Memo::Empty, Memo::Empty],
    "the recipient fixture is memo-free, so the other decrypted output proves change is empty",
);
```

It **decrypts the real transaction** and reads the memos off the outputs. My #616 note was that the only `memo` reference was a fixture's `Payment::without_memo`, which constrains the recipient's memo and says nothing about change. This observes the on-chain artifact, and the message explains why two empties prove the change memo specifically rather than leaving the reader to work it out.

### Baselines

```
check-librustzcash-compat.mjs             → EXIT=0, 11 reviewed packages, 3 shipping locks
check-librustzcash-compat.mjs --self-test → EXIT=0, 114 mutants killed   (108 at #627, 72 at #617)
```

### What I did not observe

`gate` has not completed on this head. I did not run the plugin's Rust suite — it needs the librustzcash submodule and a real `LocalTxProver` — so the 145 unit tests, the shipping-feature integration test, and the seven mutation kills in the body are the PR's account, not mine. What I verified independently is the routing, the fee derivation, the two artifact assertions, and that the item-3 regression is caught.

PR #642, opened 2026-08-24
fix(relay): share anti-replay retention guard

### LINKED ISSUE #639: f2z-relay reimplements retention_is_sound instead of calling it — the same anti-replay relationship in two places
`f2z-relay`'s config validator and `f2z-relay-proto`'s `SeenSet::retention_is_sound` express the *same* anti-replay relationship in two places, and nothing ties them together.

```rust
// rs/crates/f2z-relay-proto/src/replay.rs:196
pub const fn retention_is_sound(&self, clock_skew_ms: u64) -> bool {
    match clock_skew_ms.checked_mul(2) {
        Some(required) => self.retention_ms >= required,
        None => false,
    }
}
```

```rust
// rs/crates/f2z-relay/src/config.rs:886
if u64::from(self.limits.antireplay_window_ms)
    < u64::from(self.limits.clock_skew_ms).saturating_mul(2)
{
    return Err(ConfigError::Invalid("limits.antireplay_window_ms", …));
}
```

`grep -rn 'retention_is_sound' rs/crates/` returns only its own definition, its own tests, and two doc references. **The daemon never calls it.**

### This is not a bug today

The two are equivalent right now, and the daemon's version is well-argued — the comment above it is one of the better ones in the tree:

> #586's finding, enforced rather than published. … The specification permits the hole; this relay refuses to open it, which is why the corresponding conformance vector is one it cannot satisfy by construction rather than one it fails.

That posture is right, and being stricter than the spec while saying so is exactly what I want to see. The concern is only that the *relationship* now has two homes.

### Why it is worth closing anyway

This repo already articulated the principle, in `f2z-relay-store`'s `QueueState::restore`:

> §8.2's anti-pre-ack rule implemented twice is §8.2's rule implemented once and violated once.

The same reasoning applies. Concretely, the boundary here has already moved once this week: #624 showed that `>= 2 × skew` is only sound because the seen-set interval is **closed**, and #631 made it closed (`retain(|_, e| *e >= now_ms)`). Had that gone the other way — making `retention_is_sound` strict instead — `replay.rs` would have changed and `config.rs:886` would have silently kept admitting the configuration that #624 demonstrated accepts a replay at exactly `t + skew`.

That is not hypothetical drift; it is the exact edit this repo considered and rejected three hours ago.

### Fix

Have the daemon ask the crate:

```rust
if !SeenSet::new(u64::from(self.limits.antireplay_window_ms), 1)
    .retention_is_sound(u64::from(self.limits.clock_skew_ms))
{
    return Err(ConfigError::Invalid("limits.antireplay_window_ms", …));
}
```

Keep the existing error message and the `#586` comment — they are the valuable part, and they explain *why* in a place an operator reading a config error will find it. Only the arithmetic should move, so there is one definition of "sound retention" and the daemon inherits any future correction to it automatically.

A test asserting the daemon rejects `antireplay_window_ms == 2 × clock_skew_ms - 1` and accepts `== 2 × clock_skew_ms` would pin the boundary from the daemon's side regardless.

Found reviewing #628 on the merged tree. Related: #624, #631, #586.

skyl:
Closes #639

The daemon now delegates the anti-replay retention/skew relationship to `SeenSet::retention_is_sound` instead of carrying a second arithmetic implementation. Its config test pins both sides of the inclusive `2 × clock_skew` boundary.

Mutation evidence (compiled and red before opening):
- bypassing/reversing the daemon guard accepts `2 × skew − 1`; the daemon test fails
- changing the shared boundary from `>=` to `>` rejects exact `2 × skew`; both daemon and protocol tests fail

Green on the exact head: relay + relay-proto all-target tests, workspace all-target tests, Clippy `-D warnings`, rustfmt, and independent adversarial review.

### DIFF
--- a/rs/crates/f2z-relay/src/config.rs
+++ b/rs/crates/f2z-relay/src/config.rs
@@ -39,6 +39,7 @@ use std::fmt::Write as _;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 
+use f2z_relay_proto::SeenSet;
 use serde::Deserialize;
 
 use crate::log::Level;
@@ -883,8 +884,8 @@ impl Config {
         // specification permits the hole; this relay refuses to open it, which
         // is why the corresponding conformance vector is one it cannot satisfy
         // by construction rather than one it fails.
-        if u64::from(self.limits.antireplay_window_ms)
-            < u64::from(self.limits.clock_skew_ms).saturating_mul(2)
+        if !SeenSet::new(u64::from(self.limits.antireplay_window_ms), 1)
+            .retention_is_sound(u64::from(self.limits.clock_skew_ms))
         {
             return Err(ConfigError::Invalid(
                 "limits.antireplay_window_ms",
@@ -1339,15 +1340,23 @@ mod tests {
     }
 
     #[test]
-    fn a_short_antireplay_window_is_refused_rather_than_published() {
+    fn antireplay_window_pins_the_shared_soundness_boundary() {
         // Issue #586: the specification as written *permits* this document, and
         // a client that checks refuses it. This relay declines to be that
         // relay, which is why the corresponding conformance vector is
         // unsatisfiable here by construction.
         let mut config = Config::default();
-        config.limits.antireplay_window_ms = config.limits.clock_skew_ms;
+        let sound_boundary = config
+            .limits
+            .clock_skew_ms
+            .checked_mul(2)
+            .expect("the u32 configuration boundary fits");
+        config.limits.antireplay_window_ms = sound_boundary - 1;
         let error = config.check().unwrap_err();
         assert!(format!("{error}").contains("antireplay_window_ms"));
+
+        config.limits.antireplay_window_ms = sound_boundary;
+        assert!(config.check().is_ok());
     }
 
     #[test]


skyl:
Reviewed on `f0282f16`. **This closes #639 exactly, and the propagation it exists to create is real. Merge when `gate` and `rs / gate` are green.**

### The delegation is the shape I asked for, including what I asked to keep

```rust
if !SeenSet::new(u64::from(self.limits.antireplay_window_ms), 1)
    .retention_is_sound(u64::from(self.limits.clock_skew_ms))
{
    return Err(ConfigError::Invalid(
        "limits.antireplay_window_ms",
        "must be at least 2 x limits.clock_skew_ms, or the seen-set has no \
         coverage for frames in the gap (WIRE.md §5.5, issue #586)".to_owned(),
    ));
}
```

Only the arithmetic moved. The error message and the `#586` comment stayed, which is the part I specifically said was valuable — they explain *why* in the place an operator reading a config error will actually find it.

### Propagation verified, which is the whole point

#639's argument was that the boundary had already moved once this week (#624 → #631) and that a second copy would not have followed. So I mutated the **shared** definition rather than the daemon:

```
rs/crates/f2z-relay-proto/src/replay.rs:198
-  Some(required) => self.retention_ms >= required,
+  Some(required) => self.retention_ms >  required,
```

```
cargo test -p f2z-relay --lib   → EXIT=101
  config::tests::antireplay_window_pins_the_shared_soundness_boundary ... FAILED
  config::tests::the_defaults_pass_their_own_check ... FAILED
```

A change in `f2z-relay-proto` now breaks `f2z-relay`'s tests. Before this PR it would not have — that is the coupling #639 was about, demonstrated rather than asserted.

The second failure is a bonus signal worth naming: `the_defaults_pass_their_own_check` fails because the shipped default (`240_000` = 2 × `120_000`) sits *exactly* on the boundary. So the defaults are not merely inside the safe region, they are the tightest configuration the rule allows — and the test suite knows it. That is precisely the case #624 showed is only sound because #631 made the seen-set interval closed.

### The new test pins both sides

```rust
config.limits.antireplay_window_ms = sound_boundary - 1;
let error = config.check().unwrap_err();
assert!(format!("{error}").contains("antireplay_window_ms"));

config.limits.antireplay_window_ms = sound_boundary;
assert!(config.check().is_ok());
```

Rejects `2 × skew − 1`, accepts exactly `2 × skew`, with `sound_boundary` derived from the config's own `clock_skew_ms` rather than written as a literal. A one-sided test would have been satisfied by a guard that rejects everything.

Baseline on the clean branch tree: **96 tests pass**.

### What I did not observe

`gate` has not completed on this head. I ran `-p f2z-relay --lib`; I did not re-run the relay-proto suite, the workspace all-target build, or clippy — those are the PR's account.

PR #643, opened 2026-08-24
test(relay): pin conformance populations

### LINKED ISSUE #641: conformance.rs: a vector silently changing its Needs is NOT visible, despite the comment saying it is
`rs/crates/f2z-relay/tests/conformance.rs:224` claims a property the file does not have:

```rust
// Stated as a number so a vector silently changing its `Needs` is visible.
assert_eq!(real.skipped(), real.outcomes.len() - runnable);
```

Nothing is stated as a number, and a vector silently changing its `Needs` is **not** visible.

## Demonstrated on current `main`

`rs/crates/f2z-relay-testkit/src/vectors.rs:424`, one word changed:

```diff
-    Nothing, "§2.5", hello_must_be_the_first_frame;
+    Faults,  "§2.5", hello_must_be_the_first_frame;
```

That removes a vector from the set run against the real relay and moves it to the skipped set — precisely the silent coverage loss the comment says it catches.

```
$ cargo test -p f2z-relay --test conformance
test result: ok. 2 passed; 0 failed
EXIT=0
```

Green.

## Why both assertions compensate

`runnable` is derived from the same declarations the mutation edits (`:211-214`):

```rust
let runnable = vectors::suite().into_iter()
    .filter(|vector| matches!(vector.needs, Needs::Nothing)).count();
```

So when a vector moves `Nothing → Faults`:

| | before | after |
|---|---|---|
| `runnable` | N | N−1 |
| `real.passed()` | N | N−1 |
| `real.skipped()` | S | S+1 |
| `outcomes.len() − runnable` | S | S+1 |

`assert_eq!(real.passed(), runnable)` holds. `assert_eq!(real.skipped(), real.outcomes.len() - runnable)` holds. Both sides are functions of the same edited value, so the equation is preserved by construction and can only fail if the *relay* misbehaves — never if the *suite* shrinks.

## This matters more here than it would elsewhere

The rest of this file is unusually good and gets the surrounding doctrine exactly right — there is a section literally headed **"A `Skipped` is never a pass"** (`:23`), and `:175-198` collects `wrongly_skipped` and asserts it is empty, so a fault-driven vector that *fails* instead of skipping is caught. The gap is narrow and specific: it guards how a vector reports, not whether it is still asked to run.

Conformance suites are exactly where this kind of erosion is expensive, because a shrinking suite looks identical to a passing one. It is the same shape as #597/#609/#612/#614, where the fix each time was to enumerate the population and fail if the enumeration shrinks.

## Fix

Pin the runnable count as an actual literal, so the mutation above goes red naming what stopped being covered:

```rust
assert_eq!(
    runnable, 37,   // <- the real number
    "the handle-free vector count changed; a vector's `Needs` moved, \
     which silently removes it from the real-relay run"
);
```

`vectors::suite().len()` is worth pinning too, so a vector being deleted outright is equally visible. Either number changing should be a deliberate edit with a reviewer, which is the whole point.

Found reviewing #628 on the merged tree. Related: #597, #609, #612, #614.

skyl:
Closes #641

Pin the reviewed conformance population at 52 total vectors and 39 vectors runnable against the real relay. The real-relay test now fails if a vector is deleted or silently changes its `Needs` classification.

The counts were independently derived from the actual suite: 39 `Nothing`, 12 `Faults`, and one `Clock`. The issue body’s suggested 37 was stale and was not copied.

Compiled mutation evidence (red before opening):
- change `hello_must_be_the_first_frame` from `Nothing` to `Faults`: runnable 38 vs reviewed 39
- delete a real vector: total 51 vs reviewed 52
- alter either reviewed literal: actual population disagrees

Green exact-head evidence: genuine real relay reports 52 outcomes / 39 passed / 13 skipped / zero failed; full workspace all-target tests, Clippy `-D warnings`, rustfmt, and independent adversarial review.

### DIFF
--- a/rs/crates/f2z-relay/tests/conformance.rs
+++ b/rs/crates/f2z-relay/tests/conformance.rs
@@ -52,6 +52,12 @@ use f2z_relay_testkit::config::{RelayConfig, TESTKIT_POW_DIFFICULTY_BITS};
 use f2z_relay_testkit::fake::{FakeRelay, WebSocketEndpoint};
 use f2z_relay_testkit::vectors::{self, Needs, Report, Status};
 
+// Reviewed literals, not values derived from `vectors::suite()`. Changing a
+// vector's `Needs` or deleting a vector must stop this test and require an
+// explicit coverage review instead of shrinking the real-relay run invisibly.
+const REVIEWED_SUITE_VECTOR_COUNT: usize = 52;
+const REVIEWED_REAL_RELAY_VECTOR_COUNT: usize = 39;
+
 /// A configuration that matches the FakeRelay's published policy wherever the
 /// policy is what a vector observes.
 ///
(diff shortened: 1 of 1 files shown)


skyl:
Reviewed on `cd46dee0`. **This closes #641 exactly. Merge when `gate` and `rs / gate` are green.**

I ran both halves of #641's reproduction rather than reading the diff, since the whole issue was that the file *looked* like it had this guard.

### My reproduction, verbatim, now goes red

```diff
 rs/crates/f2z-relay-testkit/src/vectors.rs:424
-    Nothing, "§2.5", hello_must_be_the_first_frame;
+    Faults,  "§2.5", hello_must_be_the_first_frame;
```

```
EXIT=101
assertion `left == right` failed: the reviewed handle-free vector population changed;
changing a vector's `Needs` silently changes what the real relay is asked to prove
  left: 38   right: 39
```

That is the exact edit that left conformance **green** on `main` when I filed #641. Both halves proven: green without the pin, red with it.

### The second population is pinned too, and it needed to be

Deleting a vector outright:

```
assertion `left == right` failed: the reviewed conformance-suite population changed;
adding or removing a vector requires an explicit coverage review
  left: 51   right: 52
```

Two independent anchors — `REVIEWED_SUITE_VECTOR_COUNT = 52` and `REVIEWED_REAL_RELAY_VECTOR_COUNT = 39` — catching two different erosions. A `Needs` reclassification moves the second without moving the first; a deletion moves both. Pinning only the total would have missed the case that prompted the issue.

Neither is circular: both are literals compared against values derived from `vectors::suite()`, so the suite cannot move the goalposts with it.

### It did not copy my number, which is the right instinct

The body notes: *"The issue body's suggested 37 was stale and was not copied."*

`37` was an illustrative placeholder I wrote without counting. Copying it would have produced a pin that was wrong on arrival and would have had to be "fixed" to match reality — which is expectations generated from the implementation, the failure mode this whole class of anchor exists to prevent. Deriving 39 `Nothing` + 12 `Faults` + 1 `Clock` = 52 independently is the correct response to a reviewer's hand-waved constant, and I would rather be corrected than copied.

### Baseline

Clean branch tree: `cargo test -p f2z-relay --test conformance` → **2 passed, 0 failed**.

### What I did not observe

`gate` has not completed on this head. I ran the conformance test only; the workspace all-target build, clippy and the full relay suite are the PR's account.

PR #645, opened 2026-08-24
ci(zuuli): narrow wallet change selection

### LINKED ISSUE #637: Two 20k-line PRs merged with no gate verdict on the merged tree — rs / gate is not a substitute for the wallet gate
Two large PRs merged within 22 minutes with only `rs / gate` green and **no `gate` verdict on the merged tree**. One instance is a slip; two is a pattern, so recording it durably rather than leaving it in PR comments that get swept.

## The evidence

**#628 — `feat(rs): f2z-relay — the relay daemon`, +20026, merged `ab656ca` at 15:34:09Z**

```
15:28:31  zuuli run 32745073784 created on final head 841d8ebd
15:28:31  rs    run 32745073964 created on 841d8ebd  → completed/success
15:34:09  merged
          zuuli run 32745073784 still `queued` — never started
```

**#629 — `feat(kt): f2z-kt log server and f2z-witness cosigning daemon`, +18076, merged `1ad9fb4` at 15:56:05Z**

```
15:49:18  zuuli run 32747184479 created on final head edc9bb20
15:49:19  rs    run 32747184543 created on edc9bb20  → completed/success
15:56:05  merged
          zuuli run 32747184479 still `queued` — never started
```

In both cases the last green zuuli run is on an **earlier head** — `fc849ea2` for #628, `f3d894df` for #629 — and in both cases the change between that head and the merged one is exactly the rebase that resolved a crate collision (#628 dropping its own `f2z-relay-store` after #632 landed; #629 dropping its own `f2z-authority` after #626 landed).

Those rebases were the **right** resolution and I said so at the time. But they are also precisely the diff no CI has ever judged. Green on a superseded head is not evidence about the current one; that is the entire reason `strict: true` is set.

## Why `rs / gate` is not a substitute

`rs / gate` covers the `rs/` workspace. The zuuli `gate` awaits `changes`, `frontend`, `rust_fmt`, `rust_deny`, `rust_clippy`, `rust_native_clippy`, `rust_native_tests`, `rust_plugin`, `rust_android_32`, `rust_app`, `zuuallet_schema`, `rust_crypto_targets` — the whole wallet side, including the macOS and Windows native matrices. Nothing in `rs / gate` speaks to any of it.

Both of these PRs touch `rs/` almost exclusively, so the practical risk here is low. That is luck about *these* changes, not a property of the process — the same sequence on a wallet-touching PR lands unverified money-path code.

## The doctrine this crosses

`enforce_admins=false` exists so `--admin` can bypass the **human** review requirement, which is unsatisfiable here anyway since every PR is authored by the same account that would approve it. The repo's own words: merge "after the CI `gate` is confirmed green — admin bypasses the *human* check, never the *QA* check." These two merges bypassed the QA check.

## What makes this worse right now, and the likely cause

The zuuli `gate` on `ab656ca` was created at **15:34:12** and was **still `queued` 22 minutes later**, while main's newer run on `1ad9fb4` had already started. Three zuuli runs are queued repo-wide.

That backlog is the incentive: waiting for `gate` currently costs 25-60 minutes, and #625's `wallet/*` broadening means even a two-comment-line PR (#635) pulls the full macOS/Windows matrix. When the honest path costs an hour, skipping it stops feeling like a shortcut. That is worth fixing on its own terms — a narrower positive selector (`wallet/**/Cargo.toml`, `wallet/**/*.rs`, plus the explicit file list) would cut most of it without returning to the hand-maintained list #608 removed.

## Ask

1. Leave a PR in draft until its **final** head exists, then let `gate` go green on that head, then merge. Not the previous head.
2. If a merge is genuinely urgent, say so explicitly and name the specific job that passed on the exact blob — that argument is defensible when stated, and I have used it four times tonight. It is not defensible silently.
3. Consider narrowing the `wallet/*` selector so the honest path is affordable.

Main's runs on `ab656ca` and `1ad9fb4` are the only verdicts these two trees will ever get. I am watching both; a `gate` failure on either is a revert of a 20k-line change with no fallback evidence.

Related: #628, #629, #625.

skyl:
Closes #637

Remove the over-broad `wallet/*` change-detector arm that made unrelated wallet documentation queue the full native matrix, while retaining behavior-tested selection for arbitrary-depth Rust sources, root/nested Cargo manifests and locks/config, and every explicit security/release-sensitive path.

The checker executes the workflow detector in temporary Git repositories rather than reading expected values from the workflow. Positive Rust/Cargo cases select; wallet Markdown and unrelated `docs/`/`rs/` changes do not.

Compiled/executed mutation evidence (red before opening):
- delete each generalized Rust/root-manifest/nested-manifest arm: its real positive case goes false
- broaden each arm back to `wallet/*`: real wallet Markdown goes true
- restore the stale Android selector digest or use an all-zero digest: the unconditional Android policy fails

Green exact-head evidence: GitHub Actions policy self/live, Android gate self/live, workflow-gates self/live, actionlint, independent adversarial review. This directly makes exact-final-head QA affordable; it does not relax the requirement to wait for both `gate` and `rs / gate`.


skyl:
Reviewed on `f39237c2`. **This closes #637 item 2 without reopening #608. Merge when `gate` and `rs / gate` are green.**

My whole worry with narrowing was that it would quietly restore the gap #608 existed to close — a new crate under `wallet/` becoming invisible to the gate. It does not, and the reason is checked rather than argued.

### The over-broad arm is gone, and only that arm

```diff
-  …|wallet/.clippy.toml|wallet/*|wallet/*.rs|wallet/*/Cargo.toml|…
+  …|wallet/.clippy.toml|wallet/*.rs|wallet/*/Cargo.toml|…
```

`wallet/*` removed; `wallet/*.rs`, `wallet/*/Cargo.toml`, `wallet/*/Cargo.lock`, `wallet/*/.cargo/*`, the clippy configs, `wallet/zuuli/*`, `wallet/plugins/*` and every explicit security-sensitive path all stay. Since `*` crosses `/` in a `case` pattern, `wallet/*.rs` and `wallet/*/Cargo.toml` still reach arbitrary depth — the same glob property that made `wallet/*` too broad, now used deliberately rather than incidentally.

### The #608 property is behaviour-tested, and I confirmed the test is load-bearing

Deleting the nested-manifest arm:

```
EXIT=1
.github/workflows/zuuli.yml:95: wallet/ owner selector must actively select
  "wallet/nested/future/Cargo.toml" as its effective last value
.github/workflows/zuuli.yml:95: native clippy input must select at least one
  native lint path: wallet/future-crate/Cargo.toml
```

Two things worth naming. The checker **executes the detector** in temporary Git repositories rather than comparing the selector against an expected string, so it cannot be satisfied by a rewrite that merely still spells the right pattern. And "as its **effective last value**" means it evaluates the whole `case` chain, so an arm cannot be shadowed by an earlier one and left decorative.

The fixtures are hypothetical future paths — `wallet/nested/future/Cargo.toml`, `wallet/future-crate/Cargo.toml` — which is the right choice. A fixture drawn from today's tree would keep passing after someone deleted the general arm and left the specific `wallet/zuuli/*` and `wallet/plugins/*` ones.

### Baseline

`check-github-actions-pins.mjs --self-test` → **EXIT=0**, 74 source-policy / 10 gate-verdict / 134 current-workflow / **83** Rust-root ownership cases (up from 78).

### On the tradeoff itself

This is the half of #637 I said was worth doing on its own terms. The gate cost is what made skipping the wait tempting, and #638 had already shown the gated path takes about six minutes when a branch is settled — so cutting documentation and Zuuallet-only changes out of the macOS/Windows matrix removes the incentive rather than just the complaint. The body says it plainly: *"This directly makes exact-final-head QA affordable; it does not relax the requirement to wait for both `gate` and `rs / gate`."* Agreed, and worth having in writing.

### What I did not observe

`gate` has not completed on this head. I ran the pins self-test and the live checker plus one mutation; the Android gate self/live, workflow-gates self/live and actionlint runs are the PR's account.

skyl:
Exact-head independent replay: **APPROVE `82e421f1c4dca58621a46cdb4a70a24ac7678fcb`**.

- Stable patch ID remains `22db161679ff901a0e51ecf4da3ba9b859545971`; all three changed blobs are byte-identical to the previously approved head.
- The intervening #647/#648 changes are documentation-only and have no executable overlap.
- Pins live, Android self/live, workflow-gates self/live, and actionlint all pass on this exact tree.
- Delete/shadow the Rust selector: red across ordinary, whitespace, newline, backslash, dash, and nested paths.
- Restore broad `wallet/*`: README/docs negative controls go red.
- Restore a stale Android digest: rejected by the reviewed-selector-digest guard.

Merge only after the exact-head `gate` and `rs / gate` are both green.

Issue #554, opened 2026-08-23
E2EE Phase 1: ADR 0013 (adopt akd) + correct §9.2's consistency-proof claim + KT.md


skyl:
Part of #305. Unblocked by #544, which chose `akd`. **Docs only** — the code follows in a separate PR.

## Scope

**1. ADR 0013 — key transparency log construction.**

Record the decision from #544: adopt [`akd`](https://github.com/facebook/akd) (Meta's Auditable Key Directory, the SEEMless/Parakeet-lineage append-only zero-knowledge set backing WhatsApp Key Transparency) rather than hand-rolling the Merkle machinery. `ARCHITECTURE.md` §9.2 stands unamended — this is the construction it already specified.

Must record:

- **The hard floor `akd >= 0.13.0`**, and *why it is not enforceable by advisory tooling*. facebook/akd#495 (merged 2026-08-12, released in 0.13.0 on 2026-08-13) fixed an **auditor append-only bypass**: a malicious server could rewrite a label's value and still produce a valid append-only proof. **There is no RustSec/OSV advisory** — an OSV query returns `{}` — so `cargo audit` and `cargo deny advisories` pass on a stale `0.12.0` pin silently. The floor must therefore be a `[[bans.deny]]` entry in `rs/deny.toml` with a comment naming #495, not a version requirement anyone trusts tooling to police.
- **Audit status:** NCC Group Cryptography Services, August 2023, 3 consultants / 20 person-days, public 40-page report, scope v0.9.0; 1 Medium / 8 Low / 6 Informational, all fixed and retested into v0.11.0. Same standard we applied to OpenMLS.
- **The reasoning that sent this to a spike**, because it is the general rule and will be needed again: a bug in a simple cleartext log leaks metadata, which is loud and discoverable by inspection; a soundness bug in an append-only ZKS lets the log substitute a key **undetectably**, which does not weaken the guarantee but inverts it. That asymmetry argues for a maintained implementation over a simpler one we own. #495 makes it concrete — that bug existed in an audited, production-deployed codebase, and we would not have found it ourselves.
- **What `akd` is not.** It is the ZKS only. No signed tree heads, no cosigning, no gossip, and only an in-memory `Database`. Everything in §9.3 plus a durable backend is ours.

**2. Correct `ARCHITECTURE.md` §9.2's consistency-proof bullet.** It currently reads:

> **Consistency proofs** between successive epoch roots (RFC 6962-style), so a client can verify the log is append-only across every root it has ever seen.

[...]

PR #647, opened 2026-08-24
docs(e2ee): specify `HandleAssertion`; close the first-entry authorization hole

### LINKED ISSUE #594: KT.md: nothing authorizes a handle's FIRST directory entry — anyone can claim any unregistered handle

skyl:
`KT.md` §4.4 authorized every entry by reference to state that a *previous* entry
established, and rule 3 explicitly permitted `entry_version == 1`. A conforming
log therefore accepted a first `DirectoryEntry` for any unregistered handle
**from whoever submitted one** — first-come-first-served handle claiming, in the
document whose entire purpose is that the log cannot lie about who you are
talking to. [#611](https://github.com/free2z/zuu/pull/611) named the hole and
deliberately left it open.

This closes it, and closes the narrower questions #611 listed as what a decision
would have to settle.

## Why this is not #551 happening again

#611 declined to answer on the ground that naming a vouching authority is a
product-and-trust decision a specification does not get to take, and that
inventing one would repeat [#551](https://github.com/free2z/zuu/issues/551).
That reasoning was right and the premise was wrong: **the decision had already
been taken.** The owner decision recorded on
[#305](https://github.com/free2z/zuu/issues/305) is a short-lived signed
handle-ownership assertion issued by the party that runs the user directory,
with the log verifying against a configured authority key and a self-hoster
plugging in their own. [#554](https://github.com/free2z/zuu/issues/554)'s brief
did not name that artifact, so `KT.md` inherited the gap rather than introducing
it. What §4.5 does is **write down a decision that was taken and never
specified**, and it says so where a reader would otherwise assume otherwise.

The encoding is not invented either. It is transcribed from **two independent
implementations that already interoperate byte-for-byte** — the merged Rust
crates (#626, #629) and the issuing endpoint operated by the party that runs the
user directory, which has a pinned cross-implementation test vector. Where the
two disagreed on a *default* rather than a rule, the spec fixes the rule and
says the default is not one.

## What landed

**`KT.md` §4.5 — `HandleAssertion`.** `HandleAssertionTBS`, `HandleAssertion`
and `AssertionBindingTBS` in the TLS presentation language; the exact
`tls_codec` encoding down to the byte widths (`196 + len(handle)` for the body,
`260 + len(handle)` signed); both domain-separated signing transcripts; `intent`
and where each value is legal; and **seventeen numbered validity rules**
including the ones that are load-bearing rather than obvious:

- **A9, the log-side cap.** `expires_ms - issued_ms` is capped by a value *the
  log* holds and publishes, never one the assertion carries. An issuer that
  could choose its own validity would, once compromised, mint one assertion good
  for a decade and every later rotation of the authority set would leave it
  working.
- **A4, `log_id` pinning.** Refused, not ignored: an assertion for one log is
  worthless on another, and a `log_id` treated as advisory makes every log a
  replay target for every other log's assertions.
- **A17, single-use `(authority_id, nonce)`.** Keyed on the pair rather than the
  digest, so it also catches an issuer that reused a nonce across two *different*
  assertions. Refuse, never evict; retention derived from the assertion's own
  expiry so it cannot be configured shorter than the window it protects.
- **A16, the identity self-signature, as normative text.** The submission must
  also be signed by the identity key the assertion is *about*. This is what makes
  a stolen assertion worthless — producing the second signature needs the very
  key the thief was trying to replace — and it was previously only a property of
  the code. A log that checks the authority signature without the binding has
  implemented a takeover primitive and the spec now says it MUST NOT be called
  conforming.

**§4.6 — the no-authority mode, and it must be reported.** A self-hoster with no
user directory MAY run a log; it MUST be an explicit configuration and never an
empty set (the two must fail differently, so a log cannot be downgraded by a
typo); and the log MUST serve a signed `AuthorityPolicyTBS` at
`/.well-known/free2z-kt/v1/authority` so a client can see *before it resolves
anything* that handles there are unvouched. A client MUST NOT render a vouched
and an unvouched handle identically — §8.3's witness-count rule applied to the
same failure. Two limits on that reporting are stated rather than glossed: the
policy is log-wide rather than per-handle, and it is signed by the party the
client is trying to hold at arm's length, so it makes a downgrade attributable
rather than impossible.

**§4.7 — what a compromised authority can and cannot do.** The honest paragraph
the brief asked for. It **cannot** silently swap a live user's key (`same_key`
and `key_change` refuse an assertion outright; rotation still requires the
outgoing ISK per ADR 0014), cannot perform a reset on its own (that needs the
*separately pinned offline* reset authority key, plus cooldown, plus
`reset_count`, plus the alarm — §4.5 only *adds* a requirement to that path), and
cannot use a stolen assertion. It **can** claim any unregistered handle: it mints
the assertion and produces the binding signature because it chose the key, and
nothing detects it, because the person whose name was taken is not a user and has
no self-audit to fire. That is the genuine residual, there is no prior key to
check a first claim against, and it does not go away by specifying it better.

**§4.4** gains rules 11 and 12, splits the `same_key` table row by version
(stating in the table that the version-1 row's middle column authorizes nothing
on its own), and carries the closing correction. **§9.2** corrects
`/kt/v1/submit`'s request body to `SubmissionEnvelope`. **§6.2**'s "closed" label
set gains three signing labels and separately names three hash-domain labels.
§1.1, §1.2, §2, §8.1, §8.5, §9.5 and §12 follow.

## Where I did not bless the implementation

**Two rules are stricter than what ships, stated at the point of use rather than
softened to match.**

1. **`account_epoch` MUST be a durable per-account counter and MUST NOT be
   derived from a clock** (§4.5.4). A15 requires it to be strictly greater than
   the retained value; a monotonic wall clock satisfies that unconditionally and
   forever. That does not weaken A15, it **deletes** it while leaving a field in
   place that looks like it is doing the work. One reference issuer derives it
   from Unix seconds where no counter is supplied, and is non-conforming by this
   rule. (It also exhausts the `uint32` in 2106.)
   **[#651](https://github.com/free2z/zuu/issues/651)**.
2. **A17's replay ledger MUST survive a restart** (§4.5.5). `f2z-authority`'s own
   `NonceSeen` contract says "a restart must not reopen the window"; `f2z-kt`
   holds the ledger in memory and rebuilds it empty on startup, replaying the
   journal through an accept-everything ledger. §4.2's version chain independently
   refuses the replays that matter today, but a rule whose enforcement depends on
   another rule is not the rule that was specified.
   **[#650](https://github.com/free2z/zuu/issues/650)**.

**One property is stated as not holding, which is the most important finding
here.** A `HandleAssertion` travels in the `SubmissionEnvelope`. It is **not part
of `DirectoryEntry`**, so it is not inside
`AkdValue = H("free2z/kt/v1/value", tls_codec(DirectoryEntry))`, is not covered
by the tree, and is never served. Every other row of §4.4's table is the opposite:
a `RotationProof` and a `ResetAuthorization` live inside `EntryAuthorization`,
are hashed into the committed value, and are re-verified by every client on every
lookup. §4.5 is **admission control the log performs on itself, and nobody can
check that it happened.** So §4.5 closes the first-entry hole **against third
parties and not against the log**; §8.1 step 6 still verifies nothing at
`entry_version == 1` and §8.5's table still says so.

I specified what the two implementations do rather than inventing a third wire
format, and filed the fix instead: move the assertion and the binding signature
into `EntryAuthorization` as a fourth case, with the binding committing to
`H(tls_codec(DirectoryEntryTBS))` rather than the full entry digest to avoid the
circularity that would otherwise create. It is `ARCHITECTURE.md` §13-S′ and
**[#649](https://github.com/free2z/zuu/issues/649)**.

**Two further gaps named rather than papered over:** what authorizes the
*issuance* of an assertion is unspecified (an issuer that treats `intent = reset`
like `intent = bind` makes session theft the first step of a handle takeover, and
every mitigation then lives in the log rather than the issuer), and the authority
key's own distribution and rotation is the same unanswered question §1.2 already
carries for the reset authority key.

**And one tension I found while reconciling.** §4.5's binding signature is by
`identity_pk` and is required on **every** submission, including `same_key`. That
contradicts §4.4's key-separation rationale — "a routine update (adding a device,
rotating a contact endpoint) then never touches the key peers pin and display as
a safety number." Rotating a contact endpoint now does. I specified the stricter
behaviour rather than relaxing it, because the alternative design has a branch
where a verifier holds a checked assertion and no binding, which is exactly the
primitive A16 exists to forbid — and recorded the narrowing as a dated correction
on the sentence it narrows, rather than letting the claim be restated without its
new scope.

## Dated-correction style

Every change to merged normative text leaves the original visible with a pointer,
per `ARCHITECTURE.md` §4.9 / §5.4:

- `KT.md` §4.4 — a **third** correction closing the second one, kept verbatim
  because it is the record of what the rules admitted between revisions; plus a
  correction on the key-separation note.
- `KT.md` §6.2 — the label set was called closed; it is now closed *as of this
  revision*, and no existing label was renamed.
- `KT.md` §9.2 — `/kt/v1/submit` took a bare `DirectoryEntry`.
- `THREAT-MODEL.md` §3.1 — a **fourth** correction. Its previous one ended "what
  closes the gap is a decision that has not been taken," which is now false; the
  scoping it defends is still right, and for a sharper reason. §5's two counters
  ("three times" → "four times", "three corrections" → "four corrections") and
  the summary row follow, so the summary does not restate the claim without its
  scope.
- `ARCHITECTURE.md` §13-S moves to §13.1 Closed; §13-S′ is the narrower
  remainder. `KT.md` §1.2 and §12 keep the closed bullets struck through rather
  than deleted, per the convention that a reader arriving from an old citation
  should find the answer and not a hole.

## Rebased onto #648, and reconciled by hand

[#648](https://github.com/free2z/zuu/pull/648) landed on `main` while this was
open and touched the same three documents. This branch is rebased onto it, and
the overlaps that a textual merge gets wrong were fixed deliberately rather than
left to auto-merge:

- **`KT.md` §1.2 carried the first-entry bullet twice** — #648's copy of the
  original alongside this branch's struck-through replacement. The stale one is
  removed; #648's separate "a proof that a handle is *unregistered*" bullet is
  kept untouched.
- **`KT.md` §8.1's step numbering moved.** This branch inserts a step 7 (fetch
  and verify §4.6's authority policy), so pinning became step 8. #648's new §8.1
  correction says a client "MUST NOT let [an unproved absence] weaken or discard
  a pin it already holds (step 7 above)" — repointed to step 8.
- **`KT.md` §5.3 and `ARCHITECTURE.md` §13-T both cite §13-S**, which this branch
  closes and moves to §13.1. Both now cite **§13-S′**, the narrower remainder,
  and say where §13-S went. #648's observation still holds and is sharper: a
  handle's first entry is both the one whose authorization a *client* cannot
  check and the one whose non-inclusion cannot be proved.
- **`THREAT-MODEL.md` §3.1 said "Three corrections below narrow this bullet"**
  and there are now four. Updated, along with §5's preamble counter and the
  summary row — the counters exist precisely so this cannot drift.
- `KT.md` §9.2, §9.5 and §12, and `README.md`'s status narrative, took additions
  from both sides with no duplication.

## Verification

- `node scripts/check-hash-domain-labels.mjs` — pass. **49 labels, 7 registered
  documents.** Every label §4.5/§4.6 names already existed in the crates, so the
  union is unchanged; note in particular that `free2z/kt/v1/handle-id`,
  `free2z/kt/v1/handle-assertion` and §3.3's `free2z/kt/v1/handle:` are mutually
  prefix-free, which §6.2 now says out loud.
- `node scripts/check-hash-domain-labels.mjs --self-test` — 17 cases pass.
- `scripts/check-public-repo-boundary.sh` and `--self-test` — pass. The issuer is
  referred to obliquely throughout; no private repository, path, hostname or
  credential appears.
- Every relative link and `#anchor` across `docs/e2ee/` resolves, checked
  mechanically, including the new `#45-handleassertion--what-authorizes-a-handles-first-entry`,
  `#47-what-a-compromised-authority-can-and-cannot-do` and `#131-closed` targets.
- Docs only. No crate, no workflow, no `WIRE.md` touched.
- Dates stamped `2026-08-24`, the local date of this commit.

Filed against the crates rather than blessed into the spec: #649 (move the
assertion into `EntryAuthorization` so a client can verify it), #650 (the nonce
ledger does not survive a restart), #651 (`account_epoch` must be a counter, not
a clock).

Closes #594. Refs #305, #551, #554, #584, #593, #611, #626, #629.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -1220,23 +1220,23 @@ deliberate.
   portable evidence rather than a local error dialog are all unspecified.
   `KT.md` defines the evidence structures but deliberately does not invent the
   protocol that exchanges them.
-- **S. What authorizes a handle's first directory entry.** Opened 2026-08-24 by
-  [`KT.md` §4.4](./KT.md#44-what-authorizes-an-entry), and **blocking**. §9.2
-  above specifies an append-only log of `(handle → identity key, …)` and does not
-  say who is entitled to a handle; [ADR 0014](./decisions/0014-directory-key-rotation.md)
-  settles authorization for a handle that already has an entry — same-key update,
-  key change, lost key — and takes no position on the first one. `KT.md` §4.4
-  inherited that silence, and as its rules are enumerated they accept a first
-  entry for an unregistered handle from **whoever submits one**. Closing it means
-  deciding who may vouch that a handle belongs to a submitter, whether a log may
-  run with no such authority (a self-hosted log must be able to, under
-  [ADR 0005](./decisions/0005-federation.md)) and whether clients must be told
-  when it does, what binds a vouching artifact to the key being installed, and
-  what a compromise of the vouching key costs. That is a product-and-trust
-  decision of the same weight as ADR 0014's reset authority — which is what
-  [#551](https://github.com/free2z/zuu/issues/551) is open about — so `KT.md`
-  states the option space and deliberately answers none of it.
-  [#594](https://github.com/free2z/zuu/issues/594).
+- **S′. Client verification of a handle's first directory entry.** Opened
+  2026-08-24 by [`KT.md` §4.7](./KT.md#47-what-a-compromised-authority-can-and-cannot-do),
+  as the narrower remainder of **S** below. §4.5 now says what authorizes a first
+  entry, and the log is the only party that checks it: a `HandleAssertion`
+  travels in the submission envelope, is not part of `DirectoryEntry`, is not
+  inside the value the tree commits to, and is not served with a lookup. Every
+  other authorization in §4.4 is the opposite — a `RotationProof` and a
+  `ResetAuthorization` sit inside `EntryAuthorization` and are re-verified by
+  every client on every lookup. So the first-entry hole is closed against a
+  stranger and **not against the log**, `KT.md` §8.1 step 6 still verifies
+  nothing at `entry_version == 1`, and §8.5's table still says so. Closing it
+  means a fourth `EntryAuthorization` case carrying the assertion and the
+  identity countersignature, with the binding committing to
+  `H(tls_codec(DirectoryEntryTBS))` rather than to the whole entry digest —
+  a wire change to a structure two implementations have already frozen, so it is
+  filed rather than made. [#594](https://github.com/free2z/zuu/issues/594).
+
 - **T. Proving that a handle is unregistered.** Opened 2026-08-24 by
   [`KT.md` §8.1](./KT.md#81-lookup) and
   [§9.5](./KT.md#95-error-codes), and unlike most of this list it is not a
@@ -1258,14 +1258,44 @@ deliberate.
   saying so. The knock-on is in
   [`KT.md` §5.3](./KT.md#53-the-submission-receipt): a submission receipt for a
   handle's **first** entry cannot be turned into portable evidence either, which
-  is the same handle-with-no-entry case as §13-S above.
+  is the same handle-with-no-entry case as **§13-S′** above — and as §13-S, which
+  closed on 2026-08-24 and moved to §13.1.
   [#634](https://github.com/free2z/zuu/issues/634).
+**S** — what authorizes a handle's first directory entry — was opened and closed
+on 2026-08-24 and has moved to §13.1.
 
 ### 13.1 Closed
 
 Kept here rather than deleted, so that a reader who arrives via an old citation
 finds the answer instead of a hole.
 
+- **S. What authorizes a handle's first directory entry — opened and closed
+  2026-08-24.** §9.2 above specifies an append-only log of
+  `(handle → identity key, …)` and does not say who is entitled to a handle;
+  [ADR 0014](./decisions/0014-directory-key-rotation.md) settles authorization
+  for a handle that already has an entry — same-key update, key change, lost key
+  — and takes no position on the first one.
+  [`KT.md` §4.4](./KT.md#44-what-authorizes-an-entry) inherited that silence, and
+  as its rules were first enumerated they accepted a first entry for an
+  unregistered handle from **whoever submitted one**.
+  [#594](https://github.com/free2z/zuu/issues/594) found it while implementing
+  against the document.
+  **It is answered by [`KT.md` §4.5](./KT.md#45-handleassertion--what-authorizes-a-handles-first-entry)**:
+  a short-lived, log-pinned, single-use `HandleAssertion` from a configured
+  authority, countersigned by the identity key it names so that a stolen
+  assertion is worthless, with the validity cap held by the **log** rather than
+  the issuer. §4.6 fixes the no-authority mode a self-hosted log needs under
+  [ADR 0005](./decisions/0005-federation.md) and **requires it to be reported**
+  in a signed policy document, so a client can see that handles there are
+  unvouched. §4.7 states what the trust root costs.
+  **It was not decided in the specification**, which is the objection
+  [#551](https://github.com/free2z/zuu/issues/551) is open about: the decision is
+  the owner decision recorded on
+  [#305](https://github.com/free2z/zuu/issues/305), which
+  [#554](https://github.com/free2z/zuu/issues/554)'s brief did not name, so the
+  documents inherited the gap rather than introducing it, and §4.5 transcribes
+  what two independent implementations already agreed on byte-for-byte. What it
+  does **not** close is above as **S′**.
 - **A. OpenMLS audit status — closed 2026-08-08.** An independent assessment by
   SRLabs, sponsored by the Sovereign Tech Agency, was published
   [2026-05-27](https://blog.openmls.tech/posts/2026-05-27-independent-audit/):
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -1120,6 +1120,8 @@ rather than to the full entry digest to avoid the circularity that would
 otherwise create. That is a wire change to a structure two implementations have
 already frozen, it is filed rather than done here, and until it lands this
 section describes a property that does not hold.
+[#649](https://github.com/free2z/zuu/issues/649),
+[§13-S′](./ARCHITECTURE.md#13-open-questions), §12.
 
 ---
 
@@ -2300,6 +2302,7 @@ Deliberately, and listed rather than invented.
   it means moving the assertion and its binding signature into
   `EntryAuthorization`, which is a wire change to a structure two
   implementations have frozen. Filed, not invented here.
+  [#649](https://github.com/free2z/zuu/issues/649),
   [#594](https://github.com/free2z/zuu/issues/594).
 - **What authorizes the *issuance* of a `HandleAssertion`.** §4.5 fixes
   everything a log does with an assertion and nothing about how a user gets one.
@@ -2308,7 +2311,10 @@ Deliberately, and listed rather than invented.
   identity key — is unspecified, and an issuer that treats them alike makes
   session theft the first step of a handle takeover (§4.7). This is an
   account-security decision rather than a wire encoding, which is why §4.5 does
-  not invent one, and it is a real gap rather than a tidy deferral.
+  not invent one, and it is a real gap rather than a tidy deferral. The two
+  conformance rules §4.5 states more strictly than a shipped implementation are
+  tracked as [#651](https://github.com/free2z/zuu/issues/651) (`account_epoch`)
+  and [#650](https://github.com/free2z/zuu/issues/650) (A17's ledger).
 - **The authority key's own distribution and rotation.** §4.6 publishes the
   authority set in a document signed by the log, which is not a trust root —
   exactly the caveat §9.1 states for `reset_authority_pk`, and the same
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -50,7 +50,8 @@ The `DirectoryEntry` structure and its authorization — including, since
 signature (§4.5), the no-authority mode and its mandatory reporting (§4.6), and
 an accounting of what trusting an authority costs (§4.7). An earlier revision
 said this document fixed authorization "from the second entry onward" and
-deferred the first to §1.2; §4.4's third correction records the closure and what
+deferred the first to §1.2; §4.4's **closing** correction — the last of the four
+that section carries — records the closure and what
 an implementer working from that revision must add. The `SignedTreeHead` format
 and its signing transcript; log-key rotation; epoch cadence and the maximum merge
 delay; the submission receipt; the witness poll-verify-cosign protocol, its
@@ -450,8 +451,9 @@ Verification rules the log MUST apply, in order, before an entry enters a batch:
    **The `(or 1)` case is not an authorization rule and MUST NOT be read as one**
    — it fixes the version number and nothing else. What authorizes a handle's
    *first* entry is **rule 11** and §4.5. (Until 2026-08-24 this document said
-   the question was open and blocking; see the second and third corrections at
-   the end of this section.)
+   the question was open and blocking; the correction that said so, and the
+   closing correction that supersedes it, are the last two at the end of this
+   section.)
 4. `prev_entry_hash` matches the published previous entry, or is all-zero when
    `entry_version` is 1 (§4.2).
 5. `authorization.kind == entry.kind`, and the table above holds.
@@ -663,7 +665,7 @@ this paragraph be restated without its scope.
 
 ### 4.5 `HandleAssertion` — what authorizes a handle's first entry
 
-> **Invented in the sense §4.4's third correction describes: the *decision* is
+> **Invented in the sense §4.4's closing correction describes: the *decision* is
 > the owner decision on [#305](https://github.com/free2z/zuu/issues/305), and
 > the *encoding* is written down here for the first time.** Every structure,
 > transcript and rule below is transcribed from two implementations that already
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -817,10 +817,13 @@ handle. It increments when the account changes hands or is recovered, which is
 the only thing that justifies a `reset`. The log retains the value from the last
 admitted assertion for the handle and requires the next one to be **strictly
 greater** (A15), so a `reset` assertion cannot be spent twice on one
-account-ownership event.
+account-ownership event. At a **first** entry there is no retained value and A15
+has nothing to compare against; the value in a `bind` assertion is simply the
+baseline the handle's first `reset` must exceed.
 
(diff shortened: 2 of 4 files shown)


skyl:
Reviewed with one question ahead of all others: **is this settling #594, or transcribing a decision that was already taken?** I have been treating #594 as the owner's to decide and declining to design it all session, so a PR titled "close the first-entry authorization hole" needed that answered before anything else.

**It is transcription. I verified the claim rather than accepting it.**

### The decision exists, and it says what this PR says it says

`#305` carries a lettered decision list — **D-a** through **D-g** — and **D-d** is verbatim:

> **D-d — The KT submission authority is a tiny signed-assertion endpoint.** Django authenticates the user and issues a short-lived Ed25519-signed handle-ownership assertion; the Rust KT server only verifies a signature against a configured authority public key and has **no** knowledge that Django exists. A self-hoster plugs in their own authority, or none.

That is the same shape this PR writes into §4.5: short-lived, signed by the party running the user directory, verified against a configured authority key, self-hosters supply their own or none. So the body's central claim — *"the decision had already been taken… `KT.md` inherited the gap rather than introducing it"* — holds.

**#611's reasoning was right and its premise was wrong**, exactly as the body says. Declining to invent a vouching authority was correct; what nobody checked was whether one had already been chosen. It had, on #305, and #554's brief did not carry it across. That is worth stating plainly because I repeated #611's framing myself, most recently on #629 — I was guarding a decision that was already made.

### The encoding is not invented either, and that is checkable

`HandleAssertionTBS` and `AssertionBindingTBS` exist in `rs/crates/f2z-authority/src/assertion.rs:95` and `:37`, merged via #626 and reviewed then. The spec is transcribing merged code, not proposing a format.

More importantly, the crate pins its encoding with `assertion_digest_matches_the_independent_nonuniform_canonical_vector` (`:560`) — an **independently derived** expectation rather than one generated from the implementation. I flagged that as the right shape when I reviewed #626, and it is what makes "transcribed from two implementations that interoperate byte-for-byte" a claim with something behind it rather than an assertion.

### What I like about how it handles the seam

> Where the two disagreed on a *default* rather than a rule, the spec fixes the rule and says the default is not one.

That is the correct resolution of a spec/implementation disagreement, and the inverse of the failure I have filed repeatedly this session — a value silently closed by fiat because one implementation happened to pick it.

### What the owner should still eyeball

I can confirm this matches **D-d**. I cannot confirm it matches what the issuing endpoint actually ships, because that lives outside this repository and I have deliberately kept private-repo detail out of public issues all session. The cross-implementation vector is asserted in the body; the half I verified is the Rust half.

Two things worth a decision-owner's glance rather than a reviewer's:
1. **`account_epoch`** (`assertion.rs:112-115`) — "increments whenever the account changes hands or is recovered, which is the only thing that justifies a reset." That is a policy statement about account recovery, adjacent to ADR 0014's reset authority. It follows from D-d but is not literally in it.
2. **Expiry semantics** — "short-lived" is D-d's word; the concrete window is the spec's. Worth confirming the number is the one you want, since #624 was exactly a case where an exact boundary mattered more than the adjective.

### Not reviewed

`gate` has not completed. I checked the §4.5 encoding against the merged crate and the decision provenance; I did not read the full +718 of `KT.md` line by line, nor the `ARCHITECTURE.md` / `THREAT-MODEL.md` / `README.md` deltas.

skyl:
Re-checked at `fe3a7c98` — my approval above was against `2ed2a421`, and green-on-a-previous-head is not evidence about the current one, so here is the delta I looked at. It is `+8/-5` in `KT.md`, all three changes are improvements, and none of them touches the D-d provenance argument I verified.

**1. It answers one of the two things I flagged for the owner.** I said `account_epoch`'s semantics follow from D-d without being literally in it. This pins down its first-entry behaviour:

> At a **first** entry there is no retained value and A15 has nothing to compare against; the value in a `bind` assertion is simply the baseline the handle's first `reset` must exceed.

That resolves the case I would have asked about. The remaining item for the owner is unchanged: "short-lived" is D-d's word, the concrete expiry window is the spec's.

**2. It corrects an over-claim in its own prose**, which is worth naming because it is the failure mode I have spent this session filing:

```diff
-> **The one place this document is stricter than a shipped implementation…**
+> **One of the two places this document is stricter than a shipped
+> implementation…** — the other is A17's ledger durability (§4.5.5)
```

A document asserting uniqueness it does not have is exactly the kind of claim that goes stale silently. Finding and fixing it before merge is the right instinct.

**3. It disambiguates an error mapping that was genuinely ambiguous:**

```diff
-| A1, A3, A5–A12, A14, A15, A17 | ERR_BAD_AUTHORIZATION (4) |
-| A17 ledger full | ERR_RATE_LIMITED (9) |
+| A1, A3, A5–A12, A14, A15, and A17 **as a replay** | ERR_BAD_AUTHORIZATION (4) |
+| A17 refused because the ledger is **full of unexpired entries** | ERR_RATE_LIMITED (9) |
```

"A17 ledger full" appearing in both rows left an implementer to guess which condition produced which code — the same shape as `DELETE_QUEUE`'s "both addresses" contradiction I raised on #588. Now each row names its condition.

Approval carries to `fe3a7c98`. Merge when `gate` and `rs / gate` are green **on this head**.

skyl:
Conflicting since #648 (`ed4827a`) landed — it added +243 to `KT.md`, and this adds §4.5 to the same document.

Flagging one thing for the rebase, because a clean textual merge here would not necessarily be a correct one. #648's method is deliberate: it leaves each erroneous claim **in place**, marked inline, with a dated correction adjacent — for example `WIRE.md:1665-1673`, where the original bullet stays and the correction says *"left in place as the record of what was specified."* The point is that a reader can see what implementations were built against.

So the resolution needs both halves to survive intact:

- #648's dated corrections must keep their **original claims and inline markers**, not be tidied into clean prose while resolving.
- This PR's §4.5 must land whole, since it is the transcription of D-d that closes the first-entry hole.

The two should compose without argument — they touch different sections — but "resolved cleanly" and "both intentions preserved" are different properties, and only the second one matters. This is the same shape as the #556/#560 hazard I caught earlier, where git merged 4 + 4 into 6 while the assertion demanded exactly 4.

My approval stands on content: I verified the D-d provenance on `2ed2a421` (comment 5399782871) and the `+8/-5` clarifications on `fe3a7c98` (comment 5399808716). Re-request when the rebase is up and I will check the merged `KT.md` shows both #648's corrections and this §4.5 rather than one having quietly displaced the other.

skyl:
Rebase verified at `5b09a406`. **Both intentions survived — approval carries.**

I said "resolved cleanly" and "both intentions preserved" are different properties and only the second matters, so I checked the second rather than trusting the first:

```
#648's dated corrections in KT.md
  on main   : 7
  on branch : 10        (main's 7 kept, plus 3 this PR adds)

comm -23 main_corrections branch_corrections
  (empty — none of main's were lost)

§4.5 HandleAssertion references on branch : 13
```

The count going *up* is not by itself evidence — a correction could have been swapped rather than dropped — so I diffed the headlines individually. Every one of main's seven is still present by name, `WIRE.md`-style marked originals intact, and §4.5 landed whole alongside them.

That is the outcome I wanted and it is worth saying why it was not automatic: #648's corrections are prose blocks sitting next to the claims they correct, and §4.5 is a large insertion into the same document. A resolution that "tidied" the corrections while making room would have merged without conflict and looked correct in review.

Approval carries to `5b09a406`. Merge when `gate` and `rs / gate` are green **on this head** — noting this is the third head I have reviewed for this PR (`2ed2a421` → `fe3a7c98` → `5b09a406`), so verify by run ID rather than by any earlier green.

PR #652, opened 2026-08-24
test(kt): bind rejected VRF seed to akd_core

### LINKED ISSUE #644: vrf.rs: forbid_hardcoded's copy of akd_core's test key is correct today, but its test feeds itself and cannot notice drift
`FileVrf::forbid_hardcoded` is a good guard against a real mistake, and it is **correct today**. But nothing keeps it correct, and its test cannot notice if it stops being.

## The shape

`rs/crates/f2z-kt/src/vrf.rs:26-31` holds a hand-copied literal:

```rust
/// `akd_core::ecvrf::HardCodedAkdVRF`'s private key, verbatim from the library
/// source. Present here for exactly one purpose: refusing it.
const HARDCODED_TEST_VRF_SEED: [u8; 32] = [
    0xc9, 0xaf, 0xa9, 0xd8, … 0x67, 0x21,
];
```

and the test feeds that same constant back in:

```rust
use super::{FileVrf, HARDCODED_TEST_VRF_SEED};

#[test]
fn the_libraries_hardcoded_test_key_is_refused_at_startup() {
    let error = FileVrf::from_seed(HARDCODED_TEST_VRF_SEED).unwrap_err();
    assert!(format!("{error}").contains("hardcoded test key"));
}
```

So the test proves `forbid_hardcoded` rejects `HARDCODED_TEST_VRF_SEED`. That is true by construction. What it does not prove — and cannot — is that `HARDCODED_TEST_VRF_SEED` is still `akd_core`'s key.

## It is correct right now

I checked rather than assuming. `akd_core-0.13.0/src/ecvrf/mod.rs:80`:

```rust
hex::decode("c9afa9d845ba75166b5c215767b1d6934e50c3db36e89b127b8a622b120f6721")
```

The local constant matches byte for byte. **This is not a live hole.** It is an invariant with no guard.

## Why it is worth closing here specifically

The stakes are stated in the module's own doc, and they are not small:

> Using it in a real log would make every label derivable by anyone who has read the crate, which destroys the zero-knowledge property the whole construction is for.

And the drift event is routine for this repo rather than exotic: `AGENTS.md` doctrine is to stay on upstream HEAD and contribute fixes rather than pin to old commits, `rs/Cargo.toml:74` already carries a note that "`akd_core` 0.13 requires a reviewer's attention", and #565 set an `akd >= 0.13.0` floor precisely because these versions move. If a future bump changes that test key, this guard stops matching, `the_libraries_hardcoded_test_key_is_refused_at_startup` stays green, and the only signal that the log is running a world-readable VRF key is gone.

This is the same shape as #616's fee literal: production and test each holding their own copy of a constant that lives somewhere else, so they drift together and the test cannot see it. That one was fixed by deriving both sides from `Zip317FeeRule::standard()`.

## Fix

`HardCodedAkdVRF` implements `VRFKeyStorage`, and `retrieve()` is public and returns the seed — so the copy can be pinned to the library at test time:

```rust
#[tokio::test]
async fn the_copied_seed_still_matches_akd_cores_own() {
    let library = akd_core::ecvrf::HardCodedAkdVRF
        .retrieve()
        .await
        .expect("akd_core's hardcoded VRF key is retrievable");
    assert_eq!(
        library.as_slice(),
        &HARDCODED_TEST_VRF_SEED,
        "akd_core changed its hardcoded test key; forbid_hardcoded is now refusing a key \
         nobody ships and permitting the one that makes every label public",
    );
}
```

Keep the existing constant and the existing refusal — the constant is what lets `forbid_hardcoded` stay synchronous and infallible. The new test just stops it silently going stale, and names the consequence when it does.

Found reviewing #629 on the merged tree. Related: #616, #565.

skyl:
Closes #644.

## What changed

Adds an async `f2z-kt` test that retrieves `akd_core::ecvrf::HardCodedAkdVRF` through the public `VRFKeyStorage` API and compares that independent library value with the copied production refusal seed. This binds the refusal guard to the real upstream registry without a source-shape or locally duplicated fixture.

## Verification

- `cargo +1.97.1 test --locked -p f2z-kt --all-targets` — 61 tests pass
- `cargo +1.97.1 test --locked -p f2z-kt --doc` — pass
- `cargo +1.97.1 fmt --all -- --check` — pass
- `cargo +1.97.1 clippy --locked -p f2z-kt --all-targets -- -D warnings` — pass

Mutation controls, replayed independently on exact head `0bc5d4f1`:

- Corrupt one production seed byte (`0xc9 -> 0xc8`) — RED: independent retrieval remains `201`, copied seed becomes `200`.
- Delete the refusal body — RED: hardcoded key unexpectedly loads.
- Invert the equality predicate — RED: forbidden seed is accepted and both ordinary-key tests reject valid seeds.

Independent exact-head review: APPROVE. Stable patch ID: `d75a96a631d5050c27acc9350ad3dc615c535ca1`.

### DIFF
--- a/rs/crates/f2z-kt/src/vrf.rs
+++ b/rs/crates/f2z-kt/src/vrf.rs
@@ -124,7 +124,23 @@ impl core::fmt::Debug for FileVrf {
 
 #[cfg(test)]
 mod tests {
+    use akd_core::ecvrf::VRFKeyStorage as _;
+
     use super::{FileVrf, HARDCODED_TEST_VRF_SEED};
+
+    #[tokio::test]
+    async fn the_copied_seed_still_matches_akd_cores_own() {
+        let library = akd_core::ecvrf::HardCodedAkdVRF
+            .retrieve()
+            .await
+            .expect("akd_core's hardcoded VRF key is retrievable");
+        assert_eq!(
+            library.as_slice(),
+            &HARDCODED_TEST_VRF_SEED,
+            "akd_core changed its hardcoded test key; forbid_hardcoded is now refusing a key \
+             nobody ships and permitting the one that makes every label public",
+        );
+    }
 
     #[test]
     fn the_libraries_hardcoded_test_key_is_refused_at_startup() {


skyl:
Reviewed on `0bc5d4f1`. **This closes #644 exactly. Merge when `gate` and `rs / gate` are green.**

#644's point was that the guard compared against a copy and its test fed that same copy back, so it could only prove a tautology. The fix has to make the *library* the authority — so that is what I tested, rather than that a test now exists.

Corrupting the local constant's first byte:

```
vrf.rs:29
-    0xc9, 0xaf, 0xa9, 0xd8, …
+    0xc8, 0xaf, 0xa9, 0xd8, …
```

```
EXIT=101
vrf::tests::the_copied_seed_still_matches_akd_cores_own panicked at vrf.rs:137
  left:  [201, 175, 169, 216, …]   ← retrieved from akd_core at runtime
  right: [200, 175, 169, 216, …]   ← the local copy
```

`left` comes from `HardCodedAkdVRF.retrieve()` through the public `VRFKeyStorage` API; `right` is the constant. They are now independent quantities, so the copy cannot drift without the test noticing. Baseline on the clean branch: **36 tests pass**.

The failure message is the one that matters when this eventually fires on an akd bump — it names the consequence rather than the mismatch: *"forbid_hardcoded is now refusing a key nobody ships and permitting the one that makes every label public."* Whoever hits that in six months does not have to reconstruct why a byte array matters.

### One note on copying, since it cuts the other way from #643

This PR uses the code from #644's body essentially verbatim, and that is correct — what I supplied there was a **mechanism** (compare against `retrieve()`), not a value. Contrast #643, where the author explicitly refused to copy my illustrative `37` and derived the real population instead. That is the right distinction and both calls were right: copy a reviewer's mechanism, never a reviewer's guessed constant.

### What I did not observe

`gate` has not completed on this head. I ran `-p f2z-kt --lib` plus the mutation; the `--all-targets` run, doctests, fmt and clippy in the body are the PR's account.

PR #653, opened 2026-08-24
fix(kt): restore nonce ledger across restart

### LINKED ISSUE #650: f2z-kt: the assertion nonce ledger does not survive a restart, contrary to `NonceSeen`'s own contract
Found while writing `KT.md` §4.5 (#647, closes #594).

## The contract

`f2z-authority::nonce::NonceSeen` states the requirement itself:

> A trait so a log with durable storage can keep the ledger in its database (**a
> restart must not reopen the window**: unlike a relay's seen-set, nothing here
> is invalidated by a lost TLS session) while a client verifying a served
> assertion uses `NonceLedger` in memory.

`KT.md` §4.5.5 now states it normatively: *"The ledger MUST survive a restart: a
log that rebuilds an empty ledger on startup has reopened the window for the
length of one validity period."*

## What `f2z-kt` does

- `State.nonces` is a plain in-memory `NonceLedger`, constructed fresh in
  `LogService::open` (`rs/crates/f2z-kt/src/log.rs`).
- Journal replay runs every stored envelope back through `admit_submission` with
  `ReplayLedger` — an accept-everything `NonceSeen` — which is correct in itself
  (otherwise the log would refuse its own history) but leaves the **live** ledger
  empty.
- So after a restart, `(authority_id, nonce)` pairs already spent are forgotten
  for good, and A17 is not enforced for them.

## How bad is it, honestly

Not very, today — which is why this is a correctness issue rather than a
vulnerability report. The replays that a forgotten nonce would enable are
independently refused by §4.2's chain rules:

- Re-presenting the same envelope fails §4.4 rule 3 — the handle now has a
  published entry, so `entry_version` must be 2.
- Pairing a stolen assertion with different entry bytes fails A16 — the binding
  commits to `entry_digest` and the thief does not hold the ISK.
- A replayed `intent = reset` assertion fails A15 on `account_epoch` — **but see
  #651**, which is exactly the rule that is currently weak.

The last bullet is the reason not to leave this. A17 and A15 are the two
independent guards on assertion replay, and both currently lean on something
else. A rule whose enforcement depends on another rule is not the rule that was
specified.

## Fix

Persist the ledger — either its own journal, or reconstructed at startup from
`submissions.log` by observing each admitted assertion's `(authority_id, nonce,
expires_ms)` into the *live* ledger while `ReplayLedger` handles the admission
path. The second is nearly free: the journal already holds every envelope, and
entries older than `expires_ms + clock_skew_ms` are dropped on the first
`expire()` sweep anyway.

Refs #594, #647.

skyl:
Closes #650.

## What changed

Reconstructs the live A17 `NonceLedger` while replaying canonical stored assertions after restart. The existing `ReplayLedger` still reruns full admission; only successfully admitted records restore their `(authority_id, nonce, expiry)` into the production ledger before published/tail state is installed. Both epoch and tail loops share the same `readmit` path.

The decisive restart test uses independently signed first-entry assertions for Alice and Bob with different handles, identities, entries, bindings, and signatures. Bob succeeds on a fresh control log, then is rejected only for reusing Alice's `(authority_id, nonce)` after Alice is journaled and the log is reopened.

## Verification

- `scripts/check-rust-fmt.sh --root rs` — pass
- `cargo +1.97.1 test --locked -p f2z-kt --all-targets` — 61 tests pass
- `cargo +1.97.1 test --locked -p f2z-authority nonce::tests` — 6 tests pass
- `cargo +1.97.1 clippy --locked -p f2z-kt --all-targets --all-features -- -D warnings` — pass

Mutation controls, independently replayed on exact head `50a76fa6`:

- Remove restore call — RED: restart accepts Bob.
- Key by nonce only — RED: distinct authority is falsely rejected.
- Key by authority only — RED: distinct nonce is falsely rejected.
- Change expiry release `>` to `>=` — RED at the exact boundary.
- Evict on capacity instead of refusing — RED: third insertion returns `Ok` instead of `LedgerFull`.

Independent exact-head review: APPROVE. Stable patch ID: `9ebcb1f4a744da23462f86de10be08e6b038ce1a`.

### DIFF
--- a/rs/crates/f2z-kt/src/log.rs
+++ b/rs/crates/f2z-kt/src/log.rs
@@ -154,19 +154,46 @@ impl core::fmt::Debug for Pending {
 /// first restart — the nonces were checked when they were first admitted, and
 /// the record of that decision *is* the journal.
 ///
+/// It does capture the nonce observation produced by that successful check.
+/// [`readmit`] moves the captured values into the separate live ledger only
+/// after admission succeeds, so startup both verifies its history without
+/// rejecting it and closes the replay window that history already spent.
+///
 /// It is a distinct type rather than a flag so that it cannot be reached from
 /// the live path: [`LogService::submit`] names [`NonceLedger`] and nothing
 /// else.
-struct ReplayLedger;
+struct ReplayedNonce {
+    now_ms: u64,
+    authority_id: f2z_authority::types::AuthorityId,
+    nonce: f2z_authority::types::AssertionNonce,
+    expires_ms: u64,
+}
+
+#[derive(Default)]
+struct ReplayLedger {
+    observed: Option<ReplayedNonce>,
+}
+
+impl ReplayLedger {
+    fn take_observed(&mut self) -> Option<ReplayedNonce> {
+        self.observed.take()
+    }
+}
 
 impl f2z_authority::nonce::NonceSeen for ReplayLedger {
     fn observe(
         &mut self,
-        _now_ms: u64,
-        _authority_id: f2z_authority::types::AuthorityId,
-        _nonce: f2z_authority::types::AssertionNonce,
-        _expires_ms: u64,
+        now_ms: u64,
+        authority_id: f2z_authority::types::AuthorityId,
+        nonce: f2z_authority::types::AssertionNonce,
+        expires_ms: u64,
     ) -> core::result::Result<(), f2z_authority::AuthorityError> {
+        self.observed = Some(ReplayedNonce {
+            now_ms,
+            authority_id,
+            nonce,
+            expires_ms,
+        });
         Ok(())
     }
 }
@@ -309,7 +336,7 @@ impl LogService {
     ///   rather than a log serving proofs against a root nobody cosigned.
     async fn replay(&self, journal: Journal) -> Result<()> {
         let mut state = self.state.lock().await;
-        let mut ledger = ReplayLedger;
+        let mut ledger = ReplayLedger::default();
 
         let mut cursor = 0usize;
         for stored in &journal.epochs {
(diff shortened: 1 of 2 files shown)


skyl:
Reviewed on `50a76fa6`. **This closes #650 properly. Merge when `gate` and `rs / gate` are green.**

### The property is real, and the failure mode is worse than "accepts a duplicate"

Neutralising the restore call:

```rust
log.rs:982
-    restore_replayed_nonce(&mut state.nonces, record, replay)?;
+    let _ = (&mut state.nonces, record, replay); // MUTANT: restore removed
```

```
a_restart_does_not_reopen_an_admitted_assertion_nonce ... FAILED
called `Result::unwrap_err()` on an `Ok` value: SubmissionReceipt {
  receipt: SubmissionReceiptTBS { label: free2z/kt/v1/receipt, handle: Handle(bob), … },
  signature: Signature(<redacted>) }
```

The log does not merely accept the replay — it hands back a **signed receipt** for it. That is the log putting its name on an assertion it had already consumed, which is precisely what a nonce ledger exists to make impossible, and it is only reachable across a restart. Baseline on the clean branch: **6 durability tests pass**.

### The test has a control, which is what makes it mean something

> Bob succeeds on a fresh control log, then is rejected only for reusing Alice's `(authority_id, nonce)` after Alice is journaled and the log is reopened.

Bob's submission is independently signed with a different handle, identity, entry, binding and signature. So the post-restart rejection is attributable to the restored ledger and to nothing else — without the fresh-log control, a test like this passes just as happily when Bob is malformed for an unrelated reason. Same shape as #632's `0o666` probe and #626's universal-forgery fixture check.

### Two mutations in the list worth naming

- **"Key by nonce only — RED: distinct authority is falsely rejected"** and **"key by authority only — RED: distinct nonce is falsely rejected."** Both directions of the composite key, so the ledger is neither too permissive nor too strict. A one-sided test would have been satisfied by a ledger that rejects everything after a restart, which would look secure and break every legitimate second submission.
- **"Change expiry release `>` to `>=` — RED at the exact boundary."** That is the same class as #624, where `>=` versus `>` at exactly `t + skew` was the whole defect. Good that it is pinned here at the outset rather than found later.

### What I did not observe

`gate` has not completed on this head. I ran `-p f2z-kt --test durability` plus the restore mutation; the `--all-targets` run, the `f2z-authority nonce::tests`, fmt and clippy are the PR's account, as are the four mutations I did not replay.

Issue #649, opened 2026-08-24
kt: a handle's first-entry authorization is not client-verifiable — move `HandleAssertion` into `EntryAuthorization`


skyl:
Found while writing `KT.md` §4.5–§4.7 (#647, closes #594). **Not a bug in the
rules — a bug in where the artifact lives.**

## The asymmetry

`KT.md` §4.4 authorizes an entry with signed material that sits **inside**
`DirectoryEntry`:

| `kind` | Authorizing artifact | Where it lives | Committed to the tree? | Client re-verifies? |
|---|---|---|---|---|
| `key_change` | `RotationProof` | `EntryAuthorization` | **yes** | **yes** |
| `platform_reset` | `ResetAuthorization` | `EntryAuthorization` | **yes** | **yes** |
| first entry | `HandleAssertion` + binding signature | **`SubmissionEnvelope`** | **no** | **no** |

`AkdValue = H("free2z/kt/v1/value", tls_codec(DirectoryEntry))` covers the entry
*including its authorization* — §3.3 says so explicitly, and that is the whole
reason "an entry cannot be re-authorized after publication." The
`SubmissionEnvelope` is outside it. `LookupResponse` carries `entry`, `proof` and
`bundle`, and no assertion. The log journals the envelope for its own replay, and
that journal is private.

## Consequence

§4.5 is **admission control the log performs on itself**, and no client, witness
or auditor can check that it happened. A log that skipped it entirely would
publish entries indistinguishable from ones that passed it. So the fix for #594
closes the hole **against a stranger** and **not against the log** — which is the
adversary `THREAT-MODEL.md` §3.1 is about and the one the whole KT machinery
exists for.

`KT.md` §8.1 step 6 therefore still says a client verifies nothing at
`entry_version == 1`, and §8.5's table still lists it under "cannot verify."
That is stated honestly in `KT.md` §4.7 and as `ARCHITECTURE.md` §13-S′, rather
than fixed, because it is a wire change to a structure two implementations have
frozen.

## Proposed fix

A fourth `EntryAuthorization` case:

```
struct {
    HandleAssertion assertion;      /* empty on a no-authority log (§4.6) */
    opaque          identity_signature[64];  /* by identity_pk            */
    opaque          auth_signature[64];      /* by this entry's directory_auth_pk */
} InitialBind;
```

The circularity that blocks a naive move — `AssertionBindingTBS.entry_digest` is
the `AkdValue`, which would then depend on the signature inside it — is resolved
[...]

PR #654, opened 2026-08-24
fix(kt): A15 could not fail — `account_epoch` must move like a counter

### LINKED ISSUE #651: kt: `account_epoch` must be a durable counter, never a clock — A15 is currently satisfiable unconditionally
Found while writing `KT.md` §4.5 (#647, closes #594). The **rule** is right in
`f2z-authority`; the **value** an issuer supplies can void it.

## The rule

`KT.md` §4.5.5 A15 requires an assertion's `account_epoch` to be **strictly
greater** than the value retained for that handle. Its purpose (§4.5.4) is that a
`reset` is a distinct account-ownership event, so an assertion for an epoch
already seen is one that has already been spent. `f2z-authority` enforces exactly
that (`AuthorityError::AccountEpochRegression`) and is correct.

## The hole

The rule is only a rule if the value is a **counter**. If an issuer derives
`account_epoch` from a clock — Unix seconds, say — then "strictly greater than
the last one" is satisfied **unconditionally and forever**, because time moves
forward. That does not weaken A15; it **deletes** it, while leaving a field in
place that looks like it is doing the work. One reference issuer currently does
exactly this where no counter is supplied.

Compounding: a `u32` holding Unix seconds exhausts the field in 2106, at which
point encoding raises rather than degrading.

## What `KT.md` now says

§4.5.4, as a normative MUST, and marked as one of the two places the
specification is deliberately stricter than a shipped implementation:

> `account_epoch` **MUST** be a durable per-account counter held by the
> authority, and **MUST NOT** be derived from a clock.

## What is left to do here

1. A conformance test that a `reset` assertion cannot be spent twice on one
   account-ownership event — i.e. one that would fail against a clock-derived
   epoch. `f2z-authority`'s own tests check the rule; nothing checks the
   *supply* of the value, which is where it goes wrong.
2. The issuing side needs a durable per-account counter column, incremented on
   the events that justify a reset (account recovery, transfer), and it should
   refuse to issue `intent = reset` rather than fall back to a clock when the
   counter is unavailable. That change belongs to whoever runs the user
   directory and is tracked on their side; this issue exists so the public rule
   and the test are not lost.
3. `f2z-assert` (the by-hand issuing binary) defaults `account_epoch` to `0`,
   which is fine for `intent = bind` and wrong to reach for on `intent = reset`.
   It should require the value explicitly for a reset.

Related: #650 notes that A17, the other independent guard on assertion replay,
currently leans on this one.

Refs #594, #647.

skyl:
Closes #651.

> **Rebased.** This branch also carried a fix for #650, which [#653](https://github.com/free2z/zuu/pull/653) landed first with an equivalent design. That work is dropped; what survives from it is one test #653 does not have — see *The ledger's retention* below.

## The rule that could not fail

`KT.md` §4.5.5 **A15** requires an assertion's `account_epoch` to be strictly
greater than the value retained for the handle, so a `reset` assertion cannot be
spent twice on one account-ownership event. **That is only a rule if the value is
a counter.** Derived from a clock — Unix seconds, which is what one shipped
issuer does — "strictly greater than the last one" holds unconditionally and
forever. The field is present, the check runs, and it enforces nothing.

§4.5.4 already said the value **MUST** be a durable per-account counter and
**MUST NOT** come from a clock. No verifier applied it. A requirement no verifier
checks is a requirement, not a rule — and a security control that is visibly
present and functionally absent is worse than a missing one, because it stops
anybody looking.

## A18, in two halves because one is not enough

Run **with A15**, before A17, so a submission that fails it never burns a nonce.

| | catches | misses |
|---|---|---|
| **A18a — ceiling.** `account_epoch < 2^20` | Unix seconds (over the ceiling since 1970-01-13); Unix ms does not fit `uint32` at all. Fires **on the value's face at a first `bind`**, where A15 has nothing to compare against | a coarser clock — days since the epoch is ~20,700 |
| **A18b — step bound.** at most `previous + 16` | a coarser clock: two resets a month apart differ by 30, and that is not thirty ownership events | a clock ticking 1–16 times between two issuances |

Both are **fixed constants, not published policy numbers** (unlike A9's validity
cap): the same bound on every log is one a client can apply without first
fetching §4.6's document, and there is no operational reason for two logs to
disagree about a counter's shape. No wire change — `AuthorityPolicyTBS` is
untouched, and the encoding vectors shared with the second implementation are
unaffected.

### What A18 cannot do is stated, not papered over

No check on a single `uint32` can prove the value came from durable storage.
§4.5.4 and `check_account_epoch`'s own documentation name the residual band
explicitly and state what the issuer **MUST** guarantee instead: a per-account
column that survives a restart, incremented **once per ownership event** (not
once per assertion issued), and a refusal to issue `intent = reset` rather than a
substituted clock. The MUST is not weakened to match what a verifier can see.

### Issuing side

`HandleAssertionTBS::new` refuses to *mint* a clock-shaped value, and
`f2z-assert` requires `--account-epoch` explicitly for `--intent reset` instead
of defaulting it to `0` (issue #651 item 3). A tool that will not mint the value
is a tool whose operator finds out at their own keyboard rather than at somebody
else's log.

## The tests are the point

**Every new test was run against the unfixed source first**, and each fails
there. That discipline matters more than usual: a rule that cannot fail is
invisible to a test that only checks the happy path, which is precisely how this
survived review.

| test | without the fix |
|---|---|
| `two_reset_assertions_from_one_ownership_event_cannot_both_be_spent` | `unwrap_err()` on `Ok(1700000000)` — **one recovery, two spendable assertions** |
| `a_clock_coarse_enough_to_clear_the_ceiling_is_caught_by_the_step_bound` | `Ok(20730)` |
| `a_clock_derived_epoch_is_refused_at_a_first_bind_too` | `Ok(By(AuthorityId(…)))` |
| `the_issuing_constructor_refuses_to_mint_a_clock` | `Ok(HandleAssertionTBS { … account_epoch: 1700000000 … })` |
| `the_log_refuses_a_clock_derived_account_epoch_end_to_end` | a signed receipt for `alice` |
| `issue_refuses_a_reset_without_an_explicit_account_epoch` | "a reset was issued with no counter" |
| `issue_refuses_a_clock_shaped_account_epoch` | "a clock was minted into an assertion" |

The first test is the shape the issue asks for: a conforming issuer increments
once, both assertions carry the same value, and A15 refuses the second — that is
A15 working. The clock-derived issuer, at the same two moments, hands out a
larger second value for no reason but the passage of time, and there is nothing
left for A15 to refuse. `a_clock_coarse_enough…` then asserts the **residue**
too: inside the step bound the value is still accepted, so the test states the
limit rather than overclaiming it.

`AssertionOverrides` in `f2z-kt::testing` is what makes a *badly issued*
assertion producible at all. The existing harness derives every assertion field
from the entry so a §4.4 test never trips over the assertion layer — which means
it cannot express the two rules that are about what the issuer *supplied*.

## The ledger's retention (kept from the #650 work)

`the_recovered_ledger_covers_the_longest_assertion_the_log_will_accept` is the
property neighbouring #653's restart test. A17's retention is derived from the
**assertion's own** `expires_ms + clock_skew_ms`, and `expires_ms - issued_ms` is
capped at `max_validity_ms` on admission — so it must cover the *longest*
assertion the log will admit, not the typical one. A ledger sized to the
fixture's usual minute would come back from a restart already having forgotten
the atypical assertion, which is exactly the one worth stealing. Verified capable
of failing: removing #653's `restore_replayed_nonce` call turns it RED alongside
`a_restart_does_not_reopen_an_admitted_assertion_nonce`.

## Spec

`KT.md` §4.5 carries dated corrections in `ARCHITECTURE.md` §4.9's style: §4.5's
preamble (neither of the two "stricter than a shipped implementation" notes is
still a gap against the Rust log), §4.5.4 (A18, its two halves, why the bounds
are fixed, and what it cannot prove), §4.5.5 (rule A18 in the numbered list),
§4.5.6 (the wire-code row), and the §12 open-questions bullet.

## Not fixed here, reported instead

The private Django issuer still does `account_epoch = now_ms // 1000` when no
counter is supplied, and the endpoint never supplies one — so **every** assertion
it issues, `bind` and `reset` alike, carries a clock. It is **non-conforming** by
§4.5.4, and per the workspace boundary the fix is not made in this PR. Its
docstring reasons carefully about the two ways a clock *under*-delivers (two
assertions in one second, a backwards correction) and concludes "both failures
are refusals, never acceptances" — which misses the case the rule exists for:
**two assertions minted more than a second apart for one ownership event are both
spendable**, and that is an acceptance.

**Ordering note:** once A18 is deployed, that issuer's assertions are refused
outright (its epoch of ~1.79e9 is far over the ceiling). The issuer needs its
durable counter before a log carrying A18 is put in front of it.

## Not touched

#649 — that a handle's first-entry authorization is not client-verifiable — is a
wire change to `EntryAuthorization` and is deliberately untouched. Nothing here
contradicts it: A18 is log-side admission control, which is exactly the layer
#649 observes no client can check.

`#![forbid(unsafe_code)]` intact; the AGPL/permissive boundary is unchanged
(`f2z-kt` AGPL, `f2z-authority` permissive, and A18 lives in the permissive crate
where every other assertion rule lives).

### DIFF
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -676,6 +676,27 @@ this paragraph be restated without its scope.
 > must survive a restart. Each is a case where the weaker behaviour leaves a
 > field or a rule in place that looks like it is doing work it is not.
 
+> **Correction (2026-08-24) — neither of those two is still a gap between this
+> document and the Rust log, and the `account_epoch` rule has grown a rule a
+> verifier can actually apply.**
+> [#650](https://github.com/free2z/zuu/issues/650),
+> [#651](https://github.com/free2z/zuu/issues/651).
+>
+> - **A17's ledger durability was a code defect, not a specification one**, and
+>   is fixed: the log rebuilt an empty ledger on startup, so every restart
+>   reopened the replay window for every assertion still inside its validity cap,
+>   while the trait it wrote through said in its own documentation that a restart
+>   must not do that. It now restores each admitted
+>   `(authority_id, nonce, expires_ms)` into the live ledger during journal
+>   replay. §4.5.5's sentence is unchanged because it was right.
+> - **`account_epoch`'s rule is now checkable, within a stated limit.** §4.5.4
+>   carries **A18** and says exactly which clocks it still cannot see and what the
+>   issuer must therefore guarantee. The issuing endpoint operated by the party
+>   that runs the user directory still derives the value from Unix seconds where
+>   no counter is supplied and remains **non-conforming**; that is tracked on
+>   their side, and A18 means a log now refuses those assertions rather than
+>   publishing entries authorized by a rule that could not fail.
+
 An assertion is a short-lived, single-use, log-pinned statement by a configured
 authority that a named handle belongs to a named identity key. It is **not** part
 of the `DirectoryEntry`; it accompanies one, in the submission envelope of §9.2,
@@ -837,6 +858,56 @@ baseline the handle's first `reset` must exceed.
 > event. A `u32` seconds counter also exhausts the field in 2106, which is a
 > second reason not to put a clock in it.
 
+> **Correction (2026-08-24) — the MUST above is now a rule a log applies, and
+> the honest limit of that rule is stated with it.**
+> [#651](https://github.com/free2z/zuu/issues/651). As first published this
+> section stated the issuer's obligation and stopped there, which left A15 in the
+> position the paragraph above describes: a rule that **cannot fail** against a
+> clock-derived value, in a document that says so. A requirement no verifier
+> checks is a requirement, not a rule, and this one had already been disregarded
+> by a shipped issuer.
+>
+> A log therefore now applies **A18** (§4.5.5), and it has two halves because one
+> is not enough:
+>
+> - **A18a — a ceiling.** `account_epoch` **MUST** be less than `2^20`
+>   (1,048,576). This is what a count of account-ownership events can plausibly
+>   reach: an account that changed hands or was recovered a million times is not
+>   an account. Unix time in whole seconds has exceeded it since 1970-01-13, and
+>   Unix milliseconds do not fit `uint32` at all, so the exact non-conformance
+>   named above is refused **on the value's face** — including at a first `bind`,
+>   where there is no retained value and A15 has nothing to compare against at
+>   all.
+> - **A18b — a step bound.** Where a value is retained for the handle,
+>   `account_epoch` **MUST NOT** exceed it by more than **16**. A counter
+>   increments once per ownership event, and only events whose assertion never
+>   reached the log accumulate slack, so the gap between two admitted assertions
+>   is a handful. A clock's gap is however much time passed. This is what catches
+>   a clock too coarse to trip the ceiling — days since the epoch is about
+>   20,700.
+>
+> Both are **fixed constants, not published policy numbers**, unlike A9's
+> validity cap: the same bound on every log is one a client can apply without
+> first fetching §4.6's document, and there is no operational reason for two logs
+> to disagree about a counter's shape.
+>
+> **What A18 cannot do, said plainly.** No check on a single `uint32` can prove
+> the value came from durable storage. A clock fine enough to tick between two
+> issuances but coarse enough to tick at most sixteen times, and small enough in
+> absolute terms to stay under the ceiling, is indistinguishable from a counter
+> in the value alone. A18 is therefore **necessary and not sufficient**, and the
+> MUST above is not weakened to match what a verifier can see. What the issuer
+> must guarantee instead, and what this document requires of it:
+>
+> - `account_epoch` is read from and written to a **per-account column that
+>   survives a process restart**;
+> - it is incremented **only** on the events that justify invalidating
+>   outstanding assertions — account recovery and account transfer — and once per
+>   such event, not once per assertion issued;
+> - an issuer that cannot read that column **MUST refuse to issue**
+>   `intent = reset` rather than substitute a clock, a timestamp, or any other
+>   value that happens to be increasing.
+
 #### 4.5.5 The rules, in the order a log applies them
 
 These are §4.4 rule 11, expanded. They run **after** §4.4's rules 1–10, for the
@@ -907,6 +978,13 @@ a log MUST apply all of them.
     replay, while the pair also catches an issuer that reused a nonce across two
     *different* assertions — an issuer bug, or a compromised issuer trying to
     make two claims look like one, which is the case worth failing on.
+18. **A18 — `account_epoch` moved like a counter, not like a clock.** It is
+    below `2^20`, and where a value is retained for this handle it exceeds that
+    value by at most 16. §4.5.4 states both bounds, why they are fixed rather
+    than published, and exactly what they cannot prove. Applied **with A15**
+    rather than after A17: it costs nothing, and a submission that fails it must
+    not burn a nonce. Numbered last because it was added last
+    ([#651](https://github.com/free2z/zuu/issues/651)); it is not run last.
 
 **On A17's ledger.** A log MUST refuse rather than evict: discarding an unexpired
 entry to make room silently reopens the replay window at exactly the moment the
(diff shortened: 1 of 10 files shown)


skyl:
**The #650 half has already landed — #653 merged as `ba9f46d5` about ten minutes before this opened.** That is the whole of the `CONFLICTING` state: this branch reintroduces its own `ReplayLedger` in `log.rs` against a `main` that now has one.

I diffed the two implementations rather than assuming the merged one wins. They are equivalent on the property that matters:

- `main`'s `restore_replayed_nonce` (`log.rs:990-1011`) takes the observed `(now_ms, authority_id, nonce, expires_ms)` out of the replay ledger and `observe`s it into the live one, with `expires_ms` being **the assertion's own** — so retention already tracks the longest assertion the log will accept, not a typical one.
- A failed `observe` is `Err(LogError::Storage(…))` propagated out of `open`, so a `LedgerFull` during rebuild is already a refusal to start, not a warning.
- Write-through (this PR) vs. take-and-restore (`main`) differ only if one record could carry more than one nonce observation. A `SubmissionEnvelope` carries at most one assertion, so it cannot.

So: **rebase onto `main` and drop the #650 half.** #650 is closed; leave it closed.

**Two things here are additive and should survive the rebase:**

1. `the_recovered_ledger_covers_the_longest_assertion_the_log_will_accept`. `main` has `a_restart_does_not_reopen_an_admitted_assertion_nonce` and nothing else on this path — it pins that the pair is still spent, but not that retention is derived from `max_validity_ms` rather than from a constant that happens to be long enough today. That interaction is exactly the kind that rots silently when the cap is raised. Port it.
2. The whole A18 / #651 half, which is untouched by anything merged.

**On A18, one question to answer in the PR rather than in review.** The table is honest about what each half misses, and fixing the constants rather than publishing them is well argued. But A18 is a new *admission* rule, and "no wire-format change, conformance vectors unaffected" understates it: a second implementation that encodes identically and does not implement A18 will accept assertions this log now refuses. That is a conformance divergence with no encoding change to detect it. Say explicitly whether the conformance suite ranges over admission rules or only over encodings — and if only encodings, say so in the PR so nobody later reads the green suite as agreement about A18.

Nothing above is a defect in the A18 work. The mutation table is the right shape — every test run against unfixed source first, and each row naming the *value* the unfixed code returned rather than just "it passed".

skyl:
Post-merge review, against the exact merged head `c260f4cb`. The rebase did what I asked in [my earlier comment](https://github.com/free2z/zuu/pull/654#issuecomment-5400447429) — the duplicate `ReplayLedger` is gone, `log.rs` is out of the diff, and both things I said had to survive did: `the_recovered_ledger_covers_the_longest_assertion_the_log_will_accept` is here at `durability.rs`, and the A18 half is intact.

Baseline on `c260f4cb`: `cargo test -p f2z-authority -p f2z-kt` → **EXIT=0, 176 passed**.

**Two findings, filed as #656 and #657.**

- **#656 — A18b's bound is per-handle, `account_epoch` is per-account.** §4.5.4 defines the counter's domain as the account; the log's baseline (`admit.rs:90`, `:233`) is the value retained for the handle. A18b's justification assumes the two advance together, and nothing keeps them in step: the counter advances on every account event, the retained value only when an assertion is admitted *for that handle*. `spend_reset(5, 30)` → `Err(AccountEpochStepTooLarge)` while `spend_reset(20, 30)` → `Ok(30)` — same value, and the refusal turns on the handle's dormancy rather than the value's shape. `spend_reset(5, 31)` shows the gap only grows, which makes it a permanent lockout of the handle rather than a retryable refusal, on the account-recovery path A18 exists to serve.
- **#657 — `ACCOUNT_EPOCH_CEILING` is unanchored.** `1 << 20` → `1 << 24` at `authority.rs:90` leaves the whole suite green. The number lives in three places — the constant, `f2z-assert.rs:72`'s hand-typed help text, and KT.md §4.5.4 — with nothing tying them together, and every test writes its expectation against the constant so the boundary moves with it. Same shape as #652 and #643.

**Two things I went after and could not break — both are right, and worth saying so.**

- `MAX_ACCOUNT_EPOCH_STEP` 16 → 64 **fails** `a_clock_coarse_enough_to_clear_the_ceiling_is_caught_by_the_step_bound`. That test spends a concrete scenario — two issuances a month apart, 30 ticks of a days-since-epoch clock — instead of symbolic arithmetic against the constant, which is exactly why it survives a change to the constant. This is the difference between it and the ceiling, and it is the reason #657 is about only one of the two.
- Moving `check_account_epoch` after `ledger.observe` **fails** `the_log_refuses_a_clock_derived_account_epoch_end_to_end` with `Authority(ReplayedNonce)` at `durability.rs:434`: the refused submission burned the nonce and the legitimate submission behind it was rejected as a replay. So §4.5.5's new claim — *"Applied with A15 rather than after A17 […] a submission that fails it must not burn a nonce"* — is a claim with a test under it. Worth noting it is **not** `a_failed_submission_never_burns_the_nonce`, which is anchored at the signature check and cannot see A18's placement at all; the end-to-end durability test is the one carrying it.

The mutation table in the description is the right standard — every row naming the value the unfixed code returned rather than just "it passed". #656 is the one to act on.

PR #658, opened 2026-08-24
docs(e2ee): close relay wire contract gaps

### LINKED ISSUE #586: WIRE.md: anti-replay retention is unspecified against clock skew, and §6.3/§10 contradict on send-side errors

skyl:
Closes #586.

## What changed

Closes the remaining relay/WIRE implementation gaps with shared, executable protocol rules:

- binds relay-state authorization before replay commitment and before padding-policy disclosure;
- validates challenge purpose/scope centrally;
- defines `pending` as the unacknowledged index span while keeping storage quotas tied to live stored rows;
- aligns production relay, FakeRelay, store backends, capability policy, error mapping, and WIRE corrections;
- exercises TTL expiry, resubscribe, capacity replacement, and malformed challenge behavior over real sockets.

## Verification

- pinned rustfmt for all 10 `rs/` crates — pass
- all targets for `f2z-codec`, `f2z-relay-proto`, `f2z-relay-store`, `f2z-relay-testkit`, and `f2z-relay` — pass
- pinned Clippy with `-D warnings` across those five crates — pass
- real relay/FakeRelay socket and dual-backend store suites — pass

Compiled mutation controls, independently replayed on exact head `9488debf`:

- bypass production `SUBSCRIBE` authorization — RED: unauthorized replay key commits;
- neutralize challenge purpose/scope validation — RED: malformed request is accepted over a production socket;
- derive `pending` from stored rows — RED: TTL/resubscribe reports 0 instead of the required index span 1;
- derive quota from pending span — RED: replacement append is refused after TTL frees live storage;
- move padding before relay-state authorization — RED: unauthorized malformed APPEND leaks `ERR_BAD_SIZE` instead of collapsed `ERR_UNAVAILABLE`.

Independent exact-head review: APPROVE. Stable patch ID: `4dd24d96939007bf6d1a30a511ef9dd59ca38ff9`.

### DIFF
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -14,6 +14,11 @@
 [0012](./decisions/0012-anti-abuse-v1.md),
 [0014](./decisions/0014-directory-key-rotation.md).
 
+> **Correction (2026-08-24) — the implementation claim above is no longer
+> true.** `f2z-codec` and `f2z-relay-proto` now implement the canonical wire
+> structures and the shared, side-effect-free protocol rules. There is still no
+> relay or client binary and no deployment; this remains a Phase 1 specification.
+
 This document is written to be read by a third-party security auditor and by an
 independent implementer. It fixes the wire encodings that
 [`ARCHITECTURE.md`](./ARCHITECTURE.md) deliberately left as *proposed*
@@ -106,6 +111,14 @@ construction #544 selects.
   ([ADR 0001](./decisions/0001-platform-priority.md)).
 - A **fatal** error means: send the response, then close the WebSocket
   connection. A non-fatal error means: send the response, keep the connection.
+
+> **Correction (2026-08-24) — a response is possible only when a request id is
+> recoverable.** The rule above assumes that the relay can recover a nonzero
+> `request_id` from the fixed five-byte binary header. When it can, it MUST send
+> the fatal error on that id and then close. When it cannot — including a binary
+> message shorter than five bytes, a zero id, and every text message that §4.2
+> forbids the relay to decode — it MUST close without a protocol response. It
+> MUST NOT invent id zero for a response, because §4.3 reserves zero for pushes.
 
 ---
 
@@ -469,6 +482,27 @@ Steps 2 and 3 precede signature verification because they are cheap and they are
 the flood-resistant filters; step 5 follows verification so that an unsigned
 guess cannot probe the address space.
 
+> **Correction (2026-08-24) — step 5 is not one rule for every signed
+> command.** It applies literally only to `SUBSCRIBE`, `UNSUBSCRIBE`, `READ`,
+> `ACK`, and `DELETE_QUEUE`: `signer_key` MUST equal the queue's registered
+> `recv_key`, and failure is `ERR_NO_ACCESS`. The other signed commands use these
+> rules:
+>
+> - `CREATE_QUEUE` and `CREATE_CONTACT_QUEUE` have no address yet. Their
+>   transcript address is zero, and `signer_key` MUST equal the `recv_key` in
+>   the body. Failure is `ERR_NO_ACCESS`.
+> - `BIND_SEND` acts on an unbound side, so there is no registered send key to
+>   compare. `signer_key` MUST equal the `send_key` in the body. Failure is
+>   `ERR_NO_ACCESS`. This frame-internal check reveals no relay state.
+> - `APPEND` requires `signer_key` to equal the bound `send_key`, but every
+>   state-dependent failure, including a missing address and a wrong or absent
+>   bound key, is `ERR_UNAVAILABLE` (§6.3 and §10's correction).
+>
+> Requiring the `BIND_SEND` self-check is security-significant. Without it, a
+> caller that learned `send_addr` could sign with a disposable key while binding
+> an unrelated key it does not possess. The bind-once theft described by §7.4
+> would cost the caller no write capability at all.
+
 ### 5.2 Relay identity binding — the attack it closes
 
 ```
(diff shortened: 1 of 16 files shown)


skyl:
Post-merge review, against the exact merged head `9488debf`. Baseline `cargo test -p f2z-codec -p f2z-relay-proto -p f2z-relay-store -p f2z-relay-testkit -p f2z-relay` → **EXIT=0, 468 passed** across 31 targets.

The ordering argument is right and `verify_authorized` is the correct shape for it — running the closure after the signature and before both padding validation and replay commitment, with the seen-set untouched on refusal. Two findings, both about reach rather than correctness: **#659** and **#660**.

**#659 — the ordering guarantee is tested for 1 of 6 relay-state commands.** Neutering the closures at `engine.rs:807/831/908/943/1020` (UNSUBSCRIBE, READ, ACK, DELETE_QUEUE, APPEND) to `|_| Ok(())` while leaving SUBSCRIBE at `:768` intact → **EXIT=0**, everything green. Neutering only SUBSCRIBE → **EXIT=101**, `denied_queue_authorization_does_not_commit_the_replay_key` fails with *"denied command entered the relay seen-set"*. Same suite, same reader, opposite verdicts, so the green result is coverage rather than a broken probe. Worth being precise about what it means: access control still holds in all six — each handler still does its `queue_by_recv`/`queue_by_send` lookup afterwards — so what the mutation removes is exactly the property this PR adds, and nothing else. Every existing test asserts the eventual refusal; none asks *when* the seen-set was touched. APPEND is the one I would want covered first, because the PR's own justification names the consequence: without the closure running first, an unauthorized APPEND gets `ERR_BAD_SIZE` instead of the collapsed send-side refusal, which is a padding-policy oracle to an unrelated key.

Related and in the same family: `verify` is now literally `verify_authorized(…, |_| Ok(()))` and its doc says a relay admitting a command **MUST** use `verify_authorized`. Nothing enforces that MUST, and `|_| Ok(())` is already the correct argument at three legitimate call sites, so it reads as normal in review.

**#660 — the new delegation walker has no self-test.** `delegation_reaches_strict_base` is correct: pointing a `DelegatesTo` target at a `NotASignatureCheck` row fails with the right message. But replacing its body with `return true` leaves all 11 scanner tests green, and it is the only load-bearing predicate in that file without a negative self-test — `calls_path`, `calls_method_on` and `registration_matches` all have one. It also replaced an inline check that was structurally impossible to fool.

**Checked and sound**, so nobody re-derives them: `retention_is_sound`'s body is untouched and #631's `>=` at `replay.rs:234` survived — only the doc comment changed, correctly retiring the "`WIRE.md` does not state this" framing now that §5.5 derives the relation; `capabilities.rs:793` still asserts that a document with `antireplay_window_ms == clock_skew_ms` validates while `ClientPolicy` refuses it, with only the message rewritten, and the new message explains the reason better than the old one; the `.verify(` negative controls in `capabilities.rs` and `command.rs` are intact.

One process note, unrelated to the code: this landed on `main` as `docs(e2ee): close relay wire contract gaps`, but 646 of its 846 lines are Rust, and they change relay authorization ordering, `pending` semantics, error mapping and a verification scanner. The history and anything derived from it now describe a behavioural security change as documentation. `fix(relay):` or `feat(relay):` would have been the honest type.

Issue #660, opened 2026-08-24
workspace_strict_verification_scan: the new delegation walker is the only predicate in the file without a self-test


skyl:
Found reviewing #658 (merged as `7107905`), against head `9488debf`. Minor, and narrow.

#658 needed `command.rs::verify` → `command.rs::verify_authorized` → `key.rs::verify` to pass the registry contract, which is two hops. Before, `Strictness::DelegatesTo`'s target had to be registered as `CallsVerifyStrict` **directly**, inline in `every_verify_function_in_the_workspace_is_registered_as_strict`. That inline check was replaced by a new transitive walker, `delegation_reaches_strict_base` (`workspace_strict_verification_scan.rs:923`).

**The walker is correct.** Pointing `command.rs::verify`'s target at a `NotASignatureCheck` row (`f2z-kt-core/src/verify.rs::verify_lookup`) fails with exactly the right message:

```
f2z-relay-proto/src/command.rs::verify delegates to f2z-kt-core/src/verify.rs::verify_lookup,
whose registered chain does not terminate in a strict verifier
```

**Nothing proves it stays correct.** Replacing its body with an unconditional `return true`:

```
cargo test -p f2z-codec --test workspace_strict_verification_scan
=> EXIT=0, 11 passed; 0 failed
```

The registry contract's central assertion — that a row claiming "it only delegates" delegates to something that really does reach `verify_strict` — becomes unfalsifiable, and the scan reports success.

## Why this file in particular

Every other predicate here carries a negative self-test proving it can say no:

| predicate | its self-test |
|---|---|
| `calls_path` | `a_free_function_call_is_matched_only_when_it_is_really_a_call` (`:1019`) |
| `calls_method_on` | `a_comment_or_unrelated_identifier_cannot_claim_strict_verification` (`:1035`), `the_live_registry_contract_rejects_the_right_method_on_the_wrong_receiver` (`:1069`) |
| `registration_matches` | `duplicate_verify_definitions_cannot_share_one_registry_row` (`:1113`) |
| `delegation_reaches_strict_base` | — |

`delegation_reaches_strict_base` has exactly two references in the file: its definition at `:923` and its single use at `:1316`. It is the only load-bearing predicate without one, and it replaced a check that was structurally impossible to fool — the inline version could not be wrong, because it had no traversal to get wrong.

[...]

Issue #659, opened 2026-08-24
f2z-relay: #658's authorization-before-commit ordering is tested for 1 of 6 relay-state commands


skyl:
Found reviewing #658 (merged as `7107905`), against its exact head `9488debf`.

#658's thesis is an **ordering** guarantee, and it is a good one. `CommandVerifier::verify_authorized` runs the caller's relay-state authorization closure *after* the signature and frame-internal checks but *before* padding validation and replay commitment, so — in the PR's own words — "a correctly signed request under an unrelated key is authenticated but unauthorized, and must not consume the bounded seen-set", and an unauthorized APPEND cannot "distinguish an invalid padding size from the collapsed send-side refusal".

The production relay wires that closure at six call sites (`f2z-relay/src/engine.rs`): `768` SUBSCRIBE, `807` UNSUBSCRIBE, `831` READ, `908` ACK, `943` DELETE_QUEUE, `1020` APPEND. **The guarantee is tested at one of them.**

## Reproduction

Worktree at `9488debf`. Baseline `cargo test -p f2z-codec -p f2z-relay-proto -p f2z-relay-store -p f2z-relay-testkit -p f2z-relay`: **EXIT=0, 468 passed** across 31 targets.

**Mutation A — neuter five of six, leaving SUBSCRIBE (768) intact.** Lines 807, 831, 908, 943, 1020 each replaced with `|_| Ok(()),`:

```
cargo test -p f2z-relay -p f2z-relay-proto  =>  EXIT=0
```

Every target green. UNSUBSCRIBE, READ, ACK, DELETE_QUEUE and APPEND are all back to committing the replay key and disclosing padding refusals before relay-state authorization — exactly the pre-#658 behaviour — and nothing notices.

**Mutation B — the control. Neuter only SUBSCRIBE (768), restore the other five:**

```
cargo test -p f2z-relay  =>  EXIT=101
test denied_queue_authorization_does_not_commit_the_replay_key ... FAILED
assertion `left == right` failed: denied command entered the relay seen-set
```

Same suite, same reader, opposite verdicts. That is what makes A attributable: the detection mechanism works and measures the live relay-wide seen-set through the production WebSocket listener — `denied_queue_authorization_does_not_commit_the_replay_key` is a well-built test. It just ranges over one command.

## Why the green result is easy to misread

[...]

PR #661, opened 2026-08-24
fix(e2ee): close residual wire contract gaps

### LINKED ISSUE #585: spec(e2ee): five WIRE.md gaps surfaced by hand-deriving byte vectors

skyl:
Closes #585.

## What changed

- Restates the PoW counter at point of use as exactly eight-byte unsigned big-endian with no length prefix.
- Restricts exactly the five operator and three provenance capability text fields to printable ASCII `0x20..=0x7e`, with empty allowed; binary `ShortBytes` values such as challenge scopes remain unrestricted.
- Makes `ReadResponse.has_more` decode canonically: only literal `0` and `1` are accepted.
- Proves semantic attachment through both signed client policy and production relay capability construction.

## Verification

- full locked workspace tests, all targets — pass
- full locked workspace Clippy with `-D warnings` — pass
- repository fmt/Clippy gates across all 10 `rs` crates — pass

Compiled mutation controls, independently replayed on exact head `82938cdd`:

- omit each of the eight text fields from validation — all eight RED;
- widen to arbitrary bytes — RED;
- narrow to alphanumeric — RED on valid space/quote/backslash boundaries;
- remove strict `has_more` check — RED;
- normalize nonzero as true — RED (`NotCanonical` instead of required decode refusal);
- remove `ClientPolicy::accept` validation — RED after cryptographic verification;
- remove production `caps::build` validation — RED before publication;
- change PoW counter BE to LE — RED against the independent known-answer vector.

Independent exact-head review: APPROVE. Stable patch ID: `d4e531495c6bc410d2f74efa0e86d3c6405456c4`.

### DIFF
--- a/docs/e2ee/WIRE.md
+++ b/docs/e2ee/WIRE.md
@@ -1544,6 +1544,11 @@ struct {
 } SignedCapabilities;
 ```
 
+The five operator fields and three provenance fields are inert human-readable
+text, not markup. Each byte MUST be printable ASCII (`0x20..0x7e`); the empty
+value remains valid. A renderer MUST display the bytes as text and MUST NOT
+interpret HTML, Markdown, terminal escapes, or any other embedded language.
+
 > **Correction (2026-08-24) — the two anti-replay fields were published as
 > independent values, but they are not.** Section 5.5 derives the required
 > relation and retention origin. A relay MUST publish
@@ -1950,6 +1955,8 @@ A stamp is valid iff
 `H("free2z/relay/v1/pow", challenge || salt || counter)` has at least
 `difficulty_bits` leading zero bits, the challenge is unconsumed and unexpired,
 and (for `contact_append`) the challenge was issued for that `contact_addr`.
+At this point of use, `counter` is exactly an eight-byte unsigned `uint64` in
+network (big-endian) order, concatenated directly with no length prefix.
 Verification is one hash; the relay marks the challenge consumed. The choice of a
 verify-cheap, GPU-friendly function and its consequences are argued in §12.4.
 
--- a/rs/crates/f2z-codec/src/commands.rs
+++ b/rs/crates/f2z-codec/src/commands.rs
@@ -14,7 +14,9 @@
 use alloc::format;
 use alloc::vec::Vec;
 
-use tls_codec::{TlsDeserializeBytes, TlsSerializeBytes, TlsSize};
+use tls_codec::{
+    DeserializeBytes, Error as TlsError, TlsDeserializeBytes, TlsSerializeBytes, TlsSize,
+};
 
 use crate::error::CodecError;
 use crate::pow::{PowParams, PowStamp};
@@ -434,12 +436,23 @@ pub struct QueuedMessage {
 }
 
 /// `READ` response (§6.2). `READ` never mutates.
-#[derive(Clone, Debug, PartialEq, Eq, TlsSize, TlsSerializeBytes, TlsDeserializeBytes)]
+#[derive(Clone, Debug, PartialEq, Eq, TlsSize, TlsSerializeBytes)]
 pub struct ReadResponse {
     /// The messages, `<0..2^24-1>` bytes of them.
     pub messages: VecU24<QueuedMessage>,
-    /// Nonzero when more messages remain past the last one returned.
+    /// 1 when more messages remain past the last one returned; 0 otherwise.
     pub has_more: u8,
+}
+
+impl DeserializeBytes for ReadResponse {
+    fn tls_deserialize_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), TlsError> {
+        let (messages, rest) = VecU24::<QueuedMessage>::tls_deserialize_bytes(bytes)?;
+        let (has_more, rest) = u8::tls_deserialize_bytes(rest)?;
+        if has_more > 1 {
+            return Err(TlsError::UnknownValue(u64::from(has_more)));
+        }
+        Ok((Self { messages, has_more }, rest))
+    }
 }
 
 /// `ACK` request (§6.2). Cumulative, monotone, idempotent, never selective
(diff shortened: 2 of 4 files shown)


skyl:
Post-merge review, against the exact merged head `82938cdd`. Baseline `cargo test -p f2z-codec -p f2z-relay-proto -p f2z-relay` → **EXIT=0, 379 passed**.

The printable-ASCII rule is well built. `read_response_rejects_non_boolean_has_more_bytes` is the standout — hand-authored bodies with the comment saying why (*"so the expected value does not come from the encoder under test"*), both accepted values and both rejected values, and a real manual `DeserializeBytes` rather than a post-hoc check. That is the right shape.

**One finding, filed as #662, and it is a coverage regression rather than a defect in the new rule.**

Changing `an_operator_string_cannot_break_the_document_it_appears_in` from a newline to a backslash was *necessary* — after this PR a newline is refused before it reaches the renderer, so the old input would have failed for the wrong reason. But `json_string` (`caps.rs:499-518`) escapes `"`, `\`, `\n`, `\r`, `\t` and has a `\u{04x}` fallback, and its eight call sites are exactly the eight fields this PR restricts. So the last four branches now have no coverage from anywhere.

The same mutation on both trees, breaking only the newline branch:

| tree | result |
|---|---|
| `7107905` (main, pre-#661) | **EXIT=101** — `an_operator_string_cannot_break_the_document_it_appears_in` FAILED |
| `82938cdd` (this PR) | **EXIT=0**, all 379 green |

Same test, same name, same mutation; it stops failing. The fix is cheap and does not touch the new rule: unit-test `json_string` directly — it is a private free function over `&[u8]`, so the upstream validator cannot get in the way — and leave the document-level test exactly as this PR left it.

Keeping the escaper is right; it is defence in depth, and the point of a second layer is that it holds when the first is relaxed or reused. That is also the argument for testing it.

**Checked and not filed**, so nobody re-derives them:

- **The eight-field enumeration is anchored, indirectly but effectively.** `HUMAN_TEXT_FIELDS` and the `human_text` array in `validate` are two independent lists, so my first instinct was that a ninth field could be added to neither. It cannot go unnoticed: `the_capability_document_encodes_field_for_field_as_section_11_1_declares` walks the `Capabilities` encoding byte for byte and `assert_encodes` compares against the full expected length, so a new field breaks it and lands the author in exactly the file where the omission is visible.
- `set_human_text`'s exhaustive `match` over `HumanTextField` means a new *enum* variant cannot be silently skipped either.
- `json_string`'s call sites really are only those eight — the restriction does not accidentally reach challenge scopes or any other `ShortBytes`, which the code comment claims and which is true.

One note carried over from #658's review: the `Independent exact-head review: APPROVE` line in the description did not catch this, and did not catch #659 or #660 on the previous PR either. It reads as a verification claim in a list of verification claims, which is where it is least useful.

Issue #657, opened 2026-08-24
ACCOUNT_EPOCH_CEILING can be changed with the whole suite green: three copies of a normative constant, no cross-check


skyl:
Found reviewing #654 (merged as `a4a2282c`), against its exact head `c260f4cb`. Sibling of #656; independent of it.

`ACCOUNT_EPOCH_CEILING` is a **normative, client-applied protocol constant**. That is #654's own argument for fixing it rather than publishing it in §4.6:

> Both are **fixed constants, not published policy numbers**, unlike A9's validity cap: the same bound on every log is one a client can apply without first fetching §4.6's document, and there is no operational reason for two logs to disagree about a counter's shape.

The number exists in three places and nothing ties them together.

| where | form |
|---|---|
| `f2z-authority/src/authority.rs:90` | `pub const ACCOUNT_EPOCH_CEILING: u32 = 1 << 20;` |
| `f2z-authority/src/bin/f2z-assert.rs:72` | hand-typed in the CLI help: *"A value at or above **1048576** is refused outright as a clock"* |
| `docs/e2ee/KT.md:873-874` | *"`account_epoch` **MUST** be less than `2^20` (1,048,576)"* |

## Reproduction

Worktree at `c260f4cb`. Baseline `cargo test -p f2z-authority -p f2z-kt`: **EXIT=0**, 176 passed.

Mutation — `authority.rs:90` only:

```rust
-pub const ACCOUNT_EPOCH_CEILING: u32 = 1 << 20;
+pub const ACCOUNT_EPOCH_CEILING: u32 = 1 << 24;
```

`cargo test -p f2z-authority -p f2z-kt`: **EXIT=0**. Whole suite green.

The log now admits `account_epoch` values that §4.5.4 says a log MUST refuse, the CLI still tells operators the threshold is 1048576, and nothing anywhere goes red.

## Why nothing catches it

Every assertion about the ceiling is written against the constant — `ACCOUNT_EPOCH_CEILING - 1`, `== ACCOUNT_EPOCH_CEILING` (`adversarial.rs:1625`, `:1633`, `:1668`) — so the boundary moves with it and the tests keep passing on both sides of the move. `assertion.rs:164`, the issuing constructor, reads the same constant, so the mint side moves in lockstep too. **Self-consistency is what hides this**: the repo agrees with itself while diverging from the published number and from any second implementation applying 2^20.

[...]

PR #664, opened 2026-08-24
feat(zuuli): messaging client contract and messages surface

Jhonattan2121:
Implements the TypeScript half of `docs/e2ee/CLIENT-CONTRACT.md`: all 46 commands, all 11 events, and the `/messages` route.

## What is here

`src/lib/messaging/` mirrors `src/lib/wallet/` file for file. `types.ts` declares each domain type once as a schema with the static type derived from it; `bridge.ts` is the only place the frontend talks to the engine; `mock.ts` is a self-contained in-memory engine; `events.ts` is one subscription interface over both runtime modes.

Enrollment is a separate export whose `invoke` helper carries no `plugin:` prefix, per §2.2, with the reason written next to it so nobody normalises the inconsistency later.

`src/features/messages/` mounts at `/messages/*` with the navigation entry §2.4 specifies.

## The contract layer, after review

`WIRE_COMMANDS` is the module's single population: `BridgeMethod` is its keys, `RESULTS` satisfies `Record<BridgeMethod, …>`, and both `invoke` helpers take the method key and resolve the wire name themselves. A renamed command is a compile error in every consumer.

Two tests cover two different directions, and neither subsumes the other:

- `parity.test.ts` pins the mock against the bridge — a mock returning the wrong values, arity drift, a dead method left by a rename. Its count assertion is a tripwire on the module shrinking.
- `scripts/messaging-contract.node-test.mjs` reads §3's command tables and §5.1's event table out of `CLIENT-CONTRACT.md` and compares sets against `WIRE_COMMANDS` and `EVENTS`. This is what catches contract *growth* — a command added to the document and wired nowhere fails here until someone wires it. Proven by mutation: adding a row to §3 fails with a `missing` entry; renaming a wire command fails with both `missing` and `extra`; both reverted, clean.

`checked()` parses every response in development, and in every build for the small, low-frequency ones — `getEngineStatus`, `getDeviceInfo`, the enrollment trio, `startEngine`, `stopEngine` — via `ALWAYS_PARSED: ReadonlySet<BridgeMethod>`. Every event payload is parsed unconditionally. The hot list paths (`listConversations`, `listMessages`, …) stay cast, on the bundle-cost argument §4.1 makes.

## Three deliberate choices, please shoot them down if you disagree

**The mock is single-witness.** `independentWitnesses: 1`, threshold unmet, and `startEngine` settles in `degraded` rather than `running`. A mock reporting a met threshold would let every screen be built against a state that does not exist at launch, and rule 5 in §9 would go untested exactly where it matters.

**`mergedAtEpoch` stays null in the mock.** Enrollment is a directory submission; filling it in immediately would let the UI be built against an "active" state the real log never reaches that fast.

**The echo peer generates traffic.** It advances delivery through all four protocol states and replies with real `parents`, `epoch` and `senderLeafIndex`, so §7's total order is exercised by data rather than by a fixture that happens to be sorted.

## Two things I did not invent, and need you to settle

**The typed confirmation strings** for `f2zmsg_unenroll` and `acknowledge_alarm`. §3.2 and §3.10 fix the *shape* — "as `discard_unrecoverable_send` already uses" — and `send.rs:2150` shows that one is `"I CHECKED WALLET HISTORY"`, but neither string for messaging appears anywhere in `docs/e2ee/`. The client does not hold a copy: the mock checks only that a confirmation was collected, and the engine owns the comparison.

**`zod` is the only new dependency.** §4.1 names zod or valibot and is explicit that the bundle cost is real. I picked zod because the parse failing well — naming the field and the expected type — is where §4.1's value comes from, and zod's issue paths are what make that readable. Worth a measured bundle delta rather than a preference, if you care about the number.

## Verification

Run on this machine, each just now:

- `npx vitest run src/lib/messaging/` — 97/97
- `node --test scripts/messaging-contract.node-test.mjs` — 3/3, proven by mutation in both directions
- `npx playwright test` (the full suite, not a subset) — 132/132
- `tsc --noEmit` — clean
- `npm ci` — exits 0

`npm test` chains these with `&&`, and on this machine 15 pre-existing `origin/main` failures (`TopBar.signed-in`, `article-tags-contract`, `donation-http`, `private-live-contract` — environmental, likely the `pt-BR` locale shape of #449) short-circuit it before Playwright ever runs. So "ran `npm test`" is not a claim worth making from here; the three suites above are reported separately because that is what actually ran.

## Not done

`src/lib/messaging/` is the contract layer and the `/messages` screen is a working surface over it, not a finished design. The conversation UI covers transcript, composer, delivery state, gap marker and the §11.1 browser guarantee; settings surfaces for retention, receipts, relays and witnesses have commands and mocks but no screens yet.


skyl:
**This head is red. `npm test` fails 2 of 793 on `575bfe95`, and `zuuli / frontend` runs `npm run test` (`.github/workflows/zuuli.yml:212`), so the gate will land on it.** Flagging now rather than at the end of my review so you can fix it in parallel.

```
FAIL  src/components/layout/navigation.test.ts > exposes secondary and account routes coherently
  expected [ 'articles', 'messages', …(2) ] to deeply equal [ 'articles', 'profile', …(1) ]
  +   "messages",

FAIL  src/components/layout/navigation.test.ts > drives grouped desktop routes from the same config
  expected [ 'home', 'live', 'articles', …(5) ] to deeply equal [ 'home', 'live', 'articles', …(4) ]
  +   "messages",

Test Files  1 failed | 53 passed (54)
     Tests  2 failed | 786 passed | 5 skipped (793)
```

`npm ci` exits 0; this is the suite, not the install. And `navigation.test.ts` is byte-identical to `dd67eac`'s — I diffed it — so the PR adds the nav entry in `navigation.ts` without updating the test that enumerates the ids.

**Worth saying plainly: this is the navigation test doing exactly its job.** It pins the full id list for both surfaces and fails when the population changes, which is the same anchor shape as #597/#609/#612/#614. A new top-level destination in a wallet is not a detail that should land without someone re-reading that list, and the test is what forces the re-read. The fix is to add `"messages"` at the intended position in both expectations — the position is the decision, not the presence.

One process note, meant constructively. The Verification section reads:

> 96 parity tests, `tsc --noEmit` clean, `prettier --check` clean, `ui-copy-truncation` clean

Each of those is true and none of them is `npm test`. A list of green subsets reads like a green suite, and this is the case where the difference matters — the two failures are in neither the parity file nor the type check. Running what the gate runs is the cheaper habit.

Continuing the review of the contract layer; more shortly. Nothing else so far is merge-blocking.

skyl:
Contract-layer review, against `575bfe95`. `zuuli / frontend` has since gone `FAILURE`, which is the navigation test from my previous comment.

**The parity test is good, and it is good for the right reason.** It closes a real loop: `bridgeMethods` is derived from `RESULTS`, and `ALL_BRIDGE_METHODS`, `SAMPLE_ARGS` and `mockMessaging` are each asserted equal to it. Four artefacts, mutually pinned, so a rename in any one of them fails somewhere.

I tested your headline claim rather than taking it. Adding a bridge method with no schema:

```ts
export const ALL_BRIDGE_METHODS = {
  ...messaging,
  ...enrollment,
  brandNewCommand: async () => ({ nope: true }),
} as const;
```

```
npx tsc --noEmit                       => exit 0        (passes)
npx vitest run parity.test.ts          => 1 failed | 95 passed
   × declares a schema for every bridge method and no others
```

So it works — with one correction worth making because it connects to the failing gate. `ALL_BRIDGE_METHODS` is `{...messaging, ...enrollment} as const`, not `Record<BridgeMethod, …>`, so this is **not** a type error; `tsc` is happy. "Fails the build rather than review" is true only if the build includes `npm test`. The guarantee lives entirely in the suite that wasn't run before pushing — which is the same gap that produced the two navigation failures.

**One finding: the loop is closed, but not against the contract.**

All four artefacts it pins are defined in this PR, so the test proves they agree with *each other* and nothing checks them against `CLIENT-CONTRACT.md`. I extracted §3's command tables and §5.1's event table mechanically and compared:

```
spec: 46   impl: 46
in spec, MISSING from bridge: []
in bridge, NOT in spec:       []
```

(with the app-crate trio mapped: `f2zmsg_enroll`/`f2zmsg_enrollment_status`/`f2zmsg_unenroll` → `enroll`/`getEnrollmentStatus`/`unenroll`.) Events match name for name, all 11.

**So "all 46 commands, all 11 events" is true today, and nothing in the suite would notice it stopping being true.** A 47th command added to §3 and wired nowhere passes every assertion in `parity.test.ts`, because the population it ranges over is `Object.keys(RESULTS)` — the implementation's own list. That is expectations generated from the implementation, one level up. §4.1 asks this test to be the thing that stops the client drifting from the contract, and against the contract itself it currently cannot fail.

The cheap floor is a literal anchor, the same shape as #643 and #657 — a literal compared against a derived value, citing the section:

```ts
it("covers the whole contract surface", () => {
  // CLIENT-CONTRACT.md §3 declares 46 commands, §5.1 declares 11 events.
  // These literals are the contract's numbers, not this module's — when the
  // document grows a command, this fails until someone wires it.
  expect(bridgeMethods.length).toBe(46);
  expect(Object.keys(EVENTS).length).toBe(11);
});
```

A count survives a rename, so the stronger version parses `CLIENT-CONTRACT.md` and compares sets — roughly what I did above, and there is precedent for a source-scanning test in this workspace (`workspace_strict_verification_scan.rs`). Your call which; the counts are worth having either way, and they are two lines.

**On your three deliberate choices — I agree with all three**, and the reasoning for the single-witness mock is the strongest of them: a mock reporting a met threshold would let every screen be built against a state that does not exist at launch, and §9 rule 5 would go untested exactly where it matters. Same for `mergedAtEpoch` staying null. The echo peer generating real `parents`/`epoch`/`senderLeafIndex` is what makes §7's ordering exercised by data instead of by a pre-sorted fixture — that is the difference between a test and a tableau.

**On the two questions you put to the owner:** the confirmation-string one answers itself and you have already answered it correctly. A client-side copy of a typed confirmation phrase is a second source of truth that drifts silently, and "the mock checks only that a confirmation was collected, the engine owns the comparison" is the right split regardless of what the strings turn out to be. I would not hold the PR for it. The zod/valibot question is genuinely the owner's, and the way you framed it — a measured bundle delta rather than a preference — is the right way to ask.

**One thing to raise, not a defect.** `checked()` is dev-only (`bridge.ts:71`) and `listenMessaging` parses only under `import.meta.env.DEV` (`events.ts`). Both are disclosed in comments, so this is a judgment call you have already made and not something hidden. But it is worth naming: the symptom the file opens by describing — *"a renamed engine field ships as `undefined` and renders nothing until a user sees a blank row"* — is a production symptom, and in a production build neither guard runs. The parity test covers mock mode, so it cannot see this either. If the bundle cost is the reason, it may be worth parsing in production for the small, low-frequency responses (`getEngineStatus`, `getEnrollmentStatus`, the event payloads) and casting only on the hot list paths. Worth a sentence in the PR body among the deliberate choices, since it is the same "parse, don't cast" thesis being spent differently in the shipped artefact.

Nothing above blocks except the navigation test.

Jhonattan2121:
All three addressed in `99abc54`.

**The navigation test.** You are right, and it is the test doing its job — a new top-level destination in a wallet should not land without someone re-reading that list. Both expectations updated with the position documented: desktop `tools` order 1, directly after AI and before the money group; mobile `more` order 0, signed-in only.

One thing that fix surfaced, worth your call rather than mine: `articles` already holds `mobile: { area: "more", order: 0 }`, and §2.4 places messaging at order 0 too. They tie, and declaration order decides — deterministic, but implicit. I left `articles` where it is because moving it is a product decision this document does not authorize. If messaging is meant to be first in More, the renumber is yours to approve.

**The parity loop.** This is the sharpest finding in the review and you are right that it could not fail against the contract. Added the literal anchors with the reasoning in a comment. I did not go as far as parsing `CLIENT-CONTRACT.md` — the counts are the cheap floor and a set comparison is the better version, but it is a bigger piece of work than this PR should carry.

**Dev-only parsing.** Also right, and it was the wrong place to spend the "parse, don't cast" thesis. `get_engine_status`, `get_device_info`, the enrollment trio and every event payload are now parsed in all builds; the list paths stay cast.

**On process.** Fair, and taken. I listed green subsets and none of them was `npm test`.

Running the full suite turned up something you may not have seen: `origin/main` alone fails 15 tests on my machine — `TopBar.signed-in`, `article-tags-contract`, `donation-http`, `private-live-contract` — and my branch fails the same 15 and no others. You saw 2 on yours, so it is environmental; this machine is `pt-BR`, which is the shape of #449. Happy to open an issue with the output if that is useful.

skyl:
Re-ran both halves of my reproduction against the new head `99abc546`.

**The gate failure is fixed, and fixed the right way.**

```
npm ci   => exit 0
npm test => Test Files  54 passed (54)
                 Tests  789 passed | 5 skipped (794)
```

More to the point, `messages` was placed rather than appended. Both expectations carry the reasoning — `{ area: "more", order: 0 }` from §2.4 with the tie against `articles` resolved by declaration order, and `tools` order 1 directly after `ai` and before the `money` group. That is the decision being recorded, which is what makes the list worth pinning in the first place.

**The count anchor is live** — flipping `46` to `47` fails `the implementation covers the contract` (1 failed | 96 passed). But it does not catch the case its own comment describes, and that is worth fixing before this lands:

> a command added to the contract and wired nowhere would pass all of them

`expect(bridgeMethods.length).toBe(46)` passes in exactly that case too. If §3 grows to 47 commands and the bridge stays at 46, the assertion reads `46 === 46` and goes green. The literal is a tripwire on **implementation shrinkage** — someone deleting a bridge method — not on contract growth, which is the direction the comment is about.

That is my fault for offering the cheap floor and the real thing as equal options; the floor is worth having, and the comment above it now claims the ceiling. Two ways out, either is fine:

1. **Reword it.** Say plainly that the number is the contract's, maintained by hand, and that nothing checks it against the document — same honesty as `f2z-assert.rs`'s help text, which is #657 and has the identical shape: a hand-copied constant, correct today, with nothing that would notice it drifting.
2. **Parse the document.** `CLIENT-CONTRACT.md` §3's command tables are regular (`| \`command\` | args | Returns |`) and §5.1's event table likewise, so extracting both and comparing sets is a dozen lines. I did exactly that while reviewing — spec 46 / impl 46, `MISSING []` and `EXTRA []` — with the app-crate trio mapped `f2zmsg_enroll`/`f2zmsg_enrollment_status`/`f2zmsg_unenroll` → `enroll`/`getEnrollmentStatus`/`unenroll`. That mapping is the only fiddly part and it is three entries.

**On the production parse — this is better than what I suggested.** Naming the always-parsed set and keeping the hot list paths cast is the right split, and moving events to parse unconditionally is right because the payloads are small and the symptom is a silently blank row.

One thing to watch, not blocking: `ALWAYS_PARSED` is a fourth hand-maintained list of command names, and it is the only one in the module written in **wire** names (`"get_engine_status"`) rather than bridge names (`getEngineStatus`). I checked it — all seven entries correspond to commands actually invoked, no dead strings today. But it is an untyped `Set<string>`: rename a wire command and that command silently stops being parsed in production, with no test able to see it, because the parity test runs against the mock and never reaches `checked()`. Typing it as `Set<WireCommand>` — or deriving it from the same place the `invoke` call sites get their names — would make a rename a compile error instead of a silent downgrade.

Nothing here is merge-blocking. The suite is green on this head; I have not merged because the count-anchor comment is a two-line edit and it is cheaper to correct now than to file.

skyl:
**Correction to my last comment: I was wrong that the suite is green on `99abc546`. It is not, and I misread my own run.**

What I reported as a clean result was the *vitest* summary (`Test Files 54 passed (54)`). The same `npm test` invocation then runs Playwright, and my local run finished `TEST_EXIT=1` with 110 Playwright failures. I quoted the sub-summary and missed the exit code — which is precisely the "a list of green subsets reads like a green suite" mistake I raised against the PR description two comments ago. Apologies for the noise.

Most of those 110 were environmental on my machine and not usable as signal. **CI's are.** `zuuli / frontend` step 15 "Test frontend contracts" on `99abc546`:

```
2 failed
  tests/navigation.pw.ts:81:5 › signed in five-item navigation works at 320px
  tests/navigation.pw.ts:81:5 › signed in five-item navigation works at 360px
130 passed (7.4m)
```

**The navigation population is pinned in two places, and only one was updated.**

`src/components/layout/navigation.test.ts` enumerates by **id** (`"articles"`, `"messages"`, `"profile"`, …) and is now correct. `tests/navigation.pw.ts:132-136` enumerates the same More-sheet population by **visible label**, through the rendered dialog:

```js
const expectedNames = signedIn
  ? ["Articles", "Edit profile", "Revenue share"]
  : ["Articles", "Log in"];
await expect(rows).toHaveCount(expectedNames.length);
```

Three names, four links now. `"Messages"` belongs in the signed-in arm, between `"Articles"` and `"Edit profile"` — the same position and the same tie-break you already justified from §2.4, just expressed as `label` rather than `id`.

Two details that confirm the diagnosis rather than merely fitting it: **only the signed-in cases fail**, and the signed-out arm (`["Articles", "Log in"]`) is untouched — which is exactly what `auth: "signed-in"` on the entry predicts. And both failing widths are the 320px/360px mobile viewports, where `mobile: { area: "more", order: 0 }` puts it in the More sheet rather than the primary bar.

Worth noting for its own sake: this is the *second* independent anchor on the same population catching the same omission, one at the config layer and one through the rendered DOM. That redundancy is doing real work here — the id list and the label list can drift apart, and the browser test is the one that knows what a user actually sees.

I have not merged. `rs / gate` is green; `gate` will need the frontend job, and that needs this line.

Jhonattan2121:
Both in `c966083`.

**The count anchor.** You were right and the comment was the problem, so I took the second option rather than reword it. `scripts/messaging-contract.node-test.mjs` reads §3's command tables and §5.1's event table out of the document, reads `WIRE_COMMANDS` and `EVENTS` out of the source, and compares sets. Wired into `npm test`.

It is in `scripts/` and not `src/` for a reason worth recording: I wrote it as `src/lib/messaging/contract.test.ts` first and `tsc` rejected `node:fs` — not in the frontend tsconfig's scope, and `@types/node` is not installed. Nothing under `src/` reads a file in this repo; all sixteen source-scanning tests are `scripts/*.node-test.mjs`. Following that seemed better than adding a types dependency to make one test compile.

Proven by mutation rather than asserted. A 47th row in §3 fails with `missing: ['probe_engine_health']`; renaming `get_engine_status` in `WIRE_COMMANDS` fails with both a missing and an extra; both reverted, `pass 3 fail 0`, and `git status docs/` is clean.

The count assertion stays, with the comment now saying what it actually covers — a tripwire on the module shrinking, and explicitly not on the document growing.

**`ALWAYS_PARSED`.** Also right, and the fix generalises. `WIRE_COMMANDS` is now the module's single population: `BridgeMethod` is its keys, `RESULTS` is `satisfies Record<BridgeMethod, …>`, `ALWAYS_PARSED` is `ReadonlySet<BridgeMethod>`, and both invoke helpers take the method key and resolve the wire name themselves. A renamed command is a compile error in every consumer, and the contract test catches a rename that compiles because the document disagrees.

Full suite on this head: 15 failures, identical to `origin/main` alone on this machine, and `navigation.test.ts` green.

skyl:
Checked `c9660830`. **Both of the points from my earlier comments are properly closed, and the contract test is the real thing rather than the floor I offered.**

`scripts/messaging-contract.node-test.mjs` reads §3 and §5.1 and compares sets. I tested it the only way worth testing it — by growing the document, not the module. Adding a 47th command to a §3 table:

```
| `get_engine_status` | — | `EngineStatus` |
+| `pause_engine` | — | `EngineStatus` |
```

```
not ok 2 - the bridge implements exactly the commands §3 declares
    missing: [ 'pause_engine' ]
# pass 2  # fail 1
```

That is the exact direction the count anchor could not see, and it now fails until someone wires it. Clean on the unmutated tree (3/3 pass), and it is in `npm test`, so the gate carries it.

The count anchor's comment is now honest about being one-directional, which is the other half of what I asked for. Keeping both is right: the literal catches deletion, the document parse catches growth, and neither subsumes the other.

**`WIRE_COMMANDS` is a better fix than the one I suggested.** I proposed typing `ALWAYS_PARSED` as `Set<WireCommand>`; collapsing the wire names into one exported map with `BridgeMethod = keyof typeof WIRE_COMMANDS` and `ALWAYS_PARSED: ReadonlySet<BridgeMethod>` removes the second namespace entirely rather than type-checking across it. One population, camelCase everywhere, and the wire string appears once per command.

**Still outstanding: `tests/navigation.pw.ts`.** This head does not touch it, and it is what `zuuli / frontend` is failing on — my previous comment has the detail, but briefly: `navigation.pw.ts:132-136` pins the same More-sheet population by visible label rather than by id, and needs `"Messages"` in the signed-in arm between `"Articles"` and `"Edit profile"`. The entry declares `label: "Messages"` and `accessibleLabel: "Encrypted messages"`, so the name the `getByRole("link", { name })` lookup wants is `"Messages"`.

**One small thing:** the new comment in `parity.test.ts` points at `contract.test.ts`, and the file is `scripts/messaging-contract.node-test.mjs`. A cross-reference that does not resolve is worse than none — somebody will grep for it and conclude the coverage was dropped.

**And a correction I owe you on process, not code.** My "the suite is green on this head" two comments ago came from reading the vitest summary and missing that the same `npm test` goes on to run Playwright. That is the same subset-for-suite error I raised about the description, and I made it while raising it. I have stopped using my local Playwright run as evidence at all — it produces ~110 environmental failures on this machine and cannot distinguish a real failure from its own noise. CI's frontend job is the measurement for anything Playwright-shaped; the vitest and `node --test` halves I can and do run locally.

skyl:
Reviewed and merging. This tracks the contract closely, and two things in it are better than what I would have asked for.

**The parity test is mechanical, which the wallet contract's equivalent is not.** `Object.keys(mockMessaging).sort()` deep-equalling the bridge's methods — "no dead, no uncovered" — plus a `RESULTS` schema manifest and sample args per method, catches the class `tsc --noEmit` cannot: a mock that returns the wrong *shape*. `invoke<T>()` is an unchecked cast, so the compiler is being lied to at exactly that boundary. Good.

**`BrowserGuarantee` is correctly scoped and honestly worded.** Gated on `platform !== "browser"`, so it never appears in ZUULI native — which matters, because showing a browser caveat in the signed app would misstate the guarantee in the *other* direction. The copy matches ADR 0001 and `THREAT-MODEL.md` §3.6 without overclaiming, and it carries the no-ACK durability notice from §11.2, which is the subtle part most implementations would drop.

Your audit is accurate, including the part I most wanted to see flagged: **rule 9's pin-contradiction alarm has nothing to violate yet**, because no screen calls `resolveHandle`. Noting an obligation that is not yet exercised, rather than quietly counting it as covered, is the right instinct.

## Yes — keep going on the screens

The seven command groups with no UI are the right next scope, and they are genuinely yours. Two things to know while you work, both from the contract rather than preference:

**1. The mock must drive *inbound* traffic, via an echo peer.** Without it, the delivery-state machine (§6.3's four states), the gap-repair screen (§3.5) and the alarm surface (§3.10) get built blind — there is no way to see a message arrive, a state advance, or an alarm fire. `CLIENT-CONTRACT.md` specifies a scenario switch mirroring `zuuli.mock.wallet-scenario`; the failure scenarios (`gap`, `unrecoverable-gap`, `kt-key-change`, `insufficient-witnesses`, `relay-down`) are what make those screens designable at all. If the mock does not already do this, it is worth doing before the screens rather than after.

**2. You are in mock mode until we ship the Tauri plugin, and that is expected.** `bridge.ts` invokes `plugin:f2zmsg|…`, which does not exist yet. The Rust side — the MLS engine, seed-derived identity, the message DAG, and the plugin that bridges them — is ours and is starting now. Nothing you build is blocked on it, and the two-runtime-mode split is exactly what lets these proceed in parallel. Real mode will simply start working.

## Two things you should know about, because they affect the UI

**Not every account can have a messaging handle.** Handles are `[a-z0-9_]{1,30}`, ASCII, and roughly 10% of existing accounts do not qualify — uppercase folds fine, but `.`, `@`, `+`, `-` and over-30-characters do not. **Never derive a handle locally from a username**; always call `check_handle_eligibility`, which is callable before enrollment and before the engine runs precisely so the UI can decide what to render without provoking a failure. Whether excluded users get an opt-in separate handle is still an open product question.

**Absence is not proved.** `CLIENT-CONTRACT.md` §3.10 previously shipped `found: boolean // false is a PROVED non-membership`. That was wrong and has been corrected — the underlying library cannot produce an absence proof, so a "no such handle" answer is an *assertion* labelled unproved on the wire. If you are on a copy from before that correction, re-pull. The UI must not present "no such user" as a verified fact, and the normative client rules for a pin contradiction (fail closed, alarm, and be honest that the contradiction cannot be proved to a third party) are in the same section.

Merging as-is. Ping me when the screens are up, or sooner if a contract section is ambiguous — several have been corrected already because an implementer pushed back, and that is the process working.

skyl:
@Jhonattan2121 — session handoff, and two things that affect your next screens.

Full state: https://github.com/free2z/zuu/issues/305#issuecomment-5448003049

## The backend is built

Since your PR merged: the MLS engine, seed-derived identity, the DAG, the key-transparency client, and `tauri-plugin-f2zmsg` all landed, and **ZUULI links the plugin** (42 grants in `mobile.json`). Two separate processes now exchange real MLS messages over a real relay in CI. So real mode is close — but see the blockers at the bottom.

## Your `compareMessages` had a real bug, and the fix went somewhere unexpected

Building §7's ordering independently in Rust surfaced it: **`compareMessages` never read `parents`.** It implemented §7's *tie-break*, not its order. Concretely — Bob (leaf 1) sends `A`; Alice (leaf 0) replies `B` with `parents=[A]`, same epoch; keys `(7,1)` vs `(7,0)` put **`B` above `A`**. A reply above the message it answers, in roughly half of all 1:1 conversations.

**This is not a criticism of your work** — the tie-break itself agreed with Rust field-for-field, including `msgId` as single-case hex, which matters because base64 would break the comparison across 255 adjacent byte pairs. It was a scope bug, and the same mistake was independently made **twice more**: once in the plugin's Rust, once in a third place. Three implementations, all wrong the same way.

**And the fix isn't a better comparator.** A JS `sort` comparator *cannot* express a topological order — `Array.prototype.sort` needs a total order over pairs, and causal precedence is a property of the whole graph. So the architecture changed instead: **the engine now owns display order.** `CLIENT-CONTRACT.md`'s line saying display order is *"recomputed by the UI, always"* was the root cause — it mandated a second implementation of a normative protocol rule in another language, which ADR 0001 exists to prevent.

**What this means for you:** `list_messages` returns messages **already in §7 order**; the UI renders what it is given; `compareMessages` is deleted. On `message-received`, **re-read the visible window** — do not sort locally, and do not expect an insertion index. There is no insertion index deliberately: filling a gap can **reorder rows already on screen** (hold `A` and `C` with `C`'s parent missing; when `B` arrives, `A → B → C` forces a swap), and one index cannot express that. It costs one IPC round-trip per event, which §5.2 already required anyway.

## Two things to know for the remaining screens

**Not every account can have a handle.** Handles are `[a-z0-9_]{1,30}`, ASCII. Roughly **10% of existing accounts do not qualify** — uppercase folds fine, but `.`, `@`, `+`, `-` and over-30-characters do not, and every excluded account belongs to someone who has actually logged in. **Never derive a handle locally**; always call `check_handle_eligibility`, which is callable *before* enrollment so the UI can decide what to render without provoking a failure.

**Absence is not proved.** §3.10 once shipped `found: boolean // false is a PROVED non-membership`. That was wrong — the underlying library cannot produce an absence proof — and it is corrected. A "no such handle" answer is an **assertion labelled unproved**. The UI must not present it as a verified fact, and there is a new `AlarmKind` for the pinned-handle-asserted-absent case.

## What still blocks real mode

- **#769** — KeyPackage publication, currently conflicting with no CI. Until it lands, two people can find each other and one can accept the other, but **neither can open a conversation**.
- **#770** — the shipped witness set and threshold `t` are undecided, so the plugin ships `NoDirectory` and `resolve_handle` fails closed. That is an owner decision, not code.

Rule 9's pin-contradiction alarm — which you correctly flagged as having nothing to violate yet — becomes live once §3.3 has a screen and #770 is settled. Your instinct to note it rather than count it as covered was right.

skyl:
## Heads-up: the messaging client is moving to its own app (#904)

@Jhonattan2121 — flagging this on your PR since it built the messaging client contract and the messages surface, and that work is directly in scope of an architecture change.

**Your work is not being discarded or rewritten** — it moves. ZUULI is being split into three apps:

- `cash.free2z.zuuli` — wallet authority, holds the seed, renders no third-party content
- `cash.free2z.free2z` — articles, livestreams, AI, creator
- `cash.free2z.e2e2z` — **messaging**, holding only device keys and a credential

**Why this is possible without moving the seed:** `docs/e2ee/ARCHITECTURE.md` §4.2 already separates account keys (seed-derived) from device keys (CSPRNG, never seed-derived, never exported), bound by a `DeviceCredential` that is already implemented in `rs/crates/f2z-msg-identity/src/credential.rs`. So ongoing messaging never needs the seed — only enrollment does, and that becomes a one-shot bridge call.

**What changes for the messages surface you built:** it runs in `e2e2z` rather than inside the wallet app, and `f2zmsg_enroll` stops being an in-process call to `account_keys()` and becomes a credential request to ZUULI over a verified deep link. The client contract itself, the message list, the conversation surface — all still stand.

If you want to run with `e2e2z`, the two issues to start from are **#905** (the intent bridge — the keystone, and the thing everything else depends on) and **#906** (scaffolding the apps). The driver for all of it is **#367**.

PR #665, opened 2026-08-24
fix(e2ee): the relay cannot be probed in Kubernetes, and the witness restarts on someone else's outage

### LINKED ISSUE #677: f2z-relay is unprobeable in Kubernetes — it needs a healthz subcommand, not a wider admin listener
Found by running the published image against its Kubernetes manifest (see the deployment repo's activation measurements). This is a **gap in the binary**, not in the manifest.

## The problem

`f2z-relay` is **unprobeable in Kubernetes**, and the reason is a deliberate security property we should keep.

- Port 8080 speaks the WebSocket protocol at `/relay/v1`. A plain `GET /healthz` there returns an empty reply — it is not an HTTP health surface.
- `/healthz` and `/metrics` live on the **admin listener**, and the binary **refuses to bind that listener off loopback**. That refusal is correct: `WIRE.md` and `docs/e2ee/ARCHITECTURE.md` require that operational metadata never be publicly reachable, and an unconsidered metrics endpoint is a metadata leak in a system whose threat model turns on the relay revealing nothing.
- The kubelet probes the **pod IP**, not loopback. So an `httpGet` probe cannot reach the admin listener by construction.
- The image is `distroless/cc`, so there is **no shell, no wget and no curl** for an `exec` probe to use.

The result is a workload that cannot express liveness or readiness to Kubernetes at all, while every one of those individual decisions is right.

## Do not fix this by widening the admin listener

Binding the admin port to `0.0.0.0` would make `/healthz` and `/metrics` reachable from anywhere in the cluster. **That trades a deployment inconvenience for the exact metadata exposure the design forbids**, and it would do so quietly. Reject that direction explicitly so it is not proposed again later.

## Proposed fix — a `healthz` subcommand

`f2z-witness` already solves this shape: `f2z-witness healthz --state <path>` is a subcommand the kubelet can `exec`, needing no listener and no shell. Give `f2z-relay` the same:

```
f2z-relay healthz [--admin-addr 127.0.0.1:PORT]
```

It runs inside the container, connects over **loopback** to the admin listener, and exits 0/non-zero. The admin listener stays loopback-only; Kubernetes gets a probe; nothing new is exposed.

Design points to settle deliberately:

- **What "healthy" means.** Follow the witness's precedent of distinct, documented verdicts rather than a bare boolean — an operator reading a probe failure should learn *which* failure. Keep it cheap: `docs/k8s/f2z-servers.md` forbids probes that depend on an upstream, and the existing `/healthz` is deliberately constant and does **not** report queue depth or connection counts. The subcommand must not report more than the endpoint does.
- **It must not become a metadata side channel.** The exit code and any stderr are visible in pod events and logs. `THREAT-MODEL.md` and the relay's own logging rules forbid a queue address, payload, key or IP reaching a log line — that applies here too.
- **Startup semantics.** A relay that has not finished opening its store is not yet healthy but is not failed. Note `f2z-witness` exits 1 by design before its first poll, which is precisely why its manifest needs a startup probe; make sure the relay's answer is consistent and that the manifest side knows which verdict to expect when.

## Also in scope: an env var that kills the process

The manifest sets `F2Z_RELAY_DATA_DIR`, and the binary **hard-errors at boot**:

```
f2z-relay: environment: F2Z_RELAY_DATA_DIR names no configuration key
```

Decide which side is wrong and fix that one. Failing closed on an unknown `F2Z_RELAY_*` variable is defensible — a typo'd security-relevant setting silently ignored is worse than a refusal to start — but if that is the intent it should be documented as intentional, with the accepted key names discoverable from `--help` or `--print-config`. If the binary simply lacks a key it should have, add it.

## Acceptance

- [ ] `f2z-relay` can express liveness/readiness to Kubernetes **without** the admin listener leaving loopback.
- [ ] Verified by **running the published container** and executing the probe argv — not by reasoning. This gap existed because the manifest and the binary were written from the same document and never run against each other.
- [ ] The `F2Z_RELAY_DATA_DIR` mismatch resolved on whichever side is wrong, with the reasoning recorded.
- [ ] No new information is exposed: no listener widened, nothing added to logs or exit output that the existing `/healthz` does not already reveal.
- [ ] Coordinated with the manifest change in the deployment repo; each PR names the other.

skyl:
Closes free2z/zuu#677.

> **On #677's proposed fix, stated up front because this PR does something different.** #677 asks for a `healthz` subcommand and says — correctly — not to widen the admin listener. **The admin listener is not widened here**; it is byte-for-byte unchanged and a new test restates its loopback refusal. But a subcommand satisfies the *kubelet* and does nothing for the *GCE load balancer*, which is a separate blocking item on the activation epic (free2z/tuzi#1401: *"Even a working kubelet probe leaves a `Ready` pod behind a 502"*). So this adds a **health-only listener** that serves `/healthz` alone — `/metrics` on it is a 404 — which closes both halves. Full reasoning, and the evidence, in [the comment on #677](https://github.com/free2z/zuu/issues/677#issuecomment-5402527873). If the subcommand is still wanted it is a small addition, but it cannot substitute.

---

## What this is

The k8s manifests for `f2z-relay`, `f2z-kt` and `f2z-witness` (in the private
tuzi repo) were written **in parallel with these binaries and before any image
existed**, from the same design document. Nobody ran one against the other until
tuzi#1396, which found the manifests describe an interface the binaries do not
have — activating would have replaced three `ErrImagePull` pods with three
`CrashLoopBackOff` pods on stage.

Most of that is manifest bugs and is fixed on the tuzi side. **Two of them are
real gaps in these binaries**, and they are this PR.

## 1. `f2z-relay` cannot be probed in Kubernetes at all

`/healthz` lives on the admin listener, which **refuses to bind off loopback**
because `/metrics` is on it too. That refusal is correct and is unchanged here —
`/metrics` reachable off-host is exactly the metadata disclosure the component
exists to refuse.

But it left a workload Kubernetes cannot run:

| | |
| --- | --- |
| `httpGet` probe | kubelet dials the **pod IP** → loopback listener unreachable |
| GCE health check | load balancer dials the **pod IP** (NEG-backed Service) → same |
| `exec` probe | image is distroless: no shell, no `wget`, no `curl` |
| `healthz` subcommand | satisfies the kubelet, does **nothing** for the load balancer |

That last row is the trap. A subcommand makes the pod go `Ready` while every
NEG endpoint is unhealthy and `relay.<prefix>free2z.cash` serves 502 — a state
that reads healthy in `kubectl get pod` and is not.

**So: a second listener that serves `/healthz` alone.**

```
--health-listen ADDR     /healthz ONLY, any address    [health.address]
--no-health              Do not serve it at all        [health.enabled]
```

Off by default, so no existing configuration changes. `F2Z_RELAY_HEALTH_ENABLED`
/ `F2Z_RELAY_HEALTH_ADDRESS` for the environment layer. `/metrics` on it is a
**404**, not a 403 — a 403 confirms the endpoint is there and merely refused,
which is one bit more than a listener whose whole claim is that it says nothing.

`Config::check` refuses a health address that collides with the protocol
listener or with the admin listener; the second collision would put `/metrics`
wherever `/healthz` is.

### Why not on the protocol port

This was the tempting option and I want the reasoning on the record, because
the usual argument against it is weaker than it looks: **port 8080 already runs
a full HTTP request parse** (`tokio_tungstenite::accept_hdr_async`) on
unauthenticated input, for the WebSocket handshake. §2.2's "a second transport
is a second parser" has already been spent there.

It is rejected for a different reason: **§13.1 layer 1's connection guard sits
on that port.** A health check answered there consumes a connection permit and
can be *refused* (`Refusal::Backpressure`, `ConnectingTooFast`) under load — so
a traffic spike would fail the kubelet's liveness probe and restart the only
replica. A separate listener cannot do that. It also keeps GCE's documented
rule easy to satisfy: the health-check port must name a `containerPort` of a
serving pod, which 8081 is.

## 2. `f2z-witness healthz` makes someone else's outage into a restart loop

`STALE` (the log is unreachable) and `UNPINNED` (it never was) both exit 1, and
both are statements about the **upstream log**, not about this process. Wired to
a `livenessProbe` — which is what the manifest does — that restarts the witness
on a schedule for as long as KT, DNS, TLS or egress is down, *during exactly the
incident in which its state file is the evidence a human needs*. The startup
probe asks the same question, so it cannot pass either: a ~20-minute restart
cycle for the duration of somebody else's outage. A longer `--stale-after`
delays that and cannot remove it, because any finite window expires.

```
f2z-witness healthz --state FILE [--liveness yes]
```

Passes on `Following`, `Stale` and `Unpinned`; still fails on `Halted`, so a
halted witness is visible as `0/1` rather than quietly reported as fine. The
verdict text is printed unchanged either way — a liveness probe that passes on
`STALE` must still **say** `STALE`, or the one place an operator reads the
witness's own account of itself lies by omission.

Takes an explicit `yes` for the reason `--allow-cleartext` does.

## Verification — against running containers, not only `cargo test`

This gap existed because both sides were written from one document and neither
was run against the other, so "I reasoned about it" is not an acceptable
standard here.

`scripts/check-f2z-images.mjs` gains `--probe-relay IMAGE`, wired into
`f2z-images.yml` for both the PR build and the publish job. It starts the relay
with **exactly** the configuration the Deployment sets and asserts three things:

* `/healthz` → 200 `ok\n` on the health port;
* `/metrics` → 404 there;
* `/healthz` is **not** answerable on the protocol port (the measurement that
  started all of this, pinned so a future "simplification" has to change the
  line and read why).

`--probe` gains the witness's `--liveness yes` contract.

Both were checked in both directions:

```
$ node scripts/check-f2z-images.mjs --probe-relay ghcr.io/free2z/f2z-relay@sha256:77957ea…
…: the relay exited instead of serving: "f2z-relay: unknown flag --health-listen; try --help"
$ node scripts/check-f2z-images.mjs --probe-relay f2z-relay:health-local
f2z-relay:health-local: a Kubernetes probe can reach /healthz.

$ node scripts/check-f2z-images.mjs --probe ghcr.io/free2z/f2z-witness@sha256:14b8f9a…
…: `healthz --liveness yes` on a witness that has never polled exited 1; UNPINNED is a
   statement about the LOG and must not restart this pod
$ node scripts/check-f2z-images.mjs --probe f2z-witness:liveness-local
f2z-witness:liveness-local: the f2z-witness exec probes work.
```

And the manifest's environment block, verbatim, against a locally built image:

```
$ docker run --rm -e F2Z_RELAY_LISTEN_ADDRESS=0.0.0.0:8080 -e F2Z_RELAY_LISTEN_INSECURE=true \
    -e F2Z_RELAY_STORE_PATH=/data/relay.sqlite -e F2Z_RELAY_IDENTITY_PATH=/data/identity.key \
    -e F2Z_RELAY_HEALTH_ENABLED=true -e F2Z_RELAY_HEALTH_ADDRESS=0.0.0.0:8081 \
    f2z-relay:health-local --print-config
…
[health]
enabled = true
address = "0.0.0.0:8081"
…
$ curl -o /dev/null -w '%{http_code}' http://<pod>:8081/healthz   → 200
$ curl -o /dev/null -w '%{http_code}' http://<pod>:8081/metrics   → 404
```

The check uses `node:http` with `agent: false` rather than `fetch`. The
operational listener answers **one request per connection and closes**, and
`fetch`'s socket pool reuses the closed socket — the second request failed about
one run in five. A flaky check is worse than no check; found by running it six
times rather than once.

Also: `--rm` is deliberately **not** passed to the probe container, so a relay
that exits on its own arguments still has readable logs. That is the failure an
activation PR is most likely to hit (a digest pinned from a commit that predates
`--health-listen`), and "docker did not publish both ports" would have sent
somebody looking at Docker networking instead of at the digest.

New integration test `rs/crates/f2z-relay/tests/health.rs`, including one that
pins the exact Kubernetes environment block so the two repos cannot drift
silently, and one that restates the admin listener's loopback refusal so a
future edit that "harmonises" the two listeners fails a test that says why.

Local gate: `check-rust-fmt.sh`, `check-rust-clippy.sh` (10 crates clean),
`check-rust-deny.sh`, `cargo test --locked --all-targets` (29 test binaries,
0 failures), `cargo test --locked --doc`, `check-f2z-images.mjs --self-test`
(19 mutations rejected), `check-workflow-gates.mjs`,
`check-github-actions-pins.mjs`, `check-hash-domain-labels.mjs`.

## Dependency

**Nothing consumes this until the images are republished.** The companion
manifest PR — **free2z/tuzi#1399** (private) — corrects every mismatch on that side and explicitly
records that it needs a digest built from a commit containing this change —
against the currently published digests the relay would exit on an unknown flag
and the witness's probe would fail permanently. Both manifests stay unpinned and
out of `//:deploy` regardless, so nothing deploys either way.

## Related, found in the same exercise and filed rather than fixed here

* **#669** — `f2z-kt`'s `/kt/v1/cosign` is fail-open with no `witness_pk`: any key's cosignature is accepted, journalled and replayed. Measured with four unrelated witness keys against a real log. Its comment thread also carries the second half — the rate limiter is global and is spent before the body is decoded, so the same endpoint can be starved as well as stuffed.
* **#671** — the health listener this PR adds is one the GCE load balancer will now believe, and `Server::start`'s tasks are unsupervised: a panic in the protocol listener leaves a process that still answers `/healthz`. The fix is supervision, not a deeper probe — a deep probe would reintroduce exactly the failure the shallow one was chosen to avoid.


skyl:
Reviewed against the exact head `7ef6d90f`, then merged — both `gate` and `rs / gate` were green on it. Baseline `cargo test -p f2z-relay -p f2z-witness`: **EXIT=0, 164 passed**.

Both diagnoses hold up, and the second one is the kind of bug that only shows up when somebody actually runs the two halves against each other. Restarting the observer does not fix the observed, and doing it on a schedule during a KT outage destroys the state file that is the evidence — that reasoning is right and the fix matches it.

**One finding, filed as #666. Latent, not blocking, which is why I merged.**

`is_alive` is written as `!matches!(self, Self::Halted { .. })` while `is_healthy` is written as `matches!(self, Self::Following { .. })`. The two default in opposite directions for a variant that does not exist yet: `is_healthy` is fail-closed, `is_alive` is fail-open. Adding a fifth local fault to `Health` makes it report alive.

Adding one:

```
error[E0004]: non-exhaustive patterns: `&Health::EvidenceUnreadable` not covered
   --> crates/f2z-witness/src/health.rs:139:15
```

Line 139 is `message()`. The only exhaustive match on `Health` is the one that renders a **string**, so the compiler asks the author for display text and never asks whether the pod should live. Adding just that arm:

```
PROBE is_healthy (readiness) => false
PROBE is_alive   (liveness)  => true
PROBE exit_code=1 liveness_exit_code=0

cargo test -p f2z-witness -p f2z-relay  =>  EXIT=0
```

A new local fault fails readiness for free and passes liveness, and nothing in the suite moves. Fix is to write `is_alive` as an exhaustive match so the fifth variant is a compile error *there* — at the predicate that frames the question better than any test could.

`a_liveness_probe_ignores_the_upstream_log_and_not_a_local_fault` is a good test, incidentally — all four variants, both exit codes each. It just enumerates by hand, so it grows only when someone remembers.

**What I attacked and could not break**, since the useful half of a review is what survived:

- `the_health_listener_does_not_serve_metrics` asserts 404 on the health listener **and 200 on the admin listener** in the same test. That control is what makes the 404 mean "scoped" rather than "metrics is broken" — the same shape as #626's forgery fixture.
- The collision test covers both the protocol and the admin listener, which is the one that matters.
- I checked the `health.ip() == listen.ip()` exact comparison for a `0.0.0.0:9101` vs `127.0.0.1:9101` overlap. It slips past the check, but fails at bind with `EADDRINUSE` rather than silently sharing a port, and the health-only surface 404s `/metrics` either way. Fail-closed. Not a defect.
- `route`'s catch-all is 404, so a future endpoint added without a scope guard is not accidentally served off-host.
- `--liveness` requires the literal `yes`, and the verdict text prints unchanged on both paths — a probe that passes on `STALE` still says `STALE`, which was the right call to make explicit.

The verification section deserves a note in the other direction from the one I gave #664: probing the actual container in both directions, and finding the `fetch` socket-reuse flake by running it six times instead of once, is the standard. "Found by running it six times rather than once" is the sentence that makes the rest of the section credible.

PR #667, opened 2026-08-24
test(kt): pin published epoch bounds

### LINKED ISSUE #657: ACCOUNT_EPOCH_CEILING can be changed with the whole suite green: three copies of a normative constant, no cross-check

skyl:
Closes #657

Pins the account-epoch ceiling to an independent literal contract, renders the CLI help from the production constant, and verifies the real process output against an exact golden, numeric census, and independently checked FNV-1a checksum. Both help and no-argument paths share the same fallible writer path, including the shipping wrapper.

Validation:
- `cargo test -p f2z-authority -p f2z-kt --all-targets`
- `cargo test -p f2z-authority -p f2z-kt --doc`
- rustfmt and all-target clippy
- independent exact-head review on `b6e56b337b8d47a539d78065c6d7bb549e3b05e6`

Mutation evidence (compiled and RED): production constant drift; step constant drift; formatter bypass; hard-coded template/wrapper branches; help/no-args routing divergence; swallowed writer errors; coordinated production+golden contradictions; split-line/spelled-out alternate limits; the prior numeric-reallocation survivor; FNV helper-prime corruption; and checksum-literal corruption.

### DIFF
--- a/rs/crates/f2z-authority/src/authority.rs
+++ b/rs/crates/f2z-authority/src/authority.rs
@@ -979,6 +979,15 @@ fn check_account_epoch(account_epoch: u32, previous: Option<u32>) -> Result<()> 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn the_published_bounds_are_the_ones_this_crate_applies() {
+        // KT.md §4.5.4 states these to clients as fixed constants. A client
+        // that hard-codes the specification and a log that applies something
+        // else disagree about admissibility, not merely about an encoding.
+        assert_eq!(ACCOUNT_EPOCH_CEILING, 1 << 20, "KT.md §4.5.4 A18a");
+        assert_eq!(MAX_ACCOUNT_EPOCH_STEP, 16, "KT.md §4.5.4 A18b");
+    }
     use crate::assertion::HandleAssertionTBS;
     use crate::key::SigningKey;
     use crate::nonce::NonceLedger;
--- a/rs/crates/f2z-authority/src/bin/f2z-assert.rs
+++ b/rs/crates/f2z-authority/src/bin/f2z-assert.rs
@@ -33,8 +33,8 @@ use std::process::ExitCode;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use f2z_authority::{
-    AssertionNonce, DEFAULT_MAX_VALIDITY_MS, Handle, HandleAssertionTBS, Intent, LogId, SigningKey,
-    authority_id,
+    ACCOUNT_EPOCH_CEILING, AssertionNonce, DEFAULT_MAX_VALIDITY_MS, Handle, HandleAssertionTBS,
+    Intent, LogId, SigningKey, authority_id,
 };
 use f2z_codec::canonical::Canonical as _;
 use f2z_codec::types::PublicKey;
@@ -69,7 +69,8 @@ NOTES:
   last-one forever, which deletes the rule that stops a reset assertion being
   spent twice on one account-ownership event. It defaults to 0 for --intent
   bind, which is the baseline a handle's first reset must exceed, and is
-  REQUIRED for --intent reset. A value at or above 1048576 is refused outright
+  REQUIRED for --intent reset. A value at or above \
+  {account_epoch_ceiling} is refused outright
   as a clock.
 
   --validity-ms defaults to 900000 (15 minutes) and is ignored when --expires-ms
@@ -77,9 +78,18 @@ NOTES:
   the log's own cap is refused however it was signed.
 ";
 
+fn usage_with_account_epoch_ceiling(account_epoch_ceiling: u32) -> String {
+    USAGE.replace(
+        "{account_epoch_ceiling}",
+        &account_epoch_ceiling.to_string(),
+    )
+}
+
 fn main() -> ExitCode {
     let args: Vec<String> = env::args().skip(1).collect();
-    match run(&args) {
+    let stdout = std::io::stdout();
+    let mut stdout = stdout.lock();
+    match run_published(&args, &mut stdout) {
         Ok(()) => ExitCode::SUCCESS,
         Err(message) => {
             eprintln!("f2z-assert: {message}");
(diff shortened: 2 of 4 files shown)


skyl:
Reviewed against the exact head `b6e56b33`, then merged — both `gate` and `rs / gate` were green on it. **#657 is closed, and I checked it by running my own reproduction verbatim rather than reading the diff.**

Baseline: `cargo test -p f2z-authority -p f2z-kt` → **EXIT=0, 182 passed**.

**Half one — the mutation from the issue.** `authority.rs`, `1 << 20` → `1 << 24`:

```
cargo test -p f2z-authority -p f2z-kt  =>  EXIT=101
  authority::tests::the_published_bounds_are_the_ones_this_crate_applies ... FAILED
```

On the pre-fix tree that same edit left EXIT=0 with all 176 green. That is the issue closed.

**Half two — the CLI, which was the other half of #657 and the part I care more about.** The single lib failure above is just cargo stopping after the first failing target, so I ran the integration test separately under the same mutation:

```
cargo test -p f2z-authority --test cli every_usage_path_publishes_the_library_epoch_ceiling  =>  EXIT=101
  left:  "… A value at or above 16777216 is refused outright as a clock."
  right: "… A value at or above 1048576  is refused outright as a clock."
```

That output is the proof I wanted. The *real binary* emitted `16777216` — the help text is genuinely rendered from `ACCOUNT_EPOCH_CEILING` and no longer hand-typed, so the CLI can no longer state a threshold the code does not apply. A diff read could not have distinguished this from a template that merely looks parameterised.

**Two things in the test design worth naming, because they are the reason it holds.**

The ceiling is injected — `run(args, account_epoch_ceiling, stdout)` with `run_published` supplying the real constant — so `every_dispatch_usage_path_formats_the_supplied_epoch_ceiling` can drive two different ceilings (42 and 2,000,000) through **both** the `--help` and the no-args paths without touching the constant. Testing the formatter and the production wiring separately is what makes each one falsifiable on its own.

And `threshold_lines == [expected_threshold]` asserts the line is *unique*, not merely present. A second hard-coded copy of that sentence elsewhere in `USAGE` would fail it. That is the assertion most people leave out.

**On the belt-and-braces.** The FNV checksum is redundant with `assert_eq!(stdout, EXPECTED_HELP)` in the normal case — if the strings are equal, the checksums are. It is not theatre, though, and it earns its place for a specific reason: it catches the *lazy* fix. Regenerating `fixtures/f2z-assert-help.txt` from the binary's output makes the golden pass again while the checksum literal and the `ascii_numbers` census both still fail, so a blind regeneration cannot quietly launder a changed constant. Same for the census, which pins `"1048576"` as a token. That is #643's "literals compared against derived values" applied to a golden file, and it is the right instinct.

Including the published FNV-1a test vectors (`fnv1a64(b"") == 0xcbf29ce484222325`, `fnv1a64(b"a") == 0xaf63dc4c8601ec8c`) before relying on the helper is the positive control that makes the checksum assertion mean something. Without it a broken helper would produce a stable wrong number and pass forever.

**One cosmetic note, nothing more.** `the_published_bounds_are_the_ones_this_crate_applies` is inserted between `use super::*;` and the rest of the `use` block, so the module now has an item in the middle of its imports. Valid Rust, and rustfmt evidently does not mind; it just reads oddly next to the tidiness of everything else here. Worth moving below the imports next time that file is touched — not worth a commit of its own.

PR #400, opened 2026-08-19
fix(zuuli): secure card checkout launch

skyl:
## Summary

- gate card checkout on the validated ZUULI session, showing a concise Sign in to buy action; successful classic, Zcash, and social login returns to Buy
- accept only the exact internal `/buy` post-login destination and default every absent, external, protocol-relative, query, fragment, encoded, object, throwing, or tampered value to `/`
- preserve that allowlisted destination across mobile social-OAuth cold starts without accepting arbitrary history or persisted state
- validate the backend hosted URL against HTTPS plus an exact hostname allowlist before opening it; `checkout.stripe.com` is always allowed and documented Stripe Custom Domains can be configured explicitly
- open packaged-app checkout through the Tauri OS opener, navigate directly on web, and distinguish authentication, transport, server, malformed/untrusted-link, and opener failures
- preserve the narrow-layout fixes from #397 and add a 320 × 568 Playwright flow covering signed-out Buy → login → authenticated return to Buy

## Verification

- `npx vitest run src/lib/auth/login-destination.test.ts src/features/buy/card-checkout.test.ts` — 39 passed
- `npm test` — 236 Vitest tests, 5 safe-area contract tests, and 9 Playwright viewport tests passed
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- semantic rebase onto `5943e7ad` (#397); lease-pushed from verified prior PR head `e61159a0`

## Adversarial review

The URL boundary uses exact normalized host equality and rejects suffixes, credentials, explicit ports (including `:443`), fragments, non-HTTPS schemes, whitespace, parser-control input, and malformed authorities. Login state accepts only the literal `/buy`; hostile, encoded, throwing, and tampered values resolve to home. The rebase did not reverse #397 layout changes.

Earlier reviews caught and fixed a browser false-failure from `window.open(..., noopener)` and a dead return contract where login methods ignored `returnTo`.

## Remaining backend/release dependencies

This advances #388 but does not close it and is not end-to-end Stripe completion:

- free2z/tuzi#1258 must provide the validated hosted Checkout URL contract consumed here.
- free2z/tuzi#1257 must provide server-controlled native success/cancel return and authoritative balance refresh.
- free2z/tuzi#1253 must provide authoritative exact-price, verified fulfillment, and exact-credit integrity.

Signed staging and store-build verification remain after those dependencies land.

PR #668, opened 2026-08-24
docs(kt): record the pinned akd audit evidence

### LINKED ISSUE #559: docs(e2ee): ADR 0013's audit-scope claim is disprovable — and the truth is a stronger argument

skyl:
Closes #559

Records the reproducible `akd` 0.13 audit evidence, pins the exact upstream/source/release identities behind the measurements, and wires self-testing/live verification into the Rust change-selection gate.

Validation:
- exact AKD evidence self-test/live check (32 observed red mutations)
- immutable issue-comment, NCC PDF, PR identity/patch, and release-source verification
- compiler-bound `akd::HistoryProof` negative on Rust 1.97.1
- action-pins, workflow-gates, image policy, and actionlint self/live matrices
- all affected tests/docs plus all 10 Rust-root fmt/clippy checks
- independent exact-head review of `518f72e4514ebcaef49077ca52ae16a214dc8941` on `c7a155481cc01336f171cc70667abd3fae73ce1a`

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -144,7 +144,7 @@ jobs:
           rs=false
           while IFS= read -r -d '' file; do
             case "$file" in
-              rs/*|scripts/check-rust-fmt.sh|scripts/check-rust-clippy.sh|scripts/check-rust-deny.sh|scripts/check-rust-toolchain.sh|.github/workflows/rs.yml)
+              rs/*|docs/e2ee/*|scripts/check-akd-doc-evidence.mjs|scripts/check-rust-fmt.sh|scripts/check-rust-clippy.sh|scripts/check-rust-deny.sh|scripts/check-rust-toolchain.sh|.github/workflows/rs.yml)
                 rs=true
                 ;;
             esac
@@ -285,6 +285,20 @@ jobs:
       - name: Run the rs/ doc tests
         working-directory: rs
         run: cargo test --locked --doc
+
+      # The normative AKD type, benchmark figures, and audit-scope correction
+      # share one falsifiable evidence record. Its self-test first compiles the
+      # invented HistoryProofV2 type and requires that compile to fail, mutates
+      # each displayed metric independently, coordinates false values across
+      # prose and JSON, re-points both cited fix PRs, aliases both releases to
+      # one source, and changes the real pinned partition body. Report evidence
+      # is parsed with Node alone, including when PATH contains no host tools.
+      # Only then is the current documentation trusted.
+      - name: Mutation-test AKD documentation evidence
+        run: node scripts/check-akd-doc-evidence.mjs --self-test
+
+      - name: Verify AKD documentation against locked executable evidence
+        run: node scripts/check-akd-doc-evidence.mjs
 
       # `crash-injection` is off by default because it compiles an `abort()`
       # into f2z-relay-store's commit path, so the default run above skips
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -2079,7 +2079,7 @@ and the directory is a different service from a relay with a different lifecycle
 | `GET` | `/kt/v1/sth/{epoch}` | that epoch's `SignedTreeHead` + cosignatures | anyone |
 | `GET` | `/kt/v1/audit?from={e0}&to={e1}` | `AppendOnlyProof` (akd protobuf, §9.4) + both tree heads | witnesses in practice (§10) |
 | `POST` | `/kt/v1/lookup` | `{handle}` → `DirectoryEntry` + `LookupProof` + tree head + cosignatures | anyone |
-| `POST` | `/kt/v1/history` | `{handle, params}` → `DirectoryEntry<>` + `HistoryProofV2` + tree head + cosignatures | anyone |
+| `POST` | `/kt/v1/history` | `{handle, params}` → `DirectoryEntry<>` + `HistoryProof` + tree head + cosignatures | anyone |
 | `POST` | `/kt/v1/submit` | `SubmissionEnvelope` → `SubmissionReceipt` | the handle's owner |
 | `POST` | `/kt/v1/cosign` | `WitnessCosignature` → empty | witnesses |
 
@@ -2249,9 +2249,17 @@ size the system.
 Scaling from 10,000 to 1,000,000 entries costs 1.3 KB of proof and no measurable
 time — the cost is dominated by three constant-cost ECVRF verifications.
 
-**Client verifier bundle**, size-optimised `cdylib`, `akd_core` +
-`vrf,whatsapp_v1,protobuf`: 118 KB raw, 47.9 KB gzip, **39.6 KB brotli**. At
-`opt-level=3`: 183 KB raw, 48.1 KB brotli — take `opt-level=3`.
+**Client verifier bundle**, `cdylib`, `akd_core` +
+`vrf,whatsapp_v1,protobuf`: the size-optimised profile produced
+**118,314 B**<!-- akd-metric:verifier_size_raw --> raw,
+**47,940 B**<!-- akd-metric:verifier_size_gzip --> gzip -9, and
+**39,631 B**<!-- akd-metric:verifier_size_brotli --> brotli -q11. The recommended
+`opt-level=3` profile produced
+**183,375 B**<!-- akd-metric:verifier_recommended_raw --> raw and
+**48,116 B**<!-- akd-metric:verifier_recommended_brotli --> brotli -q11. Take
+`opt-level=3`; the exact measurements and provenance are checked in at
+[`evidence/akd-benchmark.json`](./evidence/akd-benchmark.json) and mechanically
+bound to every published bundle-size value.
 
 **Append-only proof, 100,000-entry directory, 5 epoch transitions, ~1,000 new
 entries per epoch:**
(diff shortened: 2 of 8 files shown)


skyl:
Reviewed against `518f72e4`. Ran the checker both ways in a worktree:

```
node scripts/check-akd-doc-evidence.mjs --self-test
  → AKD documentation evidence self-test observed every mutant fail:
    {"displayedMetrics":8,"benchmarkArtifacts":5,"coordinatedEvidence":14,
     "claims":2,"pinnedSource":2,"compileApi":1}      SELFTEST_EXIT=0

node scripts/check-akd-doc-evidence.mjs
  → AKD documentation evidence verified                LIVE_EXIT=0
```

8+5+14+2+2+1 = 32, which is the number in the description.

**The evidence pinning is the right design and I want to say why before the finding.** Every external artefact is SHA-256 compared, not merely fetched: the NCC report bytes, the issue-comment *body*, both PR patches, and the raw upstream source at each pinned commit. "The bytes we audited are still the bytes served" is a different and much stronger claim than "the URL still resolves", and it is the one that makes a documentation claim checkable at all. `assert.notEqual(audit.upstream.audited.raw_sha256, audit.upstream.selected.raw_sha256)` is a good touch too — it stops a fixture in which the audited and shipped versions collapse to the same file, which would make the whole distinction vacuous.

**The finding is operational: this puts four third parties on the critical path of the `rs` gate, unauthenticated.**

`rs.yml` runs both steps unconditionally, and `rs=true` is set by `rs/*` **or** `docs/e2ee/*` — so essentially every Rust or spec PR now depends on:

| host | calls | via |
|---|---|---|
| `api.github.com` | 3 | the benchmark comment, `facebook/akd` PRs #400 and #495 |
| `www.nccgroup.com` | 1 | the report PDF |
| `github.com` | 2 | both `.patch` bodies |
| `raw.githubusercontent.com` | 2 | the two pinned release sources |

`fetchPinnedBytes`/`fetchPinnedJson` send only a `User-Agent`. There is no `Authorization`, no retry, and no `catch` anywhere in the file — a 30-second timeout and then `assert(response.ok, …)`.

The GitHub API limit for unauthenticated requests is 60 per hour per source IP. Measured just now rather than recalled:

```
$ curl -sSI -H "User-Agent: free2z-zuu-akd-evidence-check" \
    https://api.github.com/repos/facebook/akd/pulls/400
HTTP/2 200
x-ratelimit-limit: 60
x-ratelimit-resource: core
```

Hosted runners share egress addresses, so that 60 is not per-repository. And `--self-test` fetches too (`selfTest()` calls `fetchPinnedJson(benchmarkAnchor.apiUrl)`), so both steps are exposed, not just the live one.

**The sharpest part is the error taxonomy, not the flake rate.** A rate-limited response fails as:

```
unable to fetch pinned evidence https://api.github.com/... : HTTP 403
```

That is the same shape of failure as evidence that has actually changed, and it reads like tampering. The check exists to distinguish "the audited artefact was altered" from "everything is fine"; right now a third party's quota, outage or TLS hiccup lands in the first bucket. #665 made exactly this argument two PRs ago about `STALE` and `UNPINNED` — a statement about somebody else's availability must not be reported as a statement about us — and the same reasoning applies to a gate.

Three changes, in the order I would do them:

1. **Send the token when there is one.** `Authorization: Bearer ${process.env.GITHUB_TOKEN}` when set, omitted locally. Actions' built-in token raises the limit to 1,000/hour and removes the dominant failure mode. Nothing here reads private data, so the default read-only token is enough.
2. **Separate transport failure from digest mismatch.** A digest mismatch must stay a hard, loud failure — that is the whole point. An unreachable host is a different verdict and deserves a different message and a different decision. Whether it should block is yours; conflating them means it can never be decided.
3. **Retry transient failures** (timeout, 5xx, 429) a couple of times with a short backoff. Cheap, and it removes most of what is left.

Nothing above is a defect in what the checker verifies, and I would not want the SHA-256 pinning weakened to buy availability — the fix is in how failures are classified, not in what is checked.

`rs / gate` was still in progress when I looked, so I have not merged.

PR #672, opened 2026-08-24
test(kt): guard queried-handle log privacy

### LINKED ISSUE #655: f2z-kt: the "no handle in the operator log" invariant on /lookup and /history is stated twice and tested nowhere
`rs/crates/f2z-kt/src/api.rs`'s module doc states the reason `/kt/v1/lookup` and `/kt/v1/history` are `POST`, and then makes a claim about this code:

> `/kt/v1/lookup` and `/kt/v1/history` are `POST`. §9.2: so the queried handle does not land in the log's access logs, an intermediary's logs, or a `Referer`. **This module holds up its end by never putting a handle in a URL, and by never writing one to the operator log on those two paths** […]

`api.rs:171` repeats it at the call site: `// Deliberately no log line naming the handle. See the module note.`

**The claim is true today. Nothing enforces it, and one line undoes it.**

## Reproduction

Worktree from `ba9f46d5`.

Baseline — `cargo test -p f2z-kt`: **EXIT=0**, 62 passed (36 + 16 + 6 + 4).

Mutation — one line at the top of `LogService::lookup` (`log.rs:712`):

```rust
 pub async fn lookup(&self, handle: &Handle) -> Result<LookupResponse> {
+    log::info!("lookup {}", String::from_utf8_lossy(handle.as_slice()));
```

`cargo test -p f2z-kt`: **EXIT=0**, 62 passed; 0 failed.

Every queried handle is now written to the operator log on the exact path whose method was chosen to prevent that, and the whole suite is green.

## Why this one is worth a test rather than a comment

`tests/redaction.rs` is a good file and it does not cover this. It reasons about `Debug` renderings of in-process state, and it opens by saying the opposite thing about handles — *"a server that refused to name a handle in its own operator log would be unoperable for no privacy gain"*. That is correct for the submit path and wrong for these two, and the distinction currently lives only in prose.

**The leak path is already wired, which is what moves this from hygiene to a real gap.** `error_response` (`api.rs:260-263`) logs the full error detail for anything `is_log_fault()` — `Signer | Storage | Akd | Config` (`error.rs:110-115`). And `history` builds a `LogError::Storage` from a `format!` inside a function that has `handle` in scope:

```rust
// log.rs:780-781
return Err(LogError::Storage(format!(
    "history: no stored entry for version {version}"
)));
```

Adding the handle to that message is the single most natural debugging edit anyone will ever make to this file. It reaches `log::error!` with no test in the way.

## The control already exists

The test does not need a new mechanism, and it does not need to be a source scan. A capturing `log` implementation, one submission and one lookup:

- **positive control** — after `submit`, the captured records contain the handle. `log.rs:536` logs `"admitted {handle}@v{version} ({vouch})"` deliberately, and the module doc explains why. If the control does not fire, the capture is broken and the negative half proves nothing.
- **the assertion** — after `lookup` and `history` for that same handle, no captured record contains it.

Same shape as #626's forgery fixture and #653's fresh-log control: the negative result is attributable because a positive one fired through the same reader.

Two details worth pinning while there: assert against the handle bytes in the encodings `assert_absent` already ranges over (`redaction.rs:41-77`) rather than a plain substring, and drive `history` through its `Storage` error branch so the `error_response` path is the one under test, not just the happy path.

## Not in scope

`GET /kt/v1/sth`, `/kt/v1/sth/{epoch}` and the two `.well-known` routes take no rate-limit class at all, unlike the four in `Class::ALL`. I checked KT.md §9.3 (line 2069) before filing: it mandates a separate bucket for `/kt/v1/audit` and nothing about the others, and the module doc is explicit that an ingress terminates TLS and sits in front. Not a defect — noting it only so the next reader does not re-derive it.

skyl:
Closes #655

## What changed
- exercise the real `/lookup` and `/history` router paths under a global capturing logger
- require queried handles to stay absent in UTF-8, hex, decimal-byte, and base64url forms
- cover both successful responses and an authenticated-tree/storage-failure path that reaches production error logging
- retain positive controls for production info logs and trace-level capture so the privacy assertion cannot pass with inert logging

## Verification
- `cargo test --locked -p f2z-kt --all-targets`
- `cargo clippy --locked -p f2z-kt --all-targets -- -D warnings`
- 20 repeated parallel-thread redaction runs

## Mutation evidence
Each guard was broken on the exact candidate and the focused test went red before restoration:
- add a Debug-level queried-handle leak in `LogService::lookup`
- add a Trace-level queried-handle leak in `LogService::lookup`
- narrow the global max level from Trace to Info
- narrow the capture logger enabled filter from Trace to Info

Exact reviewed head: `3d456c6e9686d1a597c86f87009ee0b80178ca09`
Stable patch: `5f1d8d9b84abe68afd1fbff9b646ec04ca0edfd4`

### DIFF
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -525,6 +525,7 @@ dependencies = [
  "protobuf",
  "tls_codec",
  "tokio",
+ "tower",
 ]
 
 [[package]]
--- a/rs/crates/f2z-kt/Cargo.toml
+++ b/rs/crates/f2z-kt/Cargo.toml
@@ -77,6 +77,7 @@ async-trait = { workspace = true }
 # `tests/`, and never for the binary".
 f2z-kt = { path = ".", version = "0.1.0", features = ["testing"] }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros", "time"] }
+tower = { version = "0.5.3", default-features = false, features = ["util"] }
 
 [[bin]]
 name = "f2z-kt"
--- a/rs/crates/f2z-kt/src/log.rs
+++ b/rs/crates/f2z-kt/src/log.rs
@@ -799,6 +799,19 @@ impl LogService {
         })
     }
 
+    /// Remove the stored entry bytes while leaving the authenticated tree
+    /// untouched, so an acceptance test can exercise the production history
+    /// fault path through the HTTP error logger.
+    ///
+    /// This is available only to test builds. A real process can reach the
+    /// same state after storage corruption; it must report that fault without
+    /// turning the queried handle into operator-log data.
+    #[cfg(any(test, feature = "testing"))]
+    pub async fn forget_entry_bytes_for_test(&self, handle: &Handle, version: u32) {
+        let mut state = self.state.lock().await;
+        state.entries.remove(&(handle.as_slice().to_vec(), version));
+    }
+
     /// **`GET /kt/v1/audit?from&to`.**
     ///
     /// Carries every head in `[from, to]` alongside the proof — see
--- a/rs/crates/f2z-kt/tests/redaction.rs
+++ b/rs/crates/f2z-kt/tests/redaction.rs
@@ -1,5 +1,5 @@
 //! **The log is public by construction — so log it normally, but never a
-//! signing key and never an unpublished submission.**
+//! signing key, never an unpublished submission, and never a queried handle.**
 //!
 //! The first half is deliberate: a `DirectoryEntry` is a public record, and a
 //! server that refused to name a handle in its own operator log would be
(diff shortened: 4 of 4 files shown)


skyl:
Reviewed against `3d456c6e`. **#655 is closed, and I checked it by running my own reproduction verbatim rather than reading the diff.** Baseline `cargo test -p f2z-kt`: **EXIT=0, 65 passed**.

**Half one — the mutation from the issue**, unchanged from what I filed: `log::info!("lookup {}", String::from_utf8_lossy(handle.as_slice()));` at the top of `LogService::lookup`.

```
cargo test -p f2z-kt  =>  EXIT=101
  lookup_and_history_never_put_the_queried_handle_in_the_operator_log ... FAILED
  the queried lookup handle leaked as UTF-8 in: lookup queryhandleprivacy
```

On the pre-fix tree that same line left EXIT=0 with all 62 green.

**Half two — the near-miss, which is the part I most wanted covered and which is not in your mutation list.** #655 named a specific future edit as the likely way this breaks: `history` builds a `LogError::Storage` from a `format!` inside a function with `handle` in scope, and `error_response` logs the full detail of anything `is_log_fault()`. So I made that edit:

```rust
 return Err(LogError::Storage(format!(
-    "history: no stored entry for version {version}"
+    "history: no stored entry for {} version {version}",
+    String::from_utf8_lossy(handle.as_slice())
 )));
```

```
cargo test -p f2z-kt --test redaction  =>  EXIT=101
  panicked at crates/f2z-kt/tests/redaction.rs:281
```

Line 281 is `assert_absent(&fault, …, "the history handle on the storage-fault path")`. The most natural debugging edit anyone will ever make to that function is now caught, which is more than I asked for.

**The controls are the reason this holds, and there are three of them.**

1. `log::trace!("query-log-privacy trace capture control")` with an immediate assertion that it was captured — proving the logger is installed *and* at `Trace` before any negative assertion runs. Without it, a logger that silently failed to install would make every later `assert_absent` pass vacuously, which is exactly the failure mode a redaction test is prone to.
2. The production submission log as a positive control: after `submit`, asserting the capture *contains* `"queryhandleprivacy"`. That is the one I specified in the issue, and it is what makes the negative attributable — the reader demonstrably can see a handle, so its absence later means something.
3. On the fault path, asserting the capture *contains* `"history: no stored entry for version 1"` before asserting the handle is absent from it. A third control, and it is the one that proves the test actually drove production error logging rather than an empty string.

**Two choices worth naming.** Driving `f2z_kt::api::router(state)` through `tower::ServiceExt::oneshot` rather than calling `LogService` directly means the assertion covers `error_response` and the real HTTP path — better than the service-layer test I suggested. And `forget_entry_bytes_for_test` is gated `#[cfg(any(test, feature = "testing"))]` with a comment saying a real process reaches the same state through storage corruption, so an otherwise-unreachable fault path becomes testable without shipping the hook.

Extending `assert_absent` to check UTF-8 was the right call and not something I flagged: the original ranged over hex, decimal and base64url because it was written for *keys*, and a handle is text. The mutation above leaked as UTF-8 and nothing else, so without that addition none of this would have fired.

`gate` is green; `rs / gate` was queued when I looked, so I have not merged yet.

PR #675, opened 2026-08-24
test(ci): mutation-test delegation walker guards

### LINKED ISSUE #660: workspace_strict_verification_scan: the new delegation walker is the only predicate in the file without a self-test

skyl:
Closes #660

## What changed
- add an executable self-test for the strict-verification delegation walker
- cover direct and mixed `DelegatesTo` / `DelegatesToFn` chains into strict verification
- fail closed on malformed targets, missing files/functions, cycles, non-signature terminals, exhaustion, and ambiguous duplicate occurrences
- require exactly one `(file, name)` registry entry at every hop so occurrence #0 cannot vouch for an unsafe duplicate

## Verification
- focused delegation-walker suite: 18/18
- `cargo test --locked -p f2z-codec --all-targets`: 138/138
- `cargo clippy --locked -p f2z-codec --all-targets -- -D warnings`

## Mutation evidence
Every production guard was broken and observed red before restoration: first-match/occurrence aliases, file-only and name-only matching, whole-walker true/false, both delegation variants true/false, malformed target acceptance, missing-entry acceptance, strict-terminal rejection, non-signature acceptance, and exhaustion acceptance. Reversed-order and second-hop ambiguity probes also pass only under the exact uniqueness rule.

Exact reviewed head: `9d5904c7091994e5755c58a4a12a4e901bd15419`
Stable patch: `01e4245ee5f21e20ed0fa5b3f79384a86aff8bc0`

### DIFF
--- a/rs/crates/f2z-codec/tests/workspace_strict_verification_scan.rs
+++ b/rs/crates/f2z-codec/tests/workspace_strict_verification_scan.rs
@@ -920,17 +920,23 @@ fn registration_matches(entry: &VerifyFn, file: &str, name: &str, occurrence: us
     entry.file == file && entry.name == name && entry.occurrence == occurrence
 }
 
-fn delegation_reaches_strict_base(mut target: &str) -> bool {
-    for _ in 0..WORKSPACE_VERIFY_FNS.len() {
+fn delegation_reaches_strict_base(mut target: &str, registry: &[VerifyFn]) -> bool {
+    for _ in 0..registry.len() {
         let Some((target_file, target_name)) = target.rsplit_once("::") else {
             return false;
         };
-        let Some(entry) = WORKSPACE_VERIFY_FNS
+        let mut matching = registry
             .iter()
-            .find(|entry| entry.file == target_file && entry.name == target_name)
-        else {
+            .filter(|entry| entry.file == target_file && entry.name == target_name);
+        let Some(entry) = matching.next() else {
             return false;
         };
+        // A delegation target names only file and function. If that pair has
+        // multiple registered occurrences, choosing either one would let one
+        // body vouch for another; ambiguity is therefore a refusal.
+        if matching.next().is_some() {
+            return false;
+        }
         match entry.strictness {
             Strictness::CallsVerifyStrict { .. } => return true,
             Strictness::DelegatesTo {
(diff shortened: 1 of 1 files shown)


skyl:
Reviewed against `9d5904c7`. **#660 is closed, and I checked it by running my own mutation verbatim in both directions.** Baseline: **18/18, EXIT=0**.

**The mutation from the issue** — replacing the walker's body with `return true`:

```
cargo test -p f2z-codec --test workspace_strict_verification_scan  =>  EXIT=101
  12 passed; 6 failed

  a_malformed_delegation_target_is_refused
  a_delegation_target_with_multiple_registered_occurrences_is_refused
  a_delegation_chain_that_terminates_in_a_nonsignature_check_is_refused
  an_unregistered_function_in_a_registered_file_is_refused
  a_delegation_target_in_an_unregistered_file_is_refused
  a_mixed_delegation_cycle_is_refused
```

On the pre-fix tree that same edit left all 11 green. Six independent refusals now depend on the walker actually walking — including `a_mixed_delegation_cycle_is_refused`, which was the case I asked for specifically because the `0..registry.len()` bound is the kind of thing that looks obviously right and is never executed.

**And the other direction, which is the half that makes the six mean something.** A walker stuck closed would satisfy every one of those refusals while breaking the scan completely, so I ran `return false` too:

```
  16 passed; 2 failed
  a_delegation_chain_that_terminates_in_strict_verification_is_accepted
  every_verify_function_in_the_workspace_is_registered_as_strict
```

So the predicate cannot be satisfied by saying yes to everything *or* no to everything, and the live registry contract is wired to the same function the synthetic cases exercise. That two-sidedness is the property; one direction alone would have been the same defect in a new costume.

**Taking the registry as a parameter is what made this possible**, and it is worth naming as the design move rather than an implementation detail. `delegation_reaches_strict_base(target, registry)` lets eighteen synthetic registries drive it — cycles, unregistered files, non-signature terminals, exhaustion — none of which can be expressed in `WORKSPACE_VERIFY_FNS`, because that table describes a workspace where those situations do not exist. Same shape as #667 injecting the ceiling into `run`: the thing under test stops being reachable only through production data.

**The occurrence-uniqueness rule goes past what I asked for.** `a_delegation_target_with_multiple_registered_occurrences_is_refused` closes a hole I had not seen: with two registry rows sharing a `(file, name)`, a first-match walker would let occurrence #0 vouch for a duplicate that might delegate somewhere else entirely. Requiring exactly one entry per hop means a duplicated function name is a refusal rather than a coin flip, and the PR's note that "reversed-order and second-hop ambiguity probes also pass only under the exact uniqueness rule" is the right way to show it — order-dependence is exactly how a first-match bug hides.

`gate` is green; `rs / tests` was still running when I looked, so `rs / gate` has not reported and I have not merged.

Issue #671, opened 2026-08-24
f2z-relay: a panicking protocol task leaves a process that still answers /healthz


skyl:
Raised while wiring the Kubernetes probes and the GCE backend health check onto the new health-only listener (#665). It is not a regression from that PR — it is a property that PR makes something *rely on*.

## The shape

`Server::start` spawns three detached tasks — the protocol listener, the (new) health listener and the expiry tick — and `main` then does:

```rust
wait_for_signal().await;
server.shutdown().await;
```

Nothing observes those `JoinHandle`s. `listener::serve` is robust against transient accept errors (it logs and `continue`s, deliberately, so a descriptor limit does not end the listener), but a **panic** in that task ends it silently. The process keeps running.

## Why it now matters more

`/healthz` answers from process state only. That is deliberate and correct: a probe that queried the store would fail during exactly the backpressure it exists to survive, and the deployment repo's kube-linter has a custom `probe-path-is-root` check because of an outage caused by a probe on an expensive path.

The consequence is that after a protocol-task panic:

* the startup, readiness and liveness probes all pass (they GET `/healthz` on the health port),
* the **GCE health check** passes too, so the load balancer keeps the NEG endpoint in rotation,
* and `relay.<prefix>free2z.cash` accepts TCP and completes nothing.

Every availability signal is green over a relay that cannot serve a single client, and the deployment is `replicas: 1` — there is no second pod to compare against.

## Suggested fix: supervision, not a deeper probe

Making `/healthz` deep would trade this for the failure mode the shallow probe was chosen to avoid. The cheaper and more precise fix is for the process to stop when a task it needs has stopped: `select!` over the join handles alongside the signal, and exit non-zero if any of them completes before shutdown was requested. A crash-looping pod is the correct outcome there, and it is one the deployment already tolerates.

The expiry tick deserves the same treatment for a different reason: silently ending it means TTLs stop being enforced, which is a §7.7 correctness property, not just an availability one.

https://claude.ai/code/session_01EF5qmzNm91R75ujoFuaTku

Issue #669, opened 2026-08-24
f2z-kt: /kt/v1/cosign is fail-open with no witness_pk, and every accepted record is permanent


skyl:
Found while running the published `f2z-kt` and `f2z-witness` images against each other for the k8s manifests (companion: #665, and the tuzi-side manifest PR). **Measured, not inferred.**

## The behaviour

`LogService::accept_cosignature` checks:

```rust
if !self.known_witnesses.is_empty()
    && !self.known_witnesses.contains(&cosignature.statement.witness_pk)
{
    return Err(LogError::NotAWitness);
}
```

With an **empty** `witness_pk` list — which `Config::parse` accepts, since the key is optional — that condition is short-circuited and the check never runs. `POST /kt/v1/cosign` is public and unauthenticated, so **any** correctly self-signed `WitnessCosignature` over a head the log actually published is:

* pushed onto `state.cosignatures[epoch]` (in memory, for the life of the process),
* `append_cosignature`'d to the journal (fsynced, permanent),
* replayed into memory at every startup, and
* returned inside **every** `/kt/v1/sth` and `/kt/v1/sth/{epoch}` bundle.

## Reproduction

Four `f2z-witness` instances, four freshly generated keys, none of them named anywhere in the log's configuration, each polling a real log once:

```
$ docker run --rm --network f2znet f2z-witness poll --log-url http://kt:8080 \
    --log-id <id> --log-pk <pk> --key /data/w.key --state /data/state.bin --allow-cleartext yes
INFO  witness 006e71462244767d01ebc5459ee0563745b259ee6171c918abbb7d74e15a3bc8
INFO  cosigned epoch 1

$ docker logs kt | grep -c "cosignature accepted"
4
```

Add `witness_pk = <one key>` to the log's configuration and the same unlisted witness gets HTTP 400 (`ERR_NOT_A_WITNESS`) and nothing is stored. Also measured.

## Why it matters

Ed25519 keys are free and `RateLimiter` admits `Class::Cosign` continuously, so an anonymous client can grow the append-only journal, the resident set and every client's tree-head response without bound. The records are replayed at startup, so an OOMKill is followed by an OOMKill — and the deployment gives that journal `replicas: 1` on a ReadWriteOnce volume with, by design, **no backup**. The recovery is hand-editing a security-sensitive append-only journal.

[...]

PR #680, opened 2026-08-24
fix(relay): bound the operational listener in time and concurrency

skyl:
Follow-up to #665, which merged while this was being written. It is the fourth finding of the codex adversarial review on the companion manifest PR (free2z/tuzi#1399) — the review noticing that a listener written for loopback had just been pointed at the cluster.

## The defect

#665 added a health-only listener that may bind off loopback, because the kubelet and the GCE health checker both dial the pod IP. That made a listener written for a loopback population reachable from **every pod in the cluster** — and `admin::answer` has a bound on request **bytes** with no bound on **time**, inside an accept loop that spawns a task per connection with no cap:

```rust
let Ok((stream, _peer)) = accepted else { continue; };
tokio::spawn(async move { let _ = answer(relay, scope, stream).await; });
```

```rust
let read = stream.read(&mut buffer).await?;   // no deadline
```

A peer that connects and never speaks holds a task, a 2 KiB buffer and a descriptor indefinitely. Enough of them grow this process until the memory limit kills it — and the deployment is `replicas: 1` on a ReadWriteOnce volume, so there is no second pod.

On loopback that population was "processes on this host" and none of it mattered. Off loopback it does.

## The fix

Two seconds per request, sixteen in flight.

`try_acquire`, not `acquire`: waiting would move the queue from the semaphore to the accept backlog and keep the connection open, which is the thing being bounded. Over the limit, the socket is closed before any read.

Two seconds is chosen to be **shorter than the probe's own `timeoutSeconds`** (3–5s in the manifests): the server should give up before the client does, so a stuck peer is the server's descriptor to reclaim rather than the probe's deadline to miss.

## Be precise about what this buys

The code says this where somebody will read it, because the overclaim is the tempting part:

> It bounds **resources** — descriptors, tasks, buffers — so a peer that opens connections and never speaks cannot grow this process without limit, and the listener recovers within `REQUEST_TIMEOUT` once the flood stops. It does **not** guarantee **availability**: a determined flood from inside the cluster can still occupy the permits and make a probe time out, and no in-process number fixes that — an attacker who can do this can also fill the accept backlog. The answer to a hostile pod on the cluster network is a NetworkPolicy on this port, not a bigger constant here.

The first version of the test asserted the availability property and **failed**, which is how that came to be written down instead of assumed.

## The test asserts exactly that and no more

64 silent peers are all disconnected on the server's own deadline — a read returning `0` at EOF rather than hanging, which is the property: the connection is the server's to reclaim, not the peer's to hold — and the listener is serving `/healthz` again afterwards.

```
test a_client_that_connects_and_says_nothing_is_disconnected_and_the_listener_recovers ... ok
```

## Gate

`check-rust-fmt.sh` (10 crates), `cargo test --locked -p f2z-relay --test health --test admin` (12 passed, 0 failed).

## Related

* free2z/tuzi#1399 — the manifest side, whose review found this.
* #669 — `/kt/v1/cosign` fail-open with no `witness_pk`, plus the global rate limiter that lets the same endpoint be starved.
* #671 — `Server::start`'s tasks are unsupervised: a panic in the protocol listener leaves a process that still answers `/healthz`, which is the *other* way this listener can lie now that the load balancer believes it.

### DIFF
--- a/rs/crates/f2z-relay/README.md
+++ b/rs/crates/f2z-relay/README.md
@@ -121,8 +121,18 @@ Two things this deliberately is **not**:
   fail the kubelet's liveness probe and restart the only replica. A separate
   listener cannot do that.
 
-`scripts/check-f2z-images.mjs --probe-relay IMAGE` asserts all three properties
-against a running container after every build.
+It is also **bounded in time and concurrency**, which the loopback listener
+never needed to be: a two-second deadline per request and sixteen in flight, so
+a peer that connects and never speaks costs an accept and a close rather than a
+task, a buffer and a descriptor held indefinitely. Be precise about what that
+buys — it bounds *resources*, not *availability*. A determined flood from inside
+the cluster can still occupy the permits and make a probe time out, and no
+in-process constant fixes that; a NetworkPolicy on the port is the answer to a
+hostile pod.
+
+`scripts/check-f2z-images.mjs --probe-relay IMAGE` asserts the three surface
+properties against a running container after every build, and
+`tests/health.rs` covers the slow-client case.
 
 ## `.well-known` is deliberately not served
 
--- a/rs/crates/f2z-relay/src/admin.rs
+++ b/rs/crates/f2z-relay/src/admin.rs
@@ -51,10 +51,11 @@
 //! relay's existence at that address is not a secret; its queue depth is.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use tokio::io::{AsyncReadExt as _, AsyncWriteExt as _};
 use tokio::net::{TcpListener, TcpStream};
-use tokio::sync::watch;
+use tokio::sync::{Semaphore, watch};
 
 use crate::config::is_loopback;
 use crate::engine::Relay;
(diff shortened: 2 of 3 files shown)


skyl:
Post-merge review against the exact merged head `3e2c20cd`. Baseline `--test health --test admin`: **EXIT=0, 12 passed**. **Two findings, filed together as #682**, both about reach rather than correctness.

The diagnosis and the fix are right, and the `MAX_CONCURRENT` doc is the best part of it — refusing to claim availability, naming a NetworkPolicy as the actual answer, and saying so where somebody will read it. That the first version of the test asserted the availability property and failed, and that this is how the note came to be written rather than assumed, is exactly the standard.

**The two bounds are not equally covered, and I isolated them rather than mutating both at once.**

Removing the deadline and leaving the semaphore → **EXIT=101**, `assertion failed: the server held connections open indefinitely`. Load-bearing.

Restoring the deadline and removing only the refusal → **EXIT=0, 7 passed**. A relay with no concurrency bound at all passes the test that exists to bound concurrency.

That is structural rather than a missing assertion: every silent peer reaches EOF within `REQUEST_TIMEOUT` whether or not it held a permit, so `closed == 64` cannot separate the mechanisms, and `assert!(silent.len() > 16)` is about the accept loop still accepting. What distinguishes them is timing, and it is already observable — with the cap, an over-limit connection is closed *before any read*, promptly; without it, only when the deadline expires. Asserting that substantially more than `MAX_CONCURRENT` reach EOF well inside `REQUEST_TIMEOUT`, with the permitted ones taking roughly the full deadline as the control, tests the documented behaviour instead of a consequence both designs share.

**The second is `REQUEST_TIMEOUT`'s justification.** "Shorter than the probe's own `timeoutSeconds`" is a genuine invariant that lives half here and half in the manifests, and if either side moves the ordering inverts silently into precisely the failure the value was chosen to avoid. #665 set the precedent for this in the same file — `the_kubernetes_shape_is_valid_configuration` exists so "the two repos cannot drift silently" — and it is still at `health.rs:151` without covering the timeout relationship. Same shape as #657.

**What I attacked and could not break:** there is no path from a refused permit to a `read`; the permit is moved into the spawned task and outlives the timeout, so a stalled peer cannot hold one past the deadline; and `closed == 64` covering both refused and permitted connections is the right assertion for the resource claim — it just cannot see the cap.

PR #681, opened 2026-08-24
fix(ci): classify AKD evidence transport failures

### LINKED ISSUE #673: check-akd-doc-evidence: the rs gate now depends on four unauthenticated third parties, and an outage is indistinguishable from tampering
Filed from my review of #668 (merged as `5a308e79`, comment 5402109818) so it is tracked rather than left in a PR thread.

**The evidence pinning #668 added is good and this issue does not ask for it to be weakened.** Every external artefact is SHA-256 compared rather than merely fetched — the NCC report bytes, the issue-comment body, both `facebook/akd` PR patches, and the raw upstream source at each pinned commit. "The bytes we audited are still the bytes served" is the claim that makes a documentation assertion checkable at all. This is about how *failures* are classified, not about what is checked.

## What is now on the critical path of the `rs` gate

`rs.yml` runs two unconditional steps — `--self-test` and the live check — and `rs=true` is set by `rs/*` **or** `docs/e2ee/*`, so essentially every Rust or spec PR runs them. `selfTest()` also calls `fetchPinnedJson`, so both steps are exposed, not only the live one.

| host | calls | for |
|---|---|---|
| `api.github.com` | 3 | the benchmark comment; `facebook/akd` PRs #400 and #495 |
| `www.nccgroup.com` | 1 | the report PDF |
| `github.com` | 2 | both `.patch` bodies |
| `raw.githubusercontent.com` | 2 | the two pinned release sources |

`fetchPinnedBytes` and `fetchPinnedJson` send only a `User-Agent`. There is no `Authorization`, no retry, and no `catch` anywhere in the file — a 30-second timeout and then `assert(response.ok, …)`.

## The rate limit is real, not folklore

Measured rather than recalled:

```
$ curl -sSI -H "User-Agent: free2z-zuu-akd-evidence-check" \
    https://api.github.com/repos/facebook/akd/pulls/400
HTTP/2 200
x-ratelimit-limit: 60
x-ratelimit-resource: core
```

60 per hour, per source IP, for unauthenticated requests. Hosted runners share egress addresses, so that budget is not this repository's alone.

## The part that matters most is the error taxonomy

A rate-limited or unreachable host fails as:

```
unable to fetch pinned evidence https://api.github.com/... : HTTP 403
```

That is the same shape of failure as evidence that has genuinely changed, and it reads like tampering. The check exists to separate "the audited artefact was altered" from "everything is fine"; as written, a third party's quota, outage or TLS hiccup lands in the first bucket and turns a red gate into an investigation of the wrong thing.

This is the argument #665 made two PRs earlier about `STALE` and `UNPINNED` — a statement about somebody else's availability must not be reported as a statement about us — pointed at CI instead of at a liveness probe. #665 got it right for the witness; the same reasoning applies here.

## Fix, in the order I would do it

1. **Send the token when there is one.** `Authorization: Bearer ${process.env.GITHUB_TOKEN}` when set, omitted when running locally without one. Actions' built-in token raises the limit to 1,000/hour and removes the dominant failure mode. Nothing here reads private data, so the default read-only token suffices.
2. **Separate transport failure from digest mismatch.** A digest mismatch must stay a hard, loud failure — that is the entire point of the check. An unreachable host is a different verdict, deserves a different message, and whether it should block the gate is a decision somebody should get to make. Conflating them means it can never be made.
3. **Retry transient failures** (timeout, 5xx, 429) two or three times with a short backoff.

A useful test for (2), in the shape this repository already uses: point one anchor at a URL that cannot resolve and assert the failure is reported as unreachable rather than as altered evidence — with a real digest mismatch as the positive control, so the two verdicts are shown to be distinguishable rather than asserted to be.

### LINKED ISSUE #674: ci(rs): authenticate and classify AKD evidence transport failures
Follow-up to #559 / PR #668. The evidence checker was merged with live third-party fetches on the required `rs` gate but without authenticated GitHub API access, retries, or a transport-specific error verdict.

The byte/digest pinning must remain fail-closed and unchanged. This issue is only about making availability failures operationally distinct from evidence drift.

Acceptance criteria:
- pass the read-only Actions token to both AKD evidence steps
- send `Authorization: Bearer` only to the GitHub API when a token is present; never leak it to NCC Group or other evidence hosts
- retry bounded transient failures: network/timeout exceptions, HTTP 429, HTTP 5xx, and GitHub quota-exhaustion 403
- report exhausted/unretryable transport failures as `EvidenceTransportError` / `pinned evidence unavailable`, distinct from hash or semantic mismatch
- mutation-test each auth, host-scope, retry, classifier, taxonomy, and workflow-wiring guard by breaking it and observing red

Review context: https://github.com/free2z/zuu/pull/668#issuecomment-5402109818

skyl:
Closes #673
Closes #674

## What changed
- authenticate only exact `https://api.github.com` evidence requests when `GITHUB_TOKEN` is available
- bound transient retries for timeouts, network failures, 429s, quota 403s, and 5xx responses, while leaving permanent 400/401/ordinary 403/404 failures single-shot
- create a fresh 30-second abort signal per attempt with exact 250ms/500ms backoff
- keep response-body consumption inside the retry boundary and classify exhausted transport failures as `EvidenceTransportError` rather than AKD drift
- prove the inverse taxonomy too: authentic digest drift cannot be mislabeled as transport unavailability
- wire the token into both workflow evidence steps without exposing it

## Verification
- focused self-test: 31 transport guard paths green
- live authenticated evidence check green
- workflow policy live check and actionlint green
- adversarial review replayed production mutants red, including default 300-second timeout and removed scheduler delay
- reviewer digest-drift-to-transport mutant red at the intended negative-control assertion
- exact final-head review pending at `83d568b37abb8a9a42a4cc3dd36d8a2754ffaadc`

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -293,11 +293,18 @@ jobs:
       # prose and JSON, re-points both cited fix PRs, aliases both releases to
       # one source, and changes the real pinned partition body. Report evidence
       # is parsed with Node alone, including when PATH contains no host tools.
+      # Transport controls scope the Actions token to api.github.com, retry
+      # transient failures, and keep unavailability distinct from evidence
+      # drift; their self-test breaks each guard before trusting the live check.
       # Only then is the current documentation trusted.
       - name: Mutation-test AKD documentation evidence
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
         run: node scripts/check-akd-doc-evidence.mjs --self-test
 
       - name: Verify AKD documentation against locked executable evidence
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
         run: node scripts/check-akd-doc-evidence.mjs
 
       # `crash-injection` is off by default because it compiles an `abort()`
--- a/scripts/check-akd-doc-evidence.mjs
+++ b/scripts/check-akd-doc-evidence.mjs
@@ -327,22 +327,80 @@ function extractRustFunction(source, signature) {
   assert.fail(`unterminated body for ${signature}`);
 }
 
-async function fetchPinnedBytes(url) {
-  const response = await fetch(url, {
-    headers: { "User-Agent": "free2z-zuu-akd-evidence-check" },
-    signal: AbortSignal.timeout(30_000),
-  });
-  assert(response.ok, `unable to fetch pinned evidence ${url}: HTTP ${response.status}`);
-  return Buffer.from(await response.arrayBuffer());
+class EvidenceTransportError extends Error {
+  constructor(url, detail) {
+    super(`pinned evidence unavailable ${url}: ${detail}`);
+    this.name = "EvidenceTransportError";
+  }
 }
 
-async function fetchPinnedJson(url) {
-  const response = await fetch(url, {
-    headers: { Accept: "application/vnd.github+json", "User-Agent": "free2z-zuu-akd-evidence-check" },
-    signal: AbortSignal.timeout(30_000),
+function evidenceHeaders(url, accept, token) {
+  const headers = { "User-Agent": "free2z-zuu-akd-evidence-check" };
+  if (accept) headers.Accept = accept;
+  if (token && new URL(url).origin === "https://api.github.com") {
+    headers.Authorization = `Bearer ${token}`;
+  }
+  return headers;
+}
+
+function isTransientResponse(response) {
+  return response.status === 429 ||
+    response.status >= 500 && response.status <= 599 ||
+    response.status === 403 && response.headers.get("x-ratelimit-remaining") === "0";
+}
+
+const delay = (milliseconds) => new Promise((resolveDelay) => setTimeout(resolveDelay, milliseconds));
+const timeoutSignal = (milliseconds) => AbortSignal.timeout(milliseconds);
+
+async function fetchPinnedValue(url, {
+  accept,
+  attempts = 3,
+  consume,
+  fetchImpl = fetch,
+  signalForTimeout = timeoutSignal,
+  sleep = delay,
+  token = process.env.GITHUB_TOKEN,
+} = {}) {
+  let lastFailure;
+  for (let attempt = 1; attempt <= attempts; attempt += 1) {
+    try {
+      const response = await fetchImpl(url, {
+        headers: evidenceHeaders(url, accept, token),
+        signal: signalForTimeout(30_000),
+      });
+      if (response.ok) {
+        return await consume(response);
+      } else {
+        lastFailure = `HTTP ${response.status}`;
+        if (!isTransientResponse(response) || attempt === attempts) {
+          throw new EvidenceTransportError(url, lastFailure);
+        }
+      }
+    } catch (error) {
+      if (error instanceof EvidenceTransportError) throw error;
+      lastFailure = error instanceof Error ? error.message : String(error);
+      if (attempt === attempts) {
+        throw new EvidenceTransportError(url, lastFailure);
+      }
+    }
+    await sleep(attempt * 250);
+  }
+  throw new EvidenceTransportError(url, lastFailure ?? "request failed");
+}
+
+async function fetchPinnedBytes(url, options) {
+  return fetchPinnedValue(url, {
+    ...options,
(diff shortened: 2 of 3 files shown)


skyl:
Reviewed against `58ec709d`, then re-checked against `5d69f218` — the two differ only by #680's rebase, and `check-akd-doc-evidence.mjs` is byte-identical between them, so everything below applies to the current head.

**#673 is substantively addressed.** `--self-test` → EXIT=0 with `transportGuards: 31` alongside the original 32; live check → EXIT=0. The token scoping, the bounded retries with exact backoff, and a real distinct `EvidenceTransportError` class are all there.

**The token scoping is better than I asked for, and I checked it the hard way.** `new URL(url).origin === "https://api.github.com"` is an exact origin comparison through the URL parser, so `https://api.github.com.example/` does not match. I went looking for a self-test on the *leak* direction — token present, non-GitHub host — could not find one in the block around the two positive assertions, and was about to file it. Removing the origin restriction answered that properly:

```
AssertionError: https://www.nccgroup.com/media/.../_ncc_group_...pdf received the GitHub token
```

So the guard is covered, my concern was unfounded, and the mutation was the only honest way to find that out.

**One finding: the classification is one-sided.**

Every assertion about the new error class is positive — `assert(error instanceof EvidenceTransportError, …)` at `:909`, `:945`, `:974`, `:995`, and `assert.rejects(…, EvidenceTransportError, …)` for the permanent-status cases. Nothing anywhere asserts that a failure which is *not* transport is not reported as transport.

So the conflation this PR exists to end still passes in the other direction. Making a digest mismatch throw the transport verdict:

```js
-assert.equal(sha256(content), entry.raw_sha256, `${entry.release} raw source hash changed`);
+if (sha256(content) !== entry.raw_sha256) {
+  throw new EvidenceTransportError(entry.release, "raw source hash changed");
+}
```

```
node scripts/check-akd-doc-evidence.mjs --self-test  =>  EXIT=0
  {…, "pinnedSource":2, …, "transportGuards":31}
```

Altered upstream evidence is now announced as a network problem, and the suite reports every mutant killed — including the two `pinnedSource` ones, which *did* die, just with the wrong verdict. The drift mutants use `assert.throws`/`assert.rejects` without an expected class, so any error satisfies them.

This is the half #673 asked for by name: *"with a real digest mismatch as the positive control, so the two verdicts are shown to be distinguishable rather than asserted to be."* One assertion closes it — take one existing drift mutant and assert its error is **not** an `EvidenceTransportError`. Same shape as #675, where `return true` and `return false` were both run because a one-sided guard is satisfied by one that refuses everything.

**An aside that is not a defect in this PR, but is thematically its subject.** My first self-test run failed in a cold worktree:

```
Error: AKD 0.13 API fixture did not compile:
    at compileApiFixture (…:653:11)
```

with **empty stderr**. That is `compileApiFixture`'s 180-second bound expiring on a from-scratch `cargo test --no-run` of `f2z-kt` into a dedicated `CARGO_TARGET_DIR`, not a compile failure — warming the directory first and re-running gives EXIT=0. I confirmed it rather than reporting it, because reporting an environment artefact as a PR failure is the mistake I made on #664.

Worth mentioning here because it is the same error-taxonomy problem in a different guise: a timeout is reported with the words "did not compile", and the empty stderr is the only thing distinguishing it. If `result.status !== 0` also carried "or timed out after 180s" when `result.error?.code === "ETIMEDOUT"`, the next person to hit it on a cold cache would not go looking at the fixture.

skyl:
Addressed the one-sided taxonomy finding on exact head `83d568b37abb8a9a42a4cc3dd36d8a2754ffaadc`.

The existing real pinned-source digest mutant now requires both the expected digest verdict and `!(error instanceof EvidenceTransportError)`. I then applied the reviewer mutation to the production `validatePinnedSource` digest branch. The self-test exited 1 at the intended assertion:

> AssertionError: pinned-source drift was mislabeled as transport unavailability

After restoring production, authenticated `--self-test` is green with `pinnedSource: 2` and `transportGuards: 31`; authenticated live evidence and the GitHub Actions policy check are green. Fresh exact-head CI has restarted.

I agree the cold-fixture timeout message is misleading. I have not folded that separate compile-runner taxonomy change into this transport PR without its own falsifiable timeout control.

PR #686, opened 2026-08-25
fix(deps): clear yanked spin release

skyl:
Refs #351

## Summary
- update the two stale wallet lockfiles from spin 0.9.8 to 0.9.9
- remove the obsolete yanked-spin allowance comment from wallet/deny.toml
- leave the advisory-backlog umbrella open for its remaining upstream items

## Verification
- scripts/check-rust-deny.sh advisories: all five wallet Cargo roots pass under the repository policy
- plugin, classic Zuuallet, and ZUULI locked builds/tests and Rust fmt/clippy passed on the candidate
- cargo tree resolves spin 0.9.9 in every affected shipping graph

## Mutation proof
The repository deliberately keeps yanked = warn so a registry-side yank cannot break unrelated PRs. Therefore the normal policy command reports an old yanked lock but exits zero. For the mutation proof only, I used the same pinned cargo-deny binary, real lockfiles, registry advisory data, and deny.toml with the CLI lint override -D yanked:

- candidate plugin and classic lockfiles at spin 0.9.9: both strict invocations exit 0
- restoring only the plugin lock to spin 0.9.8: strict invocation exits 1 with error[yanked] naming spin 0.9.8
- restoring only the classic Zuuallet lock to spin 0.9.8: strict invocation exits 1 with the same exact yanked finding
- restoring both candidate lockfiles returns both strict invocations to exit 0; the tracked tree and patch ID are unchanged

The CLI override is verification-only and does not change the repository warning policy.

### DIFF
--- a/wallet/deny.toml
+++ b/wallet/deny.toml
@@ -38,8 +38,6 @@ db-urls = ["https://github.com/rustsec/advisory-db"]
 # upstream between two runs of CI with no commit of ours in between, which would
 # turn an unrelated pull request red for reasons its author cannot act on. It is
 # reported on every run and triaged deliberately instead.
-# Currently reported: spin 0.9.8 (via flume -> zcash_client_backend, and via
-# lazy_static -> bellman/orchard/tao).
 yanked = "warn"
 
 ignore = [
--- a/wallet/plugins/tauri-plugin-zcash/Cargo.lock
+++ b/wallet/plugins/tauri-plugin-zcash/Cargo.lock
@@ -4965,9 +4965,9 @@ dependencies = [
 
 [[package]]
 name = "spin"
-version = "0.9.8"
+version = "0.9.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
+checksum = "3763264f6b73151db08c50ff20d7d8a0b8796e021cdea7ceedad07b80155fa0e"
 dependencies = [
  "lock_api",
 ]
--- a/wallet/zuuallet/src-tauri/Cargo.lock
+++ b/wallet/zuuallet/src-tauri/Cargo.lock
@@ -4912,9 +4912,9 @@ dependencies = [
 
 [[package]]
 name = "spin"
-version = "0.9.8"
+version = "0.9.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
+checksum = "3763264f6b73151db08c50ff20d7d8a0b8796e021cdea7ceedad07b80155fa0e"
 dependencies = [
  "lock_api",
 ]


skyl:
Reviewed against `3e82b48e`. The lockfile updates are right and the mutation proof is the right *shape* — reverting one root and showing that root go red while the other four stay green is per-root attribution rather than "something turned red". Two things about the description and the policy, neither of which is a defect in the dependency bump.

**1. What was removed from `wallet/deny.toml` is a comment, not an allowance.**

```diff
-# Currently reported: spin 0.9.8 (via flume -> zcash_client_backend, and via
-# lazy_static -> bellman/orchard/tao).
 yanked = "warn"
```

There is no `ignore` entry for spin, before or after, and `yanked = "warn"` is unchanged. The summary line — "remove the obsolete yanked-spin allowance" — reads as though an exemption was lifted and enforcement tightened. Nothing was exempt; a bookkeeping note about what the run currently reports went stale and was deleted. Worth correcting in the body, because "allowance removed" is the kind of phrase a future reader will take as evidence that this class is now gated.

**2. The mutation claim and the policy comment three lines above it cannot both be true.**

`deny.toml` states the reasoning for `warn` deliberately:

> Yanking is registry state, not repository state: a crate can be yanked upstream between two runs of CI with no commit of ours in between, which would turn an unrelated pull request red for reasons its author cannot act on. It is reported on every run and triaged deliberately instead.

The PR says:

> restoring only the plugin lock to spin 0.9.8 makes the real cargo-deny command fail at that exact root with `warning[yanked]`

If a yanked crate does fail the run, then the promise in `deny.toml` — that yanking is reported and never turns an unrelated PR red — is not what the tooling does, and the comment should be corrected. If it does not fail the run, then whatever went red in that experiment was something else, and the mutation proof is not evidence for the property it is offered for. Either way one of the two needs fixing, and it matters more than the wording: the comment is the stated policy for a whole category of supply-chain finding.

**Nothing in the repository settles it, which is the part worth acting on.** `check-rust-deny.sh` passes no `--deny warnings` and relies on cargo-deny's own exit code (`:283`), and `--self-test`'s four cases are all licence and policy-plumbing — a missing policy, a permitted licence, a forbidden licence, a missing `--config`. None exercises `advisories`, and none exercises a yanked crate. So the question "does `yanked = "warn"` fail this gate?" has no answer in the tree, and the file's own doctrine is the argument for giving it one: *"A check nobody has watched fail is not a check."* A fifth self-test case with a fixture pinning a known-yanked version would pin the category's behaviour once, in the place that already exists for exactly this purpose.

**I could not run it to find out, and I am not going to guess.** Reverting the plugin lock and running `./scripts/check-rust-deny.sh advisories` in a worktree gives me:

```
error: failed to get `zcash_address` as a dependency of package `tauri-plugin-zcash`
3 crate(s) failed the supply-chain policy: …
```

— a path-dependency resolution failure, because the worktree has no `z/` submodules for the wallet's path deps to point into. That is my environment, not the PR, and it tells me nothing about `warning[yanked]`. You ran it somewhere that resolves; the exact exit code and whether cargo-deny reported the yanked crate as a warning or an error is the thing to paste.

skyl:
Correction to the initial mutation claim: the reviewer was right that scripts/check-rust-deny.sh advisories itself exits 0 for a yanked crate, because wallet/deny.toml deliberately sets yanked = warn. My earlier wording incorrectly described that warning as a failing gate.

I replayed both lockfile mutations with cargo-deny 0.20.2 strict verification mode (-D yanked), without changing the production policy. Exact results:

- candidate spin 0.9.9: plugin exit 0; classic exit 0
- plugin-only rollback to 0.9.8: exit 1, error[yanked], exact spin package named
- classic-only rollback to 0.9.8: exit 1, error[yanked], exact spin package named
- restored exact head 3e82b48: both exit 0; clean tree; stable patch still 38387e3504c6f759d0fa0d83196b1d57575ecd73

The PR body now makes the normal warn-policy result and the strict mutation-only override explicit.

skyl:
On the suggested fifth persistent self-test: I prototyped a standalone real Cargo fixture pinned to spin 0.9.8. Today it proves the desired split exactly: normal wallet policy exits 0 with warning[yanked], and the same pinned cargo-deny invocation with -D yanked exits 1 with error[yanked].

I am not adding that fixture to the required repository gate in this PR. A persistent test whose green state depends on crates.io continuing to mark a particular release yanked would make an unrelated PR fail if that mutable registry state changes (including an un-yank), which is precisely the failure mode the yanked = warn policy is designed to avoid. The independent mutation replay provides the evidence for this lockfile change without converting mutable registry state into a permanent gate dependency.

PR #688, opened 2026-08-25
test(ci): pin yanked warning verdict

### LINKED ISSUE #687: check-rust-deny: whether `yanked = "warn"` fails the gate is unpinned, and deny.toml's policy comment contradicts #686's mutation proof
Raised on #686 (merged as `158f945c`, comment 5403710859) and filed so it does not sit in a merged thread. Not a defect in that PR — the lockfile bump is correct and its per-root mutation proof is the right shape.

## Two statements in the tree contradict each other

`wallet/deny.toml` states the policy for the whole yanked category, with its reasoning:

> Yanking is registry state, not repository state: a crate can be yanked upstream between two runs of CI with no commit of ours in between, which **would turn an unrelated pull request red for reasons its author cannot act on**. It is reported on every run and triaged deliberately instead.

```toml
yanked = "warn"
```

#686's verification says:

> restoring only the plugin lock to spin 0.9.8 **makes the real cargo-deny command fail** at that exact root with `warning[yanked]`

Both cannot hold. If a yanked crate fails the run, the promise above is not what the tooling does and the comment is misleading about a whole category. If it does not, then whatever went red in that experiment was something else and the mutation proof is not evidence for the property it is offered for.

## Nothing in the repository decides it

- `scripts/check-rust-deny.sh:283` invokes cargo-deny with `--manifest-path`, `--config` and `--locked` and no `--deny`/`-D`, so the verdict is cargo-deny's own exit code and the repo expresses no opinion about warnings.
- `--self-test` has four cases (`:198-245`) and they are all licence and policy plumbing: a tree with no policy, a permitted licence, a forbidden licence under a different `--config`, and a `--config` that does not exist. It proves the **policy file decides the verdict**, which is a good property and a real one. None of the four exercises `advisories`, and none involves a yanked crate.

So the question "does `yanked = "warn"` fail this gate?" has no answer in the tree, and it is a question about the gate, not about spin.

## Why it is worth pinning rather than reasoning about

The script's own header makes the argument better than I can:

> A check nobody has watched fail is not a check.

A fifth self-test case, in the fixture that already exists, pinning a known-yanked version and asserting the verdict either way, settles the category once. Whichever way it comes out, one of the two texts above then gets corrected to match, and the next person who reads `yanked = "warn"` learns what it actually does to a pull request.

The same fixture would also give `advisories` its first self-test coverage; today the four cases are all `licenses`, so the check most likely to turn a PR red is the one whose plumbing is least pinned.

## A smaller, separate point

#686's summary describes removing "the obsolete yanked-spin allowance from `wallet/deny.toml`". What it removed was a **comment** recording what the run currently reported:

```diff
-# Currently reported: spin 0.9.8 (via flume -> zcash_client_backend, and via
-# lazy_static -> bellman/orchard/tao).
 yanked = "warn"
```

There is no `ignore` entry for spin, before or after. Deleting a stale note is right; calling it an allowance suggests an exemption was lifted and enforcement tightened, which is the reading a future reviewer will take from the merge message.

## What I could not check

I could not run cargo-deny to settle this myself: reverting the plugin lock in a worktree and running `./scripts/check-rust-deny.sh advisories` fails with `error: failed to get zcash_address as a dependency of package tauri-plugin-zcash`, because a fresh worktree has no `z/` submodules for the wallet's path dependencies to resolve against. That is my environment and says nothing about the yanked verdict — recording it so nobody reads the absence of a reproduction here as evidence either way.

skyl:
Closes #687

Pins the actual yanked-warning gate contract through the real cargo-deny wrapper. The self-test creates a private local Cargo registry containing one synthetic yanked release, then proves the shipping wallet policy reports it and exits successfully while an otherwise-identical deny policy reports it and exits nonzero.

Verification:
- Bash 3.2 syntax and focused shellcheck green
- self-test green with normal paths and TMPDIR containing Unicode, spaces, #, and %
- all five shipping wallet advisory scans green
- independent exact patch review approved

Mutation evidence, each restored before commit:
- wallet shipping policy warn to deny: RED with error[yanked]
- synthetic registry yanked true to false: RED because the warning disappeared
- strict deny to warn: RED because the negative control exited successfully
- invalid strict policy: RED because failure attribution was not error[yanked]
- raw unencoded special path bytes: RED on a space-containing TMPDIR
- unmasked Bash 3.2 UTF-8 ordinal: RED on a Unicode TMPDIR

The fixture uses only scratch paths and a local file-backed registry; live crates.io yank state cannot change its verdict.


skyl:
Reviewed against `85e8e744`. **#687 is closed, and the contradiction is resolved in the direction I could not determine.** `yanked = "warn"` reports and exits successfully; `deny.toml` was right and #686's "makes the real cargo-deny command fail" was not.

```
./scripts/check-rust-deny.sh --self-test  =>  EXIT=0
  self-test: wallet yanked = "warn" reports a known-yanked crate and exits successfully.
  self-test: yanked = "deny" fails on the same known-yanked crate.
  self-test: 6 case(s) passed; cargo-deny 0.20.2.
```

Both directions, on **one** synthetic crate, through the real wrapper. That is the shape that makes the pair mean something: the same fixture under two policies, so neither verdict can be explained by the fixture rather than the policy.

**The thing I most wanted to check was whether the "warn passes" case actually observes anything.** An assertion that a run exits 0 is satisfied by every clean run, including one with no yanked crate anywhere — a guard that cannot fail. It is not that. Each half greps for three things: the exit status, the diagnostic class (`warning[yanked]` / `error[yanked]`), and the crate name and version. So I flipped the fixture's registry entry from `"yanked":true` to `"yanked":false`:

```
./scripts/check-rust-deny.sh --self-test  =>  EXIT=1
  advisories ok
  All 1 Rust crate(s) … satisfy wallet/deny.toml.
```

The case fails when there is no warning to report, which is exactly what it needs to do. Verified rather than taken from the mutation list.

**Building a private file-backed registry is the right call and worth naming.** A live crates.io release can be unyanked after the commit lands, which would make the self-test quietly stop testing what it says. A synthetic release that is yanked by construction cannot drift, and it also sidesteps the thing that stopped me reproducing any of this on #686 — a wallet root needs the `z/` submodules to resolve its path dependencies, and this fixture needs nothing but scratch paths.

**The corrected comment says what the tooling does**, and points at the test that holds it:

> This lint level reports that event but deliberately exits successfully, so it cannot turn an unrelated pull request red. The self-test pins both halves of that contract against a synthetic crate in a local yanked registry: warn reports and passes, while deny reports and fails.

That is the half of #687 that mattered more than the plumbing — the policy comment is now a claim somebody can check.

**One note, not a request.** The assertions match cargo-deny's diagnostic text (`warning[yanked]`, `error[yanked]`), which is normally the thing I would flag as not being an API. Here it is fine, and specifically because the version is pinned and the self-test prints it: a cargo-deny bump that changed the diagnostic format breaks this loudly at the moment of the bump, rather than silently degrading into an assertion that always passes. Worth leaving as it is, and worth not "improving" into a looser match later.

Neither gate had reported when I looked — 16 checks still running — so I have not merged.

PR #689, opened 2026-08-25
fix(kt,relay,witness): fail closed on the rest of the supervision cluster

### LINKED ISSUE #666: f2z-witness: Health::is_alive is fail-open on a new variant, and the compiler only catches the display arm
Found reviewing #665 (merged as `80e8bf78`), against its exact head `7ef6d90f`. Not a defect on the current tree — a fail-open default in the one predicate that decides whether a pod keeps running.

`Health::is_alive` (`f2z-witness/src/health.rs`) is written as a negation:

```rust
pub const fn is_alive(&self) -> bool {
    !matches!(self, Self::Halted { .. })
}
```

Its sibling is written as a positive match:

```rust
pub const fn is_healthy(&self) -> bool {
    matches!(self, Self::Following { .. })
}
```

**Those two default in opposite directions for a variant that does not exist yet.** `is_healthy` is fail-closed: anything new is not `Following`, so readiness fails. `is_alive` is fail-open: anything new is not `Halted`, so the liveness probe passes.

That is the reverse of what `is_alive`'s own documentation promises:

> `Health::Halted` still fails. […] so that a halted witness is never reported as fine by the one check an operator looks at first.

`Health` has four variants and is not `#[non_exhaustive]`. Adding a fifth local fault — a state file that exists but cannot be read, a signature that does not verify, a second halt-like condition — makes it report alive.

## Reproduction

Worktree at `7ef6d90f`. Baseline `cargo test -p f2z-relay -p f2z-witness`: **EXIT=0, 164 passed**.

Add a fifth variant, exactly as a future change would:

```rust
     /// No state file: this witness has never completed a poll.
     Unpinned,
+    /// A future local fault -- the state file exists but cannot be read.
+    EvidenceUnreadable,
 }
```

`cargo build -p f2z-witness` **does** fail, which is the good news and also the whole problem:

```
error[E0004]: non-exhaustive patterns: `&Health::EvidenceUnreadable` not covered
   --> crates/f2z-witness/src/health.rs:139:15
```

**Line 139 is `message()`** — the function that renders a string for the probe's output. It is the only exhaustive match on `Health`. So the compiler tells the author to write a display string; it says nothing about whether the pod should live.

Adding only that display arm, as an author fixing the error would:

```
PROBE is_healthy (readiness) => false
PROBE is_alive   (liveness)  => true
PROBE exit_code=1 liveness_exit_code=0
```

and the full suite:

```
cargo test -p f2z-witness -p f2z-relay  =>  EXIT=0
```

A brand-new local fault fails readiness (correct, and free — `is_healthy` is fail-closed) and passes liveness. `f2z-witness healthz --liveness yes` exits 0, the kubelet is satisfied, and the pod shows `1/1` while the witness is in a fault state it cannot leave. Nothing in the suite moves.

## Why the current test cannot catch it

`a_liveness_probe_ignores_the_upstream_log_and_not_a_local_fault` is a good test — it covers all four variants and asserts *both* exit codes for each, which is more than most. But it enumerates the variants by hand, so it grows only when someone remembers to grow it, and the compiler never asks.

## Fix

Make `is_alive` exhaustive, so the decision is forced at the predicate rather than at the display string:

```rust
pub const fn is_alive(&self) -> bool {
    match self {
        // Statements about the upstream LOG: a restart cannot repair either,
        // and restarting on a schedule during someone else's outage destroys
        // the state file that is the evidence a human needs.
        Self::Following { .. } | Self::Stale { .. } | Self::Unpinned => true,
        // Local and permanent. Visible as 0/1 with HALTED in the probe output.
        Self::Halted { .. } => false,
    }
}
```

A fifth variant is then a compile error **here**, where the author has to decide whether a restart could repair it — which is the question the doc comment already frames better than any test could.

Worth applying the same shape to `is_healthy` for symmetry; it is fail-closed today, so it is not urgent, but two predicates over one enum written in two different styles is how the asymmetry arose.

## Reviewed in the same pass, no findings

The rest of #665 is strong, and the relay half in particular:

- `the_health_listener_does_not_serve_metrics` asserts 404 for `/metrics` on the health listener **and 200 for `/metrics` on the admin listener** in the same test. That positive control is what makes the 404 mean "scoped", not "metrics is broken".
- `the_health_listener_may_not_shadow_another_listener` covers both collisions (protocol and admin), which is the one that would put `/metrics` wherever `/healthz` is.
- `the_admin_listener_is_still_refused_off_loopback` restates the rule the change could plausibly have relaxed.
- `the_kubernetes_shape_is_valid_configuration` pins the deployment's exact environment block so the two repos cannot drift silently — a hypothetical-future fixture in the #645 sense.
- Defaults are distinct (`127.0.0.1:8443` protocol, `127.0.0.1:9101` admin, `127.0.0.1:8081` health), so enabling health with defaults collides with nothing. I also checked the `health.ip() == listen.ip()` exact comparison, which does not catch `0.0.0.0:9101` against `127.0.0.1:9101` — that one fails at bind with `EADDRINUSE` rather than silently sharing a port, and the health-only surface 404s `/metrics` regardless, so it is fail-closed either way. Not a defect.
- `route`'s catch-all is `404`, so a future endpoint added without a scope guard is not served on the health listener by accident.
- The `--liveness` flag requires the literal `yes` and rejects anything else, and the verdict text is printed unchanged on both paths, so a probe that passes on `STALE` still says `STALE`.

### LINKED ISSUE #684: f2z-kt: the epoch scheduler is unsupervised — a dead scheduler stops publishing while /healthz stays 200
Found while fixing #671 and deliberately left out of that PR, because it is the same defect in a different binary and deserves its own fix and its own test.

## The shape

`f2z-kt`'s epoch scheduler is an **unsupervised task**. If it dies, the log stops publishing epochs — while `/healthz` keeps returning 200 and the pod stays `Ready` and in rotation.

This is exactly the shape #671 fixed in `f2z-relay`: a process whose listener outlives the task that gives it a purpose. #671's resolution was that the process goes down, because `replicas: 1` + `Recreate` makes that a brief, **visible** outage, whereas the alternative is a healthy-looking endpoint serving nobody.

## Why it is worse here than it looks

A relay that has stopped accepting is loud — senders fail immediately. **A key-transparency log that has stopped publishing epochs is silent.** Clients keep verifying against the last signed tree head, which stays valid; lookups keep succeeding; nothing errors. The log simply stops incorporating new entries, and no client can distinguish "no directory changes happened" from "the log stopped".

`KT.md` §5.1 anticipates precisely this and is the reason heartbeat epochs exist — an epoch published on cadence even with nothing to add, so that **silence is detectable**. A dead scheduler defeats that mechanism entirely: there are no heartbeats to miss, because there is no scheduler to emit them.

There is also a monitoring gap behind it. The "KT log growth stalled" alert is written against the log's *published* state, so it detects this — but the seven alert policies are **committed and not live in GCP** (see the activation epic), and this repo already has an incident where a policy existed and never fired. So today nothing catches it at either layer.

## Wanted

- Supervise the epoch scheduler the way `f2z-relay` now supervises its protocol and expiry tasks: a task ending before shutdown takes the process down with a non-zero exit, bounded so a wedged task cannot hold the port.
- **A test that fails today**, driven through a real listener rather than an in-process flag: kill the scheduler and assert the service stops presenting as healthy. Note #671's measured correction — ending a task may drop its own listener while the *health* surface keeps answering, so assert on the health surface, or the test passes on the defect.
- Audit the other long-lived tasks in `f2z-kt` and `f2z-witness` for the same shape while you are there, rather than fixing this one instance. The witness's poll loop is the obvious candidate: a witness that has stopped polling reports `ok` from `healthz` for as long as its state file is fresh.

## Related
- **#671** — the same defect in `f2z-relay`, fixed; read its PR for the supervision pattern and the test approach.
- **#669** — the other fail-open fixed in the same PR.
- The activation epic in the deployment repo tracks the alert policies not being live, which is the second layer that would have caught this.

### LINKED ISSUE #685: f2z-relay: the group-commit writer is an OS thread outside #683's supervision, so a dead write path still passes every probe
Found reviewing #683 (merged as `c8e7fccd`), against its exact head `ef7093db`. Baseline `cargo test -p f2z-kt -p f2z-relay`: **EXIT=0, 219 passed**.

#683 fixed the fail-open in `Server::start`: four long-lived tokio tasks are now held in `Supervised`, `run_until_stopped` selects over their join handles, and a task that ends before shutdown was requested closes every listener and exits non-zero. All four spawn sites go through the `supervise` closure — I checked for a missed one and there isn't any.

**The relay has a fifth long-lived worker, and it is the one that owns the entire write path.**

`commit.rs:141-144` starts the group-commit writer as an **OS thread**, not a tokio task, and discards its handle:

```rust
std::thread::Builder::new()
    .name("f2z-relay-commit".to_owned())
    .spawn(move || run(&store, &metrics, &receiver, window, max_batch.max(1)))?;
```

`Supervised` holds `tokio::task::JoinHandle`, so this worker cannot be in it. When the thread dies, `Commit::append` returns `CommitError::WriterStopped`, and `engine.rs:1095` turns that into a per-request wire code:

```rust
.map_err(|_| ProtoError::Wire(ErrorCode::Unavailable))?
```

So the process stays up and answers every probe while it cannot store a single message — which is #671's own description of the defect (*"a process that still answers `/healthz`"*), for the component #683's supervision does not reach.

## Reproduction

Make the writer exit at startup, exactly as a panic inside it would:

```rust
 .spawn(move || { /* the group-commit writer dies at startup */ })?;
```

**Nothing notices:**

```
cargo test -p f2z-relay --features testing --test supervision --test health  =>  EXIT=0
  supervision: 7 passed; 0 failed
  health:      4 passed; 0 failed
```

Every listener stays open, `/healthz` still answers 200, and no `Stopped::TaskEnded` is ever produced.

**And the write side is gone:**

```
cargo test -p f2z-relay --test configured_equivalents  =>  EXIT=101
  a_lost_ack_response_is_safe_to_retry ... FAILED
  a_full_queue_is_indistinguishable_from_an_absent_one ... FAILED
  a_cheap_command_overtakes_an_expensive_one_and_both_correlate ... FAILED
  backpressure_never_refuses_read_ack_or_delete ... FAILED
  an_expired_message_is_gone_and_announced ... FAILED
```

That is the whole finding in two commands: writes dead, probes green, supervision silent.

The blast radius is every `APPEND` and `CONTACT_APPEND` — the relay's entire write side — and #680's description notes the deployment is `replicas: 1` on a ReadWriteOnce volume, so there is no second pod to take over from one that has quietly stopped accepting messages.

## Why the existing supervision cannot cover it

This is not an oversight in the list of names; it is a type boundary. `supervise()` polls `JoinHandle` futures inside a `poll_fn`, and a `std::thread::JoinHandle` is not pollable. Two shapes work:

1. **Make it a task.** The reason it is a thread is good — `commit.rs` argues an fsync is a blocking syscall of unbounded duration and parking a Tokio worker on a 1 GB VPS is worse — so `tokio::task::spawn_blocking` is the direct translation, and its handle *is* pollable and can join the existing `Vec<Supervised>` unchanged.
2. **Watch the channel.** The writer owning the `Receiver` means its death is already observable: the `Sender` starts returning `SendError`. A supervisor could hold a liveness signal the writer drops on exit, giving `Stopped::TaskEnded(COMMIT_TASK)` without moving the thread.

Either way the test is the one the PR already established: kill it, assert the sockets close, assert the process would exit non-zero. `supervision.rs`'s module note — that the assertion turns on the health socket refusing rather than on an in-process flag, because *"an in-process assertion that a flag flipped would prove nothing"* — applies unchanged.

## What I attacked and could not break

- **All four `Server::start` spawns are supervised.** There is no bare `tokio::spawn` left in that function; the per-connection spawns in `connection.rs` and `admin::serve` are correctly outside the scheme, since a connection task dying is that connection's problem.
- **The cosign fix cannot degenerate into broken cosigning.** I made `accept_cosignature` refuse *everything* — the fix taken too far — and `a_configured_witness_is_accepted_exactly_once_and_a_stranger_is_not` FAILED, which is exactly the control the PR claims it is. A one-sided fix that simply stopped accepting cosignatures does not pass.
- **Recognition really is checked first**, before `cosignature.verify()` and before the state lock, so an unrecognised key costs a set lookup rather than an Ed25519 verification and an `fsync`.
- `ADMIN_TASK` is supervised but is the one of the four without its own abort test (`PROTOCOL`, `EXPIRY`, `HEALTH` have one each). Since `abort_task` is by name and the four are registered identically, that is a gap in symmetry rather than in coverage — noting it so it is a deliberate choice rather than an accident.

skyl:
Three fail-open defects of one shape, finished in **#683's** idiom rather than
in three new ones. Every rule fixed here is a rule that **could not fail**
before this PR — which is how all three survived review — so each lands with the
test that fails without it, and each was watched fail on unfixed code before the
fix was written.

Closes #684
Closes #685
Closes #666

---

## #684 — `f2z-kt`'s epoch scheduler was unsupervised

`server::serve` spawned the scheduler detached and then awaited the listener.
Nothing observed its join handle, so a panic ended that task and left the
process serving §9.2's endpoints correctly over a directory that had stopped
moving.

**Worse than the relay case, and measured rather than argued.** On the unfixed
shape — scheduler aborted, then a real `GET /healthz` on the bound port:

```
MEASURED after scheduler death: healthz=Some("HTTP/1.1 200 OK") epoch 1 -> 1
```

That is three seconds at a **one-second** cadence. Nothing errors. Clients keep
verifying against the last signed tree head, which stays valid; lookups keep
succeeding; and no client can distinguish *"no directory changes happened"* from
*"the log stopped incorporating them"*. `KT.md` §5.1's heartbeat epochs exist
precisely so silence is detectable — a dead scheduler does not **trip** that
mechanism, it **removes** it: there are no heartbeats to miss because nothing is
left to emit them.

### The fix

`Server::{start, run_until_stopped, supervise, abort_task}` in
`f2z-kt/src/server.rs`, the same shape as `f2z-relay`'s and deliberately not a
second idiom. `serve` now returns `Stopped`; `main` matches it **exhaustively**
(`Stopped` is not `#[non_exhaustive]`, for #683's reason) and exits non-zero on
`TaskEnded`. Shutdown is bounded by `SHUTDOWN_GRACE` so a wedged task cannot
hold the port. The scheduler also selects on the shutdown flag rather than being
`abort()`ed, so it can no longer be cancelled mid-`publish_epoch` — the one
operation in that process that writes.

### The test, and the evidence it can fail

`rs/crates/f2z-kt/tests/supervision.rs`, driven through the bound listener
because the property is what a kubelet and a load balancer observe. With
supervision removed (join handles not polled, exactly as before the fix):

```
running 4 tests
test a_requested_shutdown_is_not_a_failure ... ok
test the_scheduler_really_is_what_publishes_epochs ... ok
test a_dead_epoch_scheduler_stops_the_process_instead_of_staying_ready ... FAILED
test a_dead_http_listener_stops_the_process ... FAILED

supervision notices a dead scheduler without waiting for a signal: Elapsed(())
```

#683's measured correction applies directly and is why the assertion is on the
**health surface**: ending a task drops its own listener, so a test that only
checked the protocol port would pass on the defect. In `f2z-kt` the scheduler
owns *no* socket, which makes the point sharper — killing it changes nothing an
outside observer can see, and the assertion is that after the fix the process
stops.

`the_scheduler_really_is_what_publishes_epochs` is the positive control:
supervision of a task that published nothing would be supervision of a no-op,
and the other three tests would still pass.

### Spec

`KT.md` §5.1 carries a dated correction in `ARCHITECTURE.md` §4.9's style: the
cadence rule binds the log's **process**, not only its publishing code. Nothing
above it is withdrawn; what is added is that a conforming log MUST NOT keep
serving after whatever emits epochs has stopped, and that a *tick* that fails is
still a transient to be retried.

---

## #685 — the group-commit writer was outside #683's supervision

Not a missed name: a **type boundary**. `Supervised` holds
`tokio::task::JoinHandle`s, and the writer is an OS thread — for the reason
`commit.rs` argues at length, an fsync is a blocking syscall of unbounded
duration. Measured on the unfixed shape:

```
MEASURED after commit-writer death: healthz=Some("HTTP/1.1 200 OK") protocol_accepts=true
```

Every listener open, every probe green, and `APPEND` answering
`ERR_UNAVAILABLE`. Under delete-on-ack that is **data loss, not downtime**.

### The fix, and the trap it avoids

`CommitWriter::start` returns a `WriterStopped` alongside the handle: the thread
owns a `oneshot::Sender<Infallible>` it never sends on, so the **only** thing
that can complete the receiver is the sender being dropped — when `run` returns,
and when a panic unwinds out of it. There is nothing the writer has to remember
to do, which is the property that makes it trustworthy under a panic.

`server.rs` supervises a task that awaits exactly that, registered in the
**same** `Vec<Supervised>` as the other four. The instruction was not to solve a
supervision gap by adding an unsupervised supervisor, and this is how that is
satisfied: the watchdog's own death is reported identically, which
`the_commit_watchdog_is_itself_supervised` asserts.

**`spawn_blocking` was considered and rejected.** Its handle is pollable, but an
aborted blocking task cannot be cancelled once running: `shutdown()` awaits every
handle, so every ordinary stop would wait out `SHUTDOWN_GRACE`, and the test
that proves the supervision works could not produce the failure at all.

### The tests, and the evidence they can fail

The injection kills **the thread**, not the watchdog — aborting the watchdog
would pass identically on a build where nothing connected it to the writer,
which is precisely the defect. `Message::Stop` is behind `testing`, like
`abort_task`, and never reaches the binary. With the watchdog not registered
(`drop(writer_stopped)`, i.e. the pre-fix wiring):

```
cargo test -p f2z-relay --test supervision --test health
  supervision:  4 passed; 2 FAILED
    a_dead_commit_writer_stops_the_process_instead_of_answering_unavailable  FAILED
    the_commit_watchdog_is_itself_supervised                                 FAILED
  health:       7 passed; 0 failed        <- the issue's whole point
```

`commit.rs` also gains
`the_liveness_signal_fires_when_the_writer_thread_really_ends`, which uses **no
injection hook at all** — dropping the last `Sender` ends `run` for real — so
`stop_for_test` cannot be proving only that one testing path works.

---

## #666 — `Health::is_alive` was fail-open on a new variant

Reproduced exactly as the issue described, by adding a fifth variant to the
**unfixed** predicates:

```
cargo build -p f2z-witness   =>  one error, at message() only
PROBE is_healthy (readiness) => false  exit_code=1
PROBE is_alive   (liveness)  => true   liveness_exit_code=0
```

A brand-new local fault fails readiness for free and **passes liveness**: the
kubelet is satisfied and the pod shows `1/1` while the witness sits in a fault
it cannot leave. The compiler asked the author for a display string and never
for a verdict.

### The fix, and why the "test" is the compiler

Both predicates are now exhaustive matches, so the question — *could a restart
repair this?* — is forced at the decision point, where the doc comment already
frames it better than any test could. `Health` also loses `#[non_exhaustive]`,
for the reason #683's `Stopped` never had it: the binary is a separate crate, so
that attribute forces a wildcard arm in `main`, and a wildcard arm is how this
class recurs. Same experiment against the fixed tree:

```
error[E0004]: non-exhaustive patterns: `&Health::EvidenceUnreadable` not covered
   --> crates/f2z-witness/src/health.rs:99   (is_healthy)
   --> crates/f2z-witness/src/health.rs:153  (is_alive)
   --> crates/f2z-witness/src/health.rs:179  (message)
```

One error became three, and the two new ones are the ones that decide whether a
pod lives. No runtime test can ask this on behalf of a variant nobody has
written yet; the existing four-variant test stays as the runtime control.

---

## The audit, which is the part that was not filed

Every long-lived task and thread in the three binaries, not just the three
instances:

- **`f2z-relay`** — the four `Server::start` spawns (#683) plus the commit
  writer (this PR) are now the complete set. `listener.rs:127`, `admin.rs:190`,
  `connection.rs:54` and `engine.rs:1144` are per-connection or per-request:
  one of those dying is that connection's problem, and they are correctly
  outside the scheme.
- **`f2z-kt`** — `server.rs` was the only file with a `tokio::spawn`; both sites
  are now supervised. `signer.rs:217` is a per-signature KMS subprocess behind
  the `kms` feature, not long-lived.
- **`f2z-witness` — the obvious candidate, and not the same shape.** The poll
  loop is `main`'s own loop, not a task: every way out of it (`Halted`, a
  non-retryable error, `poll` mode) **returns from `main`**, so it cannot stop
  while the process lives, and there is no `/healthz` for a stalled loop to keep
  answering — the probe is `exec` on a separate invocation. The one wedge that
  could have made it fail open is a hung request, and `transport.rs` sets
  `timeout_global`, so a poll cannot block forever. A witness that is slow or
  cut off is `Health::Stale`, which fails readiness and deliberately passes
  liveness (§7.1: killing the observer does not fix the observed). **No fix, and
  nothing to file.**

## Gates

`check-rust-fmt.sh`, `check-rust-clippy.sh` (`-D warnings`, all 10 crates),
`check-rust-deny.sh`, `check-hash-domain-labels.mjs`,
`check-akd-doc-evidence.mjs`, `check-f2z-images.mjs`,
`check-public-repo-boundary.sh`, and `cargo test --locked --all-targets` +
`--doc` all pass locally. `#![forbid(unsafe_code)]` and the AGPL/permissive
boundary are unchanged; #649 is untouched.

### DIFF
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -1308,6 +1308,45 @@ O(entries added) (§10) and an empty epoch adds none.
 > `AppendOnlyProof` segment containing no insertions, with `audit_verify`
 > accepting it. Neither exists in `akd` 0.13. See §11.5.
 
+> **Correction (2026-08-24, later the same day) — the cadence rule binds the
+> log's *process*, not only its publishing code: a conforming log MUST NOT keep
+> serving after whatever emits epochs has stopped.** Found while fixing
+> [#684](https://github.com/free2z/zuu/issues/684). Nothing above is withdrawn;
+> what is added is the half a reader could otherwise implement around.
+>
+> §5.1's whole argument is that a heartbeat converts silence from an ambiguity
+> into a detectable fault. **That argument assumes something is still emitting
+> heartbeats.** A log whose scheduler has died — panicked, cancelled, or never
+> restarted after a transient — does not *fail* the cadence rule in a way any
+> client can see: it goes on serving §9.2's endpoints correctly, `/kt/v1/sth`
+> keeps returning the last valid signed tree head, every inclusion and
+> consistency proof still verifies, and nothing errors. The detection mechanism
+> is not tripped, it is **removed**, because there are no heartbeats left to
+> miss.
+>
+> A witness following the log does eventually see it (§7: no new head for
+> longer than the cadence), and the log's own operator sees it in
+> `KT.md`-derived monitoring — but both are *external* observers on a delay, and
+> a process-local probe answers `200` throughout. So the rule at the process
+> level is:
+>
+> - **The epoch scheduler is supervised.** If it ends for any reason before an
+>   orderly shutdown was requested, the log stops serving and the process exits
+>   non-zero, within a bounded grace period so that a wedged scheduler cannot
+>   hold the listening socket.
+> - **A tick that fails is not that**, and is unchanged: an epoch the log could
+>   not sign because a signer was briefly unreachable is logged and retried on
+>   the next tick. Publishing late is within §5.2's `max_merge_delay_seconds`
+>   promise; not publishing at all, silently and forever, is not.
+>
+> At `replicas: 1` this makes a dead scheduler a short, **visible** outage
+> instead of a `Ready` pod serving a directory that has silently frozen. This is
+> the same trade `WIRE.md`'s relay made for its listener and expiry tasks
+> ([#671](https://github.com/free2z/zuu/issues/671),
+> [#683](https://github.com/free2z/zuu/issues/683)), and it is the reason that
+> trade generalises: an availability signal that stays green over a component
+> that has stopped doing its job is worse than the outage it is hiding.
+
 ### 5.2 Maximum merge delay
 
 **`max_merge_delay_seconds` — proposed 3600, six epochs — is the log's published
--- a/rs/README.md
+++ b/rs/README.md
@@ -317,8 +317,9 @@ answers from process state and nothing else, deliberately — a probe that queri
 the store would fail during exactly the backpressure it exists to survive. That
 makes it a true statement about a relay only if the process cannot outlive the
 tasks that do its work, so `f2z-relay` supervises the protocol listener, the
-admin and health listeners and the expiry tick: if any of them ends before a
-shutdown was asked for, the relay closes its listeners, prints
+admin and health listeners, the expiry tick and — since [#685] — the
+**group-commit writer**: if any of them ends before a shutdown was asked for,
+the relay closes its listeners, prints
 `the <task> stopped while the relay was running` to stderr and **exits 1**. A
 crash-looping pod is the correct outcome. At `replicas: 1` with
 `strategy: Recreate` that is a brief, visible outage, and the alternative — a
@@ -327,7 +328,26 @@ crash-looping pod is the correct outcome. At `replicas: 1` with
 `accepted` and a relay that then loses the message have between them destroyed
 it.
 
+The commit writer is the one that is not a task. It is an **OS thread**, because
+an fsync is a blocking syscall of unbounded duration and parking a Tokio worker
+on a 1 GB VPS is worse; `Supervised` holds `JoinHandle`s, so the thread could not
+be in it, and a dead writer used to leave every listener open and every probe
+green over a relay answering `ERR_UNAVAILABLE` to every `APPEND`. It is covered
+by supervising a task that waits on the thread's liveness signal — a task
+registered in the same list as the other four, so the watchdog is not itself
+the unsupervised thing.
+
+**`f2z-kt` has the same rule and a quieter failure** ([#684]). Its epoch
+scheduler is what publishes `KT.md` §5.1's epochs, heartbeats included, and a
+log that has stopped publishing errors on *nothing*: clients keep verifying
+against the last signed tree head, lookups keep succeeding, and heartbeats
+cannot be missed because there is nothing left to emit them. So the scheduler
+and the HTTP listener are supervised the same way, and `f2z-kt serve` exits
+non-zero when either ends.
+
 [#671]: https://github.com/free2z/zuu/issues/671
+[#684]: https://github.com/free2z/zuu/issues/684
+[#685]: https://github.com/free2z/zuu/issues/685
 
 ## Working in this tree
 
--- a/rs/crates/f2z-kt/Cargo.toml
+++ b/rs/crates/f2z-kt/Cargo.toml
@@ -76,7 +76,15 @@ async-trait = { workspace = true }
 # on itself. This is the supported way to say "build my `testing` feature for
 # `tests/`, and never for the binary".
 f2z-kt = { path = ".", version = "0.1.0", features = ["testing"] }
-tokio = { workspace = true, features = ["rt-multi-thread", "macros", "time"] }
+# `net` and `io-util` for `tests/supervision.rs`, which dials the bound listener
+# rather than asserting on an in-process flag (zuu#684).
+tokio = { workspace = true, features = [
+  "rt-multi-thread",
+  "macros",
+  "time",
+  "net",
+  "io-util",
+] }
 tower = { version = "0.5.3", default-features = false, features = ["util"] }
 
 [[bin]]
--- a/rs/crates/f2z-kt/src/main.rs
+++ b/rs/crates/f2z-kt/src/main.rs
@@ -115,12 +115,26 @@ fn serve(args: &[String]) -> Result<(), String> {
         .enable_all()
         .build()
         .map_err(|error| error.to_string())?;
-    runtime.block_on(async {
+    let stopped = runtime.block_on(async {
         let state = server::build(&config).await.map_err(|e| e.to_string())?;
         server::serve(state, &config.listen)
             .await
             .map_err(|e| e.to_string())
-    })
+    })?;
+    // Exhaustive on purpose (zuu#684). `Stopped` is not `#[non_exhaustive]`, so
+    // a future stop reason is a compile error here rather than a wildcard arm
+    // that silently exits zero — the same shape zuu#666 fixed in the witness's
+    // liveness predicate.
+    match stopped {
+        server::Stopped::Requested => Ok(()),
+        // Non-zero, and loud. The pod restarts and is visibly not `Ready` while
+        // it does; a log that stays up with a dead scheduler publishes nothing
+        // and reports nothing, and no client can tell.
+        server::Stopped::TaskEnded(name) => Err(format!(
+            "the {name} ended while the log was running; the log cannot publish epochs and is \
+             stopping rather than serving a directory that has silently frozen"
+        )),
+    }
 }
 
 fn check(args: &[String]) -> Result<(), String> {
--- a/rs/crates/f2z-kt/src/server.rs
+++ b/rs/crates/f2z-kt/src/server.rs
@@ -13,8 +13,41 @@
 //! not fatal: an epoch the log could not sign because a KMS was briefly
 //! unreachable is better published late than not at all, and a process that
 //! exits on it takes the whole directory down for a transient.
+//!
+//! # Every task started here is supervised (zuu#684)
+//!
+//! A tick that *fails* is a transient. The scheduler **task** ending is not:
+//! it panicked, or it was cancelled, and from that moment the log publishes
+//! nothing at all. The listener does not care — `/healthz` answers from process
+//! state, so the pod stays `Ready`, stays in rotation, and serves §9.2's
+//! endpoints correctly forever, over a directory that has stopped moving.
+//!
+//! **That failure is silent in a way `f2z-relay`'s was not.** A relay that has
+//! stopped accepting is loud: senders fail immediately. A key-transparency log
+//! that has stopped publishing epochs errors on nothing. Clients keep verifying
+//! against the last signed tree head, which stays valid; lookups keep
+//! succeeding; and no client can distinguish "no directory changes happened"
+//! from "the log stopped incorporating them".
+//!
+//! §5.1's heartbeat epochs exist precisely so that silence is detectable — an
+//! epoch published on cadence even with nothing to add, so a client that sees
+//! none knows something is wrong. A dead scheduler defeats that mechanism
+//! rather than tripping it: **there are no heartbeats to miss, because there is
+//! nothing left to emit them.**
+//!
+//! So [`Server::run_until_stopped`] selects over the join handles alongside the
+//! caller's signal, and a supervised task that ends before shutdown was asked
+//! for takes the whole process down with a non-zero exit — bounded by
+//! [`SHUTDOWN_GRACE`], so a wedged task cannot keep the listener open by
+//! refusing to finish. This is `f2z-relay`'s supervision idiom (zuu#671,
+//! zuu#683) applied unchanged; there is deliberately not a second one.
 
+use std::future::Future;
+use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
+
+use tokio::sync::watch;
 
 use crate::api::{AppState, router};
 use crate::config::Config;
@@ -24,6 +57,18 @@ use crate::ratelimit::RateLimiter;
 use crate::signer::{FileSigner, LogSigner};
 use crate::vrf::FileVrf;
 use crate::{descriptor, now_ms, policy};
+
+/// §5.1's cadence, heartbeat epochs included. The task the log exists to run.
+pub const EPOCH_TASK: &str = "epoch scheduler";
+/// §9.2's endpoints and `/healthz` — the surface every probe reads.
+pub const HTTP_TASK: &str = "http listener";
+
+/// How long a failing log is given to close its listener politely.
+///
+/// A bound and not a promise: the point of the failure path is that the process
+/// stops, and a task that will not finish must not be able to hold the port by
+/// refusing to.
+const SHUTDOWN_GRACE: Duration = Duration::from_secs(5);
 
 /// Build every part of a running log from a configuration.
 ///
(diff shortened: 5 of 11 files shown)


skyl:
Reviewed against the exact merged head `9afb8ded` — it merged while I was testing, so this is post-merge. Baseline `cargo test -p f2z-kt -p f2z-relay -p f2z-witness`: **EXIT=0, 246 passed**. **Both of my issues are closed, and I ran both reproductions verbatim rather than reading the diff.**

**#666 — the enum probe.** My issue was that adding a fifth `Health` variant produced `E0004` **only** in `message()`, the display function: the compiler asked the author for a string and never asked whether the pod should live. Adding `EvidenceUnreadable` on this head:

```
error[E0004]: non-exhaustive patterns: `&Health::EvidenceUnreadable` not covered
   --> crates/f2z-witness/src/health.rs:99   <- is_healthy
   --> crates/f2z-witness/src/health.rs:153  <- is_alive
   --> crates/f2z-witness/src/health.rs:179  <- message
```

`is_alive` is now the second place the compiler stops, which is the whole point — the decision is forced at the predicate that frames the question. Making `is_healthy` exhaustive too goes past what I asked; I had flagged that as symmetry rather than urgency, since it was already fail-closed. Doing it means neither predicate can acquire a silent default later.

**#685 — killing the commit writer.** Before, emptying the writer thread's closure left `--features testing --test supervision --test health` at **EXIT=0** (7 and 4 passed) while `--test configured_equivalents` went red: writes dead, probes green, supervision silent. The same edit now gives **EXIT=101**, and `a_dead_commit_writer_stops_the_process_instead_of_answering_unavailable` is among the failures. My mutation kills the writer at startup, which is cruder than the test's own hook and takes the server down before the test can set up — so the attributable statement is the before/after contrast rather than that one line: the edit that used to leave everything green no longer does, and there is now a named test for the property driven through `abort_task(COMMIT_TASK)`.

**The fix is better than either shape I suggested.** I offered `spawn_blocking` or a liveness signal; the signal here is a `oneshot::Sender<Infallible>` the thread owns and never sends on, so the only thing that can complete the receiver is the sender being dropped — when `run` returns **and when a panic unwinds out of it**. The comment names the property that matters: *"it exists to be dropped when this closure ends, however it ends."* There is nothing the writer has to remember to do, which is what makes it hold under the failure it exists for. A flag set on the way out would not.

And `the_commit_watchdog_is_itself_supervised` closes the obvious next hole before anyone has to file it. Solving a supervision gap by adding an unsupervised supervisor is the standard way this fix goes wrong, and registering the watchdog in the same `Vec<Supervised>` — rather than a second mechanism beside it — is the version that does not.

**On #684, which I had not found.** The measurement is the right kind of evidence:

```
MEASURED after scheduler death: healthz=Some("HTTP/1.1 200 OK") epoch 1 -> 1
```

Three seconds at a one-second cadence, everything green. And the reasoning about why it is worse than the relay case is correct and worth keeping: §5.1's heartbeat epochs exist so that silence is detectable, and a dead scheduler does not trip that mechanism — it removes it, because there is nothing left to emit the heartbeats whose absence would be the signal. The note that the scheduler now selects on the shutdown flag rather than being `abort()`ed, so it cannot be cancelled mid-`publish_epoch`, is the detail I would have wanted to check and it is already there.

`the_scheduler_really_is_what_publishes_epochs` as the positive control is exactly right: supervising a task that published nothing would be supervising a no-op, and the other three tests would pass anyway.

Issue #381, opened 2026-08-09
security(seed): block screenshots/recents capture and clear copied mnemonics


skyl:
## Summary

Seed phrases are rendered without platform capture protection, and onboarding offers to copy the mnemonic without a timed clear. The phrase can therefore remain in screenshots, screen recordings, Android/iOS recent-app snapshots, clipboard history, keyboard/OS sharing surfaces, or another app reading the clipboard.

## Evidence

- `wallet/zuuli/src/features/wallet/SeedReveal.tsx:43-77` displays/copies a revealed seed.
- `wallet/zuuli/src/features/wallet/Onboarding.tsx:300-344` displays and copies the onboarding seed.
- Android `MainActivity` does not enable `FLAG_SECURE` while sensitive material is visible.
- The iOS app does not obscure seed content during capture/background transitions.
- Clipboard content is not cleared on a short timer or lifecycle transition.

## Required remediation

- Enable Android `FLAG_SECURE` whenever seed/key material is visible, and remove it only after leaving the sensitive view.
- On iOS, detect screen capture/recording, obscure sensitive content, and cover the app before background/recent-app snapshots.
- Prefer eliminating seed-copy functionality. If retained, require explicit confirmation, label the risk, and clear only if the clipboard still contains the exact value ZUULI placed there after a short timeout and on background.
- Never log, persist, or include the mnemonic in accessibility/analytics snapshots.
- Re-authenticate user presence immediately before revealing the phrase.

## Acceptance criteria

- Physical-device tests show screenshots/recordings and recent-app previews cannot capture a mnemonic.
- Backgrounding while a seed is visible immediately obscures it.
- Clipboard tests prove timed/lifecycle clearing does not overwrite unrelated user clipboard content.
- Seed reveal requires fresh user presence and clears renderer state when dismissed.

skyl:
Claiming this security slice from exact public main `c443ff8a053107691064ef16b20ea0d003b66a4e` in an isolated worktree. I will implement only software-verifiable native lifecycle/window-authority protections, keep physical-device capture/recents evidence explicitly open, and use `Refs #381` unless every device criterion is actually proven.

skyl:
Software hardening slice merged in #690 as `ae9de6e8b70127e7e97851b6badef0a5a18f3ab9`.

[...]

PR #690, opened 2026-08-25
fix(wallet): protect typed mnemonic entry

skyl:
Refs #381

Protects typed mnemonic entry across ZUULI restore and classic Zuuallet restore/re-link with one native lease contract. The native command returns an opaque token bound to an exact entry purpose; renderer fields stay disabled until the lease arrives, clear before release, and release on success, failure, route exit, and unmount. No typed phrase crosses into native code.

Shipping integration:
- one compiler-bound command inventory feeds the plugin build permissions and shipping router
- ZUULI and Zuuallet bridges forward the exact native token and purpose
- native begin installs the exact purpose/token state; native end clears it
- the explicit iOS/Android capability authorizes begin-sensitive-entry
- all three production mnemonic textareas disable spellcheck, autocomplete, autocapitalize, and autocorrect

Mutation evidence, all restored before commit:
- shipping invoke router reduced to create-wallet: real IPC RED
- build command population reduced to create-wallet: exact compiler-bound population RED
- decorative begin tokens and no-op end calls in both bridges: runtime RED
- native decorative begin and no-op end: managed-state RED
- production mobile permission removed: rebuilt Tauri capability RED
- each of 3 fields x 4 privacy attributes, both deletion and unsafe-default replacement: 24 mounted-route REDs
- route editability, purpose, clearing order, lifecycle attachment, and registration mutants: RED

Restored verification includes 34 behavioral tests, 99 boundary mutants, 16 ZUULI Rust tests, 150 plugin tests, production builds/typechecks, full five-crate fmt/clippy/deny, permissions, Android, workflow-policy, and actionlint gates. Independent exact-patch review approved e21f4fe.


skyl:
Reviewed against `e21f4fe5`. Baseline `node scripts/seed-capture-boundary.mjs` → **SCAN_EXIT=0**; `node --test scripts/seed-capture-boundary.node-test.mjs` → **99 passed, 0 failed**.

**One question, and a correction to something I nearly filed against it.**

`src/features/auth/RestoreIdentity.tsx` is a fourth typed-BIP39-mnemonic surface. The scanner already knows it: at `:159-163` it is one of **four** entries in the BIP39 word-count loop —

```js
[onboarding, "wallet onboarding"],
[restoreIdentity, "identity restore"],
[desktopRestore, "Zuuallet restore"],
[desktopSettings, "Zuuallet recovery re-link"],
```

— and it gets its own `requireMatch` at `:169` for rejecting unsupported word counts before native IPC. But the capture-gating loop at `:480` covers **three**, and identity restore is not among them. It has `disabled={restoring}` rather than `disabled={!sensitiveEntry.editable}` and no `useSensitiveMnemonicEntry` call.

So: **is identity restore deliberately outside the lease contract?** It plausibly is — the lease buys native-side protection for a phrase that controls funds, and a free2z identity is a different asset with a different threat model. If that is the reasoning it is worth writing down, because the two loops sit in one file over what a reader will take to be one population, and the difference between them is currently silent. The PR summary — "across ZUULI restore and classic Zuuallet restore/re-link" — is accurate but does not say which surface was left out or why.

**The correction, which is the part I would want said if I were you.** I went looking for the sharper version of this — that identity restore's four privacy attributes were unprotected — and I was wrong. Stripping all four from its textarea does leave the seed-capture boundary green:

```
node scripts/seed-capture-boundary.mjs                     => SCAN_EXIT=0
node --test scripts/seed-capture-boundary.node-test.mjs    => 99 passed, 0 failed
```

but `src/features/auth/auth-density.test.tsx:88-91` asserts all four — `autoComplete="off"`, `autoCapitalize="none"`, `autoCorrect="off"`, `spellcheck="false"` — against rendered markup, so that mutation is caught, by a file this PR does not touch and whose name gives no hint it is load-bearing for seed privacy. I had checked the scanner and the `*.test.*` files that mention `spellCheck` before concluding; the assertion is on the lowercase DOM attribute in one case, which is why my first grep missed it.

Worth knowing for a different reason: the coverage for identity restore's attributes lives in an **auth density** test. If somebody later trims that file to what its name suggests, four seed-privacy assertions go with it and nothing else notices. A pointer from `seed-capture-boundary.mjs` — even a comment naming `auth-density.test.tsx` as where that surface's attributes are pinned — would stop that.

Nothing here is a defect in the lease contract, and the `begin_sensitive_entry` registration being asserted `=== 1` in the command registry is the right kind of anchor for the command population.

skyl:
**This head is red — six failures on `e21f4fe5`, from two independent causes.** Triaged from the job logs since neither is obvious from the check names.

## 1. Five of the six: a cross-project import drags zuuallet source into zuuli's `tsc`

`zuuli / frontend` fails at **Typecheck**, and the four `ZUULI / packaging smoke` jobs (Linux, macOS, Android, iOS) fail at **Build packages** with the same errors:

```
../zuuallet/src/store/wallet.ts(45,52): error TS7006: Parameter 'set' implicitly has an 'any' type.
…13 more TS7006…
../zuuallet/src/store/wallet.ts(81,13): error TS18046: 'a' is of type 'unknown'.
…5 more TS18046…
```

The cause is `wallet/zuuli/src/lib/wallet/zuuallet-sensitive-entry-routes.test.tsx:56`:

```ts
import { RestoreWallet } from "../../../../zuuallet/src/pages/RestoreWallet";
```

That is a real static import, so zuuli's TypeScript program now includes zuuallet's source and typechecks it under **zuuli's** `tsconfig` and **zuuli's** `node_modules`. `zustand`'s generic inference does not resolve there, so every store callback parameter degrades to implicit `any` and the sort comparators to `unknown`.

**The `vi.mock` calls above it do not help, and that is the part worth noticing.** The file already mocks `../../../../zuuallet/src/store/wallet` at `:20`, along with `sensitive-entry`, `useWallet`, `tauri` and `sensitive-seed`. Those are runtime substitutions; `tsc` still follows the static import graph through `RestoreWallet` into the real modules. A test can be fully mocked at runtime and still widen the type-check surface of the whole project.

Three ways out, in the order I would try them: import the component through a mocked module boundary rather than by path; add `../zuuallet/**` to zuuli's `tsconfig` `exclude` (the test still runs, since vitest resolves independently of the build's `tsc`); or make zuuli's dependency tree able to typecheck zuuallet, which is the most work and the most coupling.

## 2. The sixth is unrelated: the plugin's test binary will not start on Windows

`Rust / native tests (windows)`, step **"Test the shared plugin on the native host"**:

```
Running unittests src\lib.rs (…\tauri_plugin_zcash-f95d7f7abaffe61b.exe)
error: test failed, to rerun pass `--lib`
Caused by:
  process didn't exit successfully: … (exit code: 0xc0000139, STATUS_ENTRYPOINT_NOT_FOUND)
```

`0xc0000139` is a dynamic-link failure at process start — the binary loads a DLL that is missing an export it was linked against. No test ran; the harness never got that far. The suspicious change is the new entry in `wallet/plugins/tauri-plugin-zcash/Cargo.toml`, since this is the shape a new native dependency produces on Windows and nowhere else. Worth checking `cargo tree --target x86_64-pc-windows-msvc` for what that pulled in, rather than re-running and hoping — the same run's Linux and macOS native tests passed, so it is platform-specific rather than flaky.

## Unchanged from my earlier review

The substance still looks right to me and neither failure touches it: `node scripts/seed-capture-boundary.mjs` → `SCAN_EXIT=0` and its node-test → 99 passed on this head, and `rs / gate` is green. My open question about `RestoreIdentity.tsx` sitting in the scanner's four-surface BIP39 loop but not its three-surface capture-gating loop stands, as does the note that its four privacy attributes are pinned only by `auth-density.test.tsx`.

skyl:
Filed the general form as #691, because the import that broke five jobs here is not a one-off.

The sweep found the same crossing **already on `main`** — `wallet/zuuli/src/lib/wallet/zuuallet-created-seed.test.ts:5` reaches `../../../../zuuallet/src/lib/sensitive-seed-session` — which has not bitten only because it lands on a leaf module with no store, no hooks and no third-party generics behind it. Same move, one import away from the same failure.

And `wallet/shared/sensitive-entry-session.ts`, which this PR introduces as the right idea, is itself consumed by counted dots from **production** files at a different depth from each side (`../../../shared/…` from zuuallet, `../../../../shared/…` from zuuli), while sitting in neither project's `tsconfig` `include`. I checked whether the two configs currently disagree in a way that could bite — they do not, both are `strict: true` with the same flags — so that half is not broken today. It is a shared module with no name and no manifest, which is the condition that makes the next divergence silent rather than loud.

Nothing in #691 asks for anything from this PR beyond the fix already discussed; the shared-package and boundary-scanner work is its own change.

skyl:
CI clean-room failure repaired without rewriting the published branch. The repair is two fix-forward commits ending at exact HEAD `08916d1623f6a220ed2af71af28c2ea24ac09146`.

What changed:
- moved the classic mounted-route test into its owning package with locked test dependencies;
- separated production and test TypeScript graphs so ZUULI production builds do not import sibling test dependencies;
- made CI clean-install/cache both package lockfiles, typecheck both graphs, and run the classic mounted suite;
- exact-bound the underlying npm script bodies and `tsconfig.build.json` semantics in the live workflow policy.

Independent adversarial review first blocked `1715b05` because three mutations survived: `typecheck:tests=true`, `test:sensitive-entry=true`, and `noCheck:true` hiding a real TS2322. At the new exact head all three are genuinely RED. The reviewer also killed all five bound-script mutations, malformed/missing/extra build-config mutations, the production-policy opt-out mutant, and all 24 DOM privacy-attribute mutants. Restored self-test, typechecks, mounted tests (31 ZUULI + 3 classic), 99 boundary mutants, builds, workflow gates, and actionlint are green locally.

Independent exact-head verdict: APPROVE. New patch ID `556f27ea5c1bafa187600e8f9b5718ccd988b703`; cumulative stable patch from the branch base `885007736bc7aa55421b98e63334a026bcfabbe4`.

skyl:
Integrated current trunk with a normal forward merge at exact HEAD `3d4c306382301449356bb38af28f5cf81cbd04f2` (no rebase/force-push). The sole conflict was `wallet/zuuli/package.json`; independent review proved the resolution is the exact semantic union of trunk's messaging test/`zod` state and #690's four guarded build/typecheck/seed-test scripts. Reverting each #690 script independently fails the live policy, and the complete 12-case frontend-build mutation suite is green restored.

The reviewer separately identified that current trunk's chat `messaging-contract.node-test.mjs` invocation is not independently pinned. That weakness predates and is neither introduced nor worsened by #690; it belongs to the separately owned chat lane. Verdict for the #690 diff against current main: APPROVE.

skyl:
The six failures are gone on `3d4c3063` — but the way they went is worth a look before this merges.

**The fix stops typechecking every test file in the project, not just the one that crosses the boundary.**

```json
// wallet/zuuli/tsconfig.build.json
{ "extends": "./tsconfig.json", "exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"] }
```

and both entry points now point at it:

```
"build":     "… tsc -p tsconfig.build.json && vite build …"
"typecheck": "tsc --noEmit -p tsconfig.build.json"     ← the CI Typecheck step
```

**Nothing else typechecks them.** `vite.config.ts`'s `test` block sets `env` only — there is no `typecheck` section, so vitest transpiles with esbuild, which strips types without checking them. After this change no tool in the repository type-checks `wallet/zuuli/src/**/*.test.{ts,tsx}`.

That is a wide trade for one import. The exposure is not theoretical here, because this repository deliberately uses tests as contracts. `src/lib/messaging/parity.test.ts` — merged an hour ago in #664 — derives `BridgeMethod` from `keyof typeof WIRE_COMMANDS` and asserts `RESULTS` covers exactly those keys; that file's **type**-level half is now unchecked. Its runtime half still works (I verified separately that adding a schema-less bridge method fails it at runtime), but a mock whose shape drifts from the interface it doubles, a `vi.mock` factory returning the wrong type, or a test holding a stale API now compile-error nowhere.

**And the boundary violation is still there.** `zuuallet-sensitive-entry-routes.test.tsx:56` still imports `RestoreWallet` from `../../../../zuuallet/src/pages/RestoreWallet`; it is simply no longer visible to `tsc`. Whatever the right long-term answer is (#691), this makes the symptom quiet rather than the cause absent — and a future test that crosses the boundary in a way that *does* break at runtime will now get no early warning at all.

Three options, in the order I would consider them:

1. **Keep tests typechecked, just not in the build's program.** `tsconfig.build.json` as written is correct for `build` — shipping code should not typecheck test files. Add a `tsconfig.test.json` that includes them and make `npm run typecheck` run both, so the CI step still covers what it covered yesterday.
2. **Exclude only the file that crosses**, so the loss is one file rather than all of them.
3. **Remove the crossing** — import the component through a boundary both projects can name — which is #691 and the only one that makes the exclusion unnecessary.

(1) is a two-line change and restores exactly what was lost; it composes with (3) rather than competing with it.

Everything in my earlier reviews still stands: `node scripts/seed-capture-boundary.mjs` → `SCAN_EXIT=0`, its node-test 99 passed, and the `RestoreIdentity.tsx` question about the scanner's four-surface BIP39 loop versus its three-surface capture-gating loop is unchanged.

skyl:
CI's merged-trunk selector failure is repaired at exact HEAD `a40a7729db4abeed3fae2df3e38027cb64c8a28e` without rewriting history. The checker-only fix binds the exact current detector and independently requires messaging changes to retain the full `wallet/zuuli/*` selector.

Independent mutation review made real shipping-workflow deletion, narrowing to `features/messages/*`, substitution to `features/chat/*`, decorative broad-selector text, and deletion of the semantic checker all RED for the intended semantic reason. Restored Android policy (40 mutants), actions policy (74/10/139/12/99), workflow gates (39), actionlint, typechecks, 31+3 mounted tests, and 99 boundary mutants are green. Exact verdict: APPROVE; patch `96fd8062a13c594c6dd19a9061198ebea82e40d1`.

skyl:
## Triage of the two remaining failures on `a40a7729`

Both come from one line. `changes` fails; `gate` fails only because it depends on `changes`.

Failing step: **Verify the required Rust/WASM build boundary** (`wallet/zuuli/scripts/wasm-boundary.mjs`), job `changes`:

```
WASM boundary failed: current tree is not a valid policy base:
wallet/zuuli/package.json script build must be exactly
"npm run wasm:build && tsc && vite build && npm run wasm:verify-dist"
```

`wasm-boundary.mjs:141` pins the `build` script string byte-for-byte, and `:243` is the self-test mutation over that same literal. This PR changed the script to `tsc -p tsconfig.build.json`, so the pin no longer matches. Mechanically the check is doing its job.

## But the mechanical fix is the wrong fix

Do **not** just update the literal at `:141`/`:243` to the new string. That makes CI green while leaving the substantive problem I raised in 5405304030 in place:

`tsconfig.build.json` is `{"extends": "./tsconfig.json", "exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"]}`, and **both** `build` and `typecheck` (`zuuli.yml:203-204`) now point at it. `vite.config.ts`'s `test:` block sets `env` only — no `typecheck` key — and vitest transpiles with esbuild, which strips types without checking them. So on this head **nothing typechecks zuuli's test files at all**. The five earlier CI failures were the typechecker reporting a real problem; excluding the files it reported on removes the reporter, not the problem.

## The actual root cause

`src/lib/wallet/zuuallet-sensitive-entry-routes.test.tsx:56`:

```ts
import { RestoreWallet } from "../../../../zuuallet/src/pages/RestoreWallet";
```

Four dots out of zuuli and into a sibling project. `tsc`'s `include` only *seeds* the file list — anything reachable by static import enters the program regardless — so this drags zuuallet's whole module graph into zuuli's typecheck. The five `vi.mock` calls above it are runtime-only and do nothing for `tsc`.

The same crossing already exists on main at `src/lib/wallet/zuuallet-created-seed.test.ts:5` (`../../../../zuuallet/src/lib/sensitive-seed-session`); it hasn't bitten only because it lands on a leaf module with a small graph. And this PR's own `wallet/shared/sensitive-entry-session.ts` is reached by counted dots from **production** files at two different depths (`zuuallet/src/lib/sensitive-entry.ts:6`, `zuuallet/src/lib/tauri.ts:2`, `zuuli/src/lib/wallet/bridge.ts:9`, `zuuli/src/lib/wallet/mock.ts:19`) while sitting in neither project's tsconfig `include`.

That is #691. Shared code between `wallet/*` projects needs to be a real module with a path alias or workspace package, not a counted-dots reach. Fixing the import is what makes the typecheck pass honestly; then `tsconfig.build.json` and the boundary-pin edit are both unnecessary and the `build`/`typecheck` scripts revert to the pinned string.

## Recommendation

1. Revert `tsconfig.build.json`, and revert `build`/`typecheck` to plain `tsc` — the boundary pin then matches again and `changes`/`gate` go green on their own.
2. Resolve the cross-project import per #691 (alias or shared package), so the test files stay inside the typechecker.
3. If (2) is too large for this PR, the minimum acceptable stopgap is that zuuli's test files remain typechecked by *something* — e.g. keep `typecheck` as plain `tsc` over the full project and scope the narrower config to `build` alone. Silently dropping them from every typecheck is the outcome to avoid.

The PR is also `BEHIND` (`strict: true`), so it needs a rebase before it can merge in any case.

skyl:
The next fail-closed CI layer exposed one remaining stale exact contract: `wasm-boundary.mjs` still pinned the old unscoped production `tsc` command. Fixed forward at exact HEAD `49e4309d4e1edade70a45b5dc9b033da2dda8d8e`.

Independent review made the real old command, old command plus a decorative current string, deleted `wasm:build`, and deleted `wasm:verify-dist` all RED for the exact package-script semantics under WASM live/self and the actions policy. Every policy command from the actual `changes` job, fresh typechecks, 31+3 mounted tests, 99 boundary mutants, production build/WASM identity, and actionlint are green. Exact verdict APPROVE; patch `925087e76d1aa12a5f21ec2c8366286aab38b3a1`.

skyl:
Windows CI compiled the candidate but its native test executable died before the harness with `STATUS_ENTRYPOINT_NOT_FOUND`. The branch-only linkage delta was the globally enabled Tauri `test` feature in both mock-runtime dev dependencies. Fixed forward at exact HEAD `2ae561debdfb815fab2f8eccc4af381adf91980e`: only the two mock-runtime dependencies/tests are target-gated off Windows; pure state/registry tests and the shipping Windows router remain compiled and policy-bound.

Independent review made all four manifest/cfg mutants RED and proved that either global-dependency mutant adds `tauri feature "test"` to the resolved Windows graph, while both restored Windows graphs exclude it. No workflow/job/selector was skipped or weakened. Plugin 150+1, ZUULI Rust 16, all wallet clippy, 103 boundary mutants, permissions, and the full actions-policy matrix are green. Exact verdict APPROVE; patch `2141aada10d17c9ebe105e3575a1bebe83ebcbda`.

skyl:
## `2ae561de` takes the mechanical fix. Here is what it costs, measured.

The gate is satisfied by teaching `wasm-boundary.mjs` the new string (+34) rather than by reverting the change that broke it. `tsconfig.build.json` is unchanged and **both** `build` and `typecheck` still point at it:

```json
{ "extends": "./tsconfig.json", "exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"] }
```

```
"build":     "npm run wasm:build && tsc -p tsconfig.build.json && vite build && npm run wasm:verify-dist"
"typecheck": "tsc --noEmit -p tsconfig.build.json"
```

I said in 5405304030 that this leaves nothing typechecking zuuli's test files. I have now measured it on this head rather than asserting it.

### The program the typechecker sees

```
$ tsc --noEmit -p tsconfig.build.json --listFiles | grep -c 'wallet/zuuli/src/'
164
$ tsc --noEmit -p tsconfig.build.json --listFiles | grep -c 'wallet/zuuli/src.*\.test\.'
0

$ find src -name '*.test.ts' -o -name '*.test.tsx' | wc -l
58
```

164 files in, 0 of the 58 test files among them.

### A type error in a test file, on this head

Appended to `src/lib/wallet/zuuallet-created-seed.test.ts`:

```ts
const reviewerTypeError: number = "not a number";
void reviewerTypeError;
```

Three runs:

| command | result |
|---|---|
| `tsc --noEmit -p tsconfig.build.json` — what `typecheck` and `zuuli.yml:203-204` run | **0 errors mentioning that file** |
| `tsc --noEmit` — the config `main` uses today | `zuuallet-created-seed.test.ts(292,7): error TS2322: Type 'string' is not assignable to type 'number'.` |
| `vitest run src/lib/wallet/zuuallet-created-seed.test.ts` | **10 passed, EXIT=0** |

vitest transpiles with esbuild, which strips types without checking them, and `vite.config.ts`'s `test:` block sets `env` only — no `typecheck` key. So after this PR there is no configuration anywhere in the repo under which that error is reported. 58 files stop being typechecked and the suite gets greener, which is the wrong direction for a change whose subject is protecting seed entry.

*(Diagnosing my own baseline, per the usual rule: the `tsconfig.build.json` run reports 5 unrelated errors — missing `zod` and `linkedom/worker` — because I symlinked `main`'s `node_modules` into a fresh worktree rather than running `npm ci`. Those 5 appear identically with and without my injection, and none of them names the injected file. The plain-`tsc` run reports those same 5 plus 5 more, one of which is the injected error. The contrast is unaffected.)*

### What I am asking for

Not a revert of the PR — the seed-entry work is good and I said so. Two lines:

```
"build":     "npm run wasm:build && tsc -p tsconfig.build.json && vite build && npm run wasm:verify-dist"
"typecheck": "tsc --noEmit"                  ← back to the full project
```

and revert the `wasm-boundary.mjs:141` pin edit for `build` only if the build genuinely needs the narrower config. That keeps every test file inside a typechecker while leaving `vite build` free of them, which is the ordinary reason a `tsconfig.build.json` exists. The five original CI failures then come back — and that is correct, because they were the typechecker reporting the real problem, which is the cross-project import at `src/lib/wallet/zuuallet-sensitive-entry-routes.test.tsx:56`:

```ts
import { RestoreWallet } from "../../../../zuuallet/src/pages/RestoreWallet";
```

`tsc`'s `include` only *seeds* the file list; anything reachable by static import joins the program regardless, so those four dots pull zuuallet's whole module graph into zuuli's typecheck. That is #691, and fixing it is what makes this PR pass honestly.

If #691 is too large to land inside this PR, say so and split it — but please do not ship the configuration that stops 58 files being checked, because nothing will notice when it stays that way.

skyl:
Published exact reviewed Windows fix-forward `4ea5f8a4b94c9aa447d1925d11474493b1adb3f0` (stable patch `69becdf317b460b9c1ff0be5ec23243b38be109b`).

Independent mutation review: APPROVE, 0 survivors. Deleting or misplacing either non-Windows cfg on the mock command imports or their argument types makes the live boundary fail for the exact import. Restored boundary corpus is 105/105; plugin tests are 150/150; production integration, rustfmt, and clippy `-D warnings` are green. Windows Cargo graphs remain free of Tauri's `test` feature. Only mock-test imports are gated; production code and pure lease/state tests remain compiled on Windows.

The preceding hosted run's sole substantive failure was the exact unused-import lint this commit removes; its Windows native tests, shared plugin, backend, and other Rust lanes were green.

skyl:
`4ea5f8a4` does not touch `package.json` or `tsconfig.build.json`, so the typecheck coverage point from 5406409963 stands unaddressed:

```
"typecheck": "tsc --noEmit -p tsconfig.build.json"     ← still the narrowed config
```

I have filed it as **#705** with the measurement, so it survives this PR's merge rather than living in a review comment. I am not going to keep holding the PR on it — the seed-entry work is good and the regression is reversible in two lines — but it should not be closed as "fixed by #690" when it lands, because it is the thing #690 introduced.

The short version for whoever picks up #705: 164 files in the typecheck program, 0 of the 58 test files; a deliberate `TS2322` in a test file is reported by plain `tsc --noEmit` and by nothing that CI runs; `vitest` passes it because esbuild strips types unchecked. Fix is `"typecheck": "tsc --noEmit"` over the full project, which brings back the five original failures — correctly, because those were the typechecker reporting `zuuallet-sensitive-entry-routes.test.tsx:56` (#691).

No other blocking findings from me here. Merge when `gate` and `rs / gate` are green on the exact head; it is currently `BEHIND` with CI in flight.

PR #694, opened 2026-08-25
feat(msg-identity): the seed-derived messaging key hierarchy (ARCHITECTURE.md §4.2)

skyl:
`rs/crates/f2z-msg-identity` — the seed-derived messaging key hierarchy of
[`docs/e2ee/ARCHITECTURE.md` §4.2](../blob/main/docs/e2ee/ARCHITECTURE.md#42-derivation-proposed).
This is Phase 1's client half of #311.

---

## ⚠️ Changing any label or personalization after this ships changes every user's identity

Every constant in `src/labels.rs` — the two BLAKE2b personalizations and the four
HKDF leaf labels — is an input to a **one-way derivation from the user's BIP-39
seed**. Moving one byte of one of them:

- moves that user's `IdentitySigningKey` to a different key;
- invalidates their handle's directory entry, which is in an **append-only** log;
- breaks every peer's pinned safety number;
- and leaves **no migration path**, because `KT.md` §4.4 requires the *outgoing*
  identity key to sign a rotation — the user would have to still hold the key
  the change just orphaned.

That is why §4.2 marks them and why they have to be right *now*, before anyone
has an identity to lose. `tests/derivation_vectors.rs` is the tripwire: any edit
to a constant turns it red.

---

## What landed

| Piece | Where |
|---|---|
| `MSK = BLAKE2b-512(personal = "Free2zMsg_MSTRv1", S)` | `src/node.rs` |
| `CKDh(node, i) = BLAKE2b-512(personal = "Free2zMsg_CKDv1_", cc_node ‖ 0x11 ‖ I2LEOSP32(i))`, **hardened only** | `src/node.rs` |
| `account_node = CKDh(CKDh(CKDh(MSK, 32'), 133'), account')` | `src/node.rs` |
| The four leaves by `HKDF-Expand(PRK = ik_account, info = label, 32)` | `src/account.rs` |
| Per-device `DeviceSignatureKey` / `QueueKey` from the OS CSPRNG | `src/device.rs` |
| `DeviceCredential` construction and signing | `src/credential.rs` |
| The personalized-BLAKE2b primitive, and the trap it avoids | `src/blake.rs` |

`#![forbid(unsafe_code)]`, `no_std` + `alloc`, no I/O, no clock, **MIT** (linked
by clients; only the server binaries are AGPL). Added to `rs.yml`'s `rs_wasm`
job and registered with `scripts/check-rust-toolchain.sh`.

## The non-negotiables, and where each is tested

| Property | Test |
|---|---|
| **No Zcash key is reachable from any messaging key path** (#311's last acceptance criterion) | `tests/zcash_separation.rs` — transcribes ZIP 32's Sapling *and* Orchard master and hardened-child derivations plus `PRF^expand`'s `ask`/`nsk`/`ovk`, walks all three trees over **one seed** at `32'/133'/0'` and four sibling indices, and asserts the two 32-byte sets are **disjoint**. Plus: identical indices do not produce identical nodes, and the seed does not survive into any derived value. |
| **Domain separation everywhere** | `src/labels.rs` — closed label set, prefix-freeness, "every personalization is exactly 16 bytes", "none is a Zcash domain". `scripts/check-hash-domain-labels.mjs` passes over the union (49 labels, 1 556 tracked files). |
| **Identity restorable from the mnemonic; device keys are not** | `tests/no_seed_derived_device_keys.rs`. There is no `DeviceSignatureKey::from_seed` and `generate<R: CryptoRng>` is the only constructor. |
| **Non-repudiation and deniable-style authentication never share a key** | Structural: **no key type exposes a general `sign(&[u8])`.** `IdentitySigningKey` signs a `DeviceCredentialTBS` and a `RotationProofTBS`; `CeremonySigningKey` signs a §11 ceremony payload; `DirectoryAuthKey` signs a `DirectoryEntryTBS`; `BackupWrapKey` signs nothing. Table in `src/account.rs`. |
| **Zeroization**, with its honest limit | `Zeroizing` on every intermediate, `ZeroizeOnDrop` on `ExtendedNode`, dalek `SigningKey` for the rest. Under WASM this is **best-effort and cannot be a guarantee** — the JS engine's GC may have copied a buffer before Rust overwrote the copy it can see. Stated in the crate note; not claimed otherwise. `src/blake.rs` also names the one thing it does *not* wipe (`block-buffer` exposes no clear) rather than glossing it. |
| **Redacting `Debug`, including the decimal trap** | `tests/redaction.rs` checks lower hex, **upper** hex, base64url, a full decimal byte list, 4-byte prefixes of each, and any 8+ character hex-looking run — for every key type, `ExtendedNode`, the issued credential, and the request. `f2z-codec`'s workspace `Debug` source scan passes over the new crate. |

## Test vectors — a deliverable, not a nicety

`tests/derivation_vectors.rs` pins the whole hierarchy from **BIP-39's published
all-`abandon` seed** (so a second implementation can start from the mnemonic
rather than from our say-so):

```
seed         5eb00bbddcf069084889a8ab9155568165f5c453ccb85e70811aaed6f6da5fc1
             9a5ac40b389cd370d086206dec8aa6c43daea6690f20ad3d8d48b2d2ce9e38e4
MSK          74f055750feeefebb248b31b88aae0b2b9b0b0db1c0f520eb831d30243c87096…
  32'        2ca48d29539f895d3566a0de26c1ea0274b5c698b0d371836b448aaeaaca8b1c…
  133'       020d1c3484f121f46c06bd68426ce3ceba7a5fd3b6f08665eb06dec300f45073…
  0'         21e92468197afc2b0b5390f00ae5dcff4a686d1f973bb15d56081d031eb63679…  ← ik_account
ISK.public   cf7011daaf3200aee965329b1976ab6e3d397402f914fdd8e27e0a4147195cdd
CSK.public   96c347ec69459beab59e76fa7571e27692da7ae85384a9546e3ef83f822eb902
DirAuth.pub  9084ebcc6f68eff30bdb3765780714fd49a67f98b0afa14a88f63aa117d44606
BackupWrap   13ace59b8f2e13e5f475bc89348328a5e6cde431ecb9dffb1529895aed44f3d7
DeviceCredentialTBS  1332 bytes, field-by-field at KT.md §4.1's declared offsets
signature    6bd3f57d6685b1bdf3ee265783feed76acc8fa89af5b0f8bcc36d2bc30222e32
             994b6286fc98bb4e78b31b172110d26970e54022b55366bcc43f2dde0b22780b
```

**Every expected value was computed by something that is not this crate** —
Python's `hashlib.blake2b(person=…)`, a hand-written RFC 5869 §2.3
`HKDF-Expand`, and `cryptography`'s Ed25519 — following the rule
`f2z-codec/tests/wire_vectors.rs` states. The full derivation transcript is
reproduced in the test file so a reviewer can re-run it. Ed25519 signing is
deterministic (RFC 8032 §5.1.6), so the credential signature is a *known
answer*, not one valid answer among many.

The node levels are each asserted **twice**: once against the shipped
`master_node`/`ckd_hardened`, once against a second transcription of §4.2's
pseudocode inside the test. Three independent sources, not one.

## The acceptance test: `f2z-kt-core` accepts what this crate issues

`tests/kt_core_agreement.rs` puts a `DeviceCredential` issued here, inside a
`DirectoryEntry` authorized by this crate's `DirectoryAuthKey`, through
**`f2z_kt_core::validate_submission`** — the same function the log server runs
before it hands anything to `akd`. §4.4 rule 8 is the rule under test. All five
cases pass:

- a v1 `same_key` entry is **accepted**;
- a credential issued by a *different* identity is **rejected** with
  `BadAuthorization` (so the accept above is not vacuous);
- two devices of one account in one entry are accepted (ADR 0002's fan-out);
- a chained v2 entry is authorized by the same seed-derived `DirectoryAuthKey`;
- a `RotationProof` signed here verifies under the **outgoing** identity key and
  not under the incoming one.

`DeviceCredentialTBS` is `f2z-kt-core`'s type, **not a second copy**. A
construction crate owning its own copy would be one field-order edit away from
issuing credentials the directory rejects, and that failure would land on users
rather than on CI.

---

## Ambiguity calls — every one, with reasoning

**1. `I2LEOSP32(i)` encodes the hardened index, not the ordinal.** §4.2 writes
`32'`, `133'`, `account'` in ZIP 32's notation and says "hardened only". ZIP 32
serializes a hardened child index as the full `i + 2^31`, so `32'` is
`0x8000_0020` → `20 00 00 80` little-endian. The alternative reading — encode
the bare ordinal, and let "hardened only" be a property of the construction
rather than of the bytes — is self-consistent and produces a **different tree**.
`HardenedIndex` takes the ordinal and sets the bit itself, so there is exactly
one spelling of a path; passing an already-hardened value is an error rather
than a silent double-hardening. **This is the vector-affecting call, and it is
the one to overrule now if it is to be overruled at all.**

**2. HKDF-Extract is skipped; only Expand runs.** §4.2 says `HKDF-Expand(PRK =
ik_account, …)` and that is implemented literally. `ik_account` is already half
of a BLAKE2b-512 output, so it is already uniform and an Extract would buy
nothing — but "we skipped Extract" is the kind of deviation a reader should stop
at, so it is written down in `src/account.rs`.

**3. `L = 32` for all four leaves.** §4.2's table gives a length only for
`BackupWrapKey` ("32 B"). The other three are Ed25519, whose secret seed is 32
bytes (RFC 8032 §5.1.5), so 32 is the only sensible reading.

**4. `handle` carries no `@`.** §4.2 writes `"@alice"`; `KT.md` §4.1 declares
`opaque handle<1..30>` over `[a-z0-9_]`. The `@` is display sugar and `Handle`'s
charset check makes it unrepresentable.

**5. ISK vs. DirectoryAuthKey — `KT.md` §4.4's narrowing is what shipped.**
§4.2's table says ISK "signs device credentials **and directory entries**" while
DirectoryAuthKey "authenticates directory updates" — those overlap. §4.4 splits
them explicitly (ISK signs credentials and rotation proofs; DirectoryAuthKey
signs the entry envelope) and gives the blast-radius reason. This crate
implements the narrowing, and the type system enforces it.

**6. Seed length is accepted over 16..=64 bytes.** BIP-39 always produces 64.
The range exists so a caller holding a truncated or non-standard seed gets a
verdict; silently accepting a 4-byte "seed" is how a test fixture becomes a
production identity.

**7. `CeremonySigningKey::sign_ceremony_payload` takes raw bytes.**
§11 has not fixed a ceremony transcript framing, and inventing one here would be
redesigning the spec. The method signs what §11 defines; what it *buys* is that
the key choice is unambiguous.

## Scope call: the `DeviceInitKey` is not in this crate

§4.2's per-device table has three rows and this crate implements two. The
missing one is `DeviceInitKey` / the KeyPackage keys, typed **X-Wing hybrid
(X25519 + ML-KEM-768)**.

It is absent deliberately. §3.1 decides the X-Wing implementation is Cryspen's
**libcrux, reached through OpenMLS's crypto provider** (`openmls_rust_crypto`
does not implement X-Wing at all), so the keypair is generated by the MLS layer
as part of a `KeyPackage`, by the same provider that will later decapsulate with
it. Generating it here would mean a **second X-Wing implementation** in the tree
for the sake of tidiness in a table — a hybrid KEM whose combiner is written
twice is exactly the implementation-divergence class ADR 0001 exists to avoid.

What this crate needs from that key is its public half, and it takes it as
opaque bytes — which is also how `KT.md` §4.1 declares it
(`opaque device_kem_pk<1..2^16-1>`, length-validated and nothing more), because
the party that *interprets* it is an MLS peer validating a `LeafNode`. Full
reasoning in `src/device.rs`'s module note.

---

## 🔍 A spec observation worth a reviewer's attention (not coded around)

**§4.2's `CKDh` does not feed the parent key into the child derivation.**

ZIP 32's hardened derivation is
`PRF^expand(c_par, [0x11] ‖ sk_par ‖ I2LEOSP_32(i))` — the parent **spending
key** is inside the hash. §4.2's is
`BLAKE2b-512(personal, cc_node ‖ 0x11 ‖ I2LEOSP32(i))`, and `msk` / `ik_node`
appear nowhere in it.

Implemented **exactly as written**, because implementing something else would
mean this crate and the specification disagree about what a user's identity
*is*. The consequences, stated so nobody has to rediscover them:

- The key half of every node above the account level is **dead** — `msk` itself
  is never read by anything.
- The hierarchy's entire secret entropy reaches `ik_account` through `cc_msk`
  alone: 256 bits of a BLAKE2b-512 output. Not a weakness, but it is half the
  state ZIP 32 carries.
- Anyone holding an intermediate **chain code** can derive that node's whole
  subtree without its key. Since a chain code is only ever held alongside its
  key, no practical capability separates — but it is a real deviation from the
  ZIP 32 idiom §4.2 invokes, and an auditor comparing the two side by side will
  ask about it.

`src/node.rs` documents it and `ckd_reads_only_the_chain_code` asserts the
behaviour, so a later change is loud. **If §4.2 intended `sk_par` to be in the
preimage, this is the last cheap moment to say so** — it changes every derived
key, and the fix is one line plus regenerated vectors. Flagging rather than
silently "improving" it, per the process that has already caught several spec
defects.

Neighbouring note: `f2z-codec`'s documented `Blake2bMac512` trap has a sibling
here. `blake2 0.10`'s obvious spelling for "BLAKE2b-512 with a personalization",
`Blake2bMac512::new_with_salt_and_personal(&[], &[], PERSONAL)`, **does not
compute unkeyed BLAKE2b**: the constructor seeds the buffer with a padded key
block unconditionally, so an empty key still prepends 128 zero bytes. The digest
looks right and agrees with no other implementation on earth. `src/blake.rs`
drives `Blake2bVarCore` directly and pins the empty-message answer against
Python's `hashlib`.

---

## Verification

| Check | Result |
|---|---|
| `cargo test --locked --all-targets` (rs workspace) | ✅ exit 0 |
| `cargo test --locked --doc` | ✅ exit 0 |
| `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-msg-identity` | ✅ |
| `scripts/check-rust-fmt.sh --root rs` | ✅ 11/11 formatted |
| `scripts/check-rust-clippy.sh --root rs` | ✅ 11/11 clippy-clean |
| `scripts/check-rust-deny.sh --root rs` | ✅ advisories/bans/licences/sources ok |
| `scripts/check-rust-toolchain.sh` + `--self-test` | ✅ 14 drift cases caught |
| `scripts/check-hash-domain-labels.mjs` + `--self-test` | ✅ 49 labels prefix-free, 17 self-test cases |
| `f2z-codec` workspace `Debug` + strict-verification scans | ✅ reached and passed the new crate |

New dependencies: `hkdf 0.12`, `hmac 0.12` (transitive), `sha2 0.10` (already in
the graph), `rand_core 0.10` (already in the lock as dalek's optional dep — its
`rand_core` **feature stays off**), `zeroize` with `derive`. All MIT/Apache-2.0;
`cargo deny` licences ok with no change to `rs/deny.toml`.

Refs #311, #305.


skyl:
Reviewed against `397195ad`. Baseline `cargo test -p f2z-msg-identity`: **EXIT=0, 62 passed**.

The crate names five properties and says *"Each has a test, and each is a thing an auditor checks first."* I took that as the review plan and went at each one.

**Two are enforced workspace-wide, and I confirmed both reach this crate by mutation rather than by reading.**

*Property 5 — redacting `Debug`.* Adding `Debug` to `BackupWrapKey`'s derives:

```
cargo test -p f2z-codec --test workspace_debug_scan  =>  EXIT=101
  no_type_derives_debug_while_holding_raw_bytes ... FAILED
  f2z-msg-identity/src/account.rs: `BackupWrapKey` derives Debug while holding `[u8;` —
  a derived Debug renders that as a decimal byte dump. Hand-write Debug and report the
  field by length, as `Payload` and `Canonicalized` do.
```

It names the file, the type, the pattern, and two existing examples to copy from.

*Property 2 — domain separation.* Making one of this crate's labels a proper prefix of an existing one:

```
node scripts/check-hash-domain-labels.mjs  =>  SCANNER_EXIT=1
  "free2z/kt/v1" is a proper prefix of "free2z/kt/v1/value": H("free2z/kt/v1", "/value" || y)
  is bit-identical to H("free2z/kt/v1/value", y) … free2z/kt/v1 at
  rs/crates/f2z-msg-identity/src/labels.rs:75
```

**Worth saying explicitly, because it is the good news:** both scanners are population-derived rather than hand-listed — `workspace_debug_scan.rs` walks `rs/crates/` and separately asserts it *reached* every crate, and the label checker reads all 1556 tracked files. So a brand-new crate is covered on the day it lands, with nobody remembering to add it. I went looking for the opposite (the label checker has a `file:` array that this crate is absent from) and was wrong — that array is a registry of documented sites, not the scan population.

**The one thing I could not find a test for is property 4.** *"Non-repudiation and deniable-style authentication never share a key … No key type in this crate exposes a general `sign(&[u8])`."* Properties 1, 2, 3 and 5 each have a named test file or a workspace scanner; property 4 has neither, so "Each has a test" is four-for-five.

The substance is fine and I am not asking for a design change. But the summary line is looser than the code it describes: `sign_ceremony_payload(&self, transcript: &[u8])`, `sign_queue_command(&self, transcript: &[u8])` and `sign_mls_content(&self, framed_content_tbs: &[u8])` all take arbitrary bytes. What separates them is the **key**, not the payload — which `sign_ceremony_payload`'s own doc says plainly:

> The transcript's framing is §11's, not this crate's … What this method is for is making the *key choice* unambiguous.

That is the honest statement of the property, and it is the one the summary should carry: **different keys for different purposes**, not "no general sign". As written, a reader who checks the claim finds three methods taking `&[u8]` and has to work out for themselves that the guarantee still holds.

If it is worth a guard rather than a wording fix, the cheap one is in the idiom the crate already relies on twice: assert that no `impl` in this crate exposes a method named exactly `sign`, next to the two workspace scanners that already reach it. That is weaker than it sounds for the reason above, which is itself an argument for fixing the sentence instead.

**Not re-derived, and worth noting as unchecked by me:** properties 1 and 3 have their own test files (`zcash_separation.rs` computing a ZIP-32 derivation independently, and `no_seed_derived_device_keys.rs` asserting two devices restored from one mnemonic are two devices), and I read them but did not mutation-test them. Saying so rather than letting their absence from this comment read as a verdict.

`rs / gate` was still running when I looked, so I have not merged.

PR #695, opened 2026-08-25
fix(release): bind Linux Cargo runtime evidence

skyl:
Refs #379

## Summary

- instrument Linux Tauri builds with pinned, checksummed `cargo-auditable` 0.7.5 and extract precise format-8 runtime evidence from each shipped AppImage/deb/rpm payload
- patch the upstream precursor conversion so the exact binary root survives same-package lib/bin deduplication, then reconcile the embedded graph against locked target/feature-filtered Cargo metadata
- bind executable evidence, dependency graph, lockfile, artifact, SBOM, and verification digests for all three Linux formats in packaging and protected release workflows
- keep the unstable Cargo SBOM unlock out of Cargo-launched rustc processes with a fail-closed wrapper
- integrate #706’s exact five-fixture selector/completion authority and fail closed on malformed/incomplete graphs, unsupported sources, missing options, producer failures, or verifier divergence

This is the Linux Cargo-runtime slice of #379. Other platform/evidence work remains tracked by the issue.

## Exact review head

- head: `3bd6274bc297e45a3e2804090d3a62689516c145`
- base: `9efb7e7b164c953eac1ae43c9df0e03fd52b0c05`
- cumulative stable patch: `5587aeb6aa8489b6bc0e13fb3f9f0233d07b23f4`

## Adversarial verification

Real RED controls include the root-patch removal, format downgrade, absent instrumentation, wrong/unknown/duplicate/unreachable graph entries, target/feature drift, every installer-stage bypass, valid installer output with producer exit 17, deletion or softening of the rustc bootstrap scrub, feature/target authority drift, and fifth-fixture rename/deletion/broaden/skip/completion mutations.

Restored local proof: host artifact suite 17 pass / 5 honest Linux skips; exact Ubuntu 24.04 + Rust/Cargo 1.97.1 + patched cargo-auditable 0.7.5 selected runner 5/5, including Cargo -> precise ELF -> deb -> finalized SBOM -> fresh verification. Hosted `Desktop / Linux packages` on the exact head is required before merge.


skyl:
## Review of `05ec982a`

Read the whole diff; ran what I could on this host. No blocking defect. One structural finding filed as #696, and one thing I want to be explicit about that I could not check here.

### What I could verify

`node --test scripts/artifact-sbom.node-test.mjs` on the PR head: **EXIT=0, 17 tests, 12 pass, 5 skipped.**

The five skips are the Linux fixture tests, gated on `process.platform === "linux" && <tools>.every(commandExists)`. That gate is honest and the tools resolve on CI — I confirmed against a completed `zuuli / frontend` job on main (`97678255136`) that the four pre-existing Linux fixture tests do execute there; the only skips in that log are the two `darwin`-only ones.

### What I could not check, plainly

`canBuildCargoRuntimeFixture` also requires `cargo-auditable`, which `zuuli.yml`'s frontend job does not install — only `zuuli-packaging.yml:391-395` does. So **test 2, "real cargo-auditable executable evidence is complete, lock-bound, and mandatory", is the load-bearing test for this entire PR and it cannot run on macOS or in the frontend job.** It runs in `ZUULI / packaging smoke`, which does trigger on PRs touching `wallet/zuuli/**`.

To make the limitation concrete rather than assumed, I deleted the fail-closed arm of `linux-cargo-runtime.mjs:83-91`:

```js
function sourceKind(source) {
  if (source === null) return "local";
  if (source === "registry+https://github.com/rust-lang/crates.io-index") return "crates.io";
  return "crates.io";                 // was: throw new Error(`...non-shipping package source: ${source}`)
}
```

Full suite still **EXIT=0, 12 pass, 5 skipped**. That is the expected result on this host, not a finding — but it means my green run says nothing at all about `linux-cargo-runtime.mjs`. The evidence for the 531 new production lines is the packaging job, and that is where I am reading the verdict from, not from anything I ran.

Your description's "independently approved" and "all seven blobs byte-identical" lines are not evidence I can act on either — I am taking the packaging job's result.

### Finding: #696 — the packaging step selects its tests by name pattern

`zuuli-packaging.yml:400` runs the fixtures via `--test-name-pattern='real AppImage, deb, and rpm|AppImage inspection|AppImage listing|real deb with an escaping|real cargo-auditable executable evidence'`. That exact string is pinned as policy at `artifact-sbom.mjs:1765`/`:1920`, and `artifact-sbom.node-test.mjs:2373-2381` proves the pin is load-bearing by substituting `'nonexistent'`. Good — the workflow cannot drift.

But nothing binds the pattern to the test names it selects, and `node --test` exits 0 when a pattern matches nothing. Dropping one word from `artifact-sbom.node-test.mjs:584`:

```
before: # tests 5   EXIT=0
after:  # tests 4   EXIT=0
workflow-contract test after the rename: # pass 1   EXIT=0
```

The newest and most important test in the set silently stops running, and every gate stays green. Details and a suggested fix in #696 — bind the direction that is currently unbound: assert each alternative in the pinned pattern matches at least one registered test name, the way `rs/crates/f2z-codec/tests/workspace_debug_scan.rs:261-269` asserts its scan reached every crate rather than trusting a hand-written list.

Not a blocker for this PR — it is a pre-existing shape that this PR extends by adding a fifth long English sentence to the pattern.

### Two smaller notes, neither blocking

- `parseFeatures` accepts a feature literally named `none` or `default`, and `cargoArguments:107-113` gives both magic meanings — `["none"]` becomes `--no-default-features` with the feature itself never passed. ZUULI has no such feature today; worth a comment if the parser is ever reused.
- `parseCargoTree:119` splits depth from body with `/^(\d+)(.+)$/`. Crates.io permits names beginning with a digit, and greedy `\d+` would absorb the leading digit of such a name. The failure mode is a wrong package identity that then fails reconciliation against the embedded evidence — fail-closed, so an error rather than a false pass. Noting it rather than filing it.

### Verdict

Approve on the packaging job's result. I will merge once `gate` and `rs / gate` are green on this exact head and `ZUULI / packaging smoke` has reported. The PR is currently `BEHIND` (`strict: true`) so it needs a rebase first, which restarts CI.

skyl:
## Blocking — `Desktop / Linux packages` fails, and the fixture is shaped so it cannot see why

Job `97692384885`, step **Build packages**:

```
error: no such command: `+1.97.1`
beforeBuildCommand `npm run build` failed with exit code 101
Error beforeBuildCommand `npm run build` failed with exit code 101
```

### Mechanism

`scripts/cargo-auditable-cargo.sh` prepends `auditable` unconditionally:

```sh
: "${ZUULI_REAL_CARGO:?...}"
exec "$ZUULI_REAL_CARGO" auditable "$@"
```

`run-with-cargo-auditable.sh` symlinks that as `cargo` at the front of `PATH`, so every `cargo` in the subtree becomes `cargo auditable <args>`.

`tauri build` runs `beforeBuildCommand` = `npm run build` = `npm run wasm:build && …`, and `scripts/wasm-build.mjs:104-107` invokes:

```js
spawnSync("cargo", [`+${channel}`, "build", "--locked", "--release", "--target", target, …])
```

Under the shim that becomes `cargo auditable +1.97.1 build …`. rustup's `+toolchain` selector is only meaningful as cargo's **first** argument; behind `auditable` it is read as a subcommand, hence the error. The wrapper breaks every `+toolchain` invocation it wraps, and the WASM build is the first one the packaging build reaches.

### Why the new fixture did not catch it

`artifact-sbom.node-test.mjs:683-692` — the one place the wrapper is exercised — calls it as:

```js
execFileSync(resolve(scriptDirectory, "run-with-cargo-auditable.sh"),
  ["cargo", "build", "--release", "--locked", "--offline"], …)
```

No `+toolchain`. So the fixture proves the wrapper works on a shape the real build never uses, and is blind to the shape the real build always uses. The 956 lines of new test go to real trouble elsewhere — a genuine crates.io dependency, an actual ELF, a real deb extraction — which makes this gap easy to miss, but it is the whole difference between the fixture and production.

This also means the description's "the real amd64 wrapper → cargo-auditable ELF → deb extraction → finalized SBOM → fresh verification path killed every production guard mutation" and this failure cannot both be true of the pushed head. Whatever was run, it was not the path `Build packages` takes.

### Fix

Hoist a leading `+toolchain` past `auditable` in `cargo-auditable-cargo.sh`:

```sh
case "${1:-}" in
  +*) toolchain=$1; shift; exec "$ZUULI_REAL_CARGO" "$toolchain" auditable "$@" ;;
esac
exec "$ZUULI_REAL_CARGO" auditable "$@"
```

and extend the fixture to call the wrapper **with** a `+<channel>` first argument, so the regression is caught by a test rather than by a 20-minute packaging job.

Worth deciding separately: the shim currently instruments *everything*, including the `wasm32-unknown-unknown` build, which ships no Linux ELF and has no runtime graph to embed. Scoping the wrapper to the Tauri Rust build — rather than to the whole `beforeBuildCommand` subtree — would remove a class of surprises like this one instead of patching the first member of it. If the shim stays global, at minimum it should pass through non-build subcommands (`metadata`, `tree`, `--version`) untouched.

### Rest of the review

Unchanged from my previous comment: the module reads well and fails closed, #696 is filed for the name-pattern coupling, and I could not run the load-bearing Linux fixture on this host. Holding the merge on this.

skyl:
The first real Linux package build found an authentic wrapper defect: Tauri's frontend hook invokes `cargo +1.97.1`, but the audit wrapper produced `cargo auditable +1.97.1`, which Cargo rejects. Fixed forward without rewriting history at exact HEAD `2ba72f035300e160f9ddf4ef7551624a1eb3bf42`.

Independent review made five wrapper mutations RED: selector handling deletion, selector/auditable reordering, ordinary audit insertion deletion, double-selector delegation deletion, and bare-`+` misclassification. Real Cargo reproduced the original status 101 and proved the restored `cargo +1.97.1 auditable ...` path. The exact pinned-image CI-selected suite is 6/6 green, including AppImage/deb/rpm, real cargo-auditable ELF→deb, invalid UTF-8, and escaping-symlink controls. Exact verdict APPROVE; patch `9fd9fc2a0e16dff63941f20e9611bd2d68cb7bc2`.

skyl:
## The `+toolchain` fix is correct and its test is load-bearing — but the job now fails deeper, in the reconciliation itself

### Verifying the fix, both halves

`scripts/cargo-auditable-cargo.sh` against a `ZUULI_REAL_CARGO` that just echoes its argv:

```
new shim, with a selector:     +1.97.1 auditable build --locked --release --target wasm32-unknown-unknown
new shim, without a selector:  auditable build --release --locked
old shim, with a selector:     auditable +1.97.1 build --locked        ← `error: no such command: +1.97.1`
```

Reverting `cargo-auditable-cargo.sh` to the five-line version takes `real cargo-auditable executable evidence wrapper preserves Cargo dispatch` from `# pass 1` to `# fail 1`, EXIT=1. The test is unskipped, so it runs in `zuuli / frontend` on every PR and not only in packaging — that is the right place for it, and it closes the gap I raised. The nested `+*` pass-through is a nice touch: `cargo +a +b` gets rejected by cargo rather than mangled by the shim.

### The new failure

`Desktop / Linux packages`, job `97697814477`, head `2ba72f03`. **Build packages now passes.** The step that fails is **Bind Linux shipped-artifact SBOMs**, on the first `finalize-artifact` call:

```
embedded Cargo runtime graph contains unknown or wrong-version package adler2 2.0.1
```

That is `reconcileRuntimePackages` — the heart of this PR — saying the graph cargo-auditable embedded in the shipped ELF contains a package the graph derived from `cargo tree` does not.

### Where `adler2` goes missing

I ran the exact arguments `cargoArguments()` builds, against `wallet/zuuli/src-tauri`:

```
cargo tree --locked --manifest-path Cargo.toml --target x86_64-unknown-linux-gnu \
           --edges normal --prefix depth --format '{p}|{f}'
```

`adler2` appears **once** in the whole 1898-line tree:

```
2tauri-macros v2.6.3 (proc-macro)|compression      ← depth 2
…
6flate2 v1.1.9|any_impl,default,miniz_oxide,rust_backend
7miniz_oxide v0.8.9|default,simd,simd-adler32,with-alloc
8adler2 v2.0.1|                                     ← depth 8, inside that subtree
```

`parseCargoTree:115-155` sets `skippedProcMacroDepth = 2` at `tauri-macros` and `continue`s every line with `depth > 2` until the depth returns to 2 or less. `adler2` is at depth 8 and has no other appearance, so it never enters `expected`.

I simulated that skip over the full tree output. **93 packages are dropped and appear nowhere else in the kept set.** Most are correctly dropped — `syn 1.0.109`, `quote`, the `*_derive` family, `darling`, `proc-macro-error`. But the list also holds a run of ordinary libraries whose only expanded appearance happens to sit under a proc-macro:

```
adler2 2.0.1      crc32fast 1.5.0   fdeflate 0.3.7    flate2 1.1.9
miniz_oxide 0.8.9 png 0.17.16       ico 0.5.0         simd-adler32 0.3.10
base64 0.21.7     itertools 0.14.0  rustc-hash 2.1.3  foldhash 0.2.0
schemars 0.8.22   cargo_metadata    camino 1.2.4      toml_edit ×3
html5ever 0.38.0  markup5ever       selectors 0.36.1  cssparser 0.36.0
```

`adler2` is simply the first one `reconcileRuntimePackages` reaches.

### Why this is a design question and not a one-line fix

Two readings, and they need different repairs:

1. **The embedded graph includes build-time dependencies while `--edges normal` excludes them.** Then the two sides are measuring different populations and the comparison is wrong in general, not just for `adler2`. `validateAuditableDocument` would need to filter on cargo-auditable's dependency kind before reconciling.
2. **The proc-macro subtree skip is unsound as written.** `cargo tree` expands each package fully at its *first* occurrence and prints `(*)` elsewhere. The parser therefore loses any package whose single expanded occurrence lands inside a proc-macro subtree — and which occurrence that is depends on cargo's traversal order, which changes whenever the dependency graph changes. A rule with that property will keep breaking at unrelated times.

Either way, the whole class deserves the treatment `f2z-codec/tests/workspace_debug_scan.rs:261-269` gives its population: derive the set and assert you *reached* everything, rather than dropping members silently. A skip that removes 93 packages should say how many it removed.

### Caveat on my numbers

My `cargo tree` ran against local `main`'s `wallet/zuuli/src-tauri/Cargo.lock`, not this branch's — this PR does not touch that lockfile, so the two should be identical, but the 93 and the depths are from main's resolution. The `adler2 2.0.1` identity and the `tauri-macros` ancestry match what CI reported.

Still holding the merge. #696 stands as filed; the wrapper fix does not touch the name-pattern coupling.

skyl:
## Root cause found: the Linux build has no `--target`, but the evidence is derived with one

My previous comment offered two readings. Neither is right, and the first was wrong on a fact I can correct: `reconcileRuntimePackages` **does** filter build-kind entries — `linux-cargo-runtime.mjs:419`, `if (entry.kind === "build") continue;` — so the embedded graph is not polluted with build dependencies. Scratch that half.

The actual cause is a flag asymmetry between the build and the evidence.

### The two commands disagree about the target

`zuuli-packaging.yml:407-409` — macOS names a target, Linux does not:

```yaml
if [[ "${{ runner.os }}" == macOS ]]; then
  ./node_modules/.bin/tauri build --target universal-apple-darwin --bundles '…' -- --locked
else
  scripts/run-with-cargo-auditable.sh ./node_modules/.bin/tauri build --bundles '…' -- --locked
fi
```

`:480-482` and `:537-539` (and `zuuli-release.yml:1212-1214`, `:1234-1236`) — every `finalize-artifact` and `verify-artifact` call names one:

```
--cargo-target=x86_64-unknown-linux-gnu
```

So the shipped Linux ELF is a **host** build, and the graph it is reconciled against is derived for an **explicit target**. Cargo resolves those two differently: with `--target` it decouples host artifacts (proc-macros, build scripts and their dependency trees) from target artifacts; without it, on a host build, it unifies them.

### Measured

I ran the parser's own proc-macro-skip logic over both trees of `wallet/zuuli/src-tauri`:

```
WITH    --target x86_64-unknown-linux-gnu   kept 410   dropped-only  93   adler2 kept? False
WITHOUT --target                            kept 316   dropped-only 114   adler2 kept? True
```

And the packages from that watchlist that survive:

```
WITH    --target:  base64
WITHOUT --target:  adler2, base64, crc32fast, fdeflate, flate2, miniz_oxide, png, simd-adler32
```

That is precisely the icon/compression chain the failure names. With `--target`, `flate2` → `miniz_oxide` → `adler2` is reachable only beneath `tauri-macros (proc-macro)` and the skip eats it. Without `--target` — the way the artifact is actually built — the same packages sit outside any proc-macro subtree and are kept.

So `cargo-auditable` is right: `adler2 2.0.1` really is in the shipped binary. `expectedRuntimeGraph` is the side that is wrong, and it is wrong because it was asked about a build that never happened.

**Caveat on my numbers:** my no-`--target` run resolved for this machine's host (`aarch64-apple-darwin`), not `x86_64-unknown-linux-gnu`, so 316/114 are the macOS graph. What is demonstrated is the *structural* effect — adding or removing `--target` moves `adler2` across the proc-macro boundary — and that is exactly the error CI reported.

### Fix

Two coherent options; they are not equivalent.

1. **Give the Linux `tauri build` its target**, `--target x86_64-unknown-linux-gnu`, matching what macOS already does. Then `--cargo-target=x86_64-unknown-linux-gnu` describes the build that happened, the graphs agree, and the proc-macro skip becomes sound because host artifacts really are decoupled. This is the one I'd take.
2. Drop `--cargo-target` for Linux and derive the expected graph the same way the build resolves. This makes the check pass while keeping the unified build.

Option 1 is worth preferring for a reason beyond bookkeeping. Under (2), `flate2`, `png`, `ico` and friends stay linked into the shipped Linux ELF purely because host/target unification put them there — cargo-auditable reports them as runtime-kind, which is what the reconciliation is reading. The macOS artifact, which does pass `--target`, does not carry them. For a PR whose subject is binding evidence to what actually ships, making the two platforms resolve the same way is the better answer, and it shrinks the Linux artifact's real dependency surface rather than documenting a larger one.

### On the proc-macro skip

Even after the target flag is fixed, #696's concern applies to `parseCargoTree` too: a filter that removes 93 packages should say how many it removed, and the reconciliation should assert it *reached* the population rather than silently dropping members. `f2z-codec/tests/workspace_debug_scan.rs:261-269` is the idiom. That is a hardening request, not a blocker — the target flag is the blocker.

skyl:
Published exact reviewed fix-forward `be434b4f26440c6cd52bd28246d91499390ffe75` (stable patch `88c32e21957c42890b49f5ad2b06d440040d76f5`).

Independent mutation review: APPROVE, 0 survivors. Real red controls verified:
- embedded format requirement `8 -> 1` recreates the hard unknown-package failure;
- deleting each of `CARGO_BUILD_SBOM`, `CARGO_UNSTABLE_SBOM`, or `RUSTC_BOOTSTRAP` produces format 1 instead of 8;
- deleting the app `custom-protocol` mapping makes the exact Cargo command fail;
- deleting the shipping build feature or reverting finalize to default makes workflow policy fail.

Restored green: exact Ubuntu 24.04/Rust 1.97.1/cargo-auditable 0.7.5 named suite 6/6, full host artifact suite 13/13 applicable, action pins, workflow gates, Rust toolchain, Linux image, and public boundary self/live checks. The precise format-8 graph excludes the inactive runtime and classifies the proc-macro helpers build-only; unknown/omitted/version guards remain fail-closed.

skyl:
## `be434b4f` — my `--target` diagnosis was wrong about which side, and your fix is the better one

I said `expectedRuntimeGraph` was the side that was wrong. It was not. The two-way fixture in this head settles it, and it settles it with better evidence than I had: the same crate tree, built twice, once with the sbom environment stripped —

```js
delete impreciseEnvironment.CARGO_BUILD_SBOM;
delete impreciseEnvironment.CARGO_UNSTABLE_SBOM;
delete impreciseEnvironment.RUSTC_BOOTSTRAP;
…
assert.equal(impreciseEvidence.document.format, 1);
assert.equal(impreciseKinds.get("inactive-runtime"), "runtime",
  "format 1 must reproduce Cargo metadata's inactive optional package false positive");
assert.equal(impreciseKinds.get("macro-helper"), "runtime",
  "format 1 must reproduce promotion of a shared build-only package");
```

That is a positive **and** a negative control on the exact defect class, in one fixture, and "promotion of a shared build-only package" is `adler2` named in general form. Format 1's inference was the broken side; precise mode fixes it at the source instead of aligning two derivations. Pinning `document.format !== 8` so a silent fallback to inference fails closed is the right shape too.

Correction noted and it is my second on this PR — worth saying plainly rather than burying: the `--target` experiment showed the two graphs disagreeing and I attributed the disagreement to the wrong side of it.

## The new failure is the same trap one layer down

`Desktop / Linux packages`, job `97714501791`, head `be434b4f`, step **Bind Linux shipped-artifact SBOMs**:

```
embedded Cargo audit evidence must have exactly one root package
```

Format 8 parsed. `validateAuditableDocument`'s **root invariant** did not, and that invariant was carried over from format 1 unchanged — only the `format` literal moved from 1 to 8. The `allowed` field set and the single-root reachability walk are still format-1 assumptions applied to a format-8 document.

The fixture cannot see it, for a reason that is now familiar:

```js
// scripts/artifact-sbom.node-test.mjs — the fixture package
'[package]\nname="zuuli"\nversion="0.1.0"\nedition="2021"'
…
resolve(project, "src/main.rs")
```

One binary target, no `[lib]`. Production:

```toml
# wallet/zuuli/src-tauri/Cargo.toml
[lib]
name = "zuuli_lib"
crate-type = ["staticlib", "cdylib", "rlib"]
```

plus `src-tauri/src/main.rs`. **One package, four compiled targets.** Precise mode records what cargo actually built, per target, so the real document does not have the one-root shape the fixture always produces — either more than one entry carries `root: true`, or the walk from a single root no longer reaches everything.

This is the third time on this PR that the fixture has been narrower than production in the one dimension that matters — first `--test-name-pattern` selection (#696), then the wrapper called without a `+toolchain` selector, now a single-target package standing in for a multi-crate-type one. **The fixture's `Cargo.toml` should mirror production's target layout**: add the `[lib] crate-type = ["staticlib", "cdylib", "rlib"]` alongside the existing `src/main.rs`, and the root invariant will be exercised by the shape that actually ships. Whatever the correct format-8 root rule turns out to be, it should be derived from cargo's SBOM semantics rather than inherited from format 1.

## `RUSTC_BOOTSTRAP=1` needs to be a stated decision, not a line in a shell script

```sh
PATH="$wrapper_directory:$PATH" \
  ZUULI_REAL_CARGO="$real_cargo" \
  RUSTC_BOOTSTRAP=1 \
  CARGO_UNSTABLE_SBOM=true \
  CARGO_BUILD_SBOM=true \
  exec "$@"
```

`grep -rn RUSTC_BOOTSTRAP` finds no other use anywhere in this repo — only this script and the test that strips it. I understand why it is there: `-Z sbom` is unstable and this is the only way to reach it from pinned stable 1.97.1.

The cost is not confined to cargo. `run-with-cargo-auditable.sh` wraps the whole `tauri build`, so the variable is inherited by **every rustc invocation in the shipped Linux build**. Nightly features are unlocked for the entire graph, which means a dependency's build script that probes "does `#![feature(…)]` compile?" gets a different answer than on a real stable toolchain and can enable a nightly code path in a release artifact. `scripts/check-rust-toolchain.sh` pins the channel and would not notice — it reads `rust-toolchain.toml`, not the environment. So the artifact this PR exists to bind evidence to is, on Linux only, no longer built by stable in the sense the toolchain pin asserts.

Worth trying: strip it back out for rustc via `RUSTC_WRAPPER`, something like a two-line wrapper that `exec env -u RUSTC_BOOTSTRAP rustc "$@"`. `-Z sbom` is cargo-side and writes JSON; rustc should not need the unlock. If that works, the blast radius drops from the whole graph to cargo itself, and it can be asserted by a test. If it does not work, the variable stays — but then it belongs on **#701** beside the advisories and the licence, because it is the same kind of thing: a supply-chain property of the shipped Linux binary that the owner should be choosing rather than inheriting.

## Smaller: `--cargo-features` still hard-codes what the build did

`--cargo-features=default` became `--cargo-features=custom-protocol`, and `--features custom-protocol` was added to the Linux `tauri build` (macOS's line is unchanged). Whether that is a real asymmetry depends on whether `tauri-cli` already passes `custom-protocol` itself for `tauri build` — if it does, the explicit flag is belt-and-braces and the platforms agree; if it does not, Linux and macOS now compile different feature sets. I have not verified which, and it is worth stating in the PR either way.

The general point survives regardless: both the features and the target are **hard-coded assertions about a build that happens in a different step**, which is the shape that produced this whole thread. `--target` is still absent from the Linux `tauri build` while `--cargo-target=x86_64-unknown-linux-gnu` is still passed to every `finalize-artifact` and `verify-artifact`. Now that precise mode removes the inference, that mismatch may simply never bite — but deriving these from the build rather than restating them is what stops the next one.

Still holding.

skyl:
Heads up on a contract change landing underneath this PR: **#706** replaces `zuuli-packaging.yml`'s `--test-name-pattern='…'` step with `node scripts/run-linux-artifact-fixtures.mjs`, which binds the Linux fixture population explicitly.

I verified it closes #696 by running both halves of my reproduction — on `main`, renaming a fixture takes the selector from `# tests 4` to `# tests 3` at EXIT=0; on #706's head the same rename is EXIT=1 in *both* the packaging run and a plain `node --test`, with `Linux artifact fixture has an unregistered title`.

What that means here concretely, once #706 merges:

- `real cargo-auditable executable evidence is complete, lock-bound, and mandatory` must be registered through `linuxArtifactFixture(...)` rather than a bare `test(...)`, or `after()`'s registered-equals-selector assertion fails.
- `LINUX_ARTIFACT_FIXTURE_TITLES`, `EXPECTED_LINUX_ARTIFACT_FIXTURE_COUNT` (4 → 5) and `EXPECTED_LINUX_ARTIFACT_FIXTURE_PATTERN` all have to move together — the literal-versus-derived check in `linuxArtifactFixturePattern()` refuses a partial edit.
- The wrapper's `skip:` is `!canBuildLinuxFixtures`, and your fixture is gated on `canBuildCargoRuntimeFixture` (which additionally needs `cargo-auditable`). Those are different predicates, so registering yours through the existing wrapper without widening the gate would change when it runs. Worth deciding deliberately rather than discovering at merge.

None of this changes my open findings on this PR — the format-8 root invariant and `RUSTC_BOOTSTRAP=1` from 5406989485 still stand. Flagging it because a rebase over #706 will otherwise turn into a surprise.

skyl:
Published exact fix-forward head `3bd6274bc297e45a3e2804090d3a62689516c145` (cumulative stable patch `5587aeb6aa8489b6bc0e13fb3f9f0233d07b23f4`) by normal push; no history rewrite.

This resolves the outstanding review findings rather than merely documenting them:

- cargo-auditable 0.7.5 is installed from the exact checksummed crate archive and patched at its precise precursor root, preserving one real root for the production lib+bin shape;
- every installer producer failure is preserved in command order, including valid `cargo install --list` output with exit 17;
- #706 is integrated as an exact five-fixture selector/completion contract;
- `RUSTC_BOOTSTRAP=1` remains Cargo-side authority for precise format 8, while a fail-closed `RUSTC_WRAPPER` removes it before every Cargo-launched rustc and preserves argv/status/stderr;
- Linux build/finalize/verify remain bound to reviewed default-host x86_64 semantics and `custom-protocol`.

Independent negative controls on this exact head were real RED: delete the launcher wrapper binding; delete the bootstrap scrub; mask rustc exit 23; delete the shipping feature; old valid-output/exit-17 installer pipeline; each curl/checksum/tar/patch/install bypass; fifth fixture rename/deletion/broaden/skip/completion. Restored host artifact suite: 17 pass / 5 honest Linux skips. Exact Ubuntu 24.04, Rust/Cargo 1.97.1, patched cargo-auditable 0.7.5 selected runner: 5/5, including Cargo → precise format-8 ELF → deb → finalized SBOM → fresh verification.

Hosted `Desktop / Linux packages` on this exact head is now the final merge authority. No merge until it and all required gates are green.

skyl:
## Review of the rebased head `3bd6274b` — both of my findings are addressed, and I verified them

Baseline on this head, after a real `npm ci`: `node --test scripts/artifact-sbom.node-test.mjs` → **EXIT=0, 22 tests, 17 passed, 5 skipped** (the five Linux fixtures on macOS — five now, matching the raised count).

### 1. `RUSTC_BOOTSTRAP` is scoped to cargo. Verified both halves.

`rustc-without-bootstrap.sh` wired in as `RUSTC_WRAPPER`, exactly the shape I suggested. Cargo invokes a wrapper as `$RUSTC_WRAPPER <rustc> <args…>`, so mirroring that against a stub rustc that prints its environment:

```
without the wrapper:  rustc sees RUSTC_BOOTSTRAP=1 ; args: --edition=2021 foo.rs
with the wrapper:     rustc sees RUSTC_BOOTSTRAP=<unset> ; args: --edition=2021 foo.rs
wrapper with no args: "ZUULI rustc wrapper requires the real rustc command", EXIT=64
```

Args forwarded intact, the unlock reaches cargo only, and it fails closed rather than exec'ing nothing. That drops the blast radius from every crate in the shipped Linux build to cargo itself, which was the whole ask.

### 2. The one-root failure is fixed at its source, and the cause is what I described

`cargo-auditable-0.7.5-root.patch` names it exactly:

> Cargo can list the root package more than once for distinct build units (for example, a Tauri binary and its same-package library).

That is the `[lib] crate-type = ["staticlib", "cdylib", "rlib"]` **plus** `src/main.rs` shape — one package, four compiled targets — that the single-target fixture could never produce. Binding `root` to `sbom.root` after deduplication is the right repair, and it is a genuine upstream bug rather than a workaround.

### 3. The fifth fixture is registered properly, and I checked it is load-bearing

`LINUX_ARTIFACT_FIXTURE_TITLES`, `EXPECTED_LINUX_ARTIFACT_FIXTURE_COUNT` 4→5 and the literal pattern all moved together, and the `linuxArtifactFixture` wrapper grew an `{ available }` option so `canBuildCargoRuntimeFixture` is used deliberately instead of inheriting `canBuildLinuxFixtures`. That is the decision I asked for made explicitly rather than by accident.

Renaming only the call-site title (`lock-bound` → `lock bound`, leaving the registry alone):

```
EXIT=1
AssertionError: Linux artifact fixture has an unregistered title:
  real cargo-auditable executable evidence is complete, lock bound, and mandatory
```

#706's mechanism extends to the new fixture correctly.

---

## Two things I would still change, neither blocking

**The post-install check can no longer fail.** `install-cargo-auditable.sh` ends with:

```sh
installed=$(cargo install --list)
printf '%s\n' "$installed" | grep -Eq "^cargo-auditable v${version}( \\([^)]*\\))?:$"
```

The parenthetical is optional, so this passes for a patched `--path` build **and** for a plain `cargo install cargo-auditable --version 0.7.5` from crates.io — the two print the same version and differ only in the `(…)` this regex makes optional. The previous head's `grep -Fx 'cargo-auditable v0.7.5:'` at least pinned "came from crates.io"; the new one pins neither. Requiring the path form is free:

```sh
printf '%s\n' "$installed" | grep -Eq "^cargo-auditable v${version} \\(/[^)]*\\):$"
```

That still is not proof the *patch* applied — the real proof is downstream, where `validateAuditableDocument`'s one-root assertion fails if it did not, and that is a good backstop. But a verification step that cannot distinguish the thing it exists to distinguish should either be tightened or dropped, because as written it reads like evidence and is not.

**The `ZUULI_REQUIRE_LINUX_ARTIFACT_FIXTURES` precondition does not know about cargo-auditable.** `artifact-sbom.node-test.mjs:133-143` throws when the platform is wrong or `linuxFixtureTools` are missing, but `cargoRuntimeFixtureTools` — which adds `cargo` and `cargo-auditable` — is not consulted. On a Linux runner where the install step somehow left `cargo-auditable` off `PATH`, the precondition passes, the fifth fixture skips, and what catches it is `after()`'s completion-record `deepEqual` rather than the precondition's clear message. So it is still fail-closed, which is what matters; the diagnostic just arrives as a set mismatch instead of "missing tools=cargo-auditable". Adding the fixture's own tools to that check is a one-line improvement.

**I could not run either of those on this host** — both need Linux — so they are from reading, and `Desktop / Linux packages` is where the verdict comes from.

## One thing that is now an owner decision

ZUULI's shipped Linux artifacts will be instrumented by a **locally patched cargo-auditable that exists nowhere upstream**. Anyone who wants to independently verify the embedded runtime graph needs the same patch to reproduce the tooling. The source pin is done properly — `curl` from `static.crates.io` with a `sha256sum --check` against a pinned digest, then `patch --fuzz=0 --batch --forward`, so the input is immutable and the patch cannot apply fuzzily — and I have no objection to the mechanism. But "we ship artifacts instrumented by our own fork of the auditor" is the same category as the advisories and the licence on **#701**, and the upstream report belongs with **#703**. Worth adding to both rather than leaving it in this PR's history.

Not holding on any of the above. `rs / gate` and `gate` need to be green on `3bd6274b`, and `Desktop / Linux packages` is the job that actually exercises the patch, the wrapper and the fifth fixture together.

skyl:
## `Desktop / Linux packages` is green on `3bd6274b`, and I checked it did the work rather than skipping it

Job `97755454800`, the three steps that matter:

```
success — Test Linux artifact inspectors against real packages
success — Build packages
success — Bind Linux shipped-artifact SBOMs
```

`Bind Linux shipped-artifact SBOMs` is the step that failed twice before — first on `+1.97.1`, then on `must have exactly one root package`. It passes now.

Green alone would not tell me the fifth fixture ran, so I read the counts:

```
✔ real cargo-auditable executable evidence is complete, lock-bound, and mandatory (8378.039024ms)
ℹ tests 5
ℹ pass 5
ℹ skipped 0
```

**Five selected, five passed, zero skipped**, and the new fixture genuinely executed — eight seconds of compiling a real crate through the patched `cargo-auditable`, not a skip reported as a pass. That is the whole of #696's second concern answered on a real runner, and it is the evidence that the patch installed, the `RUSTC_WRAPPER` did not break the build, and the format-8 one-root reconciliation works against production's actual four-target package shape.

So both findings I raised on this PR are closed by CI rather than by my reading:

- the `+1.97.1` dispatch bug — fixed, and I verified the wrapper's argv handling directly;
- the one-root invariant — fixed at source in `cargo-auditable`, and now proven against the real `[lib] crate-type = ["staticlib", "cdylib", "rlib"]` + `src/main.rs` package that the single-target fixture could never produce.

Nothing further from me. The two non-blocking notes in 5408537589 still stand as cheap improvements — tightening the post-install `grep` so it can actually fail, and teaching the `REQUIRE_LINUX_ARTIFACT_FIXTURES` precondition about `cargoRuntimeFixtureTools` — and neither needs to happen here.

Waiting on `Rust / native lints (windows)` and `gate`; zero failures across the rollup. I will merge once both required contexts are green on this exact head.

Issue #385, opened 2026-08-16
Spike: can modern Rust crypto crates build AND run on all six ZUULI targets?


skyl:
## Question

Can we build **and run** modern Rust cryptography crates on all six of ZUULI's shipping targets? We do not know, and several planned directions depend on the answer.

## Why this is foundational regardless

ZUULI ships to macOS, Windows, Linux, iOS 18, and Android (four ABIs). Any future adoption of a current crypto crate — post-quantum KEMs, formally verified primitives, threshold signatures — runs into the same question, and today nobody has tested it.

Concrete near-term cases:

- **FROST** (`z/ZcashFoundation/frost` is already a submodule but a dependency of nothing) for threshold spending authority.
- **Post-quantum primitives**, which the ecosystem is moving toward.
- Anything from the **libcrux / formally verified** family — the leading option for PQ work today, and the only provider implementing X-Wing in OpenMLS.

The specific risk is that these crates lean on SIMD backends and **runtime CPU feature detection**. iOS restricts some CPU-probing APIs; Android API 29 has its own constraints. `libcrux-ml-kem`'s SIMD backends are opt-in with a portable pure-Rust default, which *looks* safe — but detection is a runtime behaviour and **compilation success proves nothing about it**.

## The trap, stated plainly

A cross-compile that succeeds tells you the code type-checks for a target. It does not tell you the binary runs correctly on the device. This is the same class of gap as #321 (`gate` proves host-compile, not target-compile) — one level further out.

## Scope — a spike, timeboxed

1. Pick a representative modern crypto crate (suggest `libcrux-ml-kem` plus `libcrux-platform`, since they are the ones with runtime CPU detection).
2. `cargo build --target` for all six shipping targets. Record any that fail and why. Note whether anything runs a C compiler in `build.rs`.
3. **Actually execute it on a physical iPhone (iOS 18) and a physical or emulated Android API-29 device** — via a temporary debug command and a button, or a test harness. This is the step that matters and the one usually skipped.
4. Record **performance** on the slowest Android device available. A key operation that takes 50ms on a laptop and 4s on a budget phone is a design constraint, not a footnote.
5. Check MSRV interaction: we now pin **1.97.1**, which is current, so this should be comfortable — but verify rather than assume.

## Acceptance criteria

- [ ] Build results recorded for all six targets.
- [ ] Execution verified on a real iPhone and a real Android device — not just compilation.
- [ ] A latency number from the slowest available Android device.
[...]

Issue #312, opened 2026-08-08
E2EE Phase 2: MLS engine integration (Rust core + WASM) and 1:1 messaging


skyl:
Phase 2 of #305. **Blocked on** Phase 0 doc and Phase 1 identity.

## Goal
Integrate an MLS (RFC 9420) engine and ship working 1:1 messaging.

## Scope
- **One Rust crypto core**, used natively by ZUULI and compiled to WASM for the web client (D1). Do not maintain two implementations — implementation divergence is how crypto bugs happen.
- Library selection and justification: **OpenMLS** (permissive, interop-tested, X-Wing PQ merged using Cryspen's formally verified libcrux) vs **mls-rs** (AWS, WASM support, no third-party audit yet) vs **Marmot MDK**. Record audit status of whichever is chosen.
- **Hybrid post-quantum from day one** — the X-Wing ciphersuite (`MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519`), not a later retrofit.
- Message framing with **hash-links to predecessors**, so causal ordering and gap detection are transport-independent (this is what lets relay and P2P delivery interleave freely — see #131).
- Web client states its weaker trust model **in the UI**, not only in docs (D1).

## Acceptance criteria
- [ ] Library chosen, with rationale and audit status recorded in an ADR.
- [ ] Rust core builds for native and WASM from one source tree.
- [ ] 1:1 messaging works end to end in ZUULI and in the browser.
- [ ] X-Wing hybrid PQ ciphersuite in use.
- [ ] Forward secrecy and post-compromise security demonstrated by test.
- [ ] Hash-linked ordering: a test shows gap detection and re-request after simulated message loss.

Refs #305.

skyl:
**Phase 2 is complete.** Full session state: https://github.com/free2z/zuu/issues/305#issuecomment-5448003049

**Shipped:** `f2z-msg-mls` on **`openmls 0.9.0` + `openmls_libcrux_crypto 0.4.0`**, ciphersuite `MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519` — hybrid post-quantum from day one, as this issue required. Plus `f2z-msg-store` (a custom 57-method `StorageProvider`, because `openmls_sqlite_storage` collided with the repo-wide `rusqlite 0.37` `links` singleton) and `f2z-msg-dag` (§7 hash-linked ordering, gap detection, repair).

[...]

PR #697, opened 2026-08-25
feat(rs): f2z-msg-mls and f2z-msg-store — the MLS engine and its StorageProvider

### LINKED ISSUE #693: ADR 0001 says one crypto core, but openmls_basic_credential signs with ed25519-dalek outside the libcrux provider
Found by the mobile crypto spike (#385). ADR 0001's central claim is *"one Rust crypto core, built once"* — used natively by ZUULI and compiled to WASM for the web, **so that there is never a second implementation to diverge.** That is the reason the MLS library choice went to a Rust engine at all.

## It is not literally true today

`openmls_basic_credential 0.5.0` signs with **`ed25519-dalek` / `p256`**, bypassing the libcrux provider entirely. So a build that carefully selects the libcrux crypto provider for its KEM and AEAD still signs MLS leaf material with a different implementation of Ed25519.

The spike measured this rather than inferring it.

## Why it matters, precisely

It is **not** a correctness problem — both are conforming Ed25519, and the wire bytes are identical. It is a claim problem, and claims are what this project has been careful about:

- ADR 0001 says one implementation. An auditor reading that and then finding two Ed25519 implementations in the graph will ask why, and "we didn't notice" is a worse answer than "we chose to, here is the reasoning".
- It doubles the audit surface for signatures specifically — the operation that binds a device to an identity.
- It is a second dependency that must track advisories, and `cargo-deny` already reports the graph carrying two major versions of `ed25519-dalek` for an unrelated reason (`akd_core` requires `^2` while the workspace is on `3`).

## The fix is small

Roughly **20 lines**: our own `Signer` implementation routing MLS credential signing through the libcrux provider, replacing `openmls_basic_credential`.

## Decide deliberately, do not just implement

Two honest options, and the choice should be recorded rather than defaulted:

1. **Write the `Signer`** and make ADR 0001's claim true as stated. Small, and it removes a dependency.
2. **Keep `openmls_basic_credential`** and **amend ADR 0001** to say what is actually true — one MLS *engine*, with signature primitives that may come from a different audited implementation — with the reasoning for why that is acceptable.

What is not acceptable is the current state: a documented claim that the code does not honour. `THREAT-MODEL.md`'s whole posture is that a property which does not hold is stated as not holding.

My weak preference is (1), because it is 20 lines and the alternative spends an ADR amendment to preserve a dependency we do not need. But if there is a reason the upstream credential type is worth keeping — future algorithm support, upstream test coverage — that is a real argument and I would rather hear it than have it assumed away.

## Acceptance

- [ ] Either the `Signer` exists and no MLS signing path leaves the libcrux provider, or ADR 0001 is amended with the reasoning.
- [ ] If (1): a test asserting the credential's signatures verify under the provider, and the `openmls_basic_credential` dependency removed.
- [ ] Either way, `rs/README.md`'s description of the crypto core matches reality.

## Related
- **#385** — the spike that found it; its comment carries the full evidence.
- ADR 0001 (`docs/e2ee/decisions/0001-platform-priority.md`) — the claim in question.

skyl:
Phase 2 of #305 (#312): the MLS engine and the storage provider it needs.

Two crates. Neither existed before; nothing else in the tree changes behaviour.
Rebased onto `2e6b160`, so it sits on top of #694 and #698 and depends on
`f2z-msg-identity` rather than carrying a parallel identity of its own — see §5.

## What is here

**`rs/crates/f2z-msg-store`** — the `StorageProvider` OpenMLS needs.

`openmls_sqlite_storage` cannot be used and #385 reproduced the resolver error
rather than trusting the note: 0.2.0 requires `rusqlite ^0.32`, `AGENTS.md:147`
documents a repo-wide `rusqlite 0.37` singleton via `links = "sqlite3"` that
Cargo hard-errors on, and the 0.3.0-rc line that fits `^0.37` drags in an
unreleased `openmls 0.9.0-rc`.

So: all 57 methods and the one associated type, implemented **once** in
`src/storage_impl.rs` **in the upstream trait's declaration order**, so it can
be diffed against `openmls_traits::storage` and against
`openmls_memory_storage 0.5.0` side by side. Every method reduces to one of six
helpers named after upstream's. Backends are a three-method seam
(`StorageBackend`): a memory backend, and a native SQLite backend on the 0.37
singleton with WAL + `synchronous = FULL` + `secure_delete`, every pragma
verified after being set. **The browser's IndexedDB backend is deliberately not
in this PR**; the seam is what this PR buys.

**`rs/crates/f2z-msg-mls`** — the engine. Group creation, KeyPackage generation
and publication, `Welcome` processing, commits and epoch advance, `Update` for
post-compromise security, the four §5.4 exporter labels as a closed enum, and
`DeviceCredential` as the MLS `Credential` with the identity→device binding
checked at every point a credential enters a group. All application payloads are
MLS `PrivateMessage` (§5.3) and there is no public-message path.

Versions are exactly the ones #385 measured — `openmls 0.8.1`,
`openmls_libcrux_crypto 0.3.1`, `MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519`.
Nothing about the library or ciphersuite choice is re-opened here.

## Atomicity — the thing this PR is actually about

OpenMLS issues many storage calls per logical operation and has **no transaction
API**. `process_message` → `merge_staged_commit` → the application's own record
is seven or more separate writes, and a crash in the middle leaves a group whose
tree is from one epoch and whose secrets are from another. Under delete-on-ack
(`ARCHITECTURE.md` §6.4) the relay deleted its copy when this device
acknowledged, so that is data loss and not inconvenience.

`F2zStorageProvider::begin()` returns a `Transaction` that journals every write —
reads consult the journal, so OpenMLS's own read-after-write within one operation
is unchanged — and applies the whole set through one atomic `StorageBackend::apply`.
**It rolls back on drop**, because a lost operation is redelivered and a
half-applied one is not. Nesting is refused rather than reference-counted: an
inner `commit` that an outer rollback could still undo is a receipt that lies,
and a caller that acknowledged on it would have destroyed the message.

Every mutating engine method opens one. `receive()` additionally takes a
`record_key` and writes the durable "this delivery was handled" record **inside
the same transaction as the decryption**, which is what makes an `ACK` safe.

## Evidence

**Two independent engines exchange a message** — `tests/two_instances.rs`.
Separate stores, separate device keys, separate identity keys, separate
credentials; the only things that cross are byte strings that would go over a
relay. Bob publishes a KeyPackage, Alice creates the group and adds him, Bob
joins from the `Welcome`, and both directions decrypt to the exact plaintext.
Then `update()`, and both sides agree on the new epoch and keep working. All
four exporter labels produce identical bytes on both devices, different bytes
per label and per context, and different bytes after the epoch changes.
`two_engines_share_no_state` is what would notice if any of that were one handle
talking to itself.

**Crash safety** — `f2z-msg-store/tests/crash_safety.rs`, behind
`crash-injection`, killing a real child process with `abort()` at a chosen
instant and asserting on the file it left behind from a fresh connection:

- crash **before** the commit → **no trace** of a four-key operation, store empty;
- crash **after** the commit's fsync, before the caller was told → the operation
  is **whole**;
- unarmed child → runs to completion and exits 97, so a deleted injection point
  fails the suite instead of passing it.

The parent also asserts the child died *by signal*, not by exiting.

**Failure paths**, because delete-on-ack is where they turn into data loss:
a duplicate `record_key` is refused and changes nothing; an out-of-order commit
is refused and leaves the group usable, and applying the two in order still
works; a **refused delivery leaves no handled record**, so the relay's
redelivery is not swallowed as a duplicate; a truncated message and trailing
bytes are refused; a key package whose credential does not bind is not addable.

## Properties

- `#![forbid(unsafe_code)]` in both crates.
- **MIT**, stated literally in each manifest, enforced by `rs/deny.toml`.
- Redacting `Debug` on everything holding key material, with `tests/redaction.rs`
  checking hex **and decimal** — the `f2z-codec` trap where `tls_codec`'s byte
  vectors print a decimal list containing no hex, so a hex-only check passes
  while everything leaks. The negative control asserts the checks can fire.
- **`openmls >= 0.8.1` is a `[[bans.deny]]` entry**, not a comment, plus bans on
  `openmls_basic_credential` and `openmls_rust_crypto` (see below).
- `scripts/check-hash-domain-labels.mjs` passes over the new labels; the
  exporter labels and `free2z/device-credential/v1` are swept with everything
  else (49 labels, 7 documents, 1565 files).
- `scripts/check-rust-toolchain.sh`: both manifests registered in `MANIFESTS`.
- WASM: both crates on the `rs_wasm` line. See below.

## Closes #693

`DeviceSigner` implements `openmls_traits::signatures::Signer` over
`libcrux-ed25519` — the *same functions* `openmls_libcrux_crypto`'s
`CryptoProvider::sign` and `verify_signature` call — instead of
`openmls_basic_credential`, which signs with `ed25519-dalek` and `p256` outside
the libcrux provider. It adds no crate to the graph and removes two. ADR 0001's
"one Rust crypto core" is now literally true for this graph, and
`the_openmls_signer_and_the_direct_path_agree` pins it. `rs/deny.toml` bans
`openmls_basic_credential` and `openmls_rust_crypto` so it stays true.

It is **not** true of `f2z-msg-identity::DeviceSignatureKey::sign_mls_content`,
which still signs MLS framing with `ed25519-dalek`. That is the one seam left
and §5 has it.

## §13-B — stored, not solved

`ProtocolVersion` is written beside every group the engine creates or joins, so
a future codepoint relabel is a migration someone can write rather than an
orphaned store. `codepoint_is_registered()` returns `false` and a test says why.
Nothing here contradicts §13-B; #385's finding that draft-06 and draft-10
Appendix C vectors are byte-identical is the reason this is a naming risk and
not a re-key, and that is recorded in `src/version.rs`.

---

# Things a reviewer must actually decide

## 1. `openmls_libcrux_crypto 0.3.1` brings eleven RustSec advisories

**#385 did not run cargo-deny, so none of these was visible when the library
choice was made.** They are all in libcrux crates reached through
`openmls_libcrux_crypto 0.3.1`. libcrux publishes as `0.0.x`, so every release is
semver-incompatible with the next and `cargo update` cannot move any of them: the
fix has to come from `openmls_libcrux_crypto` and `hpke-rs-libcrux` bumping.

`rs/deny.toml` now carries eleven entries, each written after reading the
advisory and the dependency graph. Seven are **not reached**:

| Advisory | Why not reached |
|---|---|
| RUSTSEC-2026-0209/0210/0211 (`libcrux-aesgcm`) | Our ciphersuite is ChaCha20-Poly1305 for both the MLS AEAD and the HPKE AEAD. `provider.rs`'s `the_ciphersuite_uses_chacha20poly1305_and_not_aes_gcm` makes that a checked fact, and `deny.toml` cites the test by name. |
| RUSTSEC-2026-0073 (`libcrux-poly1305`) | The advisory says the AEAD use is unaffected; the standalone MAC API is the affected one and nothing calls it. |
| RUSTSEC-2026-0207/0208 (`libcrux-sha3`) | Both advisories explicitly exclude ML-KEM, our only SHAKE consumer. |
| RUSTSEC-2026-0075 (`libcrux-ed25519`) | The defect is in `generate_key_pair`; nothing in this tree calls it. Whoever writes `f2z-msg-identity` must not, while this version is pinned — the deny.toml entry says so. |

One is build-host only (RUSTSEC-2026-0173, `proc-macro-error2`, via libcrux's
`hax-lib` verification annotations). **Two need an owner decision:**

- **RUSTSEC-2026-0124** (`libcrux-chacha20poly1305`) — reached. `encrypt` panics
  on an *overlong output buffer*; the buffer is sized by OpenMLS, never by a
  peer or a relay, so the advisory's own impact statement does not describe this
  caller. Still a panic in a crypto core in a tree whose clippy policy denies
  `panic`.
- **RUSTSEC-2026-0212** (`libcrux-secrets`) — **this is the one reasoning cannot
  close.** Constant-time swap/select on **aarch64** compares an 8-bit selector
  with a 32-bit `cmp`, so it may return incorrect results. aarch64 is every phone
  and every Apple Silicon desktop, and the crate sits under `libcrux-aead`, which
  we do use. Against it: #385's full NIST/RFC/X-Wing KAT suite ran green on
  `aarch64-apple-darwin` and on real Android API-29 arm64, and a wrong branch
  would show as a wrong answer. That is a strong negative result on two aarch64
  environments; it is **evidence, not proof**, because the advisory says the
  fault depends on the execution environment.

### I tried to take the fix instead of accepting the risk. It cannot be taken from here.

`libcrux-secrets 0.0.6` is published and *is* the fix, so the right move was to
apply it rather than argue reachability. Three attempts, all measured against
this lockfile, all recorded next to the `ignore` entry in `rs/deny.toml`:

1. **`cargo update --precise 0.0.6` — refused.** `libcrux-aead` 0.0.6 *and*
   0.0.7 both require `libcrux-secrets = "=0.0.5"`, an exact pin.
2. **`[patch.crates-io]` at 0.0.6 — silently ignored, and this is the dangerous
   one.** The build went **green** and still linked 0.0.5: `[patch]` replaces a
   package's *source*, not its version, so 0.0.6 cannot satisfy `=0.0.5`. Cargo
   warns `patch ... was not used in the crate graph` and carries on. Trusting the
   exit code would have meant reporting the advisory fixed with the vulnerable
   code still in the binary. **Check `cargo tree -i`, not the build result.**
3. **Forcing it** — a vendored 0.0.6 whose manifest claims `version = "0.0.5"` —
   **fails to resolve, over `hax-lib` rather than over this crate's API.**
   `libcrux-secrets 0.0.6` needs `hax-lib "0.3.7"`; `libcrux-intrinsics 0.0.6`,
   `libcrux-ml-kem 0.0.8` and `libcrux-sha3 0.0.8` each need `hax-lib "=0.3.6"`.
   Every pin is exact, so the constraints are unsatisfiable together.

So it is not a one-crate override: taking 0.0.6 means simultaneously forking
`libcrux-intrinsics`, `libcrux-ml-kem` and `libcrux-sha3` off `hax-lib =0.3.6`,
and those are exactly pinned by `libcrux-aead`, `libcrux-kem` and `hpke-rs`, so
the fork propagates to `openmls_libcrux_crypto`. **That is forking the libcrux
release train**, which is a worse position than the one we are in. 0.0.6 also
carries deliberate API breaks (upstream #1499, #1446, #1484), so `=0.0.5` is a
considered pin, not an oversight.

`rs/deny.toml` now carries the exact removal condition — *"when
`openmls_libcrux_crypto` ships against a `libcrux-aead` whose `libcrux-secrets`
requirement admits `>= 0.0.6`"* — plus all three attempts, so nobody spends an
afternoon rediscovering this. The highest-leverage action available to anyone is
asking Cryspen/Celabs to relax the `libcrux-aead` pin; it is one line on their
side. Full detail in #701.

## 2. MPL-2.0 enters the graph

`hpke-rs`, `hpke-rs-crypto` and `hpke-rs-libcrux` are MPL-2.0, and HPKE is how
MLS does its asymmetric encryption — there is no alternative inside the decision
#385 settled. They are added as **per-crate `exceptions`**, not to `allow`,
because MPL-2.0 is not permissive and `allow` is the permissive list.

The argument that it sits on the correct side of the boundary: MPL-2.0's
copyleft is **per file**. §3.3 permits distributing a Larger Work under other
terms provided the MPL-covered files stay MPL; §1.10/§3.2 scope disclosure to
those files. So linking imposes nothing on ZUULI's, the web client's, or a third
party's source. The obligation we do take on is real: if we ever modify one of
those three files, that modification is MPL-2.0.

**This is a licence-policy judgement and it belongs to the owner, not to the
agent that needed the dependency.** Reverting it is a three-line edit.

## 3. An upstream `debug_assert!(false)` on an attacker-reachable path

`openmls 0.8.1`'s `private_message_in.rs:136` runs
`debug_assert!(false, "Ciphertext decryption failed")` on the AEAD-open failure
path — which any peer, or any relay that flips one byte, can reach at will. It
compiles out in release and correctly returns `MessageDecryptionError::AeadError`;
in a **debug** build, which is what `cargo test` produces, the process aborts.

So the "refuse, do not crash" property is true of what ships and cannot be
asserted by the suite CI runs in debug. `tests/two_instances.rs` gates that one
case on `#[cfg(not(debug_assertions))]` and `rs.yml` adds a
`cargo test --release` step that runs it. Gating is the honest option; the
alternative was deleting the test. **Worth reporting upstream** — drafted and tracked in #703.

## 4. A real defect in `openmls_memory_storage 0.5.0`, fixed in the port

Upstream's `clear_proposal_queue` builds the per-proposal key as
`serde_json::to_vec(&(group_id, proposal_ref))` and removes *that* — but every
write goes in under `build_key_from_vec(label, key, version)`, so the key it
removes is one no entry was ever stored at. The refs list is deleted and every
serialised proposal body is left in the store forever. Ported faithfully that is
an unbounded leak in every client's local database. Fixed here (one line: go
through the same `delete` helper every other deleter uses) and pinned by
`clearing_the_queue_removes_the_proposal_bodies_and_not_only_the_references`,
which is the test that fails if someone "restores fidelity" with upstream.

Both are drafted for upstream in #703. Upstream also `unwrap()`s a `read` in
`queued_proposals`, turning a corrupt store into a panic in the middle of message
processing. That is an error here, and
`a_dangling_proposal_reference_is_an_error_and_not_a_panic` builds the corrupt
shape a crash between `queue_proposal`'s two writes would leave.

## 5. This depends on `f2z-msg-identity`, and there is one seam left between us

`f2z-msg-identity` (#694) landed while this branch was in flight, and #698
corrected `CKDh` on top of it. This branch is rebased onto `2e6b160` and the
minimal `DeviceCredential` it had defined to unblock itself is **gone**:

- `DeviceCredential` and `DeviceCredentialTBS` are **`f2z-kt-core`'s**,
  re-exported rather than restated. One definition of the bytes, shared by the
  issuer, the directory validator and this engine.
- The identity signature is verified with `f2z_kt_core::sig::verify`
  (`verify_strict`, so small-order keys and non-canonical encodings are
  refused), not with a second verifier. An MLS peer and the transparency log
  answering differently about the same credential is a fork nobody would see.
- Every credential in `tests/` is issued by
  `AccountKeys::from_seed` → `IdentitySigningKey::issue_device_credential`, so
  `tests/credential.rs` and `tests/common/mod.rs` are the **coordination test**
  between the three crates. If they ever disagree about the structure, the
  engine stops accepting credentials the identity crate issues — in CI, not on
  a user's phone. Nothing here pins a derived constant, so #698's change to
  `CKDh` needed no fixture regeneration.

What this crate still owns is the part only an MLS peer is in a position to
check: `validate_for_leaf`, that the credential's `device_pk` is the leaf's
`signature_key`. `f2z-kt-core`'s `validate` checks the label, handle and window;
neither it nor the directory can see which leaf a credential arrived in.

**The one seam left, and it needs a decision.**
`f2z_msg_identity::DeviceSignatureKey::sign_mls_content` signs MLS framing with
**`ed25519-dalek`**. This engine signs it with **libcrux** (`DeviceSigner`),
which is what closes #693 and what ADR 0001's single crypto core requires.
`DeviceSignatureKey` does not expose its private bytes — correctly — so the two
cannot be bridged from outside, and the tests here therefore use the identity
crate for the **account** half of §4.2 and `DeviceSigner` for the **device**
half. Either `DeviceSignatureKey` signs through libcrux, or it stops offering
`sign_mls_content`. It is called out in `tests/common/mod.rs` as well as here
rather than papered over, and filed as **#702** so that closing #693 does not
leave a closed issue with a live exception. I did not change #694's crate from
this branch.

## 6. Smaller calls, each with its reasoning in the code

- **`CredentialType::Basic`** rather than a private-use codepoint. RFC 9420's
  `BasicCredential` carries an opaque application-defined identity string, which
  is exactly this. A private-use type would additionally require every member's
  `Capabilities` to advertise it and buys nothing: a peer sending a bare handle
  fails `parse` on the leading type tag either way.
- **Storage labels stay OpenMLS's**, byte-identical to `openmls_memory_storage`,
  including the key layout. That keeps the reference usable as an oracle and the
  diff honest. They are not `free2z/` labels and are deliberately not in
  `check-hash-domain-labels.mjs`'s namespace — nothing hashes them.
- **The ratchet tree travels in the `Welcome`** (`use_ratchet_tree_extension`)
  rather than out of band: `WIRE.md` gives the relay no way to carry a second
  blob beside a `Welcome`, and an out-of-band tree is one more thing a hostile
  relay can withhold to make a join fail in a way the user cannot distinguish
  from a network problem.
- **`SqliteBackend::open_in_memory()` reports `Durability::None`**, not
  `Durable`. It is a real database with real transactions and no disk, and
  `CLIENT-CONTRACT.md` §11.2 lets a client `ACK` on a durable store.
- **`--all-features` over this workspace does not build**, and the reason is
  upstream's: `extensions-draft-08` adds three methods to a trait that
  `openmls_memory_storage` — a hard dependency of `openmls_libcrux_crypto` —
  would then have to implement. No gate builds that combination (`deny.toml`
  sets `all-features = false`; clippy uses `--all-targets`), and
  `f2z-msg-store`'s manifest says so where someone will find it.
- **`openmls_memory_storage` is deliberately *not* banned**, though banning it
  was the first instinct: it is a hard dependency of `openmls_libcrux_crypto`,
  whose `Provider` owns a `MemoryStorage` field. What matters is that nothing
  hands it to OpenMLS — `F2zProvider` implements `OpenMlsProvider` itself — so
  the transaction-free store is linked and unreachable. The comment in
  `deny.toml` says this.

## WASM

Both crates build for `wasm32-unknown-unknown` and both are on the `rs_wasm`
line. Two things were needed for the engine and neither is optional:

- `--features js`, because OpenMLS emits a `compile_error!` for wasm32 without
  it — it needs JavaScript APIs for randomness and time, and refuses to build
  rather than producing a binary that traps.
- `--cfg getrandom_backend="wasm_js"`, plus a `cfg(target_arch = "wasm32")`
  target table naming **both** `getrandom` majors in the graph (0.3 via
  `rand_core 0.9`, 0.4 via `rand 0.10` inside `hpke-rs-libcrux`) purely to turn
  their `wasm_js` features on. Neither is reachable through a feature of
  `openmls` or `openmls_libcrux_crypto`.

The SQLite backend is feature-gated out; the wasm job builds both crates with
`--no-default-features`.

## What is deliberately not here

The IndexedDB backend, the Tauri plugin surface of `CLIENT-CONTRACT.md` §3, the
hash-linked causal DAG of §7 (the engine returns `(epoch, senderLeafIndex)` so
the caller can build §7's sort key, but the framing itself is not here), gap
detection, and the key hierarchy above `DeviceCredential`.

Refs #305. Part of #312; #312's remaining acceptance criteria (hash-linked
ordering, 1:1 messaging end to end in ZUULI and the browser, the ADR) are not
met by this PR.


skyl:
## Review of `fa29a279`

Baseline on the pre-rebase head: `cargo test -p f2z-msg-mls -p f2z-msg-store` → **EXIT=0**, all targets green including doctests. The work is serious — `crash.rs` spawning a child that `abort()`s rather than catching a panic is the right mechanism and is the sibling of `f2z-relay-store`'s, the `StorageBackend::apply` contract is honoured by both implementations (one lock for the whole batch in memory, a `rusqlite` transaction in SQLite), and the CI additions actually run what they claim, including the `--features crash-injection` step and the `--release` step for the case `openmls`'s `debug_assert!(false)` makes unassertable in debug.

**One blocking finding**, filed as #699. Two things below are the owner's call and not mine.

---

## Blocking — #699: a rollback after `merge_staged_commit` leaves the group ahead of the store

`Transaction` protects the store. It does not protect the in-memory `MlsGroup`, and `receive` merges into memory before the transaction commits.

Reproduced on **this head**, using the PR's own `common::issue_credential` so the credential comes from the real enrollment path, with a `StorageBackend` that wraps `MemoryBackend` and can be armed to fail its `apply`:

```
receive outcome: Err(Storage(Backend("injected disk failure")))
epoch before GroupEpoch(1), in-memory epoch after the refused commit GroupEpoch(2)
redelivery outcome: Err(OutOfOrder)
```

Read the third line. Nothing was written, so no handled record exists, so no `ACK` is sent and the relay correctly redelivers — and the live process refuses its own redelivery, permanently, because its in-memory epoch has moved past a commit that was never persisted.

The assertion this violates is yours: `two_instances.rs::an_out_of_order_commit_is_refused_and_leaves_the_group_usable` says

```rust
assert_eq!(bob_group.epoch(), epoch_before,
           "a refused commit must not have advanced the epoch");
```

That holds for a refusal raised **before** the merge. All four "refused and leaves the group usable" tests in the suite refuse before the merge, which is why the gap reads as covered.

**The route that needs no disk error is the reason this blocks.** `engine.rs:437-441`:

```rust
group.merge_staged_commit(&self.provider, *staged)?;
self.validate_members(group, now_ms)?;   // <- lands in exactly the same place
```

`validate_members` walks every member and enforces the credential validity window; the sender check at `:411` covers only the sender. So a group containing **one member whose `DeviceCredential` has expired** wedges every other member the moment anybody commits — application messages keep flowing, commits merge into memory and roll back out of the store, and each redelivery is refused. That is a clock, not an attacker. I ran the injected-backend route; the credential route I am reporting from inspection of those six lines.

Fix options are in #699. The one I'd take: reload the group from the store on the error path, so the caller's handle matches what was persisted — it is the only option that leaves every existing test's assumption true. Validating the added leaves *before* `merge_staged_commit` is worth doing regardless, because it turns the credential route into a clean refusal instead of a wedge.

## #700 — the seam the harness says is called out here, and isn't

`tests/common/mod.rs:13-26` says `DeviceSignatureKey::sign_mls_content` signs MLS framing with `ed25519-dalek` while the engine signs with libcrux, that this "needs reconciling", and that "it is called out in the pull request rather than papered over here."

It is not in the pull request. The "Closes #693" section argues one-crypto-core is true *for this graph*, which is correct as scoped, and says nothing about `sign_mls_content` — a public method named for MLS `FramedContentTBS` with zero production callers (`grep -rn 'sign_mls_content' rs/crates/` finds the definition and two lines of its own unit test). `deny.toml` bans the crate `openmls_basic_credential`, not `ed25519-dalek`, which `f2z-msg-identity` uses directly — so nothing gates the next caller who reaches for the obviously-named method. Recommendation in #700 is to delete it; it is a deletion plus its test, and it makes ADR 0001 enforced by absence rather than convention.

Not blocking this PR — but the note deferring the decision to a discussion that does not exist is how it would have been lost at merge, so please don't merge without pointing it somewhere.

## The two owner decisions — I am not making these

Both are correctly escalated in the description and I am flagging them up rather than resolving them:

1. **RUSTSEC-2026-0212 (`libcrux-secrets`)** — constant-time swap/select on **aarch64** compares an 8-bit selector with a 32-bit `cmp`. aarch64 is every phone and every Apple Silicon desktop, and the crate sits under `libcrux-aead`, which this ciphersuite uses. The description's counter-evidence is honest about its own limit: #385's KAT suite passing on two aarch64 environments is a strong negative result, and the advisory says the fault depends on the execution environment, so it is evidence and not proof. The other ten advisories I read alongside the graph and find the reasoning sound — the ChaCha20-Poly1305 argument in particular is pinned by a named test rather than asserted. This one is not closable by reasoning.
2. **MPL-2.0 entering the graph** via `hpke-rs`, `hpke-rs-crypto`, `hpke-rs-libcrux`, as per-crate `exceptions` rather than `allow`. The per-file-copyleft argument reads correctly to me and the obligation taken on is real and stated. It is a licence-policy judgement.

Reporting the `debug_assert!(false)` upstream to `openmls` is worth doing and the `#[cfg(not(debug_assertions))]` gate plus the `--release` CI step is the right handling in the meantime.

## What I did not check

The `crash-injection` suite and the wasm32 builds I read but did not run — the crash tests need the feature and a child process, and wasm32 needs the toolchain target and the `getrandom_backend` flag. CI runs both and that is where I am taking the verdict from. I also did not audit all 57 `StorageProvider` methods against upstream's declaration order; I spot-checked the six helpers they reduce to.

skyl:
Filed #701 for the three decisions in this PR that are the owner's rather than mine — the accepted `libcrux` advisories (RUSTSEC-2026-0212 above all), MPL-2.0 entering the client-linked graph, and the upstream `debug_assert!(false)` in `openmls 0.8.1`. All three are already recorded in `rs/deny.toml` and in the description; the issue exists so they are revisited deliberately rather than inherited.

skyl:
## Status on `8e95289d`: `rs / gate` is green, and I am still not merging this

No failures in the rollup and `rs / gate` is SUCCESS, so this will be mergeable shortly. Recording why I am holding rather than letting it go green-and-through.

**1. #699 is unaddressed.** The deltas since my review are `fa29a279..19356f22` (rebase onto #698 plus three lines in `two_instances.rs`) and `19356f22..8e95289d` (`rs/deny.toml` only, +47/−3). `engine.rs`'s `receive` path is untouched, so a rollback after `merge_staged_commit` still leaves the in-memory `MlsGroup` an epoch ahead of the store and still makes the relay's correct redelivery `Err(OutOfOrder)` for the life of that process. The route that concerns me needs no disk error at all — `validate_members` runs *after* the merge at `engine.rs:437-441`, so one member with an expired `DeviceCredential` wedges every other member on the next commit, silently, while application messages keep flowing.

This is cheap to fix now and expensive later: nothing depends on `f2z-msg-mls` or `f2z-msg-store` yet, so reloading the group from the store on the error path is a contained change against zero callers. Every option in #699 gets harder once ZUULI holds group handles.

**2. #701's two decisions are the owner's, and merging makes them.** RUSTSEC-2026-0212 puts a constant-time swap/select defect on aarch64 — every phone and every Apple Silicon desktop — under `libcrux-aead`, which this ciphersuite uses; and MPL-2.0 enters the client-linked graph via `hpke-rs` ×3. #701 states both fairly and I have nothing to add to it except that I should not be the one to close them by pressing merge. I did check its `libcrux_ed25519::generate_key_pair` warning: zero occurrences in `rs/crates` on `main` and on this branch, so that note is accurate and forward-looking rather than describing a live problem.

Also worth folding into #701 or its own line: `#695` is currently proposing `RUSTC_BOOTSTRAP=1` across the whole Linux `tauri build`, which is the same category of thing. Different PR, same owner.

**What I am not holding on.** #700/#702 (`sign_mls_content`) and #703 (the upstream OpenMLS reports) are follow-ups, not blockers. The engine itself reads well: `crash.rs` spawning a child that `abort()`s is the right mechanism and I said so; both `StorageBackend::apply` implementations are genuinely atomic; the CI additions run what they claim, including `--features crash-injection` and the `--release` step for the case `debug_assert!(false)` makes unassertable in debug.

**Unblock condition:** #699 addressed, or an explicit owner decision to take it as a known issue on a crate with no callers. Either is fine; inheriting it by merge is not.

PR #698, opened 2026-08-25
fix(msg-identity): CKDh must hash the parent key (ARCHITECTURE.md §4.2)

skyl:
# ⚠️ This changes every key the messaging hierarchy derives, from every seed

`ARCHITECTURE.md` §4.2's `CKDh` did not hash the parent key. It does now. **Every
node below `MSK`, every account leaf, and every `DeviceCredential` signature is a
different value than the vectors #694 shipped three hours ago.**

That is safe **only because nothing has shipped**: no user has a messaging
identity, no directory entry exists in any log, and no peer has pinned a safety
number. The cost of this correction is zero today and unbounded after the first
user. That asymmetry is the whole reason it is being made now rather than filed.

> **This is a follow-up, not an amendment.** #694 was squash-merged to `main`
> (`def7687`) before the defect was actioned, so the fix ships as its own PR
> against the merged text. `main` currently contains the defective derivation.

---

## The defect

```
CKDh(node, i) = BLAKE2b-512(personal = "Free2zMsg_CKDv1_",
                            cc_node || 0x11 || I2LEOSP32(i))     ← superseded
```

`ik_node` is absent. ZIP 32 §5.2's Sapling hardened derivation is

```
I = PRF^expand(c_par, [0x11] || sk_par || I2LEOSP_32(i))
  = BLAKE2b-512("Zcash_ExpandSeed", c_par || 0x11 || sk_par || I2LEOSP_32(i))
```

— the parent **spending key** is in the message. §4.2's own justification for the
idiom is that it is written in ZIP 32's shape *"because ZUULI already implements
that shape and auditors already know it."* It was not.

**Not directly exploitable today.** `cc_msk` is exactly as secret as `S` and is
never published. What it cost is structural:

- the key half of **every** node above the account level was dead weight; `msk`
  was never read by anything;
- all secret entropy below `MSK` reached `ik_account` through `cc_msk` **alone**
  — a two-part secret collapsed into a one-part secret;
- a bare chain code derived its **entire subtree**, so any later construct that
  treated a chain code as the less-sensitive half of a node would have been
  catastrophically wrong.

## The fix

```
CKDh(node, i) = BLAKE2b-512(personal = "Free2zMsg_CKDv1_",
                            cc_node || 0x11 || ik_node || I2LEOSP32(i))
```

**ZIP 32 §5.2 with the personalization changed and nothing else.** The dropped
term is restored in the position ZIP 32 puts it; no third construction is
invented, because auditor familiarity was the entire argument for the idiom.
All four fields are fixed-width (`32 + 1 + 32 + 4 = 69`), so the concatenation is
unambiguous with no separator.

`ARCHITECTURE.md` §4.2 carries a **dated correction (2026-08-25)** in the house
style — the superseded form stays visible, struck through in prose rather than
silently rewritten, with the consequence analysis and the "safe only because
nothing has shipped" reasoning recorded next to it.

## Regenerated vectors

Recomputed **independently in Python** again — `hashlib.blake2b(person=…)`, a
hand-written RFC 5869 §2.3 `HKDF-Expand`, `cryptography`'s Ed25519 — never
printed from the implementation. The updated transcript is in the test file's
doc comment.

```
seed (BIP-39, all-`abandon`, unchanged)
             5eb00bbddcf069084889a8ab9155568165f5c453ccb85e70811aaed6f6da5fc1
             9a5ac40b389cd370d086206dec8aa6c43daea6690f20ad3d8d48b2d2ce9e38e4
MSK          74f055750feeefebb248b31b88aae0b2b9b0b0db1c0f520eb831d30243c87096   ← UNCHANGED
             91d0cc16887876d54b585fa850dd2b2c1aecbe525e4d300b8b15dd7677de373a
  32'        288530fcef32ace05c9545252ac22fd9d8f070b0961b3bb8e955c0e86922d906
             1f9b549adf14b522e5ab8fe6658282cc5fccc1fdcef7c9f5a8a921e58fad0bd9
  133'       8b269b6624b43f1794346249cbd4c0cf6f835467f9937f997265805ad8632464
             1f0c258f743e7b01b85ed3c1c4732f66c2784a53b924a28bb74c5806b07285cf
  0'         dbd4245907a1e6e0368d41ad784262d99f52ae7e8f6ce14ab3ef46efb811c627   ← ik_account
             1462b713c286891f5298d6e625319fb7b16382c8509b18bffaa928ecd2801102
ISK.public   e73fcd0648504865a6a582473c961cab75b3e6b108bff18196ef80aa41955199
CSK.public   c5027046f3bb8b8251e060b99b52f9044210791a148c0ebae05c8c2f5e4f4c6c
DirAuth.pub  e8a99244307a22722cb61bdac80098c8c826d1fa8c064de5e685bb926cd7130b
BackupWrap   479083aada9066af10d708d474199ae99416dd80bd3f5bcefa68c65a2c30270e
DeviceCredentialTBS  1332 bytes (unchanged — the encoding did not move)
signature    24785757ba36d112cb310dd6bad39b027776ca693848c1d56557f011b4bda4ed
             196c1a0f59e14d9a100c1d10e1be9f9f779c164492e7e713b1a930080e017806
```

**`MSK` is unchanged and that is worth noticing while reading the diff** — the
master derivation never involved `CKDh`. A correction that moved `MSK` too would
have been a different and larger change.

The node levels are still asserted twice: once against the shipped
`master_node`/`ckd_hardened`, once against a second in-test transcription of
§4.2 (now updated to the corrected preimage, with a `69` length assertion).

## The test that asserted the defect is gone

`ckd_reads_only_the_chain_code` — which asserted that flipping a bit of the
parent's key left the children unchanged, and *passed*, because that was what
§4.2 then said — is replaced by:

- **`ckd_reads_both_halves_of_the_parent`**, asserting from **both** directions
  that a one-bit change to either half moves the child. Either direction alone
  would pass for a `CKDh` that read the wrong single field.
- **`the_ckd_preimage_is_chain_code_tag_key_index`**, which rebuilds the 69-byte
  preimage by hand and compares. A field-order or width mistake inside
  `ckd_hardened` is otherwise invisible to every other test in the file — they
  would all still see a deterministic, well-separated tree.

## Still true, and re-verified against the new derivation

`tests/kt_core_agreement.rs` (all five cases — the acceptance test that
`f2z_kt_core::validate_submission` accepts what this crate issues),
`tests/zcash_separation.rs` (#311's disjointness criterion),
`tests/no_seed_derived_device_keys.rs`, and `tests/redaction.rs` all pass
unchanged. None of them depends on a derivation *constant*, which is the correct
outcome: they test properties, and the vectors test values.

## The other vector-affecting call — check both in one pass

**`I2LEOSP32(i)` still encodes the hardened index whole**: `32'` is `0x8000_0020`
→ `20 00 00 80` little-endian, as ZIP 32 serializes one. The corrected
construction does not make the other reading (bare ordinal) any more or less
right, so the call stands — but it is the *other* thing that silently changes
every identity, and a reviewer looking at `CKDh` today should confirm this in the
same pass. `HardenedIndex` takes the ordinal and sets the bit itself, so there is
exactly one spelling of a path in the crate.

## Kept as-is, per review

- **The `Blake2bMac512` trap note in `src/blake.rs`.** `blake2 0.10`'s obvious
  spelling for personalized BLAKE2b prepends a padded key block even with an
  empty key: the digest looks right and matches no other implementation on
  earth. That note stays visible for the next person who tries to simplify the
  file.
- **The `DeviceInitKey` scope call.** X-Wing comes from OpenMLS's libcrux
  provider (§3.1); a second combiner here is the divergence ADR 0001 exists to
  prevent.

## Verification

| Check | Result |
|---|---|
| `cargo test --locked --all-targets` (rs workspace) | ✅ exit 0 |
| `cargo test --locked --doc` | ✅ exit 0 |
| `cargo build --locked --lib --target wasm32-unknown-unknown -p f2z-msg-identity` | ✅ |
| `scripts/check-rust-fmt.sh --root rs` | ✅ 11/11 |
| `scripts/check-rust-clippy.sh --root rs` | ✅ 11/11 |
| `scripts/check-hash-domain-labels.mjs` | ✅ 49 labels prefix-free |
| `scripts/check-rust-toolchain.sh` | ✅ |

No dependency, manifest or lockfile change — this is three files: the
specification, `src/node.rs`, and the vectors.

Refs #311, follows up #694.

### DIFF
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -193,14 +193,73 @@ MSK  = (msk, cc_msk)  = BLAKE2b-512(personal = "Free2zMsg_MSTRv1", S)
                            or Orchard key trees even at identical indices.
 
 CKDh(node, i)         = BLAKE2b-512(personal = "Free2zMsg_CKDv1_",
-                                    cc_node || 0x11 || I2LEOSP32(i))
+                                    cc_node || 0x11 || ik_node || I2LEOSP32(i))
                         ── hardened only; there is no non-hardened variant and
                            no extended public key at any level of this tree.
+                        ── BOTH halves of the parent are in the preimage. See
+                           the correction below: an earlier revision omitted
+                           ik_node, which is not what ZIP 32 does.
+                        ── I2LEOSP32(i) encodes the HARDENED index, i + 2^31,
+                           as ZIP 32 serializes one: 32' is 0x80000020 and
+                           encodes as 20 00 00 80.
 
 account_node          = CKDh(CKDh(CKDh(MSK, 32'), 133'), account')
                            purpose' = 32'   (ZIP 32 idiom)
                            coin'    = 133'  (SLIP-44 Zcash)
 ```
+
+> **Correction (2026-08-25) — `CKDh` did not hash the parent key, and now it
+> does. This changes every key this tree derives.**
+>
+> The revision of `CKDh` above that shipped in
+> [#694](https://github.com/free2z/zuu/pull/694) read
+>
+> ```
+> CKDh(node, i) = BLAKE2b-512(personal = "Free2zMsg_CKDv1_",
+>                             cc_node || 0x11 || I2LEOSP32(i))     ← superseded
+> ```
+>
+> with `ik_node` absent from the preimage. That is **not** ZIP 32's shape, and
+> this section's whole justification is that it is written in ZIP 32's idiom
+> "because ZUULI already implements that shape and auditors already know it."
+> ZIP 32's Sapling hardened derivation is
+>
+> ```
+> I = PRF^expand(c_par, [0x11] || sk_par || I2LEOSP_32(i))
+>   = BLAKE2b-512(personal = "Zcash_ExpandSeed",
+>                 c_par || 0x11 || sk_par || I2LEOSP_32(i))
+> ```
+>
+> — the parent **spending key** is in the message. The corrected `CKDh` above is
+> that construction with the personalization changed and nothing else, so a
+> reader who knows [ZIP 32](https://zips.z.cash/zip-0032) §5.2 is reading a
+> function they already know. The dropped term is restored; no third
+> construction is invented.
+>
+> **What the omission cost, stated plainly.** It was not directly exploitable —
+> `cc_msk` is exactly as secret as `S` and is never published — but:
+>
+> - the key half of **every** node above the account level was dead weight;
+>   `msk` was never read by anything;
+> - all of the hierarchy's secret entropy below `MSK` reached `ik_account`
+>   through `cc_msk` **alone**, collapsing a two-part secret into a one-part
+>   secret;
+> - a bare chain code derived its **entire subtree**, so any future construct
+>   that treated a chain code as the less-sensitive half of a node would have
+>   been catastrophically wrong.
+>
+> **This changes every key derived from every seed.** It is safe only because
+> nothing has shipped: no user has a messaging identity, no directory entry
+> exists, and no peer has pinned a safety number. The cost of this correction is
+> zero today and unbounded after the first user; that asymmetry is the entire
+> reason it is being made now rather than filed.
+>
+> The implementation and its known-answer vectors move with it —
+> `rs/crates/f2z-msg-identity`, whose `tests/derivation_vectors.rs` pins the
+> corrected hierarchy from BIP-39's published all-`abandon` seed. **Every vector
+> in that file changed.** Found while implementing §4.2 for
+> [#311](https://github.com/free2z/zuu/issues/311); reported rather than coded
+> around.
 
 From `account_node`, every leaf is an HKDF-SHA256 expansion under a distinct
 versioned label (`HKDF-Expand(PRK = ik_account, info = label, L)`):
--- a/rs/crates/f2z-msg-identity/src/node.rs
+++ b/rs/crates/f2z-msg-identity/src/node.rs
@@ -5,7 +5,7 @@
 //! S                     = BIP-39 seed
 //! MSK  = (msk, cc_msk)  = BLAKE2b-512(personal = "Free2zMsg_MSTRv1", S)
 //! CKDh(node, i)         = BLAKE2b-512(personal = "Free2zMsg_CKDv1_",
-//!                                     cc_node || 0x11 || I2LEOSP32(i))
+//!                                     cc_node || 0x11 || ik_node || I2LEOSP32(i))
 //! account_node          = CKDh(CKDh(CKDh(MSK, 32'), 133'), account')
 //! ```
 //!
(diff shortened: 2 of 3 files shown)


skyl:
## Post-merge review of `8ef91176` — independently confirmed, correct

Merged while I was reviewing. Finishing anyway, because a change that moves every key below `MSK` deserves a verdict from something other than the crate that produced it.

### The vectors are right, checked against no code from this repo

I recomputed the chain in Python — `hashlib.blake2b(person=…)` and an HMAC-SHA256 HKDF-Expand written out by hand — starting from BIP-39's published seed for the all-`abandon` mnemonic, which is exactly the property `derivation_vectors.rs:55-63` says the seed choice exists to give a second implementation.

```
MSK      74f055750feeefebb248b31b88aae0b2b9b0b0db1c0f520eb831d30243c8709691d0cc16…de373a
PURPOSE  288530fcef32ace05c9545252ac22fd9d8f070b0961b3bb8e955c0e86922d9061f9b549a…ad0bd9
COIN     8b269b6624b43f1794346249cbd4c0cf6f835467f9937f997265805ad86324641f0c258f…7285cf
ACCOUNT  dbd4245907a1e6e0368d41ad784262d99f52ae7e8f6ce14ab3ef46efb811c6271462b713…801102
```

All four are byte-identical to the constants this PR ships, with the preimage built as `cc ‖ 0x11 ‖ ik ‖ I2LEOSP32(i | 0x8000_0000)` and nothing else. Carrying `ik_account` on into `HKDF-Expand(prk = ik_account, info = "free2z/msg/v1/backup-wrap", L = 32)` reproduces `BACKUP_WRAP` = `479083aada9066af10d708d474199ae99416dd80bd3f5bcefa68c65a2c30270e` exactly. So the whole path — seed, master, three hardened steps, leaf — is confirmed end to end by an implementation sharing no code with `f2z-msg-identity`.

`MSK` being unchanged is the detail that makes the regeneration legible rather than a re-stamp: the master derivation never passes through `CKDh`, so a diff that moved it too would have been a different and larger change than the one described. The header note at `derivation_vectors.rs:21-32` says this before the reader gets to the constants, which is the right place for it.

### The new tests are load-bearing

Baseline on the head: `cargo test -p f2z-msg-identity` → **EXIT=0, 33 + 9 + 5 + 4 + 6 + 5 + 1 passed**.

Reverting the fix — `node.rs:264`, `key.copy_from_slice(node.key());` → `let _ = key;`, restoring the #694 preimage exactly:

```
EXIT=101
node::tests::ckd_reads_both_halves_of_the_parent ... FAILED
node::tests::the_ckd_preimage_is_chain_code_tag_key_index ... FAILED
test result: FAILED. 31 passed; 2 failed
```

Cargo stops after the lib target, so I re-ran with `--no-fail-fast`; the vectors catch it independently:

```
the_account_helper_walks_the_path_the_specification_writes ... FAILED
every_level_of_the_path_matches_its_vector ... FAILED
the_four_leaves_match_their_vectors ... FAILED
the_device_credential_tbs_is_the_bytes_kt_md_declares ... FAILED
the_device_credential_signature_matches_the_vector ... FAILED
the_whole_credential_re_encodes_canonically ... FAILED
test result: FAILED. 3 passed; 6 failed
```

Eight failures across two targets from one line. `ckd_reads_both_halves_of_the_parent` asserting from **both** directions is the part I want to single out: a one-direction assertion would pass for a `CKDh` that read the wrong single field, and the test says so in its own doc comment. `the_ckd_preimage_is_chain_code_tag_key_index` rebuilding the preimage by hand is the field-order oracle the other tests cannot be — they would all still see a deterministic, well-separated tree under a transposition.

`ARCHITECTURE.md:195-196` and `:218-222` carry the corrected formula and the superseded one side by side with the reason. Code, doc, and vectors agree.

### On my own #694 review

I approved #694 having verified two of its five stated properties by mutation and flagged that property 4 had no test. I did not check the derivation against §4.2's own text, which is where this was visible — the module comment cited ZIP 32 §5.2's shape while the preimage was not that shape. **When a crate ships derivation vectors, the vectors have to be checked against the specification the crate cites, not only against the crate.** That is the lesson I am taking, and it is why the independent recomputation above exists rather than another mutation run.

Nothing further blocking. Correct fix, made in the only window where it is free.

PR #706, opened 2026-08-25
test(release): bind Linux packaging fixture selection

### LINKED ISSUE #696: Packaging fixture tests are selected by name pattern; a rename silently deselects them with every gate green
## What is wrong

`.github/workflows/zuuli-packaging.yml:400` runs the real Linux artifact fixtures by **name pattern**:

```yaml
- name: Test Linux artifact inspectors against real packages
  if: runner.os == 'Linux'
  run: |
    node --test --test-name-pattern='real AppImage, deb, and rpm|AppImage inspection|AppImage listing|real deb with an escaping|real cargo-auditable executable evidence' scripts/artifact-sbom.node-test.mjs
```

That exact string is pinned as policy in `wallet/zuuli/scripts/artifact-sbom.mjs:1765` and `:1920`, and `artifact-sbom.node-test.mjs:2373-2381` proves the pin is load-bearing by substituting `--test-name-pattern='nonexistent'` and asserting the policy rejects it.

So the **workflow** cannot drift from the pattern. But nothing binds the pattern to the **test names it selects**. `node --test` exits 0 when a pattern matches nothing, so renaming a test on the other side of the coupling deselects it with every gate green.

## Reproduction

Clean tree, the packaging step's exact command:

```
$ PAT='real AppImage, deb, and rpm|AppImage inspection|AppImage listing|real deb with an escaping|real cargo-auditable executable evidence'
$ node --test --test-name-pattern="$PAT" scripts/artifact-sbom.node-test.mjs
# tests 5
EXIT=0
```

Now drop one word from one test name — `artifact-sbom.node-test.mjs:584`, `"real cargo-auditable executable evidence is complete, lock-bound, and mandatory"` → `"cargo-auditable executable evidence is complete, lock-bound, and mandatory"`:

```
$ node --test --test-name-pattern="$PAT" scripts/artifact-sbom.node-test.mjs
# tests 4
EXIT=0

$ node --test --test-name-pattern='workflow contract' scripts/artifact-sbom.node-test.mjs
# tests 1 / # pass 1
EXIT=0
```

Five selected becomes four, exit stays 0, and the workflow-contract test that pins the selector stays green. The test that proves shipped Linux executables carry a complete, lock-bound Cargo runtime graph is simply not run any more, and no check anywhere says so.

The same holds for the other four names in the pattern; `real cargo-auditable executable evidence` is just the newest and the one whose absence would be least visible.

This is not hypothetical drift bait: these names are long English sentences ("is complete, lock-bound, and mandatory"), exactly the kind that get reworded during review. Renaming a test is normally a safe refactor.

## Why the frontend job does not cover it

`wallet/zuuli` `npm run test` does run `artifact-sbom.node-test.mjs`, but the five fixture tests are gated on `process.platform === "linux" && <tools>.every(commandExists)`, and `cargoRuntimeFixtureTools` includes `cargo-auditable`. The `zuuli / frontend` job (`zuuli.yml:144-215`) installs node, the pinned Rust toolchain and npm deps — it does **not** install `cargo-auditable`, which is a `cargo install`. Only `zuuli-packaging.yml:391-395` installs it. So the packaging step selected by that pattern is the single place this test executes, and a rename silently empties it.

## Suggested fix

Bind the two sides in the direction that is currently unbound: have the node-test assert that every alternative in the pinned pattern matches at least one registered test name in `artifact-sbom.node-test.mjs`. The names are already available in the file; splitting the pinned pattern on `|` and asserting each alternative is a substring of some `test(...)` title in the source is enough, and it fails closed on both a rename and a pattern edit.

The repo already has the stronger idiom for this — `rs/crates/f2z-codec/tests/workspace_debug_scan.rs:261-269` derives its population and asserts it *reached* every member rather than trusting a hand-written list. Same shape applies here: derive the selected set, assert it is non-empty and equals the intended five.

A secondary hardening, independent of the rename problem: the packaging step is green whether those tests **run** or **skip**. If a tool ever drops off the runner image, `canBuildCargoRuntimeFixture` goes false and the step passes with five skips. Asserting the expected pass count (or failing when a Linux runner reports a skip) closes that too.

## Severity

Medium. Nothing is broken today — the tests do run on PRs touching `wallet/zuuli/**`. The defect is that the strongest evidence in the Linux release path can be removed by an ordinary refactor without any gate noticing, which is the property the surrounding policy code goes to considerable length to prevent everywhere else.

Found while reviewing #695.

skyl:
Closes #696

Replace the unbounded English-name regex workflow step with a fail-closed runner that binds the exact current Linux fixture population and proves every selected fixture body completes.

Independent mutation review: APPROVE, zero survivors. On the real Ubuntu/Node 24 runner the restored baseline is exactly 4 tests, 4 passed, 0 skipped. All four selector deletions, selector broadening, coordinated broadening, `skip: true`, completion-record deletion, selected nonfixture execution, required-env/tool/status weakening, and nonexistent/comment-only/soft-fail workflow substitutions turn red.

The fifth cargo-auditable fixture exists only on open #695; after this merges, that PR must raise the independent population to five and replay its fifth-title mutations before merge.

### DIFF
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -391,7 +391,7 @@ jobs:
       - run: npm ci
       - name: Test Linux artifact inspectors against real packages
         if: runner.os == 'Linux'
-        run: node --test --test-name-pattern='real AppImage, deb, and rpm|AppImage inspection|AppImage listing|real deb with an escaping' scripts/artifact-sbom.node-test.mjs
+        run: node scripts/run-linux-artifact-fixtures.mjs
       - run: npm run release:verify
       - name: Verify Rust compiler
         run: rustc --version | grep -F "rustc $ZUULI_RUST_VERSION "
--- a/wallet/zuuli/scripts/artifact-sbom.mjs
+++ b/wallet/zuuli/scripts/artifact-sbom.mjs
@@ -1572,6 +1572,34 @@ function requireExactRunStep(
   }
 }
 
+function requireExactCommandStep(
+  label,
+  block,
+  stepName,
+  expectedCommand,
+  failures,
+  expectedIf,
+) {
+  const marker = `\n      - name: ${stepName}\n`;
+  const count = block.split(marker).length - 1;
+  if (count !== 1) {
+    failures.push(
+      `${label}: expected one named ${stepName} step, found ${count}`,
+    );
+    return;
+  }
+  const rest = block.slice(block.indexOf(marker) + marker.length);
+  const next = rest.search(/^      - /m);
+  const step = (next === -1 ? rest : rest.slice(0, next)).trimEnd();
+  const expected = [
+    ...(expectedIf ? [`        if: ${expectedIf}`] : []),
+    `        run: ${expectedCommand}`,
+  ].join("\n");
+  if (step !== expected) {
+    failures.push(`${label}: ${stepName} executable step changed`);
+  }
+}
+
 function requireExactActionStep(
   label,
   block,
@@ -1616,6 +1644,14 @@ export function artifactSbomWorkflowFailures(packaging, release) {
   const failures = [];
   const packagingLinux = jobBlock(packaging, "desktop");
   const releaseLinux = jobBlock(release, "linux");
+  requireExactCommandStep(
+    "packaging linux",
+    packagingLinux,
+    "Test Linux artifact inspectors against real packages",
+    "node scripts/run-linux-artifact-fixtures.mjs",
+    failures,
+    "runner.os == 'Linux'",
+  );
   const prepareLinuxLines = [
     "appimages=(release-artifacts/*.AppImage)",
     "debs=(release-artifacts/*.deb)",
@@ -1763,7 +1799,7 @@ export function artifactSbomWorkflowFailures(packaging, release) {
       "packaging linux",
       jobBlock(packaging, "desktop"),
       [
-        "node --test --test-name-pattern='real AppImage, deb, and rpm|AppImage inspection|AppImage listing|real deb with an escaping' scripts/artifact-sbom.node-test.mjs",
+        "node scripts/run-linux-artifact-fixtures.mjs",
         'node scripts/artifact-sbom.mjs prepare --artifact="${appimages[0]}" --root=artifact-sbom-work/linux-appimage/root',
         'node scripts/artifact-sbom.mjs prepare --artifact="${debs[0]}" --root=artifact-sbom-work/linux-deb/root',
         'node scripts/artifact-sbom.mjs prepare --artifact="${rpms[0]}" --root=artifact-sbom-work/linux-rpm/root',
--- a/wallet/zuuli/scripts/artifact-sbom.node-test.mjs
+++ b/wallet/zuuli/scripts/artifact-sbom.node-test.mjs
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { execFileSync } from "node:child_process";
+import { execFileSync, spawnSync } from "node:child_process";
 import {
   appendFileSync,
   copyFileSync,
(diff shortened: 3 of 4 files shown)


skyl:
## Verified — #696 is genuinely closed. I ran both halves of my reproduction.

The mechanism is right, and it is stronger than what I asked for in the issue. I asked for the pattern's alternatives to be checked against registered test names; you inverted it instead — `linuxArtifactFixture(title, …)` is the only way to register a Linux fixture, it asserts membership at registration, and `after()` asserts the registered set *equals* the shipping selector. That is the `workspace_debug_scan.rs` shape: derive the population and assert you reached all of it, rather than trusting a hand-written list.

### Half A — the defect, on `main` (`ae9de6e`)

Rename one fixture so it falls out of the prefix pattern, `"AppImage listing rejects…"` → `"AppImage list rejects…"`, then run the packaging step's exact command:

```
baseline:            # tests 4   EXIT=0
after the rename:    # tests 3   EXIT=0
full suite after:    # tests 16  # pass 12  EXIT=0
```

Four selected becomes three, exit stays 0, nothing anywhere goes red. That is #696 as filed.

### Half B — the fix, on `c56836d1`

Same rename, `"…invalid UTF-8 bytes"` → `"…invalid UTF8 bytes"` at the `linuxArtifactFixture` call site, leaving the selector alone:

```
packaging selector:  EXIT=1
  AssertionError: Linux artifact fixture has an unregistered title:
    AppImage listing rejects a SquashFS member with invalid UTF8 bytes
  # tests 3  # fail 1

plain `node --test` (the frontend job):  EXIT=1, same assertion
  # tests 4  # fail 1
```

Red in **both** runs, which is the part I did not expect and like most: the registration assertion fires during file evaluation, so `zuuli / frontend` catches a drifted title on every PR, not only the ones that reach packaging. Clean baseline on the head is `# tests 17, # pass 13, # fail 0, # skipped 4` — the four skips are the Linux fixtures on macOS, correctly gated.

The two independent-authority checks are the right idiom too: `linuxArtifactFixturePattern()` derives the regex and `EXPECTED_LINUX_ARTIFACT_FIXTURE_PATTERN` is the literal it must equal, so neither side can move alone. `ZUULI_REQUIRE_LINUX_ARTIFACT_FIXTURES=1` turning a skip into a hard error on a Linux runner closes the second half of #696 that I flagged but did not demonstrate — the step being green whether the fixtures *ran* or merely *skipped*. `beforeEach` refusing a selected non-fixture closes the other direction.

### One residual, not a blocker

Nothing forces a *future* Linux fixture through the wrapper. A new `test("…", { skip: !canBuildLinuxFixtures, … }, …)` written by hand would run in the frontend job, never be selected by packaging, and leave every assertion here green — the registered set would still equal the four, because the new one never registered.

That is cheap to close now that the file has exactly one use site: `grep -c 'canBuildLinuxFixtures' scripts/artifact-sbom.node-test.mjs` finds the definition at `:126` and a single `skip: !canBuildLinuxFixtures` at `:152`, inside the wrapper. Asserting that occurrence count is exactly 1 — the `wasm-boundary.mjs` `occurrenceCount(...) !== 1` idiom this repo already uses — makes "the wrapper is the only door" a checked fact rather than a convention.

### On #695

Your note that #695 must raise the population to five and replay its fifth-title mutations is exactly right, and it is now a *mechanical* requirement rather than a remembered one: `EXPECTED_LINUX_ARTIFACT_FIXTURE_COUNT` and the literal pattern both have to move, and `after()` will fail if the fifth fixture registers without them. Worth saying on #695 directly so it is not carried only by this PR's description.

### Verdict

Approve. `rs / gate` is already SUCCESS; merge once `gate` is green on this exact head — it is currently `BEHIND` with the frontend and packaging jobs still running, and `Desktop / Linux packages` is the one I would actually watch, since it is the only place the four fixture bodies execute. I will close #696 when this lands.

PR #707, opened 2026-08-25
test(store): bind creator catalog page contract

skyl:
Refs #387

Harden the deterministic store-capture request contract against ordinary and decorative query drift for the explicit creator-catalog `page=1` request. This is a test-only slice; it does not claim the screenshots were recaptured and does not close the store-media umbrella.

Independent mutation review: APPROVE, zero survivors. Missing page, `page=2`, ignored `page_size`, a decorative `page=1` on the note or creator-detail request, and a broad `includes("page=1")` membership bypass all turn red. The restored exact request passes an independent literal oracle. All 60 store-listing tests, `store:validate`, and diff checks are green.

The attempted two-pass recapture replaced no artifacts. Current capture tooling still needs a separately reviewed digest-pinned Playwright + Rust 1.97.1/wasm32 composition and corrected wallet-root mount before the mandatory production WASM build can run reproducibly.

### DIFF
--- a/wallet/zuuli/scripts/store-screenshot-contract.node-test.mjs
+++ b/wallet/zuuli/scripts/store-screenshot-contract.node-test.mjs
@@ -244,7 +244,7 @@ test("feed capture scroll is geometry-derived and keeps a required control unobs
   assert.equal(computeFeedCaptureScroll({ currentScroll: 0, visibleTop: 60, visibleBottom: 580, controlTop: 240, contentBottom: 820, requireControl: false }), 240);
 });
 
-test("creator capture pins the authoritative first catalog page and rejects pagination drift", () => {
+test("creator capture pins the authoritative first catalog page and rejects decorative pagination", () => {
   const creatorDetail = "GET https://free2z.cash/api/creator/example_editorial/";
   const firstPage = "GET https://free2z.cash/api/zpage/?ordering=-created_at&page=1&page_size=12&username=example_editorial";
   assert.deepEqual(CAPTURE_PUBLIC_REQUESTS["creator-profile"], [creatorDetail, firstPage]);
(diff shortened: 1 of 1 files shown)


skyl:
## Approve, with a note on how the change should be described

Small and harmless. Baseline on `main` with a real `npm ci`: `node --test scripts/store-screenshot-contract.node-test.mjs scripts/store-contract.node-test.mjs scripts/store-state-audit.node-test.mjs` → **EXIT=0, 83 tests, 83 passed**.

*(My first attempt at that baseline came back 3/3 failed on `ERR_MODULE_NOT_FOUND: pngjs`. That was my environment — the checkout I symlinked `node_modules` from is older than `package.json`. Diagnosed rather than reported.)*

### What I probed

Your description reads as if the two new drift cases close something. I went looking for the mutation they close, and could not find one — every candidate is already caught on `main`:

**1. The `includes("page=1")` membership bypass you name.** Replacing the body of `capturePublicRequestAllowed` with `return key.includes("page=1");` on **`main`**:

```
node --test --test-name-pattern='creator capture pins' …
EXIT=1   # tests 1  # pass 0  # fail 1
```

Already red before this PR — main's third drift case is `…&page=1&page_size=24&…`, which contains `page=1`, so the bypass wrongly allows it and the existing assertion fires. True that it turns red on your head; it turned red before too.

**2. Endpoint widening — the one case your addition is genuinely alone on.** The creator-detail drift you added is the only assertion anywhere on that URL, so:

```js
if (key.startsWith("GET https://free2z.cash/api/creator/")) return true;
return CAPTURE_PUBLIC_REQUESTS[action]?.includes(key) ?? false;
```

survives the creator test on `main` — `EXIT=0, # pass 1`. But it does **not** survive the suite: run all three files and it is **EXIT=1, 75 passed, 8 failed**, caught by `repository capture plan, provenance, and exact entry matrix validate`, `capture records reject stale source pixels…`, `record metadata cannot waive disclosure checks or alter routes`, `the repository captured-media contract validates and publication fails closed`, and `release identity is canonical, returned, and cannot drift`.

**3. Widening the allow-list itself** — adding `"GET https://free2z.cash/api/private/analytics/?token=leak"` to `fresh` — is caught by the same 8. I had expected this one to survive, because `CAPTURE_PUBLIC_REQUESTS` is `deepEqual`-pinned only for `creator-profile` (`store-screenshot-contract.node-test.mjs:250`) and not for `fresh`, `search-privacy` or `article-reader`. It does not survive: the coverage is in those other files, and `store-screenshot-capture.mjs:398` reconciling observed requests against the same list is not the only thing holding it. Worth recording, because the inspection-only read of that file is misleading.

### The ask

Nothing to change in the diff. Please just describe it as what it is: it **extends the enumeration** of rejected creator-catalog drift, rather than closing a gap. Re-reading your description, "all turn red" is literally accurate — it is my inference that they *newly* turn red that was wrong, and I would rather that inference not be available to the next reader either.

The one thing that would make it coverage rather than documentation: the creator test is the only assertion on the `/api/creator/…` endpoint, and on its own it misses endpoint widening — the eight tests in the other two files are what actually catch that. If the intent is for `store-screenshot-contract.node-test.mjs` to stand alone as the request contract, then pinning the remaining three actions' allow-lists with the same `deepEqual` used at `:250` for `creator-profile` is the change that would earn it.

`rs / gate` and `gate` need to be green on `ad2cb19b`; it is currently `BEHIND`.

Issue #272, opened 2026-08-08
feat(zuuli): explicitly import preserved legacy wallets into multi-wallet state


skyl:
## Context

Issue #271 makes startup safe when canonical and legacy app-data directories both contain data: it opens only the canonical current wallet, preserves the legacy tree byte-for-byte, and surfaces an import-pending diagnostic. It intentionally does not merge or choose populated wallet identities implicitly.

## Required importer

Build an explicit, user-confirmed, crash-safe importer from the preserved `com.2zinc.zuuli` tree into canonical multi-wallet state.

- Inventory and validate the legacy manifest/database plus every SQLite WAL/SHM sidecar without opening or mutating canonical state prematurely.
- Support legacy single-wallet and multi-wallet manifests.
- Validate every recovered seed against the database UFVK before publication.
- Support the legacy directory’s own `.seeds/salt` encrypted-file custody and native/keyring migration rules; never assume the canonical directory salt.
- Stage destination DB names without replacement, precommit a durable generation-bound import journal, publish the canonical manifest atomically, then migrate native custody with readback before deleting any legacy material.
- Preserve both source and canonical state on cancellation, lock, corruption, mismatch, ambiguous fsync, or crash.
- Detect duplicate wallet identities/UFVKs and require an explicit non-destructive resolution.
- Provide preview, confirmation, progress, retry, and diagnostics in both ZUULI and Zuuallet.
- Cover adversarial crash boundaries and macOS/iOS, Android sandbox limits, Linux Secret Service, and Windows Credential Manager behavior.

Blocked by / follows #271. This is release-blocking for automatic legacy recovery but not for launch safety.

skyl:
Dependency status: temporarily **blocked** until #273 merges and closes #271. The explicit importer must be built on the final launch-safety and structured `importPending` contract from that PR; starting against the pre-merge plugin surface would create avoidable custody conflicts.

skyl:
Dependency cleared: #273 merged as `1e67b7b` and #271 is closed. This importer can now move forward against the stable `importPending`/launch-safety contract. Keep this issue as the tracking umbrella and implement it through small sequential PRs: read-only inventory/preview first, then journaled import engine, custody migration, frontend integration, and adversarial platform/crash tests.

skyl:
[...]

Issue #705, opened 2026-08-25
zuuli's typecheck script excludes all 58 test files, so nothing typechecks them


skyl:
`wallet/zuuli`'s `typecheck` script points at a config that excludes every test file, so nothing in the repo typechecks them any more.

```json
// wallet/zuuli/tsconfig.build.json
{ "extends": "./tsconfig.json", "exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"] }
```

```json
"build":     "npm run wasm:build && tsc -p tsconfig.build.json && vite build && npm run wasm:verify-dist",
"typecheck": "tsc --noEmit -p tsconfig.build.json",
```

`.github/workflows/zuuli.yml:203-204` runs `npm run typecheck`, so CI's typecheck is the narrowed one. `vite.config.ts`'s `test:` block sets `env` only — no `typecheck` key — and vitest transpiles with esbuild, which strips types without checking them.

## Measured

On #690's head:

```
$ tsc --noEmit -p tsconfig.build.json --listFiles | grep -c 'wallet/zuuli/src/'
164
$ tsc --noEmit -p tsconfig.build.json --listFiles | grep -c 'wallet/zuuli/src.*\.test\.'
0
$ find src -name '*.test.ts' -o -name '*.test.tsx' | wc -l
58
```

Appending a deliberate type error to `src/lib/wallet/zuuallet-created-seed.test.ts`:

```ts
const reviewerTypeError: number = "not a number";
void reviewerTypeError;
```

| command | result |
|---|---|
| `tsc --noEmit -p tsconfig.build.json` (what `typecheck` and CI run) | **0 errors mentioning that file** |
| `tsc --noEmit` (the config `main` uses today) | `zuuallet-created-seed.test.ts(292,7): error TS2322: Type 'string' is not assignable to type 'number'.` |
| `vitest run src/lib/wallet/zuuallet-created-seed.test.ts` | **10 passed, EXIT=0** |

There is no configuration under which that error is reported.

*(Baseline caveat: the `tsconfig.build.json` run also reports 5 unrelated `zod` / `linkedom/worker` resolution errors because I symlinked `main`'s `node_modules` into a fresh worktree rather than running `npm ci`. They appear identically with and without the injection and none of them names the injected file.)*

## How it got here

[...]

PR #712, opened 2026-08-25
feat(zuuli): expose safe legacy wallet preview

skyl:
Refs #272

## Scope
- wire the existing native read-only legacy wallet preview into an explicit Inspect/Retry notice
- render only aggregate counts, layout, custody state, and sidecar presence; identifiers, names, filenames, fingerprints, and native error text remain redacted
- invalidate in-flight results across authority loss, unmount, and retries; no import, deletion, or custody migration occurs
- add browser mock coverage and require the exact frontend test-discovery contract in Actions policy

This is a bounded preview slice. #272 stays open for the actual migration/custody/deletion workflow.

## Verification
- focused Vitest: 11 passed
- production and test typechecks
- production build with WASM dist verification
- Chromium: 1 passed
- live Actions policy: 187 refs / 16 workflows
- policy self-tests: 336 cases
- Zcash permission policy live + self-tests

## Mutation evidence
- removing the synchronous active-request guard makes a same-act double click invoke native twice (red)
- removing authority-transition invalidation lets an old result render after none→pending churn (red)
- narrowing the package test script to unrelated Vitest/Playwright files is rejected by the live workflow policy (red)
- publishing an old finally or stale rejection over a newer retry loses the new request state / renders stale failure UI (red)

Independent adversarial review approved exact head 25e9c4709c541548425805e02fe85d1f4e9488ae and patch 2a7ca7883d226d9106f19aa4624e59c7bf8b1c30.


skyl:
## Review of `25e9c470`

I went after the privacy claim, since that is what the PR is for. **The strong parts are stronger than the description says, and one part is weaker — and the two are grouped together in a single sentence.**

### Where redaction actually happens

Tracing `previewLegacyWalletImport` back through `bridge.ts:75` to `wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs`:

| field | how it is protected |
|---|---|
| `walletId`, `walletName` | **`[REDACTED]` at the native boundary** (`:120-121`, `:334-335`) — the real values never cross |
| native error text | **cannot be interpolated** — `preview_desktop` is `Result<_, &'static str>` (`:71`), and every `Err` is a fixed literal |
| `dbFilename` | **real value crosses**: `db_filename: entry.db_filename.clone()` (`:336`) |
| `ufvkFingerprints` | **real value crosses**: a 64-hex-char fingerprint per account |

The first two are enforced by construction. `Result<_, &'static str>` in particular is the best thing here — a future `format!("{err}")` in that function is a compile error, not a review catch. Worth keeping when this file grows.

The last two are protected only by the component choosing not to read them. That is a genuinely different strength, and the description's *"identifiers, names, filenames, fingerprints, and native error text remain redacted"* reads as one guarantee across all five.

To be fair to the wording: parsed strictly as "**render** only aggregate counts …; [the rest] remain redacted *from the render*", it is accurate, and `LegacyWalletNotice.test.tsx:188-191` proves it with canary values (`must-not-render.sqlite`, `must-not-render-fingerprint`) asserted absent from `container.textContent`. That is assert-on-the-artifact done properly. And the native side is explicit about the intent — its test is literally named `multi_layout_preserves_exact_db_names_but_redacts_manifest_identity`, so nothing here is accidental.

### The suggestion

`LegacyWalletNotice.tsx` reads exactly five things from the preview:

```
accountCount, encryptedCustodyPresent, walPresent, shmPresent, preview.diagnostics
```

It never reads `dbFilename` or `ufvkFingerprints`, and `git grep previewLegacyWalletImport` finds it is the **only** consumer outside the bridge and the mock. So those two fields could be `[REDACTED]` at the native boundary exactly like `wallet_id`/`wallet_name`, at zero cost to this feature.

That converts the weaker half into the stronger half. As it stands, the real filename and per-account viewing-key fingerprints sit in the JS heap and the React tree, where the canary test cannot follow them — an error boundary serialising props, a crash reporter, a future `console.log`, or React DevTools all see the whole object. Redacting natively means nothing downstream *can* leak them, which is the property the other two fields already have.

If the values are wanted later for the actual migration workflow (#272), that is a good reason to keep them in the **native** type and add them to the bridge type when a consumer exists — rather than shipping them across now for a screen that does not use them.

### Smaller things, all good

- The `"test"` script is now byte-pinned in `check-github-actions-pins.mjs:322-325` with a live mutation that narrows it to unrelated files. That is the right shape, and it is the same class of protection whose absence caused #705 — good to see it applied pre-emptively rather than after a regression.
- The two new scripts are additive, so `vitest run` and `playwright test` in the pinned `test` still discover `LegacyWalletNotice.test.tsx`, `legacy-import-preview-bridge.test.ts` and `legacy-import-preview.pw.ts`. Coverage is real, not just declared.
- The native secret-scan over the serialized preview (`for secret in [raw_ufvk, "secret account 0", …] { assert!(!serialized.contains(secret)) }`) is a good population check — the raw UFVK demonstrably does not cross, only the derived fingerprint.
- `#272 stays open for the actual migration/custody/deletion workflow` — accurate scoping, and the diff matches it.

### What I did not check

The `check-github-actions-pins.mjs --self-test` run (>2 min locally) and the Playwright case; both are CI's. I read the new mutation's shape against its neighbours rather than executing it.

Not blocking. The suggestion above is the only substantive one, and it is a two-line change in `legacy_import_preview.rs` plus the fixtures that pin the current values.

PR #713, opened 2026-08-25
test(release): make Cargo fixture checks falsifiable

### LINKED ISSUE #711: Two verification steps from #695 cannot fail: the cargo-auditable post-install grep, and the required-fixtures precondition
#695 merged as `8437bcb`. Two verification steps it added cannot fail in the way they are meant to. Both are cheap, neither is a live defect, and I raised both on the PR (5408537589) before it landed — filing so they are not carried only by a merged PR's comments.

## 1. The post-install check cannot tell a patched build from an unpatched one

`wallet/zuuli/scripts/install-cargo-auditable.sh`, last three lines:

```sh
cargo install --force --locked --path "$temporary/cargo-auditable-$version"
installed=$(cargo install --list)
printf '%s\n' "$installed" | grep -Eq "^cargo-auditable v${version}( \\([^)]*\\))?:$"
```

The script's whole purpose is to install a **locally patched** cargo-auditable — `cargo-auditable-0.7.5-root.patch`, which fixes the duplicate-root bug that broke `Bind Linux shipped-artifact SBOMs`. But the path parenthetical is optional in that regex, so it matches both:

```
cargo-auditable v0.7.5:                       ← plain `cargo install` from crates.io, unpatched
cargo-auditable v0.7.5 (/tmp/…/cargo-auditable-0.7.5):   ← the patched path build
```

The two differ only in the `( \([^)]*\))?` group the regex makes optional. An earlier head of #695 used `grep -Fx 'cargo-auditable v0.7.5:'`, which at least pinned "came from crates.io". The current form pins neither, so the step reads as evidence and is not.

Requiring the path form is free:

```sh
printf '%s\n' "$installed" | grep -Eq "^cargo-auditable v${version} \\(/[^)]*\\):$"
```

That still is not proof the *patch* applied — only that a path build was installed. The real proof is downstream: `linux-cargo-runtime.mjs`'s `validateAuditableDocument` requires exactly one root, so an unpatched binary fails the packaging job. That backstop is good and was demonstrated twice. The point is narrower: a check that exists should be able to fail, or it should not be there.

## 2. The required-fixtures precondition does not know about `cargo-auditable`

`wallet/zuuli/scripts/artifact-sbom.node-test.mjs:133-144`:

```js
const missingLinuxFixtureTools = linuxFixtureTools.filter((command) => !commandExists(command));
if (requireLinuxFixtures && (process.platform !== "linux" || missingLinuxFixtureTools.length > 0)) {
  throw new Error(`required Linux artifact fixtures cannot run: platform=…; missing tools=…`);
}
```

`linuxFixtureTools` is `cc`, `dpkg-deb`, `mksquashfs`, `objcopy`, `readelf`, `rpm2archive`, `rpmbuild`, `unsquashfs`. The fifth fixture is gated on `canBuildCargoRuntimeFixture`, whose `cargoRuntimeFixtureTools` adds `cargo` and **`cargo-auditable`** — and the precondition never consults it.

So on a Linux runner where `install-cargo-auditable.sh` left `cargo-auditable` off `PATH`, the precondition passes, the fifth fixture's `available` is false, and it skips. It is still **fail-closed** — `after()`'s completion-record `deepEqual` catches the missing title — but the failure arrives as a set mismatch rather than the precondition's `missing tools=cargo-auditable`, which is the message that would tell an operator what to fix.

Folding the fixtures' own tool requirements into that precondition is a small change and keeps the clear diagnostic.

## What is *not* wrong

Worth recording, because #695 got the substance right and the CI evidence is unusually good. Job `97755454800` on `3bd6274b`:

```
✔ real cargo-auditable executable evidence is complete, lock-bound, and mandatory (8378.039024ms)
ℹ tests 5   ℹ pass 5   ℹ skipped 0
```

Five selected, five passed, zero skipped — the fifth fixture genuinely executed on a real runner rather than skipping, which is exactly what #696 was about. The source pin in the installer is done properly (`curl` from `static.crates.io`, `sha256sum --check` against a pinned digest, `patch --fuzz=0 --batch --forward` so the input is immutable and the patch cannot apply fuzzily), and the `RUSTC_WRAPPER` correctly confines `RUSTC_BOOTSTRAP=1` to cargo — I verified that against a stub rustc: without the wrapper `RUSTC_BOOTSTRAP=1`, with it `<unset>`, arguments intact, and no-args exits 64.

Neither item here needs urgency. They are the two places in that pipeline where a step's outcome is independent of the thing it is checking.

skyl:
Closes #711

## What changed
- require `cargo install --list` to identify the exact temporary path build, so a registry install cannot satisfy the post-install check
- make the required Linux fixture precondition cover the union of every selected fixture tool, including cargo-auditable
- add isolated registry-style output and cargo-auditable-only missing-tool controls

## Verification
- artifact SBOM suite: 22 tests, 17 passed / 5 host-skipped on macOS
- shell syntax check
- Actions policy live: 187 references / 16 workflows
- Actions policy self-test: 334 cases
- real Cargo probe confirmed `cargo install --list` preserves the exact absolute `--path` source, including a symlink path

## Mutation evidence
- restoring the optional registry/path regex makes the focused installer test red at `a crates.io-style listing must not verify the patched path install`
- narrowing the precondition back to the old Linux-only list makes the focused runner test red because cargo, cargo-auditable, objcopy, and rustc disappear from the diagnostic

Independent adversarial review approved exact head 375029c52c9a7a9a75070cdb40a7665d7b31bb9e and stable patch 749489af837c98b8b2eb792195f8260826b5f71d.

### DIFF
--- a/wallet/zuuli/scripts/artifact-sbom.node-test.mjs
+++ b/wallet/zuuli/scripts/artifact-sbom.node-test.mjs
@@ -119,6 +119,16 @@ const linuxFixtureTools = [
   "rpmbuild",
   "unsquashfs",
 ];
+const cargoRuntimeFixtureTools = [
+  "cargo",
+  "cargo-auditable",
+  "dpkg-deb",
+  "objcopy",
+  "rustc",
+];
+const requiredLinuxFixtureTools = [
+  ...new Set([...linuxFixtureTools, ...cargoRuntimeFixtureTools]),
+].sort();
 
 function commandExists(command) {
   return (process.env.PATH ?? "")
@@ -130,7 +140,7 @@ const canBuildLinuxFixtures =
   process.platform === "linux" && linuxFixtureTools.every(commandExists);
 const requireLinuxFixtures =
   process.env[REQUIRE_LINUX_ARTIFACT_FIXTURES] === "1";
-const missingLinuxFixtureTools = linuxFixtureTools.filter(
+const missingLinuxFixtureTools = requiredLinuxFixtureTools.filter(
   (command) => !commandExists(command),
 );
 if (
@@ -244,8 +254,52 @@ test("Linux packaging runner binds every selected fixture to a real passing test
   assert.notEqual(missingToolProbe.status, 0);
   assert.match(
     `${missingToolProbe.stdout}${missingToolProbe.stderr}`,
-    /required Linux artifact fixtures cannot run: .*missing tools=cc,dpkg-deb,mksquashfs,readelf,rpm2archive,rpmbuild,unsquashfs/,
+    /required Linux artifact fixtures cannot run: .*missing tools=cargo,cargo-auditable,cc,dpkg-deb,mksquashfs,objcopy,readelf,rpm2archive,rpmbuild,rustc,unsquashfs/,
   );
+  const incompleteToolDirectory = mkdtempSync(
+    resolve(tmpdir(), "zuuli-linux-fixture-tools-"),
+  );
+  try {
+    for (const command of [
+      "cargo",
+      "cc",
+      "dpkg-deb",
+      "mksquashfs",
+      "objcopy",
+      "readelf",
+      "rpm2archive",
+      "rpmbuild",
+      "rustc",
+      "unsquashfs",
+    ]) {
+      symlinkSync(process.execPath, resolve(incompleteToolDirectory, command));
+    }
+    const missingCargoAuditableEnvironment = {
+      ...process.env,
+      PATH: incompleteToolDirectory,
+      [REQUIRE_LINUX_ARTIFACT_FIXTURES]: "1",
+    };
+    delete missingCargoAuditableEnvironment.NODE_TEST_CONTEXT;
+    const missingCargoAuditableProbe = spawnSync(
+      process.execPath,
+      [
+        "--test",
+        "--test-name-pattern=nonexistent",
+        resolve(scriptDirectory, "artifact-sbom.node-test.mjs"),
+      ],
+      {
+        encoding: "utf8",
+        env: missingCargoAuditableEnvironment,
+      },
+    );
+    assert.notEqual(missingCargoAuditableProbe.status, 0);
+    assert.match(
+      `${missingCargoAuditableProbe.stdout}${missingCargoAuditableProbe.stderr}`,
+      /required Linux artifact fixtures cannot run: .*missing tools=cargo-auditable(?:\s|$)/,
+    );
+  } finally {
+    rmSync(incompleteToolDirectory, { recursive: true, force: true });
+  }
   assert.throws(
     () =>
       runLinuxArtifactFixtures({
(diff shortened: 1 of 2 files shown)


skyl:
## Verified — both halves of both findings in #711. Approve.

### 1. The post-install check now rejects a registry install

Your fix is **stronger than what I suggested**. I proposed requiring any path form (`\(/[^)]*\):$`); you pin the exact temporary path just built:

```sh
expected="cargo-auditable v${version} ($temporary/cargo-auditable-$version):"
printf '%s\n' "$installed" | grep -Fx -- "$expected"
```

Run against synthetic `cargo install --list` output, four cases:

| listing | old regex | new `grep -Fx` |
|---|---|---|
| `cargo-auditable v0.7.5:` (registry) | **ACCEPTED — #711's defect** | rejected |
| `cargo-auditable v0.7.5 ($temporary/…):` (this build) | accepted | accepted |
| `cargo-auditable v0.7.5 (/tmp/somewhere-else/…):` | accepted | **rejected** |

The third row is the part I did not ask for and is worth having: a path install from some *other* directory no longer satisfies it either, so the check now identifies the specific artifact rather than a category.

Pinning an exact absolute path is the kind of thing that breaks on canonicalisation, so I am glad you probed it — "real Cargo probe confirmed `cargo install --list` preserves the exact absolute `--path` source, including a symlink path" is exactly the right thing to have checked before relying on `-Fx`.

### 2. The precondition now covers the union, and names the missing tool

Reproduced on `main` (`8437bcb`) with a fabricated `PATH` containing every `linuxFixtureTools` entry plus `cargo`, `objcopy` and `rustc` — everything except `cargo-auditable`:

```
main:      required Linux artifact fixtures cannot run: platform=darwin; missing tools=none
pr-713:    required Linux artifact fixtures cannot run: platform=darwin; missing tools=cargo-auditable
```

`missing tools=none` with `cargo-auditable` absent is #711's second finding, exactly. On a Linux runner main would not have thrown at all — the fixture would simply have skipped and been caught later by the completion-record `deepEqual`. Now the precondition names it.

The `requiredLinuxFixtureTools` union built with `[...new Set([...a, ...b])].sort()` is the right shape — derived rather than a third hand-written list that could drift from the two it merges. And the new isolated control that symlinks ten of the eleven tools and asserts the diagnostic names the eleventh is a real control, not just a restatement of the change.

### A failure I saw and did **not** report

My first full-suite run on your head came back `22 tests, 16 passed, 1 failed` — `macOS DMG inventory is copied without following its Applications link`, with `failed to detach artifact DMG safely`. Against your description's "17 passed / 5 host-skipped", that looked like a regression.

It was mine. `hdiutil info` showed ten mounted images on this machine, three of them `zuuli-artifact-dmg-*` left behind by my own earlier runs of this suite. Controls:

```
main, full suite:      22 tests, 17 passed, 0 failed
pr-713, full suite:    22 tests, 16 passed, 1 failed   ← first run
pr-713, full suite:    22 tests, 17 passed, 0 failed   ← re-run
```

A single-run difference between two trees is not a distinguishing case when the failing test mounts and detaches disk images on a host with stale mounts. Re-running reproduced your numbers. I have detached the images I left behind.

Recording it because the alternative — reporting a red suite on the strength of one run — would have cost you a round trip chasing a defect that was on my desk, and because it is the eighth time this session a failing baseline has been my environment rather than the repo.

### Verdict

Approve. `#711` is answered on both counts, and the fix on the first is better than the issue asked for. I will close #711 when this merges, with these reproductions attached.

Issue #710, opened 2026-08-25
json.rs: "every_container_says_which_half_is_authoritative" tests one of two branches, and not the one production uses


skyl:
`f2z-kt/src/json.rs` exists to hold one rule, and its header states the mechanism:

> Every response therefore carries `"authoritative": "tls_codec"` — a machine-readable statement of which half a client is supposed to use, **so that a client author who reaches for the convenient half has to ignore a field that says not to.**

`container()` has two branches. Only one of them is tested, and it is not the one production uses.

```rust
pub fn container(bytes: &[u8], fields: &str) -> String {
    if fields.is_empty() {
        format!("{{\"authoritative\":\"tls_codec\",\"encoding\":\"base64url\",\"tls_codec\":\"{}\"}}", …)
    } else {
        format!("{{\"authoritative\":\"tls_codec\",\"encoding\":\"base64url\",\"tls_codec\":\"{}\",\"rendered\":{{{fields}}}}}", …)
    }
}
```

```rust
#[test]
fn every_container_says_which_half_is_authoritative() {
    let body = container(b"abc", "");     // ← the empty-fields branch, only
    assert!(body.contains("\"authoritative\":\"tls_codec\""));
    assert!(body.contains("\"tls_codec\":\"YWJj\""));
}
```

"Every container" is one container.

## Measured

`cargo test -p f2z-kt --lib` on `main` (`9efb7e7`): **36 passed, EXIT=0**.

Dropping `\"authoritative\":\"tls_codec\",` from the **`else`** branch only, and running the whole crate — lib plus every integration test, `redaction.rs` included:

```
cargo test -p f2z-kt --no-fail-fast
EXIT=0        (16 + 1 + 4 + 8 + 5 + 4 + 36 passed, 0 failed)
```

## The untested branch is the one that ships

`fields` is non-empty for every response a reader would actually fetch:

- `api.rs:111-112` — the log descriptor, `crate::descriptor::render(&state.descriptor)`
- `api.rs:120-131` — the tree head, a `format!`-built rendering
- every error, because `error_container` always builds at least `"code":{code}` and hands it to `container`

The empty-`fields` call — the only one under test — is the shape used by the responses that render nothing. So the field that tells a `curl` user "do not trust the convenient half" is absent from exactly the JSON bodies that *have* a convenient half, and no test would notice if it disappeared from them.

[...]

Issue #709, opened 2026-08-25
WitnessState::advance's token guard has five clauses and every one can be deleted with the suite green


skyl:
`WitnessState::advance` (`rs/crates/f2z-kt-core/src/auditor.rs:220-247`) is what makes `AppendOnlyVerified` mean anything: it refuses to advance unless the token describes the run being accepted. Its guard has five clauses, and **each one can be deleted individually with the suite green**. Only deleting all five together turns anything red.

```rust
// The token must be about *this* range, starting where we actually are.
if first.sth.epoch != self.view.epoch()
    || first.sth.root_hash != *self.view.root_hash()
    || verified.from_epoch() != first.sth.epoch
    || verified.to_epoch() != last.sth.epoch
    || verified.to_root() != &last.sth.root_hash
{
    return Err(KtError::Malformed);
}
```

## Measured, one clause at a time

`cargo test -p f2z-kt-core --lib --no-fail-fast` on `main` (`9efb7e7`), baseline **83 passed, EXIT=0**:

| clause removed | result |
|---|---|
| `first.sth.epoch != self.view.epoch()` | 83 passed, **EXIT=0** |
| `first.sth.root_hash != *self.view.root_hash()` | 83 passed, **EXIT=0** |
| `verified.from_epoch() != first.sth.epoch` | 83 passed, **EXIT=0** |
| `verified.to_epoch() != last.sth.epoch` | 83 passed, **EXIT=0** |
| `verified.to_root() != &last.sth.root_hash` | 83 passed, **EXIT=0** (also green across `-p f2z-witness`) |
| **all five, replaced by `let _ = verified;`** | `auditor::tests::advancing_requires_a_token_for_exactly_this_range` **FAILED**, EXIT=101 |

## Why the broad mutation reddens and no narrow one does

`advancing_requires_a_token_for_exactly_this_range` hands in a token that violates **several clauses at once** — a token from a wholly different range, where the start, the end epoch and the end root all disagree. Any single surviving clause still rejects it, so the test passes no matter which one you delete. It proves the guard is not *empty*; it does not pin any part of it.

The test's name is the thing I would fix first. "Requires a token for exactly this range" reads as a claim about all four range endpoints. What it actually establishes is "requires a guard that rejects a completely unrelated token".

## Why the last clause is the one that matters

[...]

PR #717, opened 2026-08-25
fix(wallet): redact legacy preview identity

### LINKED ISSUE #714: Legacy wallet preview ships real db filenames and UFVK fingerprints across the bridge for a screen that reads neither
#712 merged as `a7c59f3` at the head I reviewed, so the suggestion from 5409277476 is not applied. Filing it so it does not live only in a merged PR's comments.

**This is a hardening suggestion, not a defect.** The current behaviour is deliberate, tested, and documented — the native test is literally named `multi_layout_preserves_exact_db_names_but_redacts_manifest_identity`. Nothing is broken.

## Two of the four are protected differently from the other two

`legacy_import_preview.rs` builds every `LegacyWalletPreview` like this (`:334-341`):

```rust
Ok(LegacyWalletPreview {
    wallet_id: REDACTED.to_owned(),
    wallet_name: REDACTED.to_owned(),
    db_filename: entry.db_filename.clone(),
    account_count,
    ufvk_fingerprints,
    …
})
```

| field | protection |
|---|---|
| `wallet_id`, `wallet_name` | **`[REDACTED]` at the native boundary** — the real values never cross |
| native error text | **cannot be interpolated** — `preview_desktop` is `Result<_, &'static str>` (`:71`) and every `Err` is a fixed literal |
| `db_filename` | real value crosses |
| `ufvk_fingerprints` | real value crosses — 64 hex characters per account |

The first two are enforced by construction, and the `Result<_, &'static str>` in particular is the best thing in this file: a future `format!("{err}")` there is a compile error, not something review has to catch. The last two are protected only by the frontend choosing not to read them.

## The frontend does not read them, and there is only one consumer

On `main` (`0948d8c`), `LegacyWalletNotice.tsx` reads exactly three fields plus the diagnostics list:

```
$ grep -n 'accountCount\|encryptedCustodyPresent\|walPresent\|shmPresent\|dbFilename\|ufvkFingerprints\|walletId\|walletName' \
      wallet/zuuli/src/components/layout/LegacyWalletNotice.tsx
21:    (total, candidate) => total + candidate.accountCount,
25:    (candidate) => candidate.encryptedCustodyPresent,
28:    (candidate) => candidate.walPresent || candidate.shmPresent,
```

No hit for `dbFilename`, `ufvkFingerprints`, `walletId` or `walletName`. And it is the only consumer:

```
$ grep -rn 'previewLegacyWalletImport' wallet/zuuli/src | grep -v '\.test\.'
LegacyWalletNotice.tsx:114 — the component
bridge.ts:75-76 — the bridge itself
mock.ts:200 — the mock
```

So `db_filename` and `ufvk_fingerprints` could be `[REDACTED]` at the native boundary exactly like `wallet_id`/`wallet_name`, with no loss of function to anything that exists today.

## Why it is worth doing

`LegacyWalletNotice.test.tsx:188-191` asserts, with canary values, that none of the four reaches `container.textContent`. That is assert-on-the-artifact done properly and I want to be clear it is good. But it can only see the DOM. The values are still in the JS heap and in React's element tree, where an error boundary serialising props, a crash reporter, a future `console.log`, or React DevTools would see the whole object — and no canary follows them there.

Redacting natively means nothing downstream *can* leak them, which is the property `wallet_id` and `wallet_name` already have. It converts a convention that every future consumer inherits into a guarantee that none of them can break.

## If the values are wanted later

They will be, for the real migration workflow in #272 — a user picking which wallet to import needs to see something identifying. The suggestion is not to delete them from the native type, only to stop shipping them across the bridge for a screen that does not use them. When a consumer exists, add them back to the bridge type deliberately, with whatever consent or gating that screen needs.

Concretely: two lines in `legacy_import_preview.rs`, plus the fixtures at `:771-772` and the multi-layout test that pin the current values.

skyl:
Closes #714

Redacts legacy wallet database filenames and UFVK fingerprints at the native preview boundary while preserving account/fingerprint cardinality for the existing UI. Strengthens the WAL-backed regression with two distinct accounts and independently derived fingerprint canaries.

Mutation evidence before opening:
- restoring the raw database filename makes the exact WAL test fail
- restoring the two raw UFVK fingerprints makes the exact WAL test fail
- collapsing the two redaction markers to one makes the exact WAL test fail

Verification:
- cargo +1.97.1 fmt --all -- --check
- cargo +1.97.1 test --locked --all-targets (150/150)
- cargo +1.97.1 clippy --locked --all-targets -- -D warnings
- independent adversarial review of exact fe5207b0cbd3b42f7039cf997b9cff3aaa1cbe69: APPROVE

### DIFF
--- a/wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs
+++ b/wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs
@@ -1085,19 +1085,45 @@ mod tests {
         connection
             .execute_batch("CREATE TABLE accounts (id INTEGER PRIMARY KEY, name TEXT, ufvk TEXT);")
             .expect("accounts schema");
+        let raw_ufvk_a = valid_ufvk(0);
+        let raw_ufvk_b = valid_ufvk(1);
         connection
             .execute(
                 "INSERT INTO accounts (id, name, ufvk) VALUES (0, 'private', ?1)",
-                [valid_ufvk(0)],
+                [&raw_ufvk_a],
             )
             .expect("uncheckpointed account");
+        connection
+            .execute(
+                "INSERT INTO accounts (id, name, ufvk) VALUES (1, 'private too', ?1)",
+                [&raw_ufvk_b],
+            )
+            .expect("second uncheckpointed account");
         assert!(db.with_extension("sqlite-wal").exists());
         assert!(db.with_extension("sqlite-shm").exists());
         let before = tree_bytes(&tree.root);
 
         let preview = preview(&tree.canonical(), CANONICAL_IDENTIFIER);
+        let serialized = serde_json::to_string(&preview).expect("serialize WAL preview");
+        let fingerprint = |ufvk: &str| {
+            let mut hasher = Sha256::new();
+            hasher.update(b"ZUULI legacy UFVK fingerprint v1\0");
+            hasher.update(ufvk.as_bytes());
+            to_hex(&hasher.finalize())
+        };
+        let fingerprint_a = fingerprint(&raw_ufvk_a);
+        let fingerprint_b = fingerprint(&raw_ufvk_b);
         assert_eq!(preview.state, LegacyImportPreviewState::Ready);
-        assert_eq!(preview.wallets[0].account_count, 1);
+        assert_eq!(preview.wallets[0].account_count, 2);
+        assert_eq!(preview.wallets[0].db_filename, REDACTED);
+        assert_eq!(preview.wallets[0].ufvk_fingerprints, [REDACTED, REDACTED]);
+        for secret in [
+            "wallet.sqlite",
+            fingerprint_a.as_str(),
+            fingerprint_b.as_str(),
+        ] {
+            assert!(!serialized.contains(secret), "WAL preview leaked {secret}");
+        }
         assert!(preview.wallets[0].wal_present);
         assert!(preview.wallets[0].shm_present);
         assert_eq!(tree_bytes(&tree.root), before);
--- a/wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs
+++ b/wallet/plugins/tauri-plugin-zcash/src/legacy_import_preview.rs
@@ -333,9 +333,9 @@ where
     Ok(LegacyWalletPreview {
         wallet_id: REDACTED.to_owned(),
         wallet_name: REDACTED.to_owned(),
-        db_filename: entry.db_filename.clone(),
+        db_filename: REDACTED.to_owned(),
         account_count,
-        ufvk_fingerprints,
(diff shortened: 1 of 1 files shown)


skyl:
## Verified — all three mutation claims reproduce independently. Approve.

Baseline on `fe5207b0`: `cargo test --lib legacy_import_preview` → **EXIT=0, 12 passed**.

Then each of your three claims, isolated and restored between runs:

| mutation | result |
|---|---|
| `db_filename: REDACTED` → `entry.db_filename.clone()` | **EXIT=101**, 3 failed — `single_layout_is_redacted_and_zero_mutation`, `multi_layout_redacts_all_bridge_identity`, `wal_snapshot_includes_uncheckpointed_accounts_without_source_mutation` |
| `vec![REDACTED; len]` → `ufvk_fingerprints` (raw) | **EXIT=101**, same 3 failed |
| `vec![REDACTED; len]` → `vec![REDACTED]` (cardinality collapsed) | **EXIT=101**, 1 failed — the WAL test |

The third is the one I care most about and the one I would not have thought to ask for. Preserving `ufvk_fingerprints.len()` while dropping the values keeps the UI's future ability to say "N keys" without shipping the keys, and strengthening the WAL fixture to **two distinct accounts** is what makes that count falsifiable. Without it, `vec![REDACTED]` and `vec![REDACTED; 2]` would be indistinguishable to every test, and the cardinality would quietly become a lie the first time someone simplified the expression.

Adding `"wallet.sqlite"` to the existing secret-scan loop is the right move too — the filename is now scanned for in the serialized preview alongside the raw UFVK, so it is checked against the artifact rather than against the assignment.

## The rename

`multi_layout_preserves_exact_db_names_but_redacts_manifest_identity` → `multi_layout_redacts_all_bridge_identity`.

Worth calling out. That old name was accurate when it was written, and it was what told me the previous behaviour was deliberate rather than an oversight — I said so on #714 and it is why that issue was filed as hardening rather than as a defect. Renaming it in the same commit that changes the behaviour is exactly right; a test whose name outlives the property it described is the failure mode I keep finding elsewhere in this tree (#709, #710).

## What this closes

#714's ask was that `db_filename` and `ufvk_fingerprints` be redacted at the native boundary, like `wallet_id`/`wallet_name` already were, on the grounds that `LegacyWalletNotice.tsx` reads neither and is the sole consumer. That is now true by construction: nothing downstream *can* leak them, which is the property the other two fields had and these two did not.

I will close #714 when this merges.

## Baseline note

A fresh worktree has no `z/zcash/librustzcash` submodule, so `cargo test` in this crate fails on path-dependency resolution until `git submodule update --init z/zcash/librustzcash` is run there. That was my environment, not this branch — recording it because a reviewer hitting it cold could easily read it as a broken tree.

Merge when `gate` and `rs / gate` are green on `fe5207b0`.

Issue #701, opened 2026-08-25
E2EE: three supply-chain decisions the MLS engine's graph forces (owner call)


skyl:
Adopting the MLS engine (#697, Phase 2 of #305) forces three decisions that are
recorded in `rs/deny.toml` but are **not the agent's to make**. #697 is green
with all three accepted; this issue exists so they are revisited deliberately
rather than inherited.

**#385 evaluated `openmls_libcrux_crypto 0.3.1` on nine targets against real
NIST/RFC known-answer vectors and did not run cargo-deny**, so none of this was
visible when the library choice was settled. Nothing below re-opens that choice —
it is the cost of it, priced.

## 1. RUSTSEC-2026-0212 — aarch64 constant-time swap/select. **The real one.**

`libcrux-secrets 0.0.5` compares an 8-bit selector with a 32-bit `cmp` on
aarch64, including 24 bits of unspecified register content, so "in certain
circumstances the swap and select instructions on aarch64 platforms returned
incorrect results". aarch64 is every phone we ship to and every Apple Silicon
desktop, and `libcrux-secrets` sits under `libcrux-aead` — the ChaCha20-Poly1305
we do use.

Against it: #385's full ML-KEM ACVP / RFC 7748 / RFC 8032 / X-Wing Appendix C
suite ran green on `aarch64-apple-darwin` and on a real Android API-29 arm64
device, and a swap/select returning the wrong branch would surface as a wrong
answer. **That is evidence, not proof** — the advisory says the fault depends on
the execution environment.

Fixed in `libcrux-secrets 0.0.6`. Unreachable from here: `libcrux-aead 0.0.6` and
`0.0.7` both pin `0.0.5`, and both arrive through `openmls_libcrux_crypto 0.3.1`.
libcrux publishes as `0.0.x`, so every release is semver-incompatible and
`cargo update` cannot move it — the bump has to come from
`openmls_libcrux_crypto` / `hpke-rs-libcrux`.

Options, in increasing cost: accept and watch for the upstream bump; fork
`openmls_libcrux_crypto` onto newer libcrux; wait.

Also accepted and lower-stakes: **RUSTSEC-2026-0124** —
`libcrux-chacha20poly1305::encrypt` panics on an *overlong output buffer*, which
OpenMLS sizes and no peer controls. Seven further advisories are accepted as
**not reached** with a per-advisory argument in `rs/deny.toml`
(AES-GCM is not in our ciphersuite; the SHAKE advisories exclude ML-KEM; the
[...]

PR #729, opened 2026-08-25
feat(rs): migrate to the OpenMLS 0.9 train, clearing nine advisories

### LINKED ISSUE #723: E2EE: migrate to the OpenMLS 0.9 train — it clears ten of the eleven accepted advisories
The whole OpenMLS 0.9 train shipped on **2026-08-25**, hours after #697 merged
against 0.8.1:

| crate | we ship | released |
|---|---|---|
| `openmls` | 0.8.1 | **0.9.0** |
| `openmls_traits` | 0.5.0 | **0.6.0** |
| `openmls_libcrux_crypto` | 0.3.1 | **0.4.0** |
| `openmls_memory_storage` | 0.5.0 | **0.6.0** |
| `openmls_sqlite_storage` | — | **0.3.0** |

Two things follow, and the first is the important one.

## 1. It clears ten of the eleven accepted advisories

`openmls_libcrux_crypto 0.4.0` moves to `libcrux-aead 0.0.9` /
`libcrux-ed25519 0.0.9`, which pulls `libcrux-secrets 0.0.6`,
`libcrux-sha3 0.0.10`, `libcrux-poly1305 0.0.6`,
`libcrux-chacha20poly1305 0.0.9` and `libcrux-aes 0.0.9`.

`cargo deny check advisories` against that graph with an **empty ignore list**
leaves exactly one finding — `RUSTSEC-2026-0173`, the build-host-only
`proc-macro-error2`. Detail and the table in #701.

That includes **RUSTSEC-2026-0212**, the aarch64 constant-time swap/select defect
accepted as a risk on #701. The removal condition recorded in `rs/deny.toml` —
*"when `openmls_libcrux_crypto` ships against a `libcrux-aead` whose
`libcrux-secrets` requirement admits `>= 0.0.6`"* — **is now met**.

**This is the reason to do the migration**, and it should be the acceptance
criterion: the ten `ignore` entries in `rs/deny.toml` come out, and only the
`proc-macro-error2` one stays.

## 2. It re-opens (but does not settle) why `f2z-msg-store` exists

`openmls_sqlite_storage 0.3.0` requires `rusqlite = "0.37"` — the repo-wide
singleton (`AGENTS.md`). The specific blocker #385 measured, and that
`f2z-msg-store`'s crate documentation cites, no longer applies to the 0.9 line:
0.2.0 needed `rusqlite ^0.32` and the `0.3.0-rc` line needed an unreleased
`openmls 0.9.0-rc`, and 0.9.0 is now released.

**That does not make `f2z-msg-store` redundant**, and the migration should not
assume it does. What it provides that `openmls_sqlite_storage` does not:

- **The transaction.** `openmls_sqlite_storage 0.3.0` exposes no transaction API
  — it cannot, because `StorageProvider` has none — so it commits each of the
  many calls in a `process_message` → `merge_staged_commit` run independently.
  That is the atomicity requirement delete-on-ack turns into a data-loss
  question (`ARCHITECTURE.md` §6.4), and it is the whole reason our provider is
  shaped the way it is.
- **The backend seam.** ADR 0001 needs one core shared with the browser;
  `f2z-msg-store`'s `StorageBackend` is what lets an IndexedDB backend be a new
  type rather than a second copy of 57 methods. A native-only SQLite provider
  cannot serve that.
- **The application namespace**, which is how the durable "handled" record lands
  in the *same* atomic write as the epoch change.

So the honest framing is: the migration should **re-evaluate the justification
and rewrite the crate docs**, which currently lean on a blocker that has expired,
rather than delete the crate. If the conclusion is still "keep it", say why in
the new terms.

## Scope

- Bump the five crates; `openmls_traits 0.6.0` will move `StorageProvider`, so
  `f2z-msg-store`'s 57 methods need re-diffing against the new trait (this is
  exactly what keeping them in upstream declaration order was for).
- Re-check `f2z-msg-mls` against 0.9's API.
- Re-run the wasm32 builds and the `--features js` / `getrandom_backend` plumbing.
- Remove the ten libcrux `ignore` entries and confirm `cargo deny` is green
  without them; keep `RUSTSEC-2026-0173`.
- Re-check the MPL-2.0 `exceptions` — the `hpke-rs` family may have moved too.
- Re-run the crash suite and the two-instance exchange.
- Ideally re-run #385's nine-triple build matrix, since its validation was
  against 0.8.1. At minimum, the KATs on aarch64.
- `rs/deny.toml`'s `openmls`/`openmls_traits` floors need raising with it.

## Not urgent, and worth saying

0.9.0 is a **major version released hours ago**. #697 shipped against the version
#385 validated on nine triples with real NIST/RFC vectors on Android hardware,
which was the right call at the time and still is. This is a deliberate,
separately-verified migration — not a version bump.

Refs #305, #312, #385, #697, #701.

skyl:
Closes #723.

`openmls 0.9.0` / `openmls_traits 0.6.0` / `openmls_libcrux_crypto 0.4.0`, taken
for a security reason and not for freshness: **it removes a real aarch64
cryptographic defect from the shipping graph.**

## The acceptance criterion, measured

`rs/deny.toml`'s `ignore` list goes from **ten entries to one**. Every entry
that came out is *fixed upstream*, not re-argued.

Measured rather than inferred — `cargo deny check advisories` over each graph
with an **empty** `ignore` list, via `scripts/check-rust-deny.sh --root rs`:

| | `origin/main` (`openmls 0.8.1`) | this branch (`openmls 0.9.0`) |
|---|---|---|
| advisories firing with an empty ignore list | **10** | **1** |

The nine that cleared, and what each is:

| advisory | crate | defect |
|---|---|---|
| **RUSTSEC-2026-0212** | `libcrux-secrets` | **aarch64 constant-time swap/select may return incorrect results** |
| RUSTSEC-2026-0124 | `libcrux-chacha20poly1305` | encrypt panics on an overlong output buffer |
| RUSTSEC-2026-0073 | `libcrux-poly1305` | panic in the standalone MAC API |
| RUSTSEC-2026-0075 | `libcrux-ed25519` | all-zero key on catastrophic RNG failure |
| RUSTSEC-2026-0207 | `libcrux-sha3` | incorrect incremental SHAKE squeeze |
| RUSTSEC-2026-0208 | `libcrux-sha3` | panic in AVX2 SHAKE-256 |
| RUSTSEC-2026-0209 | `libcrux-aesgcm` | AES-GCM AAD length limit not enforced |
| RUSTSEC-2026-0210 | `libcrux-aesgcm` | unmaintained (renamed to `libcrux-aes`) |
| RUSTSEC-2026-0211 | `libcrux-aesgcm` | non-constant-time AES-GCM tag check |

`RUSTSEC-2026-0173` (`proc-macro-error2`, unmaintained, build-host only) stays,
and it is the only entry left.

**A correction to #723's arithmetic, since the acceptance criterion was written
in terms of it.** The issue and the old `deny.toml` header both say "eleven
advisories, ten of them libcrux". The file had **ten** `ignore` entries and
**nine** of them were libcrux. Ten came out of the *count* only if you include
the one that stays. The substantive claim is unaffected and is fully met:
**every libcrux advisory is gone, RUSTSEC-2026-0212 included, and the only thing
left is the build-host proc macro.**

Resolved versions, from `rs/Cargo.lock`: `libcrux-secrets 0.0.6`,
`libcrux-sha3 0.0.10`, `libcrux-ed25519 0.0.9`, `libcrux-poly1305 0.0.6`,
`libcrux-chacha20poly1305 0.0.9`, `libcrux-aes 0.0.9`, `libcrux-aead 0.0.9`.

The per-advisory reachability reasoning has been **deleted, not archived**: a
justification for a risk that no longer exists teaches the next reader that the
risk is still being accepted. #701 and #723 hold the record. The
`[patch.crates-io]` warning is **kept** — it is a fact about cargo (a patch that
cannot satisfy an exact pin is skipped with a warning and the build goes green
on the vulnerable version), and that lesson outlives this migration.

Floors raised in `rs/deny.toml`: `openmls < 0.9.0`, `openmls_traits < 0.6.0`,
and a **new** `openmls_libcrux_crypto < 0.4.0` — redundant against today's
manifests, and written down precisely because it is the crate that resolves
`libcrux-secrets`.

## The API breaks, and the three that needed judgement

**`openmls_traits 0.6.0` did not move `StorageProvider` out from under us.**
Every method 0.6.0 adds is behind `virtual-clients-draft`, which nothing here
enables; the base trait is unchanged. `f2z-msg-store`'s 57 methods compiled
against 0.6.0 **without a single edit**. The one break was a feature rename,
`extensions-draft-08` → `extensions-draft`, renamed rather than aliased.

**The X-Wing ciphersuite moved behind a feature.** 0.9.0 gates every
post-quantum codepoint on `draft-ietf-mls-pq-ciphersuites`; without it
`Ciphersuite::MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519` does not exist.
`rs/Cargo.toml` enables it by name. The **codepoint is unchanged** (`0x004D`) —
what moved is whether the name compiles, which is exactly the relabelling risk
`version.rs` was written to survive.

**`openmls` moved to `tls_codec 0.5`** while `f2z-codec` and `f2z-kt-core` stay
on 0.4. The two `tls_codec` calls in `f2z-msg-mls` are on OpenMLS's own types,
so `tls_codec` was **dropped as a dependency of that crate** and the traits now
come through `openmls::prelude::tls_codec` — which cannot drift from what
OpenMLS derived them with. Nothing crosses the seam: a `DeviceCredential`
reaches MLS as opaque `BasicCredential` bytes via `f2z-codec`'s canonical
encoding. Same shape as the two `ed25519-dalek` majors this workspace already
carries. **The wire codec crates are deliberately untouched** — a `tls_codec`
major bump under `f2z-codec` is a wire-format change and does not belong in a
security migration.

**Two new `ProcessedMessageContent` variants, and one is a behaviour change
callers can see.** 0.9 added `OwnPrivateMessage` and `OwnPendingCommit`. Under
0.8.1, a device's own `PrivateMessage` handed back by a relay failed to decrypt
and `receive` returned `EngineError::Mls`; 0.9 surfaces it as a distinct
outcome. The engine now returns a new `Received::Own`, **writes the durable
"handled" record**, and returns no payload. That matters under delete-on-ack
(§9.4 makes relay echoes routine): an echo that always errored could never be
acknowledged, so the relay would redeliver it forever.
`a_devices_own_message_handed_back_is_an_outcome_and_not_an_error` pins all of
it. `OwnPendingCommit` is unreachable in this engine — `update`/`add_member`
merge before returning — and is handled by merging if a pending commit exists,
because assuming the invariant would silently *skip an epoch change* if anyone
ever split those two steps.

### 1. `f2z-msg-store`: keep it, and the reason has changed

**The blocker #385 measured has genuinely expired**, and the crate docs said so
in their first clause. `openmls_sqlite_storage 0.3.0` requires
`rusqlite = "0.37"` with `features = ["bundled"]` — our repo-wide singleton
exactly — and the `openmls 0.9.0` it needs is released.

The decision is still **keep**, on three reasons checked against 0.3.0's source
rather than its README:

1. **The transaction.** 0.3.0's entire public surface is `new`, `initialize`,
   `run_migrations` and the `StorageProvider` impl. No transaction API — there
   cannot be one, `StorageProvider` has none — and one cannot be supplied from
   outside either: it is generic over `ConnectionRef: Borrow<Connection>`, and
   `rusqlite::Transaction` reaches `Connection` through `Deref`, not `Borrow`.
2. **The backend seam.** `rusqlite`'s `bundled` feature links SQLite's C
   amalgamation, so it is native by construction and cannot serve ADR 0001's
   browser half.
3. **The application namespace.** `StorageProvider` has no notion of
   application data, so no implementation of it alone can put the durable
   "handled" record in the *same* atomic write as the epoch change.

The crate root, `sqlite.rs`'s note and `AGENTS.md`'s worked example were all
rewritten to say the blocker expired rather than to keep repeating it.

### 2. The `openmls_memory_storage` defect is still there

`clear_proposal_queue` still removes `serde_json::to_vec(&(group_id,
proposal_ref))` while `queue_proposal` writes through
`write::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, …)` — a key nothing was ever
stored under, leaking every serialised proposal body. **Verified present
unchanged in `openmls_memory_storage 0.6.0`**, filed as
[openmls/openmls#2188](https://github.com/openmls/openmls/issues/2188). Our fix
and `clearing_the_queue_removes_the_proposal_bodies_and_not_only_the_references`
both survive; `storage_impl.rs`'s note now names the version it was re-checked
against and says **do not resync with upstream**.

### 3. The `debug_assert!(false)` workaround is gone — verified, not assumed

`openmls 0.9.0`'s `private_message_in.rs` has no `debug_assert` on either
AEAD-open path; both are `log::error!` + a returned
`MessageDecryptionError::AeadError`. The whole of `src/framing/` contains one
`debug_assert`, on an unrelated key-package version check.

So `a_corrupted_ciphertext_is_refused_and_leaves_the_group_usable` lost its
`#[cfg(not(debug_assertions))]` gate and **now runs in the ordinary debug
`cargo test`** — confirmed by running it, not by reading the changelog. The
extra `--release` step in `rs.yml` is removed, replaced by a comment saying why
it is gone rather than leaving a silent hole.

## Re-validation

#385's nine-triple validation was against 0.8.1, so it was re-run.

**Green, on this branch, on aarch64-apple-darwin:**

- **The two-instance exchange** — `f2z-msg-mls/tests/two_instances.rs`, 18/18.
  Separate storage, separate identities, real `Welcome`, decrypted payload in
  both directions, `Update` with both sides agreeing on the epoch, exporter
  material changing with the epoch.
- **The crash-safety suite** with its real child-process abort —
  `cargo test -p f2z-msg-store --features crash-injection --test crash_safety`,
  4/4, including "a crash before the commit leaves no trace" and "a crash after
  it leaves the whole operation".
- **The full `rs/` suite** — `cargo test --locked --all-targets` and
  `--locked --doc`, everything green.
- **rustfmt**, **clippy at `-D warnings`**, and **`cargo-deny`** (advisories,
  bans, licences, sources) over all 13 crates, plus the four workflow-policy
  scripts and the toolchain-pin check.
- **Cross-compilation for all nine shipping triples**
  (`cargo check --locked --lib -p f2z-msg-mls --no-default-features`):
  `x86_64-unknown-linux-gnu`, `aarch64-apple-darwin`, `x86_64-apple-darwin`,
  `x86_64-pc-windows-msvc`, `aarch64-apple-ios`, `aarch64-linux-android`,
  `armv7-linux-androideabi`, `i686-linux-android`, `x86_64-linux-android`.
- **`wasm32-unknown-unknown`** for `f2z-msg-store --no-default-features` and for
  `f2z-msg-mls --no-default-features --features js`.

### The wasm plumbing got simpler, and that is a graph change

`--features js` is still required. **`--cfg getrandom_backend="wasm_js"` is
not, and `getrandom 0.3` is no longer declared.** Against 0.8.1 this crate's
wasm graph carried getrandom 0.3 (via `rand_core 0.9`) *and* 0.4 (via
`rand 0.10` inside `hpke-rs-libcrux`), and **0.3 needs the cfg as well as the
feature**. On the 0.9 train nothing in that graph reaches `rand_core 0.9` —
getrandom 0.3 survives in the lockfile only under `proptest` and `f2z-relay`'s
dev-dependencies — and getrandom **0.4** selects its wasm backend from the
`wasm_js` feature alone (`src/backends.rs`, the `target_family = "wasm"` arm).
Both halves came off together. If getrandom 0.3 ever re-enters, the wasm job
fails with getrandom's own `compile_error!`; it cannot regress quietly.

### What could NOT be re-validated — stated plainly

- **The KATs were not re-run.** #385's NIST ACVP ML-KEM, RFC 7748, RFC 8032 and
  draft-connolly-cfrg-xwing-06 Appendix C harness **does not exist in this
  repository** — `wallet/zuuli/crypto-target-spike` is a compile-target probe
  over `libcrux-ml-kem`, not a known-answer suite — so there was nothing to run.
  This is a gap worth its own issue rather than a line in a PR body.
- **What was run on real aarch64 is the functional suite, not KATs.** This work
  was done on Apple Silicon (`arm64`), so the two-instance exchange, the crash
  suite and the whole `rs/` suite executed the libcrux code paths natively on
  aarch64 — a real X-Wing key schedule, a real ChaCha20-Poly1305 AEAD, real
  Ed25519 framing signatures. That is meaningful evidence and it is **not** a
  known-answer test.
- **No physical Android or iOS device.** The Android and iOS triples were
  **compile-checked only**. Nothing here was run on a phone, and no simulator
  result is being reported as a device result.
- **The eight non-host triples are `cargo check`, not link-and-run.**
  `--no-default-features` (no `rusqlite`) so no cross C toolchain is needed.
  CI's own matrix is the authority on the link step.

## Also updated

- `docs/e2ee/ARCHITECTURE.md` §3.1 and §13-A: the floor is now
  `openmls >= 0.9.0`, carrying two defects rather than one.
- `AGENTS.md`'s `rusqlite` singleton worked example, which cited the expired
  `openmls_sqlite_storage 0.2.0` situation as if it were current.

### DIFF
--- a/.github/workflows/rs.yml
+++ b/.github/workflows/rs.yml
@@ -331,16 +331,21 @@ jobs:
         working-directory: rs
         run: cargo test --locked -p f2z-msg-store --features crash-injection --test crash_safety
 
-      # `openmls 0.8.1` runs `debug_assert!(false)` on the AEAD-open failure
-      # path (`private_message_in.rs:136`), which any peer or relay can reach by
-      # flipping one byte. It compiles out in release, so the engine's
-      # corrupted-ciphertext case — "refuse, do not crash" — is a property only a
-      # release build can assert, and `tests/two_instances.rs` gates that case on
-      # `not(debug_assertions)`. This step is what runs it. See #385 and the
-      # note on that test.
-      - name: Run the MLS engine's release-only refusal cases
-        working-directory: rs
-        run: cargo test --release --locked -p f2z-msg-mls --test two_instances
+      # There is deliberately NO extra `--release` step for f2z-msg-mls here.
+      #
+      # There used to be one, and this note is why it is gone rather than
+      # missing. `openmls 0.8.1` ran `debug_assert!(false)` on the AEAD-open
+      # failure path (`private_message_in.rs:136`) — reachable by any peer or
+      # relay that flips one byte — which aborted the process in a debug build
+      # and compiled out in release. The engine's corrupted-ciphertext case
+      # ("refuse, do not crash") was therefore a property only a release build
+      # could assert, `tests/two_instances.rs` gated it on
+      # `not(debug_assertions)`, and this step existed to run it.
+      #
+      # `openmls 0.9.0` replaced that path with `log::error!` plus a returned
+      # `MessageDecryptionError::AeadError`, so the case now runs in the
+      # ordinary `cargo test --all-targets` above and this step would be a
+      # second full compile of the same crate for nothing. See #723 and #701.
 
   rs_wasm:
     name: rs / wasm32
@@ -434,25 +439,26 @@ jobs:
       # and the web client, so this reaching wasm32 is a protocol-level
       # requirement rather than a nice property.
       #
-      # Two things are needed and neither is optional:
+      # `--features js` is not optional: OpenMLS emits a `compile_error!` for
+      # wasm32 without it, because it needs JavaScript APIs for secure
+      # randomness and for the current time. It refuses to build rather than
+      # producing a binary that traps at runtime, which is the right choice and
+      # is why the feature exists.
       #
-      #   * `--features js` — OpenMLS emits a `compile_error!` for wasm32
-      #     without it, because it needs JavaScript APIs for secure randomness
-      #     and for the current time. It refuses to build rather than producing
-      #     a binary that traps at runtime, which is the right choice and is why
-      #     the feature exists.
-      #   * `--cfg getrandom_backend="wasm_js"` — `getrandom` refuses to compile
-      #     for wasm32-unknown-unknown until a backend is named, and this graph
-      #     carries **two** major versions of it (0.3 via `rand_core 0.9`, 0.4
-      #     via `rand 0.10` inside `hpke-rs-libcrux`). `f2z-msg-mls`'s manifest
-      #     names both under a `cfg(target_arch = "wasm32")` target table purely
-      #     to turn their `wasm_js` features on; this flag is the other half.
-      #     RUSTFLAGS rather than `.cargo/config.toml` so it applies to this job
-      #     and cannot silently change a native build.
+      # There is deliberately **no** `RUSTFLAGS: --cfg getrandom_backend="wasm_js"`
+      # here any more. Against `openmls 0.8.1` this graph carried two major
+      # versions of `getrandom` — 0.3 via `rand_core 0.9`, 0.4 via `rand 0.10`
+      # inside `hpke-rs-libcrux` — and **0.3 needs the cfg as well as the
+      # feature**. On the 0.9 train nothing in this crate's wasm graph reaches
+      # `rand_core 0.9`, and getrandom **0.4** selects its wasm backend from the
+      # `wasm_js` feature alone, which `f2z-msg-mls`'s target table declares.
+      # See that manifest's note; #723 removed both halves together.
+      #
+      # This cannot regress silently: if getrandom 0.3 re-enters the graph, this
+      # step fails with getrandom's own `compile_error!` naming the missing
+      # backend.
       - name: Build the MLS engine for the browser target
         working-directory: rs
-        env:
-          RUSTFLAGS: --cfg getrandom_backend="wasm_js"
         run: |
           set -euo pipefail
           cargo build --locked --lib \
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -144,11 +144,15 @@ integrate, and move it forward alongside everything else.
   `wallet/zuuli/src-tauri` links that plugin together with the rest of the app,
   so **rusqlite 0.37 is a repo-wide singleton**: any new crate that needs SQLite
   must resolve to it. All three lockfiles currently agree on `rusqlite 0.37.0` /
-  `libsqlite3-sys 0.35.0`. Worked example: `openmls_sqlite_storage` 0.2.0 — the
-  latest *release* — requires `rusqlite ^0.32` and is therefore unusable here;
-  its `0.3.0-rc` line moved to `^0.37` and would fit. Check the requirement
-  before adopting, because the conflict appears only once the app links both: a
-  crate that builds fine on its own can still be off-limits.
+  `libsqlite3-sys 0.35.0`. Worked example, **and note how it ended**:
+  `openmls_sqlite_storage` 0.2.0 required `rusqlite ^0.32` and was therefore
+  unusable here, which is part of why `rs/crates/f2z-msg-store` exists; its
+  0.3.0 release (2026-08-25) moved to `rusqlite = "0.37"` and now fits. The
+  singleton did not change — upstream moved. Check the requirement before
+  adopting *and* re-check it when the upstream releases, because the conflict
+  appears only once the app links both: a crate that builds fine on its own can
+  still be off-limits, and one that was off-limits can stop being so without
+  anyone noticing.
 
 ## Guardrails that keep us honest
 
--- a/docs/e2ee/ARCHITECTURE.md
+++ b/docs/e2ee/ARCHITECTURE.md
@@ -143,12 +143,27 @@ Eight findings, one of them **High severity — "improper authentication of
 MACs."** The High finding and all but one of the rest are fixed in **0.8.1** (and
 0.7.3 on the previous line).
 
-**Floor: `openmls >= 0.8.1`.** Any earlier version contains a High-severity
-authentication defect and MUST NOT be shipped. One Low-severity finding was still
-open at publication; **review its status before release** rather than assuming the
-audit closed everything. Note also that the version floor interacts with the
-crypto provider: `openmls_rust_crypto` does not implement X-Wing, so the PQ
-ciphersuite (§5.2) requires the libcrux provider.
+**Floor: `openmls >= 0.9.0`**, raised from `>= 0.8.1` by
+[#723](https://github.com/free2z/zuu/issues/723) and carrying two defects rather
+than one. Any version before 0.8.1 contains the High-severity authentication
+defect above. Any version before **0.9.0** resolves, through
+`openmls_libcrux_crypto <= 0.3.1`, to a `libcrux-secrets 0.0.5` carrying
+**RUSTSEC-2026-0212** — a constant-time swap/select that can return incorrect
+results on **aarch64**, underneath the AEAD of §5.2's ciphersuite. aarch64 is
+every phone we ship to and every Apple Silicon desktop.
+[#701](https://github.com/free2z/zuu/issues/701) established that the fix could
+not be applied without forking the libcrux release train; `openmls 0.9.0` and
+`openmls_libcrux_crypto 0.4.0` apply it upstream, and taking them cleared nine
+accepted advisories from `rs/deny.toml` at once. Neither floor may be shipped
+under.
+
+One Low-severity finding from the audit was still open at publication; **review
+its status before release** rather than assuming the audit closed everything.
+Note also that the version floor interacts with the crypto provider:
+`openmls_rust_crypto` does not implement X-Wing, so the PQ ciphersuite (§5.2)
+requires the libcrux provider — and on `openmls 0.9.0` the X-Wing codepoint
+additionally sits behind the `draft-ietf-mls-pq-ciphersuites` feature, which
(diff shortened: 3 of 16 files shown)


skyl:
## Note on the one red run, and a defect it exposed

`rs / tests` failed once and was re-run. The failure was **not this change**:

```
EvidenceTransportError: pinned evidence unavailable
https://www.nccgroup.com/media/phzpm0qv/_ncc_group_metaplatforms_e008327_report_2023-11-14_v10.pdf: HTTP 403
    at fetchPinnedValue (scripts/check-akd-doc-evidence.mjs:376)
    at async validateAuditEvidence (…:581)
    at async selfTest (…:1091)
```

`nccgroup.com` returned 403 to the GitHub runner. The same command passes
locally against the same commit.

Worth flagging rather than shrugging at, because `rs.yml`'s own comment on that
step says its transport controls "keep unavailability distinct from evidence
drift" — and here an outage at a third party failed a **required** gate exactly
as loudly as real drift would have. That is a fail-closed choice one can defend,
but the two outcomes are currently indistinguishable to anyone reading the check
name, and the third party is one we do not control. Filing it separately rather
than fixing it here; it has nothing to do with OpenMLS.

Everything green on the re-run: `gate` ✅ and `rs / gate` ✅.

Issue #734, opened 2026-08-25
ARCHITECTURE §7: msg_id does not commit to sender_leaf_index, so third-party repair would give two receivers different transcripts


skyl:
Found by `f2z-msg-dag` (#732) while implementing §7. Not a defect today — a latent one that a plausible future optimisation turns real.

## The gap

§7's display order breaks ties by `(epoch, sender_leaf_index, msg_id)`. But **`sender_leaf_index` lives in MLS framing, not in the hashed `AppMessage`** — so `msg_id` does not commit to it.

Meanwhile §7's repair path delivers a message **outside its original framing**: the sender re-encrypts the plaintext under the *current* epoch and sends `gap_response`. The recovered message therefore arrives in framing that is not the framing it was authored in.

This is sound today for exactly one reason: §7 says *"**the sender** re-encrypts the original plaintext"*. Because only the original sender repairs, its leaf index is whatever it always was, and every receiver derives the same sort key.

## Why it will not stay sound

**§7 does not forbid third-party repair**, and third-party repair is the obvious optimisation the moment a group has more than two members: any member holding the plaintext can answer a `gap_request`, and requiring the original sender to be online defeats much of the point of having a DAG.

If that ever lands, two receivers who learned the same message by different routes — one directly, one repaired by a third party — compute **different `sender_leaf_index` values for the same `msg_id`**, and therefore different sort keys. They render **different transcripts** while agreeing on every message. That breaks §7's own guarantee that every client produces the same transcript from the same set.

It is the kind of defect that appears long after the change that caused it, in a feature nobody connects to ordering.

## Fix

**Move `sender_leaf_index` inside the hashed `AppMessage`**, so `msg_id` commits to it and it travels with the message regardless of which framing delivered it. That makes the sort key a property of the message rather than of the delivery, which is what §7's guarantee actually requires.

This is a **wire change**, so it is far cheaper now than after a release. `f2z-msg-dag` is merged but nothing has shipped and no message has ever been sent.

Points to settle when specifying:

- Does the field become authoritative, with the framing value cross-checked on direct delivery and a mismatch treated as an error? (Probably yes — otherwise there are two sources of truth.)
- What does a repairer put there — the original author's index, necessarily. Which means the repairer must know it, which is an argument for it being in the message.
[...]

Issue #733, opened 2026-08-25
ZUULI transcript ordering ignores the DAG — a reply can render above the message it answers


skyl:
Found by `f2z-msg-dag` (#732, merged) building the same ordering independently in Rust and comparing. This is the two-implementations check doing its job — the disagreement is now pinned in `rs/crates/f2z-msg-dag/tests/typescript_parity.rs` as executable fact.

## The bug

`compareMessages` in `wallet/zuuli/src/lib/messaging/types.ts` is:

```ts
export function compareMessages(a: Message, b: Message): number {
  if (a.epoch !== b.epoch) return a.epoch - b.epoch;
  if (a.senderLeafIndex !== b.senderLeafIndex) return a.senderLeafIndex - b.senderLeafIndex;
  return a.msgId < b.msgId ? -1 : a.msgId > b.msgId ? 1 : 0;
}
```

**It never reads `parents`.** `Transcript.tsx` then does `[...messages].sort(compareMessages)`.

`ARCHITECTURE.md` §7 and `CLIENT-CONTRACT.md` §7 both make the **DAG's partial order** primary — `(epoch, sender_leaf_index, msg_id)` is what breaks ties *within* that order, not the order itself. The implemented comparator is the tie-break alone.

## The counterexample, run against the shipped code verbatim

Bob (leaf 1) sends `A` at epoch 7. Alice (leaf 0) replies `B` with `parents = [A]`, same epoch.

Sort keys are `(7, 1, …)` and `(7, 0, …)`, so **`B` renders above `A`** — a reply displayed above the message it answers.

Epochs are non-decreasing along causal edges, so this is confined to a single epoch. That is not much comfort: **it fires whenever the replier holds the lower leaf index, which is about half of all 1:1 conversations.** In a two-person chat that is a coin flip on whether the transcript reads correctly.

## The fix is not a comparator change

**A JS `sort` comparator cannot express a topological order.** `Array.prototype.sort` requires a total order over pairs; causal precedence is a relation over the whole graph, and a pairwise function has no access to it. Tightening `compareMessages` cannot fix this.

What is needed is a **graph pass before the sort**: compute the causal order from `parents`, then use the existing `(epoch, senderLeafIndex, msgId)` key only to break ties among mutually-incomparable messages. `rs/crates/f2z-msg-dag` on `main` implements exactly this and its tests define the expected behaviour, including determinism across shuffled delivery orders — port the semantics rather than re-deriving them.

[...]

PR #735, opened 2026-08-25
feat(f2zmsg): tauri-plugin-f2zmsg — two ZUULI engines chat over a real relay

skyl:
`tauri-plugin-f2zmsg` — the bridge between the merged Rust messaging core and
the TypeScript in `wallet/zuuli/src/lib/messaging/`, which has been running in
mock mode because `bridge.ts` invokes `plugin:f2zmsg|<cmd>` and no such plugin
existed.

**Two independent processes exchange messages over a real WebSocket relay.**
`cargo test --features relay-harness`, ~2 s, eight consecutive green runs:

```
conversationId 587ce5be…80cf  (both sides, same MLS group)

round 1   alice sent b6b43f07…6d6c → bob received b6b43f07…6d6c   leaf 0
          bob   sent 58051ff5…13de → alice received 58051ff5…13de leaf 1
          parents [] on both — genuinely concurrent

round 2   sent after a full round trip
          alice sent f65ee1f1…1259 → bob received f65ee1f1…1259
          bob   sent 428973a3…b2c4 → alice received 428973a3…b2c4
          bob's round-2 parents = [58051ff5…13de, b6b43f07…6d6c]
                                   ↑ both round-one messages: the head set
                                     being a *set*, which was the other
                                     defect the harness found

gaps 0 on both sides.  held: the same four ids on both sides.
events: engine-state, relay-state, contact-request, conversation-updated,
        message-received ×2, message-state ×2
```

`msg_id` is BLAKE2b-256 over the canonical §7 envelope and is **recomputed by
the receiver**, so `alice.sent == bob.received` is a statement about bytes. The
leaf indices differ (0 and 1), so §7's second ordering key came from real MLS
framing.

---

## What it actually is

`f2zmsg-peer` spawned twice against `FakeRelay::listen_loopback()` — a real RFC
6455 listener reached over TCP from outside the test process. Each peer has its
own SQLite file and its own CSPRNG device keys. They run `WIRE.md` §12.5's
handshake in full: `CREATE_CONTACT_QUEUE`, `CONTACT_APPEND` behind a
proof-of-work stamp, `CREATE_QUEUE`, `SUBSCRIBE`, `BIND_SEND`, `APPEND`, `READ`,
`ACK`. The ciphertext is real MLS `PrivateMessage`s under
`MLS_256_XWING_CHACHA20POLY1305_SHA256_Ed25519`.

**Exactly one thing is substituted: the directory resolution.** §6.4 refuses to
resolve a new handle without a witness-cosigned root and §9 rule 5 forbids
proceeding silently, so the shipping build refuses. The harness supplies the
answer from a shared file, from *outside* the crate.

## §7 comes from `f2z-msg-dag`, and adopting it fixed a defect

The crate landed (#732) while this was in flight. `src/framing.rs` existed
because it did not; it is deleted and `src/envelope.rs` is the adapter.

That was not deduplication. `framing.rs` linearised by the sort key **alone**,
and the crate's `order` module documents the cost: applying `(epoch,
senderLeafIndex, msgId)` to every pair rather than only to the pairs the causal
DAG leaves incomparable **puts replies above the messages they answer**, in every
one-to-one conversation where the replier holds the lower leaf index. Two
independent readings of the same paragraph both mistook the tie-break for the
whole rule. `list_messages` now pages over `MessageDag::display_order()`.

Three decisions that required, each stated where it is made:

- **Three codepoints allocated** (8, 9, 10) for §3.8's ephemeral hint and §3.9's
  purge request and ack, which §7's `type` list does not name. Pending
  registration upstream.
- **Control payloads carry `parents` but are not DAG vertices.** A vertex must
  survive a restart or it manufactures gaps: the DAG is rebuilt from the
  persisted transcript, so a vertex with no record disappears and the next
  message referencing it dangles. The harness found exactly that with the
  in-band `queue_advert` — the joiner reported a gap for its own message.
- **`rs/crates/*` now selects the ZUULI gate.** This plugin path-depends on six
  crates there and the change detector did not name them.

### Five defects the two-process harness found

Every one was fixed rather than worked around, and each is a class of bug that
a single-process test cannot see.

1. **A padded relay payload had no framing.** §9 fixes the outer length and says
   nothing about the inner one, because the relay never looks inside. Trimming
   trailing zeros is wrong — AEAD output genuinely ends in `0x00` — so the
   payload carries a big-endian `u32`.
2. **`ERR_STALE_TIMESTAMP` was terminal.** §8's `device-clock-skew` row says to
   re-learn the relay's time from `GET_CHALLENGE`, apply the offset locally and
   retry once. It does now, and never touches the system clock.
3. **`BIND_SEND` only happened for chat messages.** The first thing a device
   writes is a `queue_advert`, and an `APPEND` to an unbound address collapses to
   the same `ERR_UNAVAILABLE` as an absent queue.
4. **The head set was collapsed to "the last message seen".** Two peers who send
   before hearing each other produce concurrent messages; overwriting the set
   dropped one from the next `parents`, and the peer would then have had no
   dangling hash to notice its loss by. The evidence above — bob's round-2
   message naming *both* round-one ids — is this fixed.
5. **`transportHealth` reported `ok` on the read side alone**, telling the UI a
   message could be sent when there was no address to send it to.

## The three rules

**Never ACK before the durable write.** `Engine::pump_inbound` runs decrypt →
commit → ACK → emit, and `engine.rs`'s header names the one window that ordering
cannot close: a crash between MLS's commit and the plugin's leaves a message MLS
believes it handled with the ratchet moved past. It is detected and rendered as
§3.4's `{ kind: "unrecoverable" }` marker, never as a silent hole (§9 rule 7).

**Not `tauri-plugin-zcash`'s blur handler.** `Exit`/`ExitRequested` only.

**The event path is proven, not assumed.** This is the app's first real
`listen()` consumer; the harness records every emission.

## Registration

All four machine-enforced points, plus the two that were not.
`scripts/check-zcash-permissions.mjs` became
**`scripts/check-tauri-plugin-permissions.mjs`** and judges every plugin under
`wallet/plugins/`, from a registry that must cover every tracked plugin manifest.

§12.3 predicted a second checker "has to be written, not parameterized". That was
true of the script as it stood and is also how two near-identical checkers drift.
Everything §12.3 correctly identified as different became a field. The document
records the prediction and why it was wrong rather than being quietly edited.

**Two new links**, the two §2.3 said were unenforced: `permissions/default.toml`
must allow exactly the registered set, and every registered capability file must
grant either `<plugin>:default` or an exact list of real commands.

`rust_msg_plugin` is a **required job** — in `gate`'s `needs:` and inspected,
since that gate is the exhaustive shape. It runs `--features relay-harness`
(without it the two-process test compiles to nothing and passes by not
existing), builds the shipping configuration separately, and rejects permission
drift under a fresh generation nonce.

`messaging-contract.node-test.mjs` compares §3's tables to the plugin registry as
well as to `bridge.ts`, both directions, and asserts the enrollment trio is
absent from the plugin.

---

### ZUULI does not link the plugin. It cannot.

```
librustzcash  bip32 = "=0.6.0-pre.1"  →  sha2 = "=0.11.0-pre.4"
messaging     ed25519-dalek 3         →  sha2 = "^0.11"        → 0.11.0
```

Cargo unifies semver-compatible versions and `^0.11` does not match a
pre-release. Reproduce by adding `tauri-plugin-f2zmsg` to
`wallet/zuuli/src-tauri/Cargo.toml` and running `cargo check`. Upstream
librustzcash `main` still carries the same requirement, so a submodule bump does
not fix it, and `transparent-inputs` — which pulls `bip32` — is load-bearing.

Consequences, none worked around:

- **`capabilities/default.json` and `capabilities/mobile.json` carry no
  `f2zmsg:` entries.** A capability naming an unlinked plugin fails
  `tauri_build::build()`. The checker records the empty grant with this reason
  and refuses an empty grant *without* one. This is the one deliverable from the
  brief I could not meet.
- **`wallet/zuuli/src-tauri/src/messaging.rs` — the enrollment trio — is written
  and not landed.** It reads the seed in process from `tauri-plugin-zcash`'s
  managed state, derives §4.2's keys and hands the plugin a public summary; the
  `prepare_device` / `install_identity` API it calls **is** in this PR and is
  exercised by the harness and by `tests/engine_lifecycle.rs`.

Resolving this is its own PR and wants an owner's eyes: either a
librustzcash/`bip32` change, or moving the messaging core off `ed25519-dalek 3`,
which `rs/Cargo.toml` chose for stated security reasons.

### No key-transparency client

`f2z-kt` is a running server and `f2z-kt-core` is the verifier, but nothing
carries a `/kt/v1/lookup` between them. `resolve_handle`, `start_conversation`,
`accept_contact_request` and self-audit fail closed with
`witness-threshold-unmet` — the truthful code, since zero independent witnesses
have cosigned anything. Manual safety-number verification, which §8 tells the UI
to offer instead, works.

---

## Contract corrections and disagreements

**The brief said `Committed<T>` is in `f2z-msg-store`. It is not** — it lives in
`f2z-relay-store` and gates the *relay's* ACK response; its constructor is
`pub(crate)` there. The client-side equivalent is `Transaction::commit()`
returning `Ok(())` plus `Durability::may_acknowledge()`, checked at the single
ACK site and nowhere else.

**~~`compareMessages` implements the tie-break, not the order.~~ Fixed upstream
in #736 while this was in flight**, and the fix went further than the report:
§5.2's *"recomputed by the UI, always"* is withdrawn, `compareMessages` is
deleted, and §9 **rule 10** is now "never order a transcript in the frontend".
This plugin had the same defect for the same reason and it is fixed here too —
which is worth saying plainly, because two people implemented §7 independently
and both mistook the tie-break for the rule. That is a better argument for one
crate than either of us could have made alone.

**`MessageType` has no codepoint for §3.8's hint or §3.9's purge request/ack.**
§7's `type` list does not name them; `CLIENT-CONTRACT.md` requires them, and they
must travel inside MLS because a hint is only meaningful in being attributable.
`src/envelope.rs` allocates 8, 9 and 10 and says so. Registering them upstream is
one diff and is not mine to make from here.

**Control payloads carry `parents` but are not DAG vertices** — a decision this
plugin had to make rather than read, and the harness is why. Insert a received
`queue_advert` as a vertex and it becomes a head; the peer's next chat references
it; the *sender*, which never persisted it, reports a gap for its own message. A
vertex has to survive a restart or it manufactures gaps.
`envelope::is_transcript_vertex` draws the line at "is this history" and states
the cost: a lost advert or hint is undetectable. If §7 wants them tracked, that
is one function plus persisting them, and it should be the contract's call.

**`set_witness_set`'s argument.** §3.11 declares
`WitnessInput { witnessId, name }`; `bridge.ts` types it `string[]` and
`parity.test.ts` passes `[["0f2c8a41"], 2]`. `models::WitnessInput` accepts both
so neither side is broken.

**There is no error envelope.** §8 says every rejection carries exactly one
member and `types.ts` declares no wrapper, so `error.rs` serializes the bare
kebab code and routes the human-readable context to `tracing`.

**A credential does not attest the key a peer encrypts to.** `device_kem_pk` and
the `KeyPackage` init key cannot be bound with the current APIs — the engine
needs the credential before it exists, and the credential needs the key.
`f2z-msg-mls`'s own tests use a stand-in for the same reason.

**Ambiguity calls, all reversible:** the crate is `tauri-plugin-f2zmsg` and not
`tauri-plugin-f2z-msg`, because `tauri_plugin::Builder` derives the permission
namespace from the crate name and `bridge.ts` hard-codes `plugin:f2zmsg|`;
`add_relay` connects with the two consentable relaxations to *read* the
capability document before judging it, and stores a `ws://` relay as `refused`
so `set_relay_trust` has a `relayId` to address; index blobs stand in for the
prefix scan `f2z-msg-store` deliberately does not offer; the local store is not
encrypted at rest (`f2z-msg-store` offers none, and a plugin cannot add it from
outside — §6.1's `locked` is implemented as a state machine with the device key
sealed under `BackupWrapKey`).

---

## Verification

- 42 unit tests, 10 lifecycle tests, the two-process relay test. Zero warnings
  in both the default and `--all-features` builds.
- `check-rust-fmt.sh`, `check-rust-clippy.sh`, `check-rust-deny.sh`,
  `check-rust-toolchain.sh` — green over 6 wallet crates.
- `check-tauri-plugin-permissions.mjs`, `check-workflow-gates.mjs`,
  `check-github-actions-pins.mjs`, `check-zuuli-linux-image.mjs`,
  `check-zuuli-android-gate.mjs`, `check-hash-domain-labels.mjs`,
  `wasm-boundary.mjs`, `mobile-webview-authority.mjs` — live and `--self-test`.
- `parity.test.ts` still passes (97 tests): mock mode is unchanged.
- `permissions/` is byte-identical to a fresh regeneration.


skyl:
Coordination from #736 (`Closes #733`, `Closes #734`), which lands two §7
corrections that this PR's `src/framing.rs` and `list_messages` are directly
affected by. Flagging now rather than after one of us merges.

**1. `f2z-msg-dag` exists.** `src/framing.rs`'s module note says the crate
"did not exist on `main`, on any remote branch, or in any open pull request
when this was written". That was true when it was written and is no longer:
#732 merged as `5bc785b` on 2026-08-25 and `rs/crates/f2z-msg-dag` is on
`main` now — `AppMessage`, `msg_id`, `Parents`, `MessageDag`, `linearise`,
`PlaintextOutbox`, `GapRequest`/`GapResponse`. The note is worth updating
whichever way the code goes, because a stale "does not exist" is what makes
a seam permanent.

**2. The `msg_id` preimages already differ, so the two are not the same
protocol.** Your note is right that this is the thing that must not diverge.
It has:

- `framing.rs` hashes an envelope carrying `ENVELOPE_VERSION: u16` and a
  `u16` type code; `f2z-msg-dag` hashes `{type: u8, parents, epoch,
  sender_leaf_index, sent_at, retention_class, body}` with no version field
  and a one-byte open `type`.
- #736 additionally moves **`sender_leaf_index` inside the hash**
  (`ARCHITECTURE.md` §7's dated correction, issue #734). §7's sort key
  breaks ties by `(epoch, sender_leaf_index, msg_id)` but `msg_id` did not
  commit to `sender_leaf_index`, while repair delivers a message *outside*
  its original framing — so two receivers who learned one message by
  different routes could compute different sort keys for it. `AppMessageTbs`
  in the crate now has the field; `DagEntry::from_delivered` cross-checks it
  against the MLS framing and returns `DagError::LeafIndexMismatch` on
  disagreement; `DagEntry::from_repair` **lost** its leaf-index parameter.

**3. `list_messages` has #733's bug, in Rust.** This is the one that matters
most. `Engine::list_messages` walks a `BTreeMap` keyed on
`SortKey::to_cursor()`, i.e. `(epoch, sender_leaf_index, msg_id)` — and
nothing reads `parents`. That is §7's **tie-break**, not §7's order. §7's
rule has two halves and the DAG is the primary one; the key decides only the
pairs the DAG leaves *incomparable*.

The counterexample, against this branch: Bob at leaf 1 sends `A` at epoch 7;
Alice at leaf 0 replies `B` with `parents = [A]`, same epoch. Cursors sort
`B` before `A`, so a reply renders above the message it answers — in about
half of all one-to-one conversations, which is where the replier holds the
lower leaf index. That is exactly the defect #733 filed against the
TypeScript comparator, and #736 fixes it by making the engine own the order.
`f2z_msg_dag::MessageDag::display_order()` / `linearise()` is the graph pass;
`rs/crates/f2z-msg-dag/src/order.rs` carries the counterexample as a test.

**4. Therefore the cursor cannot be a sort key.** `SortKey::to_cursor()` is
lexicographically ordered so `before`/`after` match the transcript order —
which works only while the transcript order *is* the key order. It is not.
Worse, positions in a topological order are not stable under arrival: hold
`A` and `C` where `C`'s parent `B` is missing, and nothing orders `A` against
`C` but the key; when `B` lands, `A → B → C` forces `A` first and the two
rows swap. `CLIENT-CONTRACT.md` §7 (as corrected) therefore says **a cursor
is a `msgId`, never an offset** — the message a cursor names is stable, its
position is not. `order.rs`'s
`filling_a_gap_can_reorder_messages_already_held` is that as a test.

**5. What `list_messages` must return.** #736 deletes `compareMessages` from
`wallet/zuuli/src/lib/messaging/types.ts` and adds `CLIENT-CONTRACT.md` §9
**rule 10**: never order a transcript in the frontend. So the plugin's
`list_messages` is now the only thing standing between the DAG and the
screen:

- `MessagePage.messages` **already in §7's order, oldest first** — a
  contiguous window of the total order induced by *every* message held for
  that conversation, not a locally sorted page, so paging backwards cannot
  contradict what is on screen.
- `MessagePage.cursor` is a `msgId`.
- `f2zmsg://message-received` carries no insertion index and must not gain
  one — one index cannot express the reorder in (4). The UI re-reads the
  visible window, which `Transcript.tsx` already did.

**Suggested resolution, and it is your call, not mine.** The cheapest correct
path looks like: depend on `f2z-msg-dag` for the envelope and the order,
delete `framing.rs`'s `AppMessage`/`MsgId`/`SortKey`, and keep whatever
plugin-local types the store needs. The two crates already share
`f2z-codec`'s canonical encoder, `f2z-msg-dag` is `no_std` and reaches
`wasm32-unknown-unknown` (ADR 0001's web client), and it brings the plaintext
outbox and the gap-repair protocol your note lists as deliberately unimplemented.
If instead this PR merges first, #736 rebases onto it and the same
consolidation has to happen anyway — one of us should do it once rather than
both of us maintaining a preimage.

Whichever order we merge in, the ordering fix in (3) is not optional and is
not a rebase artifact: a transcript keyed on the tie-break alone is wrong
today, on this branch, for half of all one-to-one conversations.

skyl:
All five taken, and the branch is rebased onto `66b235b`. Thank you — (3) and
(4) were real defects on this branch and neither was going to surface in a test
I had written.

**1, 2 — `framing.rs` is deleted.** The plugin depends on `f2z-msg-dag`;
`src/envelope.rs` is a thin adapter (hex ↔ `MsgId`, the plugin's codepoints, and
a `dag_error` → §8 mapping). The stale "did not exist" note is gone, replaced by
one that records *why* it existed and what adopting the crate fixed, because a
seam nobody explains is a seam that becomes permanent either way.

`sender_leaf_index` is now hashed: `envelope::seal` takes it,
`DagEntry::from_delivered` cross-checks it against the MLS framing, and the
repair path takes no leaf index — a repaired message records the **author's**
leaf, which is the point of #734.

**3 — `list_messages` pages over `MessageDag::display_order()`.** You were
right, and I had the same defect for the same reason: I read §7's two halves and
implemented the tie-break. What makes that worth saying out loud is that two
people wrote §7 independently and both took the tie-break for the rule, which is
a better argument for one crate than either of us could have made alone.

**4 — the cursor is a `msgId`.** `before`/`after` resolve to a position in the
*current* linearisation at read time, so a window is anchored to a message and
not to an offset. The `BTreeMap` keyed on `SortKey::to_cursor()` is gone from
`store.rs` too — it was the same defect in storage form.

**5 — the page is a contiguous window of the whole order**, oldest first, and
`f2zmsg://message-received` carries no insertion index.

---

Two things I ran into on the way that are yours to weigh, since they are
protocol-shaped rather than plugin-shaped:

**`MessageType` has no codepoint for §3.8's ephemeral hint or §3.9's purge
request and ack.** §7's `type` list does not name them and `CLIENT-CONTRACT.md`
requires them; they have to travel inside MLS, because a hint is only meaningful
in that it is confidential, authenticated and *attributable*. `src/envelope.rs`
allocates 8, 9 and 10 and says it is doing so. Registering them in the crate is
one diff and I did not want to make it from here.

**Control payloads carry `parents` but are not DAG vertices, and that is a
decision I had to make rather than read.** The two-process harness caught it:
insert a received `queue_advert` as a vertex and it becomes a head, the peer's
next chat references it, and the *sender* — which never persisted it — reports a
gap for its own message. A vertex has to survive a restart or it manufactures
gaps, and this plugin rebuilds the DAG from the persisted transcript. So
`envelope::is_transcript_vertex` draws the line at "is this history", and states
the cost: a lost advert or hint is undetectable. For the advert that is fine —
it surfaces immediately as a conversation that cannot send, which is louder than
a gap marker — and for a hint it is exactly what §3.8 says a hint is. If §7
wants control payloads to be vertices, the fix is that one function plus
persisting them, and it should be the contract's call.

---

Evidence, unchanged in shape and re-run after the rebase — two processes, a real
`ws://` listener, `cargo test --features relay-harness`:

```
round 1  alice → bob and bob → alice, parents [] on both (concurrent)
round 2  sent after a full round trip
         bob's round-2 parents = [alice's round-1, bob's round-1]
                                  ↑ the head *set*, which was the other defect
                                    the harness found
gaps 0 on both sides; four ids held on both sides; msg_ids cross-match
```

52 tests green; five consecutive harness runs.

Issue #702, opened 2026-08-25
E2EE: DeviceSignatureKey still signs MLS framing with ed25519-dalek — the last exception to #693


skyl:
#697 closes #693 by giving the MLS engine its own `Signer` over `libcrux-ed25519`
— the same functions `openmls_libcrux_crypto` calls — instead of
`openmls_basic_credential`, which signs with `ed25519-dalek` and `p256` outside
the libcrux provider. `rs/deny.toml` bans `openmls_basic_credential` and
`openmls_rust_crypto` so it stays that way.

**One exception survives that closure, and a closed issue with a live exception
is how a claim quietly drifts back out of the code.**

`f2z_msg_identity::DeviceSignatureKey::sign_mls_content`
(`rs/crates/f2z-msg-identity/src/device.rs`) signs MLS framing with
**`ed25519-dalek`**:

```rust
pub fn sign_mls_content(&self, framed_content_tbs: &[u8]) -> Signature {
    Signature::new(self.0.sign(framed_content_tbs).to_bytes())
}
```

MLS framing is exactly the domain #693 is about. If a client ever signs through
that method, the group's leaf signatures are produced by RustCrypto and verified
by libcrux (OpenMLS verifies through `openmls_libcrux_crypto`) — which is the
two-implementations hazard ADR 0001's single crypto core exists to remove, in the
one place it matters most.

Nothing calls it today. `f2z-msg-mls`'s engine uses its own `DeviceSigner`, and
`DeviceSignatureKey` does not expose its private bytes — correctly — so the two
cannot be bridged from outside. That is why #697 left it alone rather than
reaching into #694's crate from an unrelated branch.

## The options

1. **`DeviceSignatureKey` signs through libcrux.** Then there is one device
   signing key type, `f2z-msg-mls` takes it instead of `DeviceSigner`, and #693
   is true rather than nearly true. This is the right end state.
2. **`DeviceSignatureKey` drops `sign_mls_content`.** MLS framing is signed by
   the engine's signer and the identity crate stops offering an alternative. Also
   honest, and smaller.

Either way the two crates should end up with **one** device-signing type between
them. Whoever takes it should also decide where that type lives — `f2z-msg-identity`
owns §4.2's key hierarchy, so probably there, with `f2z-msg-mls` depending on it.

Refs #693, #694, #697, #305.

https://claude.ai/code/session_01EF5qmzNm91R75ujoFuaTku

skyl:
[...]

PR #750, opened 2026-08-26
fix(zuuli): link tauri-plugin-f2zmsg — the blocker was librustzcash's RustCrypto pre-release island, not our Ed25519

skyl:
ZUULI links `tauri-plugin-f2zmsg`. It could not before this, and the reason was
not the one the resolver named.

## The blocker, diagnosed

Two **independent** conflicts, both rooted in `bip32 0.6.0-pre.1` — the only 0.6
release (2025-01-27), which exact-pins three 2024-era RustCrypto pre-releases,
and which librustzcash pins with `=` in turn:

```
librustzcash  bip32 = "=0.6.0-pre.1"  →  sha2   = "=0.11.0-pre.4"
                                      →  hmac   = "=0.13.0-pre.4"
                                      →  ripemd = "=0.2.0-pre.4"

messaging     ed25519-dalek 3         →  sha2 ^0.11   → 0.11.0  (2026-03-25)
              openmls_rust_crypto     →  hmac ^0.13   → 0.13.0  (2026-03-29)
```

`^0.11` does not match a pre-release and Cargo unifies semver-compatible
versions, so neither pair can coexist.

**The `hmac` half settles the diagnosis, and it changes the answer.**
`openmls_rust_crypto 0.6.0` is an *optional* dependency of `openmls 0.9.0` that
nothing in this tree enables — and Cargo version-resolves optional dependencies
regardless of features, so it constrains the lock anyway. That conflict is
untouched by anything we could do to our own crates. **Moving
`f2z-msg-identity`, `f2z-relay-proto`, `f2z-authority` and `f2z-kt-core` off
`ed25519-dalek` onto `libcrux-ed25519` would have removed the `sha2` row and
left the `hmac` row exactly where it was.** Verified, not assumed — the
resolver output for both experiments is in the verification comment below.

There is a third pin, and it is librustzcash's own. The workspace carries
`block-buffer = "=0.11.0-rc.3"` and `crypto-common = "=0.2.0-rc.1"`, each with
the comment *"later RCs require edition2024"*, each annotated in
`zcash_primitives` as *"remove after edition2024 upgrade"*. **That upgrade has
landed** — the workspace is `edition = "2024"`, `rust-version = "1.88"` — and
every current release of those crates declares `rust-version = "1.85"`. Neither
crate is referenced anywhere in `zcash_primitives`' source. They are stale by
upstream's own stated criterion, and `=0.2.0-rc.1` cannot coexist with the
`^0.2` that `digest 0.11` requires.

So: **yes, the pin is upstream's bug, and upstream has been told.**
**zcash/librustzcash#3010** is open with exactly that 4-line deletion.

## The fix — two transient pins, per AGENTS.md's "branch, PR, pin, resume"

* **`z/zcash/librustzcash` → free2z/librustzcash `f2z/drop-stale-rustcrypto-rc-pins`.**
  Upstream `main` at the previously reviewed `330e4c0`, plus the 4-line deletion
  and its regenerated lockfile. Nothing else. Upstream PR:
  **zcash/librustzcash#3010**.
* **`[patch.crates-io] bip32` → free2z/crates `f2z/bip32-secp256k1-0.29`,
  pinned by `rev`.** iqlusioninc/crates `main` *already* moved bip32 onto the
  released RustCrypto lines — but the same branch bumped `secp256k1-ffi`
  0.29 → 0.31, and `zcash_script 0.4.5` names
  `bip32::ExtendedPublicKey<secp256k1::PublicKey>` against 0.29. No upstream
  commit has both, so this branch is upstream `main` (`eb57d49`) with that one
  requirement held at 0.29 and nothing else. `rev`, never `branch`: `--locked`
  has to mean something.

Both are transient and both say so at their sites — `.gitmodules`,
`scripts/check-librustzcash-compat.mjs`, the `[patch]` block in
`wallet/zuuli/src-tauri/Cargo.toml`, `wallet/deny.toml`'s `allow-git`, and
`wallet/zuuli/scripts/linux-cargo-runtime.mjs`. They come out when `bip32 0.6.0`
is published and the Zcash stack moves to secp256k1 0.31.

**Our Ed25519 does not change.** `ed25519-dalek 3` stays, `verify_strict` stays,
and so does `f2z-codec`'s workspace-wide strict-verification scan. That matters
beyond scope discipline: `libcrux-ed25519` is a raw RFC 8032 `sign`/`verify`
pair with no strict-verification mode, and `f2z-kt-core::sig`'s module header
explains at length why §7.2's non-repudiation argument *needs* small-order and
non-canonical-encoding rejection — "two conflicting cosignatures under one
`witness_pk` are a contradiction on their face" stops being true under malleable
verification. Migrating would have meant either weakening that or
reimplementing the strict checks by hand.

**#702 stays open, and this PR does not touch it.** It is a real issue and the
right end state is its option (1) or (2); it is simply not this blocker, and
folding a crypto-core migration into a dependency-unblocking PR would make both
unreviewable.

## The wiring

* **`wallet/zuuli/src-tauri/src/messaging.rs`** — the enrollment trio, in the
  app crate because it needs the wallet seed and the seed must never cross IPC
  (§2.2). It borrows the seed in-process from `tauri-plugin-zcash`'s managed
  state, derives §4.2's keys, issues the `DeviceCredential`, and hands the
  engine only public material plus the wrap key. Every seed-touching line is
  inside one private function.
  * `f2zmsg_enroll` **doubles as the §6.1 unlock path**, and that is the
    decision most worth reviewing: after a restart the engine is enrolled but
    `locked`, because the device key is sealed under the seed-derived
    `BackupWrapKey`. Nothing in §3's 43 plugin commands can lift `locked` and
    nothing should be able to — lifting it needs the seed. So
    enroll-on-an-enrolled-device re-derives and unlocks instead of refusing, and
    only **from `locked`**: `unlock` ends by setting the engine `stopped`, which
    would otherwise mark a live engine stopped with its relay connections still
    open.
* **`f2zmsg:default` in `capabilities/default.json`, and 42 named
  `f2zmsg:allow-…` in `capabilities/mobile.json`** — mobile gets the whole §3
  surface except `set_relay_trust`, the one grant that is a security downgrade
  (`WIRE.md` §2.3, §5.3): a store build must not let a mobile user opt in to a
  relay advertising `transport_security: "none"`, and `add_relay` already
  refuses one.
* Registered in **`scripts/check-tauri-plugin-permissions.mjs`**'s `PLUGINS`
  (the `capabilitiesPending` reason is gone; its machinery stays, still
  exercised by `--self-test`) and in
  **`wallet/zuuli/scripts/mobile-webview-authority.mjs`**'s reviewed allowlist,
  with negative controls for both directions and for the relay-trust omission.
* `on_event` is the plugin's own — `Exit` / `ExitRequested` only, never
  `Focused(false)`. Not copied from `tauri-plugin-zcash`.
* `zuuallet.yml`'s **upstream canary now fetches `zcash/librustzcash` by URL**.
  While `.gitmodules` points `origin` at our fork, `origin/main` is the fork's
  stale main — a canary testing that is a canary testing nothing.

## Upstream

1. **zcash/librustzcash#3010** — open, the 4-line deletion, rebased on their
   current `main` with the lockfile regenerated. Their own comments say to make
   it; the condition they named has been met.
2. **iqlusioninc/crates** — `bip32` has no released version compatible with the
   RustCrypto *releases*, and `main` has been ready for one since its
   "Bump all dependencies to the current stable versions" commit. The ask is a
   `0.6.0` publish. Not filed yet — it is a request, not a patch.
3. **The structural item, which is the one that matters.** Even with (1) and
   (2), librustzcash and `zcash_script` cannot *take* a released bip32 until
   they move to **secp256k1 0.31**, because bip32 moved with it. So the Zcash
   transparent HD-derivation path is pinned to an island with no exit short of a
   coordinated secp256k1 bump across `zcash_script` and librustzcash. Worth
   raising with them explicitly, because from inside either repo it looks like a
   dependency detail rather than a dead end.

## Also opened

**#753** — a consequence of linking, recorded rather than fixed here: an `Err`
from a plugin's `setup` fails `Builder::build()`, so a messaging store that will
not open now takes the whole wallet down at launch. Failing soft needs
`EngineState::Faulted` plumbed through all 43 commands and does not belong in
this PR.

Refs #735, #702, #753.


skyl:
Not a review of this change — messaging is not mine to review, and I am not asking for anything here. One consequence you will want to know about, because it lands in a file I updated a few hours ago.

**`wallet/zuuli/STATUS.md` carries an "E2EE messaging" row that this PR makes false.** I added it in #747 when `tauri-plugin-f2zmsg` first appeared under `wallet/plugins/`, precisely so the crate existing could not be mistaken for a shipping feature. Its evidence column reads:

> **None, and none is possible from a shipped build:** the plugin is absent from `wallet/zuuli/src-tauri/Cargo.toml` and its lock file, and `src-tauri/src/lib.rs` registers `opener`, `dialog`, `deep_link`, `http` and `zcash` and not this one

This PR adds both of the things that sentence rests on:

```
+tauri-plugin-f2zmsg = { path = "../../plugins/tauri-plugin-f2zmsg" }
+        .plugin(tauri_plugin_f2zmsg::init())
```

So on merge that row states as fact something the tree contradicts. It needs rewriting to whatever is then true — linked and registered, with whatever the honest disposition is for a surface that now ships in the binary but has no production evidence behind it. I wrote the row with "re-check this if the plugin's wiring changes" in mind; this is that.

**Second, a release-mechanics note that has already cost two cycles today.** This PR touches `wallet/zuuli/`, `wallet/plugins/`, `.github/workflows/`, and `scripts/check-*` — all inside the release-impacting surface — so merging it invalidates STATUS.md's audit anchor and the next release bump will fail. Not in `gate`: the freshness rule runs in the **protected release workflow**, so a bump merges green and then dies at release time. That is exactly what #735 did to build 16, and it cost a full re-derivation.

`0.1.0+16` is already pinned at `e53b9da` and is unaffected. The next one after it will need STATUS re-derived against a post-merge SHA. Before pushing any bump:

```
node wallet/zuuli/scripts/status-freshness.mjs --source-sha=<release commit>
```

which prints `{sourceSha,parentSha,auditSha,auditDate}` and exits 0 when the anchor is still good.

I will fix the STATUS row myself once this merges if nobody beats me to it.

skyl:
## Verification

**Reproduction of the blocker, before anything was changed.** Adding
`tauri-plugin-f2zmsg` to `wallet/zuuli/src-tauri/Cargo.toml` on `origin/main` and
running `cargo metadata`:

```
error: failed to select a version for `sha2`.
    ... required by package `bip32 v0.6.0-pre.1`
versions that meet the requirements `=0.11.0-pre.4` are: 0.11.0-pre.4
  previously selected package `sha2 v0.11.0`
    ... which satisfies dependency `sha2 = "^0.11"` of package `ed25519-dalek v3.0.0`
```

**Then the counter-experiment that decided the approach.** With
`ed25519-dalek = "2"` in `rs/Cargo.toml` — the cheap version of the proposed
crypto migration, and strictly more permissive than moving to `libcrux-ed25519`
because dalek 2 still resolves — the same command fails again, on a different
crate:

```
error: failed to select a version for `hmac`.
    ... required by package `bip32 v0.6.0-pre.1`
versions that meet the requirements `=0.13.0-pre.4` are: 0.13.0-pre.4
  previously selected package `hmac v0.13.0`
    ... which satisfies dependency `hmac = "^0.13"` of package `openmls_rust_crypto v0.6.0`
    ... which satisfies dependency `openmls_rust_crypto = "^0.6.0"` of package `openmls v0.9.0`
```

`openmls_rust_crypto` is optional and nothing enables it; Cargo version-resolves
optional dependencies regardless. **No change to our own crates removes that
row.** That is why this PR does not touch Ed25519.

### Builds

| What | Result |
|---|---|
| `cargo check --all-targets`, host (aarch64 macOS via the pinned x86_64 1.97.1) | pass |
| `cargo check --lib --target aarch64-apple-ios` | pass (7m13s) |
| `cargo check --all-targets`, `tauri-plugin-zcash` | pass |
| `cargo check --all-targets`, `wallet/zuuallet/src-tauri` | pass (librustzcash change reaches it too) |
| CI **`iOS / unsigned target app`** | pass |
| CI **`Android / unsigned AAB`** | pass |
| CI **`Desktop / macOS universal app`**, **`Desktop / Linux packages`** | pass |
| CI **`Rust / ZUULI backend`**, **`Rust / shared plugin`**, **`Rust / messaging plugin`**, **`Rust / Android 32-bit`**, **`Rust / modern crypto targets`** ×3 | pass |

So: **desktop and both mobile targets**, and the mobile ones as real packaged
artifacts in CI rather than a local `cargo check`.

### Gates and checkers

`gate` and `rs / gate` results are recorded in the PR checks; `rs / gate` was
green on the pre-rebase head and `rs / clippy`, `rs / rustfmt`, `rs / tests`,
`rs / wasm32` and `rs / cargo-deny` are green again after it. Locally,
`scripts/check-rust-clippy.sh` reports **all 6 wallet crates clippy-clean on the
pinned 1.97.1**, and all of
`check-librustzcash-compat.mjs`, `check-tauri-plugin-permissions.mjs` (+
`--self-test`), `check-rust-toolchain.sh`, `check-rust-fmt.sh`,
`check-rust-deny.sh` (+ `--self-test`), `check-github-actions-pins.mjs`,
`check-workflow-gates.mjs`, `mobile-webview-authority.node-test.mjs`,
`messaging-contract.node-test.mjs` and `artifact-sbom.node-test.mjs` pass.

`permissions/` regenerates byte-identically: the plugin's `build.rs` ran in every
one of the builds above and `git status --untracked-files=all` over
`wallet/plugins/tauri-plugin-f2zmsg` is empty. `wallet/zuuli/src-tauri/gen/schemas`
is gitignored for Zuuli by design and leaves nothing untracked either.

### The one CI failure this PR caused, and what it was

`Desktop / Linux packages` failed the first time with

```
Cargo metadata contains a non-shipping package source:
git+https://github.com/free2z/crates?rev=131d490…
```

That is `linux-cargo-runtime.mjs` doing its job — the shipped Linux artifact's
Cargo evidence had never contained anything but crates.io and path packages.
Fixed by admitting **one exact source string**, repository and revision, with a
test that holds the allowance equal to the revs `src-tauri/Cargo.toml` actually
pins. Not a scheme allowance and not a prefix.

## What I could **not** verify, stated plainly

1. **Two people chatting — in the app or anywhere else — is still not reachable,
   and not because of anything in this PR.** `tauri-plugin-f2zmsg` ships
   `directory::NoDirectory`: there is no HTTP client anywhere in this repository
   that carries a `/kt/v1/lookup` to the key-transparency log, so
   `resolve_handle`, `start_conversation` and `accept_contact_request` fail
   closed with `witness-threshold-unmet`. §6.4 and §9 rule 5 make that the
   *required* behaviour rather than a placeholder. **Linking was a blocker; it
   was not the last one.** The KT client is the next.
2. **I did not run enrollment inside a live ZUULI process.** What is proven is
   that the app links and builds for three targets, that the shipping routers
   route `plugin:f2zmsg|…` and the unprefixed enrollment trio (compiler-bound IPC
   probes over Tauri's mock runtime, in `src/lib.rs`), and that both capability
   files authorize the plugin. What is *not* proven is a real
   seed → `AccountKeys::from_seed` → `issue_device_credential` →
   `install_identity` → `unlock` round trip in a running app: that needs
   `tauri dev`, a created wallet behind a macOS Keychain prompt, and a human at
   the webview, and I could not drive that here. The same sequence *is* exercised
   end to end against a real relay by `src/bin/peer.rs` and
   `tests/two_instances_over_a_relay.rs` (CI `Rust / messaging plugin`, green),
   and `src/messaging.rs` performs it in the same order.
3. **`#753`** — linking made a messaging-store failure fatal to ZUULI's launch,
   because an `Err` from a plugin `setup` fails `Builder::build()`. Recorded
   rather than fixed here: failing soft needs `EngineState::Faulted` plumbing
   through all 43 commands, which does not belong in this PR.
4. **The `[patch.crates-io]` is not exercised by the `rs` tree.** `rs/Cargo.lock`
   is untouched: that workspace has no librustzcash in it and needs no patch, so
   `rs`'s bip32 is still whatever crates.io resolves. Nothing crosses.

skyl:
**Both required gates are green on `af7864a`** (rebased onto `origin/main` `64aca54`): `gate` — SUCCESS, `rs / gate` — SUCCESS. 37 checks pass, 2 skipping (`Publish immutable image`, `upstream-canary`), 0 failing.

That run includes `iOS / unsigned target app`, `Android / unsigned AAB`, `Desktop / macOS universal app` and `Desktop / Linux packages` — so ZUULI packages with the messaging plugin linked on **both mobile targets and all three desktop ones**, and the Linux job's shipped-artifact SBOM now binds the patched `bip32` as a reviewed git source rather than refusing it.

Not merging, per instruction. Ready for review.

Issue #754, opened 2026-08-26
release: six parts of the release path have never executed, and two live releases have already failed on ones that hadn't


skyl:
Two Android releases have now failed on steps that had never executed, both added by #534 after `0.1.0+14`:

- `0.1.0+15` died in `seal_verifier` ([#738](https://github.com/free2z/zuu/issues/738)).
- `0.1.0+16` fixed that, advanced, and died in the signed/unsigned payload comparison ([#751](https://github.com/free2z/zuu/issues/751)).

Each fix reveals the next unexecuted step. That is not two bugs, it is one process gap: **a release path is the only path merging cannot exercise**, so new steps in it ship untested and fail one live release at a time. The audit below was produced while fixing #751 and lists what is still unproven.

## Still never executed in a real release

1. **Everything in the Android sign step after the `cmp`** — `signed_abis`, `signing-record.json`, `CHECKSUMS`, and the entire Play upload block. #752 unblocks the `cmp`; this is what will surface next.
2. **Both `Destroy …` credential steps on the success path.** They have only ever run on the failure path. A credential-destruction step that has never run to completion after a successful upload is the one you least want to discover is broken.
3. **The whole `android-finalize` job** — SBOM unpack, scan, binding, and provenance attestation. Skipped in both 15 and 16.
4. **`release-index`'s rewritten `if:`** — a twelve-clause Android expression that **has never evaluated true**. It reads correct. The failure mode is quiet: a release that silently skips its own index, which is exactly the kind of thing nobody notices until they need the index.
5. **Every new Linux step** — Cargo audit instrumentation, per-artifact SBOM scan and binding, source-inventory labelling.
6. **Every new macOS step** — the same SBOM work plus the Keychain entitlement policy.

Points 5 and 6 have a sharper edge than "new steps are untested": they are gated on `target == desktop || target == all`, and **builds 12 through 16 were all mobile**. So Linux and macOS have not run in *any* recent release — not merely their new steps. Whatever state those paths are in, no recent evidence says they work.

## What would actually close this

Not "be careful". The release path needs a way to run without shipping:

- **A dry-run that reaches every step.** `dry_run=true` already exists for desktop and still signs and notarizes; the equivalent for the mobile path should execute the full sign → compare → record → finalize chain against the real artifacts and stop only at the upload API call. Anything the dry run cannot reach is, by construction, only testable in production.
[...]

Issue #751, opened 2026-08-26
release(android): the signed/unsigned payload comparison sorts one side and not the other, so it can never pass — and it had never run before build 16


skyl:
`0.1.0+16` got Android further than `0.1.0+15` — the credential-free universal AAB built and sealed successfully, so [#738](https://github.com/free2z/zuu/issues/738)'s fix works — and then died in the next job.

[Run 32911822458](https://github.com/free2z/zuu/actions/runs/32911822458), `Android / protected sign and Play upload`, step 7:

```
/home/runner/work/_temp/unsigned-aab-payload.sha256 /home/runner/work/_temp/signed-aab-payload.sha256 differ: byte 1, line 1
```

The check is a good one: sign the AAB, then prove signing changed only the signature and not a single payload byte. It cannot pass as written.

## Cause: the two sides iterate in different orders

The unsigned side reads a **sorted** member list (`.github/workflows/zuuli-release.yml:382-388`):

```sh
unzip -Z1 "$aab" | LC_ALL=C sort > "$RUNNER_TEMP/current-aab-members.txt"
…
while IFS= read -r member; do
  case "$member" in META-INF/*|*/) continue ;; esac
  printf '%s  %s\n' "$(unzip -p "$aab" "$member" | sha256sum | awk '{print $1}')" "$member" >> …
done < "$RUNNER_TEMP/current-aab-members.txt"
```

The signed side reads **archive order** (`:431-435`):

```sh
unzip -Z1 "$secret_dir/ZUULI-signed.aab" | while IFS= read -r member; do
  case "$member" in META-INF/*|*/) continue ;; esac
  printf '%s  %s\n' "$(unzip -p … "$member" | sha256sum | awk '{print $1}')" "$member" >> …
done
```

No `LC_ALL=C sort`. `cmp` is byte-exact, so two files carrying identical `(digest, member)` pairs in different sequences differ at the first line — which is exactly what the runner reported. The digests are almost certainly all equal; the check is comparing the same set written two different ways.

## Why nobody saw it

Same reason as #738, and this is the part worth noticing. **The comparison did not exist for `0.1.0+14`:**

```
$ git show 5d51a792:.github/workflows/zuuli-release.yml | grep -c unsigned-aab-payload.sha256
0
$ git show 3fa7fd48:.github/workflows/zuuli-release.yml | grep -c unsigned-aab-payload.sha256
4
```

[...]

Issue #738, opened 2026-08-25
release(android): seal_verifier leaks sha256sum -c output into $GITHUB_OUTPUT, and it blocked build 15 from Play internal


skyl:
Build `0.1.0+15` reached TestFlight and **did not reach the Play internal track**. The AAB built fine; the job died three seconds after it, in output handling.

```
##[error]Unable to process file command 'output' successfully.
##[error]Invalid format 'aab-members.txt: OK'
```

[Run 32885179531](https://github.com/free2z/zuu/actions/runs/32885179531), job `Android / credential-free unsigned universal AAB`, step 14 `Build source-bound unsigned universal AAB`.

## Cause

`seal_verifier` in `wallet/zuuli/scripts/android-release-artifact.sh` is contracted to print **one** thing on stdout — the sealed `CHECKSUMS.sha256` digest, from `sha256_file` at `:89`. It prints four:

```sh
# :72-78
(
  cd "$directory"
  …
  sha256sum -c CHECKSUMS.sha256      # <- writes "<name>: OK" per member, to STDOUT
) || { echo "unsigned artifact is not ready to seal" >&2; return 1; }
```

`sha256sum -c` reports each member on stdout, so the subshell emits:

```
ZUULI-android-unsigned.aab: OK
aab-members.txt: OK
source-record.json: OK
```

and the caller captures the lot:

```yaml
# .github/workflows/zuuli-release.yml:290
payload_sha256=$(scripts/android-release-artifact.sh seal-verifier unsigned-android "$verifier" a099…)
…
echo "payload_sha256=$payload_sha256" >> "$GITHUB_OUTPUT"   # :296
```

`$GITHUB_OUTPUT` takes `key=value` per line. Line one is `payload_sha256=ZUULI-android-unsigned.aab: OK`; line two is `aab-members.txt: OK`, which has no `=`, and the runner rejects the whole command.

## Why it was not caught

**This step has never run in a real release.** `seal-verifier` entered the release workflow in #534, which landed after `0.1.0+14`:

```
$ git show 5d51a792:.github/workflows/zuuli-release.yml | grep -c seal-verifier
0
```

Build 15 is its first execution, and it failed on that first execution. Whatever covers `android-release-artifact.sh` today does not exercise `seal_verifier`'s **stdout contract** — that the function prints exactly one line, and that the line is a bare 64-character digest.

## Fix

Silence the check's stdout while keeping its exit status, which is the only thing `:78` reads:

```sh
[...]

PR #534, opened 2026-08-23
ci(android): split unsigned build from protected publish

skyl:
## Summary

- build and attest the exact-source universal four-ABI AAB in a credential-free job
- verify one immutable artifact, source identity, checksums, provenance, package/version/ABIs, and pinned Bundletool before the protected job materializes signing or Play credentials
- sign without rebuilding Rust, retry only the Play bundle upload, destroy credentials before handoff, then finalize SBOM/checksum/provenance in a credential-free job
- extend the release boundary policy with current-workflow mutations for credential exposure, stale or unattested inputs, dependency/build escape, skipped or soft-failed controls, cleanup, and gate result handling

## Verification

- npm test: 672 Vitest passed / 5 skipped; 235 Node passed; 126 Playwright passed
- post-main focused release tests: 161 passed / 4 platform skips
- release identity, live credential-boundary verifier, actionlint, shellcheck, and diff check pass

Refs #293.

This intentionally does not close #293: Tauri CLI still compiles the first Android target once for plugin initialization and again through Gradle, with no supported initialization-only contract. The upstream fix is tracked at tauri-apps/tauri#15910 and documented here.

PR #755, opened 2026-08-26
docs(zuuli): re-derive the release-readiness audit for 0.1.0+17

skyl:
Re-derived against `fe60fc6b` after #752 landed in `wallet/zuuli/`, which
invalidated the previous anchor.

**The API side is unchanged.** All eight anonymous GETs still 200, the five
protected probes still 403, `/api/auth/social/mobile/providers/` still 200 with
every provider `configured: false`.

**Android is now two builds behind, not one, and the row says so.** iOS
`0.1.0+16` read back at 01:19Z with all three Booleans true. The same-hour Play
audit reports exact release 16 `present: false`; 15 never reached the track
either. **Build 14 is still the newest build on the Play internal track.**

Two different Android faults, and they share a cause worth recording rather
than treating as bad luck: both were steps #534 added to the release path, and
**neither had ever executed**. Build 15 died in `seal_verifier` (#738); build 16
got past it — proving that fix — and died in the signed/unsigned payload
comparison (#751, fixed in #752). #754 lists six further release steps that
have still never run, including the whole Play upload block, the entire
`android-finalize` job, and every Linux and macOS step, since builds 12 through
16 were all mobile.

The device checklist is unchanged and still unmet; nothing recording a
signed-device wallet operation has landed.

`npm run test:status-freshness`: 31 pass.

Claude-Session: https://claude.ai/code/session_01NMwYnLDGotB9Ph4NgDFNeD

### DIFF
--- a/wallet/zuuli/STATUS.md
+++ b/wallet/zuuli/STATUS.md
@@ -8,7 +8,7 @@ back from the named store. Authenticated, money-moving, wallet, KYC, and media
 operations are not called working without recorded evidence from that path.
 
 Last re-derived from `origin/main` at
-`72a4d9d8236583dc82b08399536788e45173516d` on 2026-08-25. Before a release,
+`fe60fc6b75438de308668fc57b98f3cb4e33f65c` on 2026-08-26. Before a release,
 update the evidence and disposition for every non-ready row; do not carry this
 commit or date forward mechanically.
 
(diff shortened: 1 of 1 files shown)


skyl:
Closing rather than fighting the conflict — this became `DIRTY`, and the right move is to redo it from scratch later rather than rebase it.

Two things happened after I opened it:

1. **#750 merged and rewrote the same STATUS row**, so the file conflicts. That is a good outcome, not a collision to resolve: their E2EE messaging row is better than the one I would have written. I recorded "linked and registered, no production evidence". They established that first contact **cannot succeed at all** — the plugin ships `directory::NoDirectory` because no `/kt/v1/lookup` client exists anywhere in the repository, so `resolve_handle`, `start_conversation` and `accept_contact_request` fail closed with `witness-threshold-unmet`. "Linked and reachable; still not usable" is the accurate disposition and I am keeping it verbatim.

2. **#750 also invalidated the anchor this PR recorded** (`fe60fc6b`), along with `wallet/deny.toml`, `z/zcash/librustzcash`, four `scripts/check-*`, and most of `wallet/zuuli/src-tauri/`.

More importantly, **the release this audit was for is on hold.** #753 — a messaging store that cannot be opened takes the whole ZUULI wallet down at launch — is a launch blocker introduced as a consequence of #750's link, and it is being fixed now. That fix lands in `wallet/plugins/` and `wallet/zuuli/src-tauri/`, so it will invalidate any anchor I set today anyway.

Re-deriving now would produce an audit that is stale before the bump. I will re-derive against the post-#753 tree and open a fresh PR then, updating only the anchor, the evidence run links, and the distribution row.

For the record, the distribution facts this PR captured still hold and will carry forward: iOS `0.1.0+16` read back at 01:19Z with `uploaded`/`processed`/`availableToInternalTesters` all true; the Play audit reports exact release 16 `present: false`; **build 14 remains the newest build on the Play internal track**, two behind.

Issue #634, opened 2026-08-24
KT.md §8.1/§9.5: 'no such user' must be proved, and akd 0.13 cannot produce that proof


skyl:
Found while implementing `f2z-kt` (branch `feature/kt-server`, PR #629). **`KT.md` requires the log to *prove* that a handle is unregistered, and `akd` 0.13 has no API that produces such a proof.**

## The requirement, and it is deliberate

§9.5, spelling out why the error table has no unknown-handle code:

> **There is no "unknown handle" code**, deliberately: an unregistered handle is answered with a non-membership proof (§8.1). "No such user" is a claim the log must prove.

That is the right design. A log that can say "never heard of them" without proving it can disappear a user from the directory and be indistinguishable from a log that is telling the truth — which is exactly the withholding attack the whole construction exists to bound.

## Why it is not implementable as adopted

`akd::Directory::lookup` **errors** for a label with no user state. It does not produce a proof of anything:

```rust
match self.storage.get_user_state(&akd_label, ValueStateRetrievalFlag::LeqEpoch(epoch)).await {
    Err(_) => {
        // Need to throw an error
        Err(AkdError::Storage(StorageError::NotFound(format!("User {name} at epoch {epoch}"))))
    }
    Ok(latest_st) => self.build_lookup_info(&latest_st).await,
}
```

The `NonMembershipProof` that *does* exist inside a `LookupProof` is about the freshness marker for a label that **is** in the tree — it proves a stale version is not the current one. It is not a proof that a label was never registered, and there is no public entry point that asks for one.

So a conforming answer to "resolve `@nobody`" cannot be produced by the library ADR 0013 adopted.

## What #629 does

Tells the truth rather than dressing it up. The lookup response carries an explicit `Presence` discriminant, and the absent case is named `AbsentUnproved` on the wire, with the type documentation stating plainly that it is an assertion and not a proof. A client can therefore *see* that it has been given something weaker than §8.1 promises, instead of inferring a proof it never received.

That is a stopgap and it is marked as one. It should not be the answer.

## Possible directions, none of them free

- **Upstream.** A `Directory::lookup` variant that returns a non-membership proof for an absent label, against the VRF-derived label. Per `AGENTS.md`'s prime directive this is the preferred shape if it is tractable — and unlike the `audit_verify` wasm wart (#544) it is not a one-line change.
- **Publish a presence commitment** the log signs per epoch, which bounds the lie to one epoch rather than eliminating it. Cheap, weaker, and would need specifying.
[...]

Issue #633, opened 2026-08-24
KT.md §5.1: an empty epoch is not producible with akd, and the numbering it implies breaks /kt/v1/audit


skyl:
Found while implementing `f2z-kt` (branch `feature/kt-server`, PR #629). **`KT.md` §5.1's empty epoch cannot be produced with `akd`, and the numbering it implies makes every audit range containing one unverifiable.**

## The rule

§5.1, and it is load-bearing rather than decorative:

> **The log publishes an epoch every `epoch_interval` seconds — proposed 600 — whether or not there is anything to publish.** An epoch with no submissions is a valid epoch: same monotonicity, same signature, same cosignatures, **an append-only proof over zero insertions**.
>
> […] A witness that sees nothing for six hours cannot distinguish "nobody changed a key" from "the log has stopped" from "the log is serving me a stale branch while it serves someone else a fresh one." A heartbeat epoch […] converts that ambiguity into a **detectable fault** with a timestamp attached.

I agree with every word of that. The problem is that the adopted library cannot express it.

## Why `akd` cannot

**`akd::Directory::publish` with an empty update set does not advance the epoch.** It filters the batch, finds nothing to insert, and returns the *current* `EpochHash` unchanged:

```rust
if update_set.is_empty() {
    info!("After filtering for duplicated user information, there is no publish which is necessary (0 updates)");
    let root_hash = current_azks.get_root_hash::<TC, _>(&self.storage).await?;
    return Ok(EpochHash(current_epoch, root_hash));   // note: current, not next
}
```

So "an append-only proof over zero insertions" does not exist: there is no epoch transition to prove.

**And a `SignedTreeHead.epoch` decoupled from `akd`'s epoch does not rescue it.** `audit_verify` requires exactly one proof segment per epoch transition, and `AppendOnlyProof.epochs` must line up with the roots being checked — `f2z_kt_core::auditor::verify_append_only` additionally pins each `proof.epochs[i]` to `heads[i].sth.epoch`, deliberately, so that a log cannot serve a valid proof for a *different, earlier* range that happens to match. If our epoch numbering ran ahead of `akd`'s, `/kt/v1/audit` would be unverifiable for every range containing a heartbeat.

The two constraints are jointly unsatisfiable as written: §5.1 wants an epoch that adds nothing, and §9.2's audit endpoint wants every epoch to be an `akd` transition.

## What #629 does, and it is an invention

[...]

PR #769, opened 2026-08-26
feat(e2ee): §12.6 — MLS KeyPackage publication, and first contact works end to end

skyl:
Closes the last functional blocker on first contact. Before this, `resolve_handle` worked, `accept_contact_request` worked, and `start_conversation` ran the whole verified directory lookup — monotonicity, threshold, inclusion proof, alarms, pin — and then **refused by name**, because `WIRE.md` §12.5 step 2 needs a `Welcome` addressed to a `KeyPackage` and nothing published one. Two people could find each other and neither could open a conversation.

## The tension, and the decision

**A `KeyPackage` is consumed on use; a key-transparency log is append-only.** RFC 9420 §10 deletes a package's init secret once a `Welcome` addressed to it is processed; `KT.md` §4.2's version chain and §4.3's one-entry-per-epoch rule exist so nothing in the log is ever withdrawn. An append-only structure cannot express consumption. That is why §4.1 says a `DirectoryEntry` carries *"no `KeyPackage`"*, and it is still why.

So the question is not *how do we get key packages into the directory*, but **what already knows how to address a device and is allowed to forget things?**

**Decision: a device's key packages live at the relay hosting its contact queue, addressed by the same `contact_addr` its directory entry already publishes.**

- `PUBLISH_KEY_PACKAGES` (`0x0032`) — signed by the **contact queue's receive key**, addressed to its `recv_addr`. Appends, skips byte-duplicates so a retry is safe, clamps to a published cap, replaces the one last-resort package.
- `CLAIM_KEY_PACKAGE` (`0x0033`) — **unsigned**, behind a proof-of-work stamp with its own challenge purpose. Serves and deletes the oldest; serves the last-resort package without deleting when the pool is empty; `ERR_UNAVAILABLE` when there is neither.

Four properties made this the shape:

- **No new identifier at the relay.** ADR 0004 forbids accounts, handles and identity keys; a `contact_addr` is an opaque value the relay generated itself and has held since `CREATE_CONTACT_QUEUE`. The directory does the handle → address binding, already covered by the entry's authorization signature.
- **No new key.** Owning the pool is owning the contact queue.
- **No wire change to `DirectoryEntry`.** `KT.md` §12 names such a change as the price of closing an *unrelated* open item; this one does not spend it.
- **The relay is a cache of self-authenticating objects.** It can withhold, exhaust, or choose among the device's own packages. It cannot substitute one, because it does not hold the device's signing key.

### Rejected, with reasons — [ADR 0015](docs/e2ee/decisions/0015-key-package-publication.md) has these in full

1. **In the directory, consumption tracked elsewhere.** One directory epoch per conversation started; §4.3 caps refill at the epoch cadence; `entry_version` becomes a public counter of how many people have contacted you. The commitment-only variant adds a directory write per refill and buys nothing the `DeviceCredential` does not already provide.
2. **A purpose-built relay endpoint.** Its address is either derived from an identity — the identifier ADR 0004 forbids — or a second opaque address that must be published in the directory, which is the frozen-structure wire change plus two addresses per device that can get out of sync.
3. **The free2z platform serves them.** Puts an account-bearing service back in the first-contact path and breaks federation: a relay-hosted pool federates, a platform-hosted one makes free2z a required participant in every conversation.
4. **Synthesize a package from the published `device_kem_pk`.** Needs no endpoint at all — and makes **every** first contact use a permanently reused init key, which is strictly worse than the last-resort package. It is what a last-resort package degenerates into if never rotated, which is why §12.6.6 bounds its lifetime.
5. **Invert the handshake — the responder creates the group.** The closest call. No pool, no exhaustion, no relay change, better forward secrecy. Rejected because it **moves an unbounded-state problem from the relay onto the initiator**: Alice must keep the init secret of every unanswered request alive indefinitely, uncapped, invisible, ungarbage-collectable. The relay's pool has the same problem in mirror image and there it is capped, refillable, observable and paid for by proof of work.

## Authentication — mandatory and structural

A claimed package is authenticated against the `DirectoryEntry` the log proved: the credential's signature under the entry's `identity_pk` (**the check the section exists for**), its handle, its membership in the published `devices` set, its absence from `revocations`, the leaf binding, the ciphersuite, and RFC 9420's own validation.

`MlsEngine::add_member` now takes a `VerifiedKeyPackage`, which has **no constructor but the verifying one**. There is no path from bytes-a-relay-handed-us to a member of a group that skips the directory — the way to break it is to delete the check, not to forget it.

## Costs, stated where they land

- `THREAT-MODEL.md` **§4.12** — a last-resort package is a reused init key: two initiators under one long-lived secret, retroactively openable. Taken because the alternative is unreachability, which an attacker can force for the price of `contact_max_key_packages` stamps.
- **§4.13** — a hostile relay can serve the reusable package while holding a full pool, and the `last_resort` byte is unsigned. The recipient can detect it; nothing in v1 requires it to.
- **§4.14** — a claim is observable even when first contact is abandoned. One new bit beyond §4.1.
- `WIRE.md` **§12.6.9** — a claim consumes and a `CONTACT_APPEND` does not, so draining a pool is cheaper than flooding a contact queue.
- The **Forward secrecy** row of §5's summary table now carries the exception.

## Evidence

**`rs/crates/f2z-kt-client/tests/first_contact.rs` — cold start, real everything.** Two independent instances open a conversation from nothing but a handle, against a real `f2z-kt` `LogService` (real `akd` trees, real VRF, real protobuf inclusion proofs, real epochs), a real `f2z-witness` that verified the append-only proof before cosigning, the shipping `KtClient::resolve` with the threshold applied, and a real relay over a real `ws://127.0.0.1:0` socket. Key packages really published, really claimed behind a stamp, real MLS `Welcome` under the X-Wing hybrid suite, messages both ways, and the claimed package is never served twice.

Plus the two required refusals: a package from a second **valid** identity is refused for `@bob` (and `@bob`'s own still verifies, so the assertion is about the substitution); and an exhausted pool falls back to the reusable package and then answers `ERR_UNAVAILABLE` rather than panicking.

It lives in that crate because `f2z-kt` and `f2z-witness` are **AGPL-3.0** and `tauri-plugin-f2zmsg` is on the client side of `rs/README.md`'s boundary — a dev-dependency on either would put them in ZUULI's lockfile. `f2z-kt-client` already dev-depends on both, by the same argument its `acceptance.rs` makes.

**The plugin's two-process harness passes** the shipping `start_conversation` / `accept_contact_request` path over a real relay, now publishing a **real signed `DirectoryEntryTBS`** so the §12.6 check it runs is the shipping one against real bytes.

Also: 7 new relay conformance vectors (run against both `FakeRelay` and the real `f2z-relay`), 7 store property tests against both stores, 10 `f2z-msg-mls` tests over the verification.

## Still open, and named

- The numbers: pool target, low-water mark, last-resort lifetime. Placeholders in the sense §12.3's caps are.
- Multi-relay redundancy — `k` is §13-G, and this build claims the first endpoint.
- Whether a recipient should alarm on a manufactured exhaustion. The evidence exists; no mechanism is specified.
- ZUULI still ships `NoDirectory`. That is now a **configuration** gap — a log identity, a signing key, a witness list, a threshold — not a code gap.

Surfaced but not re-reported: #649, #634, #633.

Refs #305.


skyl:
## Status for whoever picks this up

**This PR is `CONFLICTING` and has zero CI checks.** Two things to do before anything else:

1. **Merge `origin/main` into the branch and push.** `main` has moved to `4dfdbbf` (#767 landed after this branch was cut). Do not rebase — `AGENTS.md` forbids force-pushing a shared branch, and squash-merge collapses the merge commit anyway.
2. **Confirm the workflows actually start.** GitHub created no runs for the last two commits here — diagnosed as an Actions-side delay rather than anything in the branch, since the ref is correct. If runs still do not appear after the merge-up, an empty commit or a close/reopen will kick them.

Both required contexts must be green before merge: **`gate`** (wallet/zuuli) and **`rs / gate`**. They are easy to confuse — check the workflow, not just the name.

## What this PR is

The last functional gap. Without it, two people can find each other through the verified directory and one can accept the other, but **neither can open a conversation** — a `Welcome` must be addressed to an MLS `KeyPackage` and nothing published one.

Everything was reported green locally at the time of the last push: full `rs` workspace `--locked --all-targets`, clippy, fmt, cargo-deny, both wasm32 builds, the plugin's `--features relay-harness` suite, and the permission checker. The cold-start test (`f2z-kt-client/tests/first_contact.rs`) opens a conversation from nothing but a handle against a real log, a real cosigning witness and a real relay.

## Two findings in here worth not losing

**RFC 9420's `last_resort` extension does not work out of the box.** OpenMLS's `KeyPackage::validate` requires every key-package extension to be listed in the leaf's `Capabilities`, and `Capabilities::default()` lists none — so `mark_as_last_resort()` alone produces a package **this crate's own verifier refuses**. A pool of silently unusable last-resort packages.

**The refill rule had a trap.** Most claims never produce a `Welcome`, so consumption is *invisible to the owner*. Deciding when to refill from a local count means sitting at "I published 32" while the relay holds zero — falling back to the reusable package on every first contact, forever, with no signal. Fixed by probing with an **empty publish** that changes nothing and returns the relay's true `pool_size`; §12.6.6 is now normative on this with a conformance vector holding the relay to it.

## After this lands

#770 is the next blocker and it is a **decision, not code**: the shipped witness set and threshold `t`. The plugin will keep shipping `NoDirectory` until that is settled.

Context: https://github.com/free2z/zuu/issues/305#issuecomment-5448003049

skyl (commented):
Adversarial security review: changes requested. I did not approve.

1. **Blocking — first contact only works when the peer happens to use an already configured local relay, and the signed relay identity is discarded.** `ResolvedPeer` carries `contact_relay_url` and `contact_addr` but drops the directory entry's `relay_id` (`directory.rs:95-110, 511-533`). Both claim and contact append then do only `connections.get_mut(&peer.contact_relay_url)` and fail `RelayUnreachable` when absent (`engine.rs:3047-3050, 3133-3137`). `start_engine` connects only the user's stored relays; it never opens the peer's published endpoint. Therefore two users on different federated relays cannot start a conversation. The two-process harness and the cold-start test both put both peers on one shared configured relay, so this path remains unexercised. Opening the endpoint also must carry and enforce its signed `relay_id`, not route by URL alone.

2. **Blocking — cross-role duplicate publication can reuse a nominally single-use init key.** §12.6.3 says byte-identical duplicates are skipped. The memory store unconditionally assigns a supplied last-resort package (`memory.rs:308-310`), and SQLite unconditionally deletes/reinserts it (`sqlite.rs:963-976`), without checking whether those bytes already exist in the pooled set. The testkit duplicates the behavior (`state.rs:436-438`). Publish `packages=[X], last_resort=X` (or publish X pooled, then X as last resort): X is served once as pooled and then repeatedly as last resort. This defeats the pool's one-Welcome-per-init-key property. A last-resort candidate equal to any post-batch pooled package should be skipped; do not “fix” this by deleting the pooled copy and converting it to reusable.

3. **Blocking availability regression after 30 days — the reference client never rotates the last-resort package it deliberately expires.** `ensure_key_packages` decides replacement solely from the relay's boolean `has_last_resort` (`engine.rs:3252-3259`). It generates the 30-day package only when that boolean is false (`3275-3283`), while the relay retains expired bytes indefinitely. Thus after 30 days an online device still reports “has last resort,” never replaces it, and an attacker can drain the ordinary pool; every later claimer receives an expired package that strict verification rejects. Restarting or receiving contact traffic does not repair it. Persist the publication/expiry metadata and replace with overlap before expiry (without replacing on every top-up).

Targeted checks run: all 26 `f2z-relay-store` property tests pass and all 10 `f2z-msg-mls` key-package tests pass. Those green tests do not cover the three cases above. Strict credential/package verification itself follows the intended boundary: canonical parse, OpenMLS validation, fixed ciphersuite, strict identity signature, handle/device/revocation checks, and a constructor-closed `VerifiedKeyPackage`.

skyl (commented):
Adversarial re-review at `b427c3d`: changes still requested; I did not approve.

**Blocking — the responder pins its outbound relay to an attacker-supplied identity, not an authenticated advert.** `pump_contact_queue` parses the unsigned `CONTACT_APPEND` body and persists `contact.advert.{relay_url,relay_id,send_addr}` verbatim (`wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs:3540-3575`). `accept_contact_request` resolves the claimed identity but uses only `peer.identity_pk`; it rebuilds the introduction from those untrusted stored advert fields (`:752-795`). `join_conversation` then stores them as the outbound queue and immediately calls `send_control` (`:3060-3117`), where `endpoint_connection` faithfully pins to the attacker-provided `relay_id` (`:3302-3337`). The check runs, but its expected value came from the same unauthenticated payload as the URL, so it cannot detect substitution.

This directly conflicts with `WIRE.md` §7.2/§12.5: a `queue_advert` is confidential and authenticated **inside the MLS group**; `CONTACT_APPEND` is unsigned and is specified to carry the `Welcome`, with Bob then sending the first queue advert inside MLS. A contact-queue attacker can currently replace Alice’s advert with its own relay/queue. Bob joins Alice’s genuine Welcome, connects to and binds the attacker-chosen queue, and sends the MLS-encrypted response there. The attacker cannot decrypt it, but can black-hole first contact and gains routing metadata. The new two-relay harness uses only honest adverts, so it passes even if either `expected_relay_id` assignment/check is deleted; it does not provide mutation/fail-before evidence for the security boundary it claims.

The original three implementation gaps otherwise look materially addressed: the initiator opens the signed directory endpoint on demand; cross-role dedup is no-downgrade in memory, SQLite, and testkit with fail-before properties; and expiry metadata advances only after relay-confirmed last-resort publication, so pre-publication failure leaves the old fallback/deadline retryable. Targeted local results: relay-store properties 29/29, relay-testkit unit tests 38/38, rotation 1/1, two-relay two-process harness 1/1. CI is not approvable independently at this head either: `rs / wasm32` is currently failing and neither required aggregate gate has completed.

skyl:
CI follow-up: `rs / gate` has now completed **failure**. The two concrete failures are:

- `rs / wasm32`: `f2z-msg-mls/src/keypackage.rs:215` passes `std::time::SystemTime` to OpenMLS `Lifetime::validate_with_time`, which expects `web_time::SystemTime` on `wasm32-unknown-unknown` (`E0308`).
- `rs / tests`: `f2z-relay/tests/conformance.rs:227` trips the reviewed-suite population assertion after the new vector: actual 61, expected 60.

These are in addition to the responder-side authentication blocker in the review.

skyl (commented):
Adversarial re-review at `7f297bf`: changes still requested; I did not approve.

**Blocking — the DSK signature authenticates the route but is not bound to the `Welcome` or the group creator, so a hostile contact relay can substitute the conversation identity while retaining a valid Alice advert.** `routing_advert_digest` covers only `conversation_id` and serialized `QueueAdvert` (`wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs:2933-2944`). `accept_contact_request` verifies that signature against some active, non-revoked device in Alice’s resolved entry, then independently joins the supplied Welcome (`:785-795`). `join_conversation` validates the Welcome’s member credentials only under their own embedded identity keys; it never requires the group creator leaf/DSK or identity key to match `advert_device_pk` / Alice’s verified entry (`:3146-3166`; `rs/crates/f2z-msg-mls/src/engine.rs:595-615`).

Attack: a hostile relay observes Alice’s genuine unsigned `CONTACT_APPEND`, retains its valid `(conversation_id, advert, Alice DSK signature)`, and replaces only `welcome` with a new group it created using the same public Bob KeyPackage and chosen group id. Bob’s advert authentication succeeds, OpenMLS accepts the attacker’s internally valid credential, and the stored conversation is labeled/fingerprinted as Alice even though its MLS peer is the attacker. This is possible even for a nominally single-use package because the hostile relay sees the claimed public KeyPackage before deletion and can retain it. Bind the signature to a digest of the exact Welcome as well, or after joining prove the creator leaf is the same active directory device that signed the bootstrap.

The added mutation tests catch relay-id changes and an actual relay-id mismatch, but no test replaces the Welcome while preserving a valid signed advert; the current code and happy-path harness would accept that mutation. Tests for deleting the active-device/revocation checks are also absent, although those checks themselves look correct.

Everything else requested now checks out locally: exact CI WASM command succeeds; real/fake relay conformance 2/2; key-package/routing tests 11/11; last-resort/relay substitution tests 2/2; and the two-process, two-relay harness 1/1.

skyl (commented):
Adversarial re-review at `7189150`: implementation looks correct, but the security gate remains mutation-deletable, so I did not approve.

**Blocking test gap — replacing the complete shipping pre-join verifier with `Ok(())` leaves every focused plugin test green.** I tested this in an isolated detached worktree by changing only `Inner::authenticate_introduction` (`wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs:3115-3151`) to ignore its inputs and return success, while leaving the call from `accept_contact_request` in place. Results remained: plugin lib **54/54**, last-resort/relay substitution **2/2**, and the shipping two-process/two-relay harness **1/1**. This mutation produces no dead-code warning and would survive the current focused QA.

The two new tests prove separate helpers: `welcome_is_inside_the_device_authenticated_routing_digest` proves the digest changes, while `a_swapped_welcome_breaks_the_device_authenticated_routing_transcript` calls `MlsEngine::authenticate_routing_advert` directly with arbitrary byte strings. Neither drives a modified stored contact request through `accept_contact_request`, so neither proves that the shipping path invokes the check. This is precisely the recurring defect class identified in #305: a correct security check that can stop running while CI stays green.

Add a fail-before integration test that receives/persists a valid signed bootstrap, changes only the Welcome (and preferably separate route/signature/device/revocation mutations), then proves `accept_contact_request` refuses before `join_from_welcome` or any conversation write. That test must fail if the `authenticate_introduction` call/body is deleted.

The implementation itself closes the prior attack: the exact Welcome bytes and complete routing fields are independently domain-hashed into a fixed-width outer transcript; labels are prefix-free; verification uses the active, non-revoked DSK from the witnessed entry; and it occurs before joining or connecting. Fresh local checks are green: exact WASM build, key-package/routing tests 12/12, real/fake conformance 2/2, unmutated plugin lib 54/54, security tests 2/2, two-relay harness 1/1. Fresh required aggregate gates have not emitted yet.

skyl:
Adversarial review at `e4449f4` found three blockers despite green aggregate gates: (1) the change adds fields/purposes while still negotiating protocol v1, violating WIRE v1 compatibility; (2) production and FakeRelay handlers do not call `PublishKeyPackagesRequest::validate()` before mutation, so forbidden multi-last-resort requests succeed; and (3) the responder does not prove the signed outer `conversation_id` equals the joined MLS group ID before persistence, yielding state that cannot reload after restart. A worker is repairing these on the existing PR worktree with fail-before/conformance coverage. Do not merge this head.

skyl:
Exact-head supervision at `9155c979d2be454784013314a9953263a8455588`: independent adversarial re-review found no remaining P0/P1/P2 defect in the prior federation, cross-role deduplication, last-resort rotation, authenticated Welcome, mutation-control, v1-compatibility, request-validation, or outer-conversation/group-id boundaries. This is **not merge-ready yet**: `closingIssuesReferences` is empty / the body has no `Closes #N` or `Refs #N`; branch protection is strict and GitHub reports this head `BEHIND` current `main`; and fresh exact-head `gate` plus `rs / gate` will be required after the live owner merges current `origin/main` and resolves the STATUS overlap. Please add the intended issue linkage and update the owned branch when ready. I have not edited or rebased it.

skyl:
Exact-head merge-resolution review at `1c5af5ac1c21228ec85242815b47d8a4972c97d0`:

- The merge has parents `9155c979d2be454784013314a9953263a8455588` (the independently reviewed feature head) and `4b939558a21dc38164ff2b2fb56263ebc855acf2` (then-current `origin/main`, including #782).
- `git show --remerge-diff` shows the only manual conflict was `wallet/zuuli/STATUS.md`. Its resolution correctly preserves #769s KeyPackage/first-contact evidence and operator-configuration blocker while incorporating #782s faulted-store behavior, `get_engine_status`, and pure `check_handle_eligibility` facts.
- The incoming #782 code cleanly removes the old stateful `Engine::check_handle_eligibility` method in favor of the reviewed stateless command path; the #769 federation, relay-identity, authenticated-Welcome, cross-role deduplication, last-resort rotation, request-validation, v1-compatibility, and outer-conversation/group-id boundaries remain present.
- `git diff --check` is clean. I found no new unresolved P0/P1/P2 defect introduced by this merge resolution, and GitHub reports no unresolved review thread.

This exact head is **not merge-ready yet**. `closingIssuesReferences` is still empty, so the PR needs its intended `Closes #N` or `Refs #N` issue linkage. The prior heads gates/review do not substitute for this merge commit: fresh exact-head `gate` and `rs / gate` must both complete green, and branch protection still reports `REVIEW_REQUIRED`. At the time of this review the aggregate gates had not emitted and the fresh matrix remained in progress. I did not edit or rebase the owned branch.

skyl:
Current-base handoff after #802 merged: remote `main` is now `7bf653b16dcecedb5768f616d47a703aa14bbd38`, so #769 head `1c5af5ac1c21228ec85242815b47d8a4972c97d0` is `BEHIND` and its prior CI cannot authorize a strict-up-to-date merge.

A read-only overlap audit found #802 touches `f2z-relay-testkit` connection/server-shutdown files while #769 touches separate testkit client/config/engine/state/vector files. `git merge-tree 1c5af5ac 7bf653b1` completed cleanly (tree `f4a8b672b2a1e4a1d432be830fe794af03fcd887`) with no conflict. The live owner still needs to merge current `origin/main` into the owned branch, push the new head, and obtain fresh exact-head `gate` plus `rs / gate`; no one else should edit or rebase it.

The process blocker also remains: `closingIssuesReferences` is empty. Please add the intended `Closes #N` or `Refs #N` issue linkage before merge. I made no branch edit.

skyl:
Exact-head adversarial review finding at `d76fdcc7103b2ef049d32970f6fc9ff047bbcb45` — **P1 regression proof remains bypassable**.

`authenticated_request()` constructs `conversation_id: "conversation"`, which is neither hex nor a canonical 32-byte MLS group id. The three `assert_acceptance_refuses` cases only assert that `accept_contact_request` returns `RelayIdentityMismatch`. If the shipping `authenticate_introduction` call is removed, or its body is replaced with `Ok(())`, execution continues into `join_conversation`, where decoding that invalid outer id fails with the same `RelayIdentityMismatch`. The tests therefore stay green without exercising first-contact authentication—the exact mutation they were added to kill.

The helper comment says the refusal is before MLS join/relay I/O, but it does not assert the absence of group/conversation/relay mutation and its fixture cannot reach a valid authenticated join. Use a real canonical group id and Welcome (or an instrumented valid join fixture), assert the mutated transcript fails before any join/persistence/relay side effect, and demonstrate that deleting/bypassing `authenticate_introduction` makes the regression fail. This blocks an exact-head no-P0/P1/P2 verdict.

skyl:
Repaired the exact-head P1 at `55693d95bf16a479f692a9ae9f70833b0632b0cd`.

The acceptance fixture now uses a real MLS Welcome for the recipient KeyPackage and a canonical 32-byte group id. Each stored-transcript mutation is driven through `accept_contact_request`, and the helper proves the authentication refusal leaves the in-memory group map empty, leaves no persisted MLS group or conversation, retains the pending request, and establishes no relay connection.

Mutation evidence: replacing the shipping `authenticate_introduction` call with success made all three focused tests fail (`Internal` / `EngineLocked`, not `RelayIdentityMismatch`); the call was then restored. Unmutated exact-head results: focused authentication tests 3/3; full plugin `--features relay-harness` matrix green (58 unit plus integration/two-relay harness); all-target/all-feature clippy with `-D warnings` green; permission self-test and live check green. Fresh exact-head CI and independent review are required; prior verdicts are stale.

skyl:
Independent adversarial review of exact head 55693d95bf16a479f692a9ae9f70833b0632b0cd

P1 — the swapped-Welcome regression is still downstream-invalid and does not prove the shipping authentication gate prevents a valid hostile MLS group from being joined. authenticated_request now correctly creates a real Welcome and canonical 32-byte group id, but accept_contact_request_refuses_a_swapped_welcome immediately replaces those bytes with hex("attacker Welcome") (wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs:4747-4750). If authenticate_introduction is bypassed, this case fails with Internal while parsing/processing the malformed Welcome. It is red because the expected error class changed, not because a valid attacker Welcome would otherwise reach the forbidden MLS join. The helper claim at lines 4641-4645 that the fixture must reach MLS join after bypass is therefore false for this mutation.

Use a second, valid MLS Welcome addressed to Bobs real KeyPackage, with the same canonical outer group id but created by the attacker/alternate sender, then replace only the Welcome while retaining Alices original signed transcript. With authentication present it must refuse as RelayIdentityMismatch with the existing zero in-memory group, zero persisted group/conversation, pending-request, and zero-connection assertions. With authentication bypassed, that valid hostile Welcome must reach the prohibited join/state (or a fully join-capable acceptance fixture) so the regression is red for the security boundary rather than malformed MLS input.

The route and signature cases do not have the same problem. My isolated exact-head bypass mutation made both fail with EngineLocked only after join_from_welcome_for_group_id committed and the group was inserted into inner.groups; that is meaningful MLS state past the boundary, not a pre-join mask. The unmutated focused tests pass 3/3 and their no-state assertions are correct. The bypass mutation produced 0/3 as reported: swapped Welcome = Internal, route/signature = EngineLocked.

No additional P0/P1/P2 regression found in the latest repair or previously reviewed federation, deduplication, rotation, authenticated routing, request-validation, v1 compatibility, or group-id boundaries. Refs #305 is now present and the branch contains the normal current-main merge. CI remains in progress, so this review does not assert green required gates.

skyl:
Repaired the valid-hostile-Welcome P1 at exact head `74a6333185e7f819d82f0357dc9d1b921c2bf6a0`.

The swapped-Welcome mutation now creates a second valid MLS Welcome from an alternate sender for Bob’s exact real KeyPackage and the same canonical 32-byte group id, then changes only the stored Welcome while retaining Alice’s original signed routing transcript. The no-state assertions run before the error-class assertion.

Explicit mutation evidence: bypassing the shipping `authenticate_introduction` call made all three focused tests fail at `authentication refusal must precede the in-memory MLS join`; the hostile Welcome therefore parsed, joined, and inserted MLS group state rather than failing setup. The production call was restored. Unmutated focused tests pass 3/3; full plugin `--features relay-harness` suite, all-target/all-feature clippy with `-D warnings`, permission negative controls/live parity, and `git diff --check` all pass. Fresh exact-head CI and independent review are required; prior results are stale.

skyl:
Independent adversarial re-review of exact head 74a6333185e7f819d82f0357dc9d1b921c2bf6a0

No P0/P1/P2 finding in the second repair or the previously reviewed boundaries. The prior swapped-Welcome P1 is closed:

- The fixture creates Alice and Mallory as distinct valid MLS senders, verifies Bobs exact real KeyPackage against Bobs DirectoryEntry for both, and creates both groups with the same canonical 32-byte group id. Both resulting Welcome values are real and asserted distinct.
- The stored request retains Alices original device key/signature over Alices Welcome and routing transcript; the mutation replaces only request.welcome with Mallorys valid Welcome. Shipping accept_contact_request refuses it as RelayIdentityMismatch before joining.
- The state assertions now run before the error-code assertion and prove zero in-memory group, zero persisted MLS group, zero conversation, the request still pending, and zero relay connections.
- I independently replaced the shipping authenticate_introduction call with success in an isolated detached worktree. All three focused cases failed at the in-memory-group assertion itself (0/3), including the valid hostile Welcome. This proves each bypass reached the prohibited MLS join and cannot be masked by the later EngineLocked/error-code path. I restored the mutation, verified the scratch tree clean, and removed it.

Exact-head local evidence:

- cargo +1.97.1 test --locked --features relay-harness accept_contact_request_refuses_ -- --nocapture — PASS (3/3)
- Same command with authenticate_introduction bypassed — expected RED (0/3), every failure at “authentication refusal must precede the in-memory MLS join”
- git diff --check 55693d95..74a63331 — PASS

The latest delta is test-only and does not alter the reviewed production federation, cross-role deduplication, last-resort rotation, authenticated routing, request-validation, v1-compatibility, or outer-conversation/group-id behavior. Refs #305 remains present and the normal current-main merge remains in ancestry. GitHub reports the exact head mergeable; fresh CI is still in progress, so this review does not claim required gates are green.

skyl (commented):
**Approve.** Adversarial review on the owner's behalf, focused on correctness and security. I found no defect that would break `main` and no security hole. Notes below are nits, not gates.

### What I actually verified (not just read)

**1. KeyPackage consumption, reuse and exhaustion.** The claim's delete and its response are one transaction in both backends (`sqlite.rs` `claim_key_package`, `memory.rs` `claim_key_package`), and the `Committed` return is load-bearing rather than decorative — a crash cannot resurrect a served package. Oldest-first in both. The last-resort package is deliberately not consumed, and exhaustion is `ERR_UNAVAILABLE`, collapsed with absent-address and not-a-contact-queue, so the published address space stays non-enumerable. The cross-role duplicate rules are the sharp part and they are right: a candidate last-resort that collides with the pool is skipped **without retiring the existing fallback** (`sqlite.rs` checks against `held`, which includes the stored last-resort row; `memory.rs` checks `key_packages` and `last_resort` separately) — so a single-use init key can never be silently promoted to reusable, in either backend, including within a single request.

**2. Backend parity.** I diffed the two implementations rule by rule (dedupe set, clamp-from-end, pool count excluding last-resort, kind check, error codes, idle-timer behaviour) and could not construct an input where they diverge observably. The 7 property tests run against both. `deleting_a_queue_takes_its_pool_with_it` is the one that matters most and it is there: it recreates the queue on the same addresses and asserts an empty pool, which is what actually proves the SQLite `ON DELETE CASCADE` fires (`PRAGMA foreign_keys = ON`, `sqlite.rs:284`) rather than leaving a claimable orphan. Schema 1→2 is additive, upgraded in place, and a newer `user_version` is still a hard refusal.

**3. Replay / stamp reuse.** `CLAIM_KEY_PACKAGE_CHALLENGE_PURPOSE = 3` sits outside `ChallengePurpose::from_code`'s 0..2, so `GET_CHALLENGE` cannot mint one and a `CONTACT_APPEND` stamp cannot be spent on a claim (or vice versa). `Challenges::consume` removes before judging, compares scope unconditionally (#663's fix intact) and is scoped to `contact_addr`. The stamp is checked **before** the pool is consulted, so "nothing here" and "here you are" cost the same. `Config::check` refuses `key_packages_enabled` with zero PoW bits, a zero pool, or no contact queues; the client independently refuses an unmetered claim challenge. No fail-open path in either direction.

**4. The MITM check is structural, and that is the right call.** `VerifiedKeyPackage` has no constructor but the verifying one and `add_member` takes it, so the check cannot be forgotten — only deleted. The check itself is complete: OpenMLS `validate`, ciphersuite pin, credential parse, identity signature, `identity_pk == entry.identity_pk`, handle, `devices` membership, `revocations`, leaf binding, plus trailing-byte rejection on the framing. `last_resort` is read from the package's own signed extension, never the relay's advisory byte — and the code treats the byte as advisory everywhere it appears.

**5. #700/#702 — not reintroduced.** `sign_routing_advert` goes through `DeviceSigner` → `libcrux_ed25519::sign`; nothing here touches `f2z_msg_identity::DeviceSignatureKey::sign_mls_content` or `ed25519-dalek` signing. The signed message is a 32-byte domain-separated digest (`free2z/msg/v1/first-routing-advert` over a `conversation_id`/advert/Welcome tree), which cannot collide with MLS `SignContent` framing, and the DSK's widened scope is stated in `ARCHITECTURE.md` §5.6 in this same PR rather than left as a silent drift. Worth writing down for whoever picks up #702: this is now the one signature in the tree **produced** by libcrux and **verified** by dalek `verify_strict` (`f2z_kt_core::sig::verify`). Ed25519 signing is deterministic and libcrux emits canonical `s`, so there is no operational gap — but the produce/verify pair is split across the two cores, and #702's eventual fix should account for it.

**6. Docs vs code.** Spot-checked the claims that carry weight — §12.6's exhaustion behaviour, the one-code collapse, the owner-only pool report, the advert signature construction in §12.5, ADR 0015's rejected alternatives — and the code does what the prose says.

### Nits (mention only; none blocking)

- `src/store.rs:718-721` — `advert_device_pk` / `advert_signature` are the only new persisted fields without `#[serde(default)]`, while the sibling `peer_relay_id` has it. A store holding a pending contact request written by an older build will fail to deserialize. It fails closed and is unreachable in practice (nobody could complete `start_conversation` before this PR), so it is genuinely cosmetic — but the asymmetry inside one struct will read as an oversight later.
- `src/engine.rs` `last_resort_rotation_due_for` returns true whenever the relay reports `has_last_resort == 0`, and the receive pump consults it every tick. A relay that always answers 0 makes the client regenerate and republish a last-resort package on a 5-second loop, growing the local MLS key store without bound. Self-inflicted only, from a relay you already trust with your queue — but a backoff would be about four lines.
- `PublishKeyPackagesRequest`'s "a retry does not double the pool" holds for a byte-identical resend; `ensure_key_packages_at` regenerates fresh packages on each attempt, so a lost response does add packages (bounded by `max_pool_size`). Harmless — the doc just reads stronger than the client's behaviour.
- `sqlite.rs` `next_seq` uses `.unwrap_or(0)` on the `MAX(seq)` query. A swallowed error there would collide with an existing primary key and surface as an INSERT failure rather than corruption, so it is cosmetic, but it is the one place in this diff where an error is discarded rather than propagated.

Safe to squash-merge.

Issue #770, opened 2026-08-28
DECISION NEEDED: the shipped witness set and threshold t — the last thing blocking a working conversation


skyl:
**This is the single thing between a running ZUULI and two people holding a conversation. It is a decision, not an implementation.**

## The state

Every piece of code is built and merged. `wallet/plugins/tauri-plugin-f2zmsg` nonetheless ships `directory::NoDirectory`, so `resolve_handle`, `start_conversation` and `accept_contact_request` fail closed with `witness-threshold-unmet`.

That is **correct behaviour**, not a bug. `f2z-kt-client` refuses a root that does not carry at least *t* cosignatures from the client's **own configured** witness set, and `WitnessSet` deliberately has **no `Default`** — because `KT.md` §12 has decided neither the shipped list nor *t*. Per the owner decision on #311, proceeding silently would overclaim and must not be the default.

## What has to be decided

1. **The log's identity** — `log_id` and signing public key that clients pin.
2. **The shipped witness list** — which witness public keys a fresh install trusts.
3. **The threshold *t*.**
4. **What the UI says when the standing is what it currently is.**

## The honest problem with (2) and (3)

**free2z would operate the log *and* the only witness.** That provides **zero** independent anti-equivocation — a party that can sign tree heads with the log key can also sign cosignatures with the witness key. `ARCHITECTURE.md` §9.3 already states this plainly:

> *witnesses we operate are not independent witnesses. Until there are at least two run by parties outside free2z, `t` is effectively 0 and client gossip is the only anti-equivocation that exists. The UI must say so rather than display a reassuring witness count.*

The client is already built to honour that: `WitnessStanding::independent()` is deliberately short and `counted_including_dependent()` deliberately long, and `is_independently_witnessed()` returns `false` in the shipped configuration **on purpose**. So the code will not let the UI quietly claim more than is true — but somebody still has to choose what it does claim.

## The options, none free

- **Ship with `t = 1` and free2z's own witness**, with the UI stating in plain words that no independent party has verified this directory. Honest, works today, and buys no anti-equivocation.
- **Ship with `t = 0`** and rely on client gossip alone, which is what §9.3 says is actually operative in this configuration. Most honest, least reassuring.
- **Recruit independent witnesses first.** The property becomes real at exactly the moment two operators outside free2z exist. The witness role was deliberately made cheap for this — no inbound port, no TLS, no domain, no database, state is one file — but it is a social task with no code left to write.

[...]

PR #774, opened 2026-08-28
feat(zuuli): add fail-closed first-contact messaging UI

### LINKED ISSUE #771: ZUULI: ship the first-contact messaging UI
Child of #305 and #312.

## Problem

The native messaging engine, typed frontend bridge, event surface, and mock already expose `start_conversation`, `list_contact_requests`, `accept_contact_request`, and `reject_contact_request`, but ZUULI never calls or renders them. An enrolled user can use an already-established conversation, but cannot start the first one; inbound contact requests are invisible, including the contract's `first-contact` mock scenario.

This frontend slice is implementation-unblocked. Live first contact still depends separately on #769 (KeyPackage publication) and #770 (the shipped witness set and threshold); the UI must preserve that fail-closed boundary.

## Scope

- Add a focused first-contact component to the existing Messages feature.
- Initiate a conversation by an exact messaging handle.
- List inbound contact requests and accept, reject, or block locally.
- Re-read authoritative state after first-contact events instead of treating event payloads as state.
- Select a newly created/accepted conversation.
- Show proof-of-work as device computation, not generic network waiting.
- Preserve exact handle semantics: `/^[a-z0-9_]{1,30}$/`, byte-for-byte, with no trim, case-fold, Unicode normalization, or `@` stripping before invocation.
- Preserve fail-closed witness behavior: below threshold, both outbound initiation and inbound acceptance are unavailable; established conversations continue unaffected.

## Acceptance criteria

- [ ] An enrolled user with a met witness threshold can start a conversation; the exact handle string is passed once and the returned conversation is selected.
- [ ] Invalid, uppercase, whitespace-padded, `@`-prefixed, non-ASCII, empty, and over-length handles never invoke the bridge.
- [ ] First-contact work is single-flight and visibly described as proof-of-work computed on this device.
- [ ] Below the witness threshold, initiation and acceptance are disabled with truthful copy; no fallback lookup or silent proceed exists.
- [ ] Pending requests show the peer handle and identity fingerprint. Accept selects the returned conversation. Reject calls `block: false`; the separate block action calls `block: true` and says “on this device.”
- [ ] `f2zmsg://contact-request` and `f2zmsg://conversation-updated` cause a re-read of conversations/requests; their payload is never appended directly. Late listener setup is cleaned up after unmount.
- [ ] Bridge error codes are visible and do not erase the existing conversation/request state.
- [ ] Component tests include mutation-sensitive negative assertions for the fail-closed, exact-handle, single-flight, authoritative-re-read, and block-local boundaries.
- [ ] Frontend tests, test typecheck, production typecheck/build, and relevant messaging contract/authority guards pass.

## Out of scope

No Rust/plugin/bridge/type/contract changes; no witness-set decision; no retry/cancel, read-receipt, gap-repair, retention, or safety-number UI in this PR.

Refs #305
Refs #312

skyl:
## Summary

- add ZUULI's missing first-contact surface for exact-handle initiation and inbound requests
- keep initiation and acceptance fail-closed below the independent witness threshold while established chats and local reject/block remain available
- re-read authoritative conversation/request state after messaging events, with late-listener cleanup
- make proof-of-work, identity fingerprints, and device-local blocking explicit in the UI

## Tests

- `npx vitest run src/features/messages/FirstContact.test.tsx` — 17 passed
- `npx vitest run` — 837 passed, 5 skipped
- `npm run typecheck` — passed
- `npm run typecheck:tests` — passed
- `npm run build` — passed, including fresh WASM build/dist verification
- `node --test scripts/messaging-contract.node-test.mjs scripts/mobile-webview-authority.node-test.mjs` — 14 passed
- `node scripts/mobile-webview-authority.mjs` — passed
- five mutation probes each turned the focused suite red: witness gate, exact-handle rewrite, single-flight guard, authoritative event re-read, and block-local boolean

## Boundaries

Frontend only: no Rust, bridge, type, capability, or contract changes. Live first contact still depends on #769 and the owner witness decision in #770; this PR does not bypass or simulate either.

Closes #771
Refs #305
Refs #312

### DIFF
--- a/wallet/zuuli/src/features/messages/FirstContact.test.tsx
+++ b/wallet/zuuli/src/features/messages/FirstContact.test.tsx
@@ -1,0 +1,301 @@
+import { act } from "react";
+import type { Root } from "react-dom/client";
+import { Simulate } from "react-dom/test-utils";
+import { parseHTML } from "linkedom";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type {
+  ContactRequest,
+  Conversation,
+} from "@/lib/messaging/types";
+
+const controls = vi.hoisted(() => ({
+  startConversation: vi.fn(),
+  listContactRequests: vi.fn(),
+  acceptContactRequest: vi.fn(),
+  rejectContactRequest: vi.fn(),
+  listen: vi.fn(),
+}));
+
+vi.mock("@/lib/messaging/bridge", () => ({
+  messaging: {
+    startConversation: controls.startConversation,
+    listContactRequests: controls.listContactRequests,
+    acceptContactRequest: controls.acceptContactRequest,
+    rejectContactRequest: controls.rejectContactRequest,
+  },
+}));
+vi.mock("@/lib/messaging/events", () => ({
+  listenMessaging: controls.listen,
+}));
+
+import { FirstContact } from "./FirstContact";
+
+const REQUEST: ContactRequest = {
+  requestId: "req-1",
+  peerHandle: "newcomer",
+  peerIdentityFingerprint: "2B8F 60C1 D4A7 39E5",
+  receivedAt: 1,
+  bodyPreview: null,
+};
+
+const CONVERSATION: Conversation = {
+  conversationId: "conv-newcomer",
+  peerHandle: "newcomer",
+  peerIdentityFingerprint: REQUEST.peerIdentityFingerprint,
+  verification: { state: "unverified" },
+  epoch: 1,
+  createdAt: 1,
+  lastMessageAt: null,
+  unreadCount: 0,
+  retention: {
+    scope: "global",
+    mode: "keep",
+    ttlSeconds: null,
+    effectiveFrom: 1,
+  },
+  ephemeralHint: null,
+  receiptPolicy: { deliveryReceipts: true, readReceipts: false },
+  hasGaps: false,
+  transportHealth: "ok",
+};
+
+let root: Root;
+let container: HTMLElement;
+let restoreGlobals: () => void;
+let listeners: Map<string, () => void>;
+let unlisteners: Map<string, ReturnType<typeof vi.fn>>;
+
+async function installDom() {
+  const { window, document } = parseHTML(
+    "<!doctype html><html><body><div id='root'></div></body></html>",
+  );
+  const saved = new Map<string, PropertyDescriptor | undefined>();
+  for (const [name, value] of Object.entries({
+    window,
+    document,
+    navigator: window.navigator,
+    HTMLElement: window.HTMLElement,
+    Event: window.Event,
+    IS_REACT_ACT_ENVIRONMENT: true,
+  })) {
+    saved.set(name, Object.getOwnPropertyDescriptor(globalThis, name));
+    Object.defineProperty(globalThis, name, {
+      configurable: true,
+      writable: true,
+      value,
+    });
+  }
+  restoreGlobals = () => {
+    for (const [name, descriptor] of saved) {
+      if (descriptor) Object.defineProperty(globalThis, name, descriptor);
+      else delete (globalThis as Record<string, unknown>)[name];
+    }
+  };
+  container = document.getElementById("root") as unknown as HTMLElement;
+  const { createRoot } = await import("react-dom/client");
+  root = createRoot(container);
+}
+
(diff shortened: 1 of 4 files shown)


skyl (commented):
Adversarial UI/security review: changes requested. I did not approve.

1. **Blocking — the “authoritative reread” path has both an attach gap and stale-response races.** The initial request read starts in the first effect (`FirstContact.tsx:66-68`), while both listeners attach asynchronously in the next effect (`73-86`). A contact request committed after the initial read but before either promise installs its handler is missed until a later focus/visibility transition. The late-listener test proves cleanup only; it never emits in this window. After attachment, every event/focus/action starts an unconstrained read (`46-64, 84-90, 113-115`). An older slow `listContactRequests` can resolve after a newer read and overwrite the newer authoritative result, resurrecting an accepted/rejected request or hiding a new one. Subscribe first and perform a post-subscription reconciliation, then serialize/coalesce reads or guard commits with a generation. Add tests whose promises resolve out of order and whose event lands during listener attachment.

2. **Blocking honesty boundary — the card presents an attacker-claimed, production-unformatted key as “Identity fingerprint.”** Before acceptance, the contact envelope's handle and `identity_pk` have not been confirmed by the directory; `accept_contact_request` is the operation that performs that check. Yet the UI labels the value simply “Identity fingerprint” (`236-244`) with no “claimed/unverified” qualification. In production, `list_contact_requests` returns the stored raw identity-key hex unchanged (`engine.rs:667-679, 3111-3115`), while established-conversation fingerprints go through the five-character human grouping helper. The test fixture supplies an already pretty, four-character-grouped string (`FirstContact.test.tsx:33-38`), so it passes while testing a shape production never returns. Present this as the request's unverified claim pending directory confirmation and format the actual 64-hex production value consistently for human comparison; test raw production input.

3. **Blocking error semantics — every refusal is flattened to one generic raw-code callout.** `FirstContact.tsx:22-25, 219-223` renders `directory-proof-invalid` (fork evidence), `witness-threshold-unmet`, `directory-protocol-violation`, retryable network/PoW failures, and `internal` under the same “First contact was refused / Messaging error” treatment. That contradicts CLIENT-CONTRACT §8's required distinctions: fork evidence is fatal and not a network glitch, protocol/internal errors need a report affordance, witness failure must explain fail-closed/manual verification, and retryable failures have different actions. Successful background rereads also never clear a prior load error, leaving a stale refusal visible after recovery. Render the contract's security-relevant meanings/actions and clear or supersede stale errors from the operation that produced them.

Accessibility follow-up: the multi-second PoW state changes only button text/spinner (`188-198`) without a live status/`aria-busy`, so assistive technology is not guaranteed to announce that computation started or ended. Please cover the progress state with an accessible status test.

Focused suite passes 17/17, but none of the tests can fail for the attach gap, out-of-order read overwrite, real raw fingerprint shape, fatal-vs-retryable error collapse, or accessible status announcement.

skyl:
Addressed the blocking adversarial review in `ee23db7`.

Fail-before evidence: the new tests were first run against the prior implementation and produced failures for the attach window, stale request/conversation reads, raw production key shape, four contract-distinct error states, stale load-error recovery, and PoW accessibility.

Fix-after:
- both messaging listeners attach before the initial authoritative reconciliation; contact-request and parent conversation commits are generation-guarded, so older responses cannot resurrect stale state
- pending requests now say the key is an unverified request claim pending directory confirmation, and real 64-hex production input is rendered in the backend's five-character human groups
- fork evidence, unmet witness threshold, protocol/internal defects, and retryable failures have distinct honest actions while retaining the raw code; successful source reads clear only their own stale load failure
- multi-second local PoW exposes `role=status`, `aria-live=polite`, and form `aria-busy`

Verification on the restored fix:
- focused Vitest: 26/26 passed
- full Vitest: 846 passed, 5 skipped across 63 files
- production and test TypeScript typechecks passed
- production build/WASM binding passed (116 bytes, SHA-256 `94770ca6e67ceb6f0c6998139f3b31c5e79bdb9cd843f0ee7fa2de78d1039819`)
- messaging/mobile authority guards: 14/14 passed; live mobile authority check passed
- six independent mutation probes all failed at the intended boundary: post-subscription reread, request generation, conversation generation, raw-key grouping, security error presentation/stale clearing, and PoW live status

CI is rerunning on this head; I am monitoring it through completion. No merge requested.

PR #780, opened 2026-08-30
fix(zuuli): make Android artifact stdout explicit

### LINKED ISSUE #741: release(android): make every android-release-artifact.sh subcommand guarantee its own stdout, not its caller
#738 fixed one instance of this and proved the class exists. `wallet/zuuli/scripts/android-release-artifact.sh` has three subcommands, and each one's stdout is a contract with whoever captures it — but only one of them guarantees anything about it.

That was not a style problem. `seal_verifier` letting `sha256sum -c` write to stdout is what stopped `0.1.0+15` from reaching the Play internal track.

## What is left, measured during #738

| site | what leaks to stdout | captured today? |
|---|---|---|
| `verify` `:101` | `sha256sum -c` prints `<member>: OK` per member — four lines | **no caller captures `verify`** anywhere in the repo; the node tests read its exit status |
| `verify` `:105` | `cmp` prints its diff to stdout | same |
| `validate_archive` `:25` | `grep -E '<unsafe pattern>' "$listing"` prints the offending member names | reached from `record` and `verify`, but only on the failure path, where the function returns 1 and no digest is printed — so it cannot corrupt a captured value today |
| `record` | nothing — already exactly one digest | yes, at `zuuli-release.yml:288`, with a `>/dev/null` the caller supplies |

None of these is currently an outage. All of them are the same latent shape: **the caller is responsible for a property the function should guarantee.** `record` survived `0.1.0+15` only because its call site happens to redirect; that redirect is load-bearing and nothing says so.

## Suggested

- `verify`: `sha256sum -c … >/dev/null` and `cmp -s`, keeping exit status and stderr. Its chatty output is genuinely useful when a human runs it by hand, so if that is worth keeping, make it explicit — a `--verbose` flag, or print the detail to stderr where a captured stdout cannot pick it up.
- `validate_archive:25`: `grep -qE`. The human-readable message on the next line already goes to stderr.
- One test per subcommand asserting stdout is exactly what that subcommand promises, in the shape #738 added for `record` and `seal-verifier`. That is the part that stops this recurring.

## A correction to #738 while I am here

#738 said `>/dev/null` was preferable to `--status` because "a genuine mismatch should still say which member failed on stderr". That is wrong on this platform, and the fix's author measured it rather than taking my word: `sha256sum -c` writes the per-member `FAILED` line to **stdout** too, so `>/dev/null` silences it. What survives on stderr is `sha256sum: WARNING: N computed checksum did NOT match`, plus errors for unreadable members. The redirect is still the right choice — it keeps that warning, which `--status` would also suppress — but the reason I gave for it was not accurate.

skyl:
## Summary
- define stdout contracts for every Android release artifact subcommand
- silence archive-validation, checksum, and inventory comparison status output while retaining stderr diagnostics
- add regression coverage for success output and checksum/archive/inventory failure streams

## Tests
- `node --test scripts/android-release-artifact.node-test.mjs`
- `shellcheck wallet/zuuli/scripts/android-release-artifact.sh`
- `ZUULI_PW_PORT=1441 npm test` (846 Vitest passed; 370 Node passed, 5 skipped; 133 Playwright passed)
- `git diff --check`

Closes #741

### DIFF
--- a/wallet/zuuli/scripts/android-release-artifact.node-test.mjs
+++ b/wallet/zuuli/scripts/android-release-artifact.node-test.mjs
@@ -33,44 +33,93 @@ async function fixture() {
   return { root, aab, output, verifier, verifierSha };
 }
 
-test("records and verifies a checksum-closed four-ABI AAB", async (t) => {
+test("record, seal-verifier, and verify honor their success stdout contracts", async (t) => {
   const { root, aab, output, verifier, verifierSha } = await fixture();
   t.after(() => rm(root, { recursive: true, force: true }));
+
+  // Release workflows capture digest-producing commands in shell variables,
+  // while verify is status-only. Each helper owns that boundary itself.
   const recorded = run(["record", aab, "0.1.0+14", "a".repeat(40), "b".repeat(40), output]);
   assert.equal(recorded.status, 0, recorded.stderr);
+  assert.match(recorded.stdout, /^[0-9a-f]{64}\n$/);
   const sealed = run(["seal-verifier", output, verifier, verifierSha]);
   assert.equal(sealed.status, 0, sealed.stderr);
-  assert.equal(run(["verify", output]).status, 0);
+  assert.match(sealed.stdout, /^[0-9a-f]{64}\n$/);
+  const verified = run(["verify", output]);
+  assert.equal(verified.status, 0, verified.stderr);
+  assert.equal(verified.stdout, "");
   const source = JSON.parse(await readFile(join(output, "source-record.json"), "utf8"));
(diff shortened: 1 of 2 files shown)


skyl (commented):
Adversarial review: no blocking findings. The helper now owns each stdout contract; unsafe-name, checksum, and inventory failure cases prove stdout remains empty while actionable diagnostics remain on stderr. The focused suite, shellcheck, diff check, and full local ZUULI suite are green. Hold for both required aggregate gates and the packaging smoke matrix before squash-merge.

PR #781, opened 2026-08-30
fix(mls): restore group after receive rollback

### LINKED ISSUE #699: MLS engine: a transaction rollback after merge_staged_commit leaves the in-memory group ahead of the store, and the redelivery is then refused
`f2z-msg-store`'s `Transaction` exists to stop exactly one thing, and its module note says so:

> a crash in the middle leaves a group whose tree is from one epoch and whose secrets are from another.

The rollback protects the **store**. It does not protect the **in-memory `MlsGroup`**, and `MlsEngine::receive` merges into memory before the transaction commits. When the transaction is rolled back after a merge, the caller is left holding a group one epoch ahead of everything that was persisted — and, because no handled record was written, the relay correctly redelivers the very commit that group can no longer apply.

## Demonstration

`rs/crates/f2z-msg-mls/tests/` on #697's head (`e8b6ccb`), a `StorageBackend` that wraps `MemoryBackend` and can be armed to fail its `apply`:

```rust
impl StorageBackend for FailableBackend {
    fn get(&self, key: &[u8]) -> f2z_msg_store::Result<Option<Vec<u8>>> { self.inner.get(key) }
    fn apply(&self, ops: &[Op]) -> f2z_msg_store::Result<()> {
        if self.armed.load(Ordering::SeqCst) {
            return Err(StoreError::Backend("injected disk failure"));
        }
        self.inner.apply(ops)
    }
    fn durability(&self) -> Durability { self.inner.durability() }
}
```

Alice and Bob pair exactly as `two_instances.rs::paired()` does, except Bob's engine is on `FailableBackend`. Alice issues an `update()`. Bob's disk is armed to fail at the instant `Transaction::commit` applies the write set — after `merge_staged_commit` has already mutated the in-memory group:

```
in-memory epoch after the refused commit: GroupEpoch(2)
redelivery outcome: Err(OutOfOrder)

assertion `left == right` failed: a refused commit must not have advanced the epoch
  left:  GroupEpoch(2)
  right: GroupEpoch(1)
```

Three facts in that output:

1. `receive` returned `Err`. Nothing was written, no handled record exists, so no `ACK` is sent and the relay keeps the message. That part is correct and is what the transaction is for.
2. `bob_group.epoch()` advanced from 1 to 2 anyway.
3. The relay redelivers the commit — with the disk healthy again — and Bob's live process refuses it as `OutOfOrder`. **Permanently.** That group can never apply that commit again in this process, and its epoch-2 state will never be persisted by this path.

The assertion in (2) is not mine. It is `two_instances.rs::an_out_of_order_commit_is_refused_and_leaves_the_group_usable`, verbatim:

```rust
assert_eq!(bob_group.epoch(), epoch_before,
           "a refused commit must not have advanced the epoch");
```

It holds for a refusal raised **before** the merge, which is the only kind any test in the PR exercises. It does not hold for a refusal raised after it.

## The route that needs no disk error

`engine.rs:437-441`:

```rust
group.merge_staged_commit(&self.provider, *staged)?;
// Re-validated after the merge, because the commit may have
// *added* members whose credentials nobody has looked at.
self.validate_members(group, now_ms)?;
```

That `?` lands in the same place: the in-memory group is merged, `transaction` drops unconsumed, the store rolls back. No I/O failure required.

`validate_members` walks **every** member and calls `validate_for_leaf(key, now_ms)`, which enforces the credential validity window. The sender check at `:411` covers only the sender. So a group containing one member whose `DeviceCredential` has expired wedges every other member the moment anybody commits:

- application messages keep flowing (`validate_members` is only called on the `StagedCommitMessage` branch),
- every commit is merged into memory and then rolled back out of the store,
- the redelivery is refused as `OutOfOrder`,
- and the group silently stops advancing while appearing to work.

That is a clock, not an attacker. Credentials expire. I ran the injected-backend route; the credential route I am reporting from inspection of the same six lines, not from an executed test.

On the disk route: `SqliteBackend::apply` can return `SQLITE_BUSY` from a WAL database under concurrent access on a mobile device, which is considerably more ordinary than "disk full".

## Why this is the crate's own invariant

The PR's argument for `Transaction` is that the store must never hold a tree from one epoch and secrets from another. The error path produces the *process-level* version of that same split, and the process is still running. If the caller continues to use the group — and nothing in the API, docs, or return type says not to — the next successful `receive` writes epoch-2-derived state into a store whose tree is at epoch 1, through a transaction that commits cleanly.

## Suggested fix

The engine knows the group is unusable; the caller does not. Options, roughly in order of how much I like them:

1. **Reload the group from the store on the error path** before returning, so the in-memory object matches what was persisted. This is the only fix that leaves the caller's existing group handle correct, which is what every current test assumes.
2. **Poison the handle** — return a distinguished error (`EngineError::GroupMustBeReloaded`) and have the engine record that this `GroupId` is not usable until reloaded, refusing further operations on it. Honest, but pushes work onto every caller.
3. **Validate before merging.** For the credential route specifically, `StagedCommit` exposes the added leaves; checking them before `merge_staged_commit` removes that route entirely and is worth doing regardless of 1 or 2, because it turns a wedge into a clean refusal.

Whatever is chosen, the existing assertion deserves a sibling: a test that raises the failure **after** the merge and makes the same "must not have advanced the epoch" claim. The current suite has four "refused and leaves the group usable" tests and all four refuse before the merge, which is why this reads as covered.

Filed while reviewing #697. My scratch test is not in the PR; it is reproduced in full above.

skyl:
Closes #699

## Summary
- reload the durable pre-transaction MLS group into the caller handle after a staged or pending commit merge is rolled back
- preserve ephemeral AAD while restoring all durable group state
- prove injected StorageBackend apply failure restores the epoch and allows same-record redelivery
- prove post-merge credential validation failure can be redelivered without becoming OutOfOrder

## Verification
- cargo test -p f2z-msg-mls
- cargo test --locked --all-targets
- scripts/check-rust-fmt.sh --root rs
- scripts/check-rust-clippy.sh --root rs

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -615,15 +615,28 @@ jobs:
 
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
-          workspaces: wallet/plugins/tauri-plugin-f2zmsg
+          workspaces: |
+            wallet/plugins/tauri-plugin-f2zmsg
+            rs
           key: zuuli-msg-plugin
 
-      # The acceptance criterion. `--features relay-harness` builds the
-      # `f2zmsg-peer` binary the integration test spawns twice against a real
-      # `ws://` listener; without it the test compiles to nothing and passes by
-      # not existing.
+      # The client crate cannot link the AGPL relay daemon. Build its actual
+      # production binary separately so the rollback/cursor regression can
+      # spawn it across that licence boundary rather than substituting the MIT
+      # FakeRelay behind a WebSocket.
+      - name: Build the production relay daemon for integration tests
+        run: |
+          cargo build --locked --manifest-path rs/Cargo.toml \
+            -p f2z-relay --bin f2z-relay
+
+      # The acceptance criteria. `--features relay-harness` builds the
+      # `f2zmsg-peer` binary for the two-process exchange and enables the exact
+      # two-record rollback regression against the production relay binary.
+      # Without it both integration tests compile to nothing and pass by not
+      # existing.
       - name: Build and test the messaging plugin
         env:
+          F2Z_RELAY_BIN: ${{ github.workspace }}/rs/target/debug/f2z-relay
           TAURI_PERMISSION_GENERATION_NONCE: ${{ github.run_id }}-${{ github.run_attempt }}
         run: |
           cargo test --locked --all-targets --features relay-harness \
--- a/rs/crates/f2z-msg-mls/src/engine.rs
+++ b/rs/crates/f2z-msg-mls/src/engine.rs
@@ -22,10 +22,13 @@
 //! relay, un-acknowledged, and is redelivered — which is recoverable, and a
 //! half-applied epoch change is not.
 //!
-//! Rolling back storage is only half of that invariant after OpenMLS has merged
-//! a commit into the caller's [`MlsGroup`]. [`MlsEngine::receive`] reloads the
-//! durable pre-transaction group into that same handle before returning a
-//! post-merge error, so redelivery sees the epoch the store actually holds.
+//! Rolling back storage is only half of that invariant after OpenMLS has
+//! advanced a private-message receive ratchet or merged a commit into the
+//! caller's [`MlsGroup`]. [`MlsEngine::receive`] reloads the durable
+//! pre-transaction group into that same handle before returning an error, so
+//! redelivery sees the state the store actually holds. If that reload itself
+//! fails, [`EngineError::GroupStateUnavailable`] tells the caller to discard
+//! the possibly advanced handle and attempt a fresh durable load.
 //!
 //! # Delete-on-ack, and what `record_key` is for
 //!
@@ -683,7 +686,10 @@ impl<B: StorageBackend> MlsEngine<B> {
     /// [`EngineError::OutOfOrder`] for a message from an epoch this group is
     /// not in; [`EngineError::Credential`] if the sender's credential does not
     /// validate; [`EngineError::Mls`] if OpenMLS refused;
-    /// [`EngineError::Storage`] if the store did.
+    /// [`EngineError::Storage`] if the store did; or
+    /// [`EngineError::GroupStateUnavailable`] if rollback succeeded but its
+    /// durable state could not be reloaded. After that last outcome the caller
+    /// must discard `group` and load a fresh handle from durable storage.
     pub fn receive(
         &self,
         group: &mut MlsGroup,
@@ -702,9 +708,12 @@ impl<B: StorageBackend> MlsEngine<B> {
         let group_id = group.group_id().clone();
         let original_aad = group.aad().to_vec();
         let transaction = self.provider.store().begin()?;
-        let mut restore_group_on_error = false;
 
         let operation = || -> Result<Received> {
+            // Private-message processing advances the receive ratchet before
+            // content dispatch. Restoration therefore has to cover every
+            // failure from this call onward, not only the visibly mutating
+            // commit branches below.
             let processed = group
                 .process_message(&self.provider, protocol_message)
                 .map_err(map_process_error)?;
@@ -744,9 +753,6 @@ impl<B: StorageBackend> MlsEngine<B> {
                     Received::ProposalQueued
                 }
                 ProcessedMessageContent::StagedCommitMessage(staged) => {
-                    // Mark before the merge: OpenMLS is allowed to mutate the
-                    // caller's group before returning an error.
-                    restore_group_on_error = true;
                     group
                         .merge_staged_commit(&self.provider, *staged)
                         .map_err(|_| EngineError::Mls("merge staged commit"))?;
@@ -779,7 +785,6 @@ impl<B: StorageBackend> MlsEngine<B> {
                     // only correct action when a pending commit exists, and doing
                     // nothing is the only correct action when none does.
                     if group.pending_commit().is_some() {
-                        restore_group_on_error = true;
                         group
                             .merge_pending_commit(&self.provider)
                             .map_err(|_| EngineError::Mls("merge pending commit"))?;
@@ -805,8 +810,11 @@ impl<B: StorageBackend> MlsEngine<B> {
 
         let result = operation();
         if let Err(error) = result {
-            if restore_group_on_error {
-                self.restore_group_after_rollback(group, &group_id, original_aad)?;
+            if let Err(reload) = self.restore_group_after_rollback(group, &group_id, original_aad) {
+                return Err(EngineError::GroupStateUnavailable {
+                    operation: Box::new(error),
+                    reload: Box::new(reload),
+                });
             }
             return Err(error);
         }
--- a/rs/crates/f2z-msg-mls/src/engine.rs
+++ b/rs/crates/f2z-msg-mls/src/engine.rs
@@ -22,6 +22,11 @@
 //! relay, un-acknowledged, and is redelivered — which is recoverable, and a
 //! half-applied epoch change is not.
 //!
+//! Rolling back storage is only half of that invariant after OpenMLS has merged
+//! a commit into the caller's [`MlsGroup`]. [`MlsEngine::receive`] reloads the
+//! durable pre-transaction group into that same handle before returning a
+//! post-merge error, so redelivery sees the epoch the store actually holds.
+//!
 //! # Delete-on-ack, and what `record_key` is for
 //!
 //! [`MlsEngine::receive`] takes a `record_key`: the caller's identifier for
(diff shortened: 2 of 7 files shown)


skyl:
Adversarial review at `cde7857`: do not merge yet. Two P1 blockers remain. First, if durable reload itself fails after an in-memory merge/store rollback, `restore_group_after_rollback(...)?` masks the triggering error and returns with the caller handle still advanced; ZUULI then reinserts that handle, so later redelivery can still wedge. Add a distinguished unusable/must-reload outcome, keep the mutated handle out of the live map until a fresh durable load succeeds, and cover apply failure + one-shot restore `get` failure + successful fresh-load redelivery. Second, restoration is armed only in commit branches, but `process_message` advances private-message receive state before content dispatch; application/proposal handling followed by handled-record/apply failure can leave memory ahead of storage too. Arm restoration before processing can mutate and add an application-message apply-failure/redelivery regression. Also add non-empty AAD assertions so removal of the AAD preservation cannot stay green. Required `rs / gate` is green, but correctness blocks merge regardless.

skyl:
Adversarial review at new exact head `bb3fc199e56b54a10b490dd1847356af78368d10`: **P1, do not merge.** The lower-level changes now distinguish an unrecoverable rollback reload, cover application-message receive-ratchet restoration, and retain non-empty AAD. But the production batch caller does not actually keep the failed ciphertext at the queue head. `pump_conversation` continues its `for queued in response.messages` loop after `apply_inbound` returns `GroupStateUnavailable`. That first failure leaves `stored.next_index` unchanged and removes the group; the immediately following queued record calls `reload_group`, and if storage has recovered it can succeed or return an ordinary error whose path calls `advance(stored, index)`. `acknowledgeable` prevents cumulative relay deletion, but the next READ starts from the now-advanced local cursor, so the original failed ciphertext is never requested again. The comment promising retry on a later pass is false for a multi-message READ (`READ_BATCH=32`). Break/fail the batch on this distinguished outcome without advancing the failed index, and add a plugin/production-caller regression with at least two queued records proving one-shot apply + restore-read failure, successful fresh durable reload, and exact first-record redelivery before the later record. Lower-level manual fresh-load coverage alone cannot catch this caller bug. I have not edited or rebased the owned branch.

skyl:
Independent adversarial review of exact head c6add44a9bd9dfb7ad1b0c7ce2452b4648786f2a

P1 — the required real-relay two-record regression is still absent. The new test is titled and documented as exercising a real relay, but it imports f2z_relay_testkit::fake::FakeRelay and constructs FakeRelay::with_defaults() (wallet/plugins/tauri-plugin-f2zmsg/tests/two_record_rollback_over_a_relay.rs:21,278). Putting that fake behind listen_loopback() proves the plugin path over a real WebSocket, not against the production f2z-relay engine. The repository makes this distinction explicit: rs/crates/f2z-relay-testkit/README.md:140 says “f2z-fakerelay is not a relay”; its conformance documentation distinguishes the FakeRelay WebSocket listener from “the real relay”; and rs/crates/f2z-relay/tests/conformance.rs:101-117 separately constructs FakeRelay and f2z_relay::server::Server.

This is material because the fake and daemon have distinct server/listener, persistence, queue/group-commit, and ACK implementations. Protocol conformance does not prove this exact failed-first/two-record production-client interaction through the real daemon. Please add the requested regression against the actual f2z-relay server/daemon while still exercising the plugin shipping pump path. If linking the AGPL daemon into the plugin test would violate the intended license boundary, place the cross-crate integration test where that dependency is legal or spawn the daemon binary; a FakeRelay WebSocket test cannot be labeled or accepted as real-relay evidence.

The implementation logic itself looked correct in this scope: GroupStateUnavailable maps to AbortBatch, pump_conversation records the failed index and breaks before applying the later record, and cumulative ACK remains bounded by the successful prefix. No additional P0/P1/P2 findings. Local evidence at this exact head:

- cargo +1.97.1 test --locked --features relay-harness --test two_record_rollback_over_a_relay — PASS (1/1)
- cargo +1.97.1 test --locked --features relay-harness --lib engine::tests:: — PASS (6/6 ACK-prefix tests)

CI is still in progress, so this review does not assert a green required gate.

skyl:
Independent adversarial re-review of exact head af46b40b23ff3dcefd37442dae985d98f474f96c

No P0/P1/P2 finding. The prior real-relay P1 is closed:

- The regression contains no FakeRelay import or construction. It spawns the actual f2z-relay executable as a separate process, connects through the plugins shipping WebSocket client, selects --store sqlite with a fresh file, and sets F2Z_RELAY_COMMIT_WINDOW_MS=1. Server::start routes that configuration through SqliteStore and the production group-commit writer.
- CI builds exactly -p f2z-relay --bin f2z-relay from this checked-out rs/Cargo.toml, then supplies the absolute ${{ github.workspace }}/rs/target/debug/f2z-relay path through F2Z_RELAY_BIN. The test executes that path directly rather than resolving PATH; the child removes the harness-only selector because the daemon rejects unknown F2Z_RELAY_* keys. The workflow policy now requires this exact environment binding.
- Two application records are appended before the injected apply + restore-read failure. The first failed pump emits zero and leaves the local transcript empty. The second pump then redelivers exactly two messages with the original first/second msg ids and texts in order. That combined evidence proves no later record advanced the cumulative cursor/ACK past the failed ciphertext.
- The production GroupStateUnavailable -> AbortBatch -> break path and successful-prefix ACK calculation are unchanged by this test/workflow-only repair.

Exact-head local evidence:

- cargo +1.97.1 build --locked --manifest-path rs/Cargo.toml -p f2z-relay --bin f2z-relay — PASS
- F2Z_RELAY_BIN=<exact-worktree>/rs/target/debug/f2z-relay cargo +1.97.1 test --locked --features relay-harness --test two_record_rollback_over_a_relay -- --nocapture — PASS (1/1, actual daemon, 2.97s)
- node scripts/check-github-actions-pins.mjs --self-test and live — PASS (all mutation cases; 191 pinned references)
- node scripts/check-workflow-gates.mjs --self-test and live — PASS (39 cases; both required contexts uniquely/gate-completely wired)
- git diff --check c6add44a..af46b40b — PASS

GitHub reports this exact head mergeable. Fresh CI remains in progress, so this review does not claim the required aggregate gate is green yet.

skyl (commented):

**Core question: real state restoration, or a weakened replay check?**
Real restoration. `MlsEngine::receive` (rs/crates/f2z-msg-mls/src/engine.rs) now runs the whole receive body as a closure; on any `Err`, `restore_group_after_rollback` reloads the caller's `&mut MlsGroup` handle from `MlsGroup::load(self.provider.store(), group_id)` — i.e. from the same durable state the dropped `Transaction` rolled back to — and re-applies the ephemeral AAD (which is intentionally not persisted). `EngineError::OutOfOrder`/`Duplicate` and the redelivery-refusal logic in `error.rs` / `receive()` are untouched; nothing was relaxed there.

**Does the new test actually reproduce the bug?**
Yes — verified empirically, not just by reading. I reverted only `engine.rs` to its pre-PR state (keeping the new tests) and ran `cargo test -p f2z-msg-mls --test two_instances`: 4 of the new tests fail, e.g.
```
an_apply_failure_after_merging_restores_memory_and_redelivery_succeeds ... FAILED
assertion `left == right` failed: a commit refused after merge must restore the caller's group
  left: GroupEpoch(2)
 right: GroupEpoch(1)
```
i.e. without the fix, storage correctly rolls back but the in-memory `MlsGroup` stays one epoch ahead — exactly issue #699. With the fix restored, all 22 tests pass. This is about as strong a proof as a review can get.

**Panics / unwraps on remote-triggerable paths:** none found in non-test code in either `rs/crates/f2z-msg-mls/src/engine.rs` or `wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs`.

**Plugin-level batch handling:** `InboundApplyError::{Continue, AbortBatch}` correctly distinguishes ordinary per-record refusals (cursor still advances) from `GroupStateUnavailable` (batch aborts, cursor does not advance, group handle dropped from the live map, `reload_group` is required before the next attempt). This also fixes a latent related bug: the old code silently returned `Ok(Vec::new())` when a group was simply absent from `self.groups`; now it calls `reload_group` and propagates failure instead of masking it.

CI wiring (`.github/workflows/zuuli.yml`, `check-github-actions-pins.mjs`) is scoped support for the new production-relay regression test (`two_record_rollback_over_a_relay.rs`), not unrelated churn.

No blockers found.

skyl (commented):
Adversarial review — PR #781 (`fix/699-mls-rollback-state`)

## Verdict: APPROVE. Real state restoration, not a weakened replay check.

**#699 fix mechanism** (`rs/crates/f2z-msg-mls/src/engine.rs:479-596`): `receive()` now
runs the whole process/merge/validate/commit sequence inside a closure. On ANY
error, it calls `restore_group_after_rollback` (engine.rs:684-700), which reloads
the caller's `MlsGroup` handle from the durable pre-transaction store
(`MlsGroup::load`) and re-applies the ephemeral AAD (which OpenMLS does not
persist). This repairs the actual defect: `merge_staged_commit`/`process_message`
mutate OpenMLS's in-memory group state directly (epoch, tree, ratchets) — the
storage-transaction rollback (drop-without-commit, `f2z-msg-store`
provider.rs:456) only reverts the *durable* store, leaving the caller's live
handle ahead of it. No replay/`OutOfOrder` logic was touched (verified full diff
of error.rs and engine.rs's `map_process_error`/`OutOfOrder` path — unchanged).

**Proof it's real, not cosmetic**: reverted just engine.rs to pre-fix and reran
`cargo test -p f2z-msg-mls --test two_instances` — 4 of the new tests fail
exactly as expected without the fix, e.g.
`an_apply_failure_after_merging_restores_memory_and_redelivery_succeeds`:
`assertion left == right failed: left: GroupEpoch(2) right: GroupEpoch(1)`.
With the fix restored, all 22 tests pass. This is conclusive: the bug is real
and the new tests genuinely exercise it (not vacuous).

**Reload-failure handling**: if the reload itself fails (`GroupStateUnavailable`,
error.rs:39-54), the plugin layer (`wallet/plugins/tauri-plugin-f2zmsg/src/
engine.rs`) correctly treats the in-memory handle as poisoned: it's dropped from
`self.groups` (not reinserted), and `apply_inbound` returns `InboundApplyError::
AbortBatch`, which stops the read-batch loop before any later record in the same
`READ` can advance the cursor past the ciphertext that needs a fresh durable
load (`pump_conversation`, engine.rs:3289-3320). This closes a related latent
bug too: the pre-PR code silently returned `Ok(Vec::new())` when a conversation's
group wasn't in the in-memory map at all — now it attempts `reload_group` and
propagates failure instead of quietly treating a missing group as "handled".

**Panics/unwraps**: none found in production code paths touched by this PR (`grep
-n unwrap\|expect` on engine.rs files returns nothing outside `#[cfg(test)]`).
`reload_group`'s `hex::decode` and `MlsGroup::load` failures are mapped to
`Error::internal` rather than panicking.

**CI wiring** (`.github/workflows/zuuli.yml`, `scripts/check-github-actions-pins.mjs`):
builds the production `f2z-relay` binary and pins `F2Z_RELAY_BIN` for the new
`tests/two_record_rollback_over_a_relay.rs`, which exercises the same regression
through the real WebSocket/relay path rather than only the in-process MemoryBackend
harness. This is necessary support for the new test, not scope creep.

No findings that would block merge.

skyl (commented):
## Adversarial review — verdict: solid, real fix

**Core question: real state restoration, or weakened replay check?**
Real restoration. No changes were made to `OutOfOrder`/replay-refusal logic anywhere in `error.rs` or `engine.rs` (verified by diffing both files in full against the merge-base). Instead, `MlsEngine::receive` now wraps its whole body in a closure; on any error it calls a new `restore_group_after_rollback` (`rs/crates/f2z-msg-mls/src/engine.rs`, new fn), which does `MlsGroup::load(self.provider.store(), group_id)` to reload the durable pre-transaction group and swaps it into the caller's `&mut MlsGroup` handle (preserving the ephemeral, non-durable `aad` explicitly). The storage side was already correctly rolled back via `Transaction`'s drop-rolls-back default (`f2z-msg-store/src/provider.rs`); the bug was that OpenMLS's *in-memory* `MlsGroup` handle (epoch, tree, receive ratchet) is mutated directly by `process_message`/`merge_staged_commit` independent of that storage rollback, so the store and the caller's handle diverged. The fix addresses exactly that divergence.

**Proof the regression test is real, not decorative.** I reverted just `rs/crates/f2z-msg-mls/src/engine.rs` to its pre-PR content (keeping the new tests) and ran `cargo test -p f2z-msg-mls --test two_instances`: 4 of the new tests fail, with e.g. `assertion left == right failed: a commit refused after merge must restore the caller's group / left: GroupEpoch(2) / right: GroupEpoch(1)` — i.e. without the fix the in-memory epoch stays advanced past the rolled-back store, precisely the #699 symptom. Restoring the fixed file, all 22 tests pass. This is about as strong a confirmation as is practical without instrumenting a live relay.

**Plugin-level handling (`wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs`).** When reload itself fails (`EngineError::GroupStateUnavailable`), the plugin correctly treats the handle as unusable: drops it from the live `groups` map, aborts the rest of the relay batch via the new `InboundApplyError::AbortBatch` (rather than letting a later record in the same `READ` batch leapfrog the ciphertext that needs retrying), and leaves the queue cursor untouched. This is the correct, conservative choice — it costs availability, never correctness. Worth noting as a side benefit: the old code silently treated "no in-memory group for this conversation" as `Ok(Vec::new())` (a no-op success!) before even attempting a durable load; the new `reload_group` fallback closes that latent gap too.

**Panics/unwraps on attacker-reachable paths.** None found — `grep` for `.unwrap()`/`.expect()` in the non-test portions of both `engine.rs` files is empty. `restore_group_after_rollback` maps a missing/failed reload to a normal `Result` error, never panics.

**CI wiring** (`zuuli.yml`, `check-github-actions-pins.mjs`): builds the real `f2z-relay` binary as a separate step and pins `F2Z_RELAY_BIN` for the new relay-backed regression test (`tests/two_record_rollback_over_a_relay.rs`, gated behind `relay-harness`) — reasonable, not scope creep, since the new test needs it.

No blocking findings. Approve pending green CI.

skyl:

PR #782, opened 2026-08-30
fix(f2zmsg): validate handles without engine state

### LINKED ISSUE #762: f2zmsg: check_handle_eligibility refuses on a faulted store even though its answer is pure
Follow-up to #753 / #759 (merged as `1deb992`).

#759 made a messaging store that cannot open fail **soft**: ZUULI starts, `get_engine_status` reports `state: "faulted"` with `lastError`, and all 43 plugin commands plus the enrollment trio refuse with a §8 code instead of taking the wallet down at launch. That was the right shape for a command that genuinely needs the engine.

**`check_handle_eligibility` is the one command in that census whose answer does not need the engine at all**, and it now refuses along with everything else.

## Why it is different from the other 42

Handle eligibility is a **pure function of the handle string** — the syntactic rules a handle must satisfy before anything is looked up. It reads no store, opens no database, and contacts no directory. It is the check a UI wants to run *while the user is still typing*, precisely in the situation where the store is broken and the user is trying to understand what is wrong.

Under #759 it takes the same `engine()?` path as `list_relays`, so on a faulted store the enrollment screen cannot tell a user that `Not A Handle!` is malformed — it can only say the messaging subsystem is unavailable. The user gets a worse diagnostic than the one the code already knows how to produce.

## Wanted

Give the eligibility check an **engine-free path**, so that it answers from the handle alone and is reachable whether or not the store opened. The engine-dependent parts of enrollment (`resolve_handle`, `start_conversation`, `accept_contact_request`) must keep refusing exactly as they do now.

## Guardrails — this must not reopen #753

- **Do not weaken `SqliteBackend::open`.** #759's whole value is that `open`'s strictness is unchanged; a fix here that makes anything about store opening more permissive is the wrong fix.
- **The command census must still cover it.** `an_unopenable_messaging_store_leaves_zuuli_running_and_every_command_refusing` in `wallet/zuuli/src-tauri/src/lib.rs` pins `tauri_plugin_f2zmsg::COMMANDS.len()` to a reviewed literal and panics on any command it does not know how to call. Moving one command from "refuses" to "answers" means that test must assert the *new* expectation for it explicitly, not drop it from the census.
- **Positive control required.** A test that eligibility still *rejects* a malformed handle on a faulted store, not merely that it returns something. Without it, "works without the engine" is satisfied by a function that says yes to everything.

## Not in scope

No UI work is needed for the faulted state itself — `EngineStatusSchema` already carries `state: "faulted"` and `lastError`, so the surface exists to render. Whether the enrollment screen *uses* the engine-free eligibility answer is a separate frontend change.

skyl:
## Summary
- route `check_handle_eligibility` directly to the pure handle validator without requiring managed engine state
- keep the command registry at 43 entries and preserve the storage fault for every engine/storage-dependent command and the enrollment trio
- strengthen the faulted-store IPC census with valid and malformed handle mutation controls, and align the client/operational documentation

## Tests
- `cargo +1.97.1 test --locked handle::tests`
- `cargo +1.97.1 test --locked an_unopenable_messaging_store_routes_only_engine_free_commands -- --nocapture`
- `cargo +1.97.1 test --locked --features relay-harness`
- `cargo +1.97.1 clippy --locked --all-targets --all-features -- -D warnings`
- `cargo +1.97.1 fmt --manifest-path wallet/plugins/tauri-plugin-f2zmsg/Cargo.toml -- --check`
- `cargo +1.97.1 fmt --manifest-path wallet/zuuli/src-tauri/Cargo.toml -- --check`
- `node scripts/check-tauri-plugin-permissions.mjs --self-test`
- `node scripts/check-tauri-plugin-permissions.mjs`
- `node --test wallet/zuuli/scripts/messaging-contract.node-test.mjs`
- `git diff --check`

Closes #762


skyl (commented):
Adversarial review at `63f84ec`: no blocking findings. The implementation removes engine state from the pure command entirely, retains validation through the same handle policy, and the registry-driven faulted-store test has both mutation controls: routing back through `engine()?` breaks the valid case, while unconditional acceptance breaks the malformed case. All remaining 41 plugin commands plus enrollment stay fail-closed. Hold for both required aggregate gates and packaging smoke before squash-merge.

skyl:
Exact-head adversarial re-review at `9a9979341f581397cf1887113ffa85730667332b`: no unresolved P0/P1/P2 findings. The focused PR still closes #762 by routing only pure handle eligibility outside managed engine state; the registry-driven faulted-store census preserves both mutation controls and keeps every engine/storage-dependent command plus enrollment fail-closed. Since the earlier review, the owner merged current `main` and aligned the explanatory docs; the PR diff remains one logical issue-linked change. `rs / gate` is green, but the macOS lanes, aggregate required `gate`, and packaging smoke are still pending. Do not merge until those exact-head results are green.

PR #798, opened 2026-08-30
fix(zuuli): preserve unknown participant counts

### LINKED ISSUE #794: ZUULI Live: represent unknown participant counts without fabricating zero or +1
Parent: #784. Client-side safety child of #265.

## Problem

ZUULI synthesizes participant counts, including local `+1` behavior, and treats missing data like a real zero.

## Acceptance criteria

- [ ] Participant count is represented as `number | null` (or an equivalent explicit unknown state).
- [ ] Missing or failed count hydration renders a concise unavailable state, never fabricated zero.
- [ ] Remove local `+1` and other synthesized host/viewer counts.
- [ ] Unknown counts do not sort as real zero.
- [ ] Accessibility text matches the visible truth.
- [ ] Tests and mutation-sensitive assertions catch reintroduced constants or coercion.

Authoritative server hydration remains in #265; this issue must not invent a backend response.

skyl:
Closes #794

## Summary
- represent participant counts as explicit `number | null` values and keep absent/failed hydration unavailable
- remove local host/viewer count synthesis and sort known zero ahead of stable unknown values
- render truthful visible and accessible count copy across Live cards, rails, and rooms

## Tests
- `npm run typecheck`
- `npm run typecheck:tests`
- `npx vitest run` (856 passed, 5 skipped)
- `ZUULI_PW_PORT=14794 npx playwright test tests/participant-counts.pw.ts tests/private-live.pw.ts` (5 passed)
- `node --test scripts/ui-copy-truncation.node-test.mjs`
- `npm run build`

### DIFF
--- a/wallet/zuuli/src/features/home/LiveRail.tsx
+++ b/wallet/zuuli/src/features/home/LiveRail.tsx
@@ -10,6 +10,7 @@ import { useAsync } from "@/hooks/useAsync";
 import type { Livestream, StreamKind } from "@/lib/api/types";
 import { cn } from "@/lib/utils";
 import { coverTone } from "@/lib/cover";
+import { participantCountCopy } from "@/lib/participant-count";
 import { SectionHeader } from "./parts";
 
 const KIND_META: Record<
@@ -24,10 +25,11 @@ const KIND_META: Record<
 
 function StreamCard({ stream }: { stream: Livestream }) {
   const kind = KIND_META[stream.kind] ?? KIND_META.broadcast;
+  const participants = participantCountCopy(stream.participants);
   return (
     <Link
       to={`/live/${stream.username}`}
-      aria-label={`${stream.title} — live now, ${stream.participants} watching`}
+      aria-label={`${stream.title} — live now, ${participants.watching}`}
       className="group flex w-[280px] shrink-0 snap-start flex-col overflow-hidden rounded-xl border border-border bg-card shadow-sm transition-all hover:border-primary/40 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring sm:w-auto"
     >
       <div
@@ -47,9 +49,7 @@ function StreamCard({ stream }: { stream: Livestream }) {
         </div>
         <div className="absolute bottom-3 right-3 flex items-center gap-1 rounded-full bg-black/55 px-2 py-1 text-xs font-medium text-white backdrop-blur">
           <Users className="h-3.5 w-3.5" aria-hidden />
-          <span className="tabular-nums">
-            {stream.participants.toLocaleString()}
-          </span>
+          <span className="tabular-nums">{participants.value}</span>
         </div>
         <div className="absolute inset-0 grid place-items-center opacity-70 transition-opacity group-hover:opacity-100">
           <RadioTower className="h-8 w-8 text-white/90" aria-hidden />
--- a/wallet/zuuli/src/features/live/Discovery.tsx
+++ b/wallet/zuuli/src/features/live/Discovery.tsx
@@ -7,6 +7,7 @@ import { Tabs, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { live } from "@/lib/api/free2z";
 import { useAsync } from "@/hooks/useAsync";
 import type { StreamKind } from "@/lib/api/types";
+import { compareParticipantCountsDescending } from "@/lib/participant-count";
 import { StreamCard, StreamCardSkeleton } from "./StreamCard";
 import { GoLiveDialog } from "./GoLiveDialog";
 
@@ -51,7 +52,7 @@ export function Discovery() {
     // Live first, then most-watched.
     return [...filtered].sort((a, b) => {
       if (a.live !== b.live) return a.live ? -1 : 1;
-      return b.participants - a.participants;
+      return compareParticipantCountsDescending(a.participants, b.participants);
     });
   }, [data, filter]);
 
--- a/wallet/zuuli/src/features/live/Room.tsx
+++ b/wallet/zuuli/src/features/live/Room.tsx
@@ -41,6 +41,7 @@ import { useSession } from "@/store/session";
 import { formatTuzis, timeAgo, initials, truncateAddress } from "@/lib/format";
 import type { DyteJoinTicket, Livestream, StreamKind } from "@/lib/api/types";
 import { coverTone } from "@/lib/cover";
+import { participantCountCopy } from "@/lib/participant-count";
 import {
   parsePrivateInviteHash,
   privateInviteDisplayUrl,
@@ -136,7 +137,9 @@ export function Room() {
         title: resolved.title,
         kind: resolved.kind,
         live: true,
-        participants: 1,
+        // Starting or joining locally does not reveal an authoritative room
+        // count. Never synthesize the host/viewer into this value.
+        participants: null,
         price_tuzis: resolved.price_tuzis,
         thumbnail: null,
         started_at: new Date().toISOString(),
@@ -228,6 +231,7 @@ export function Room() {
     creatorUsername: stream.username,
     kind: stream.kind,
   });
+  const participants = participantCountCopy(stream.participants);
 
   return (
     <div className="space-y-6 animate-slide-up">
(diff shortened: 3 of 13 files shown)


skyl:
Exact-head review is bound to `b0bbfe1cc5eeb4107e97786dc46877e0f18df404`. I found no unresolved P0/P1/P2 in the full diff or the latest-head delta. The PR is linked (`Closes #794`) and remains one logical change: the contract is explicitly `number | null`, missing/partial/malformed/unsafe totals fail closed, known zero remains distinct from unknown, sorting is stable/truthful, visible and accessibility copy share the same formatter, and both local participant-count synthesis paths are removed. The new mutation controls cover the missing-creator and post-join no-increment cases. `git diff --check` is clean.

This is not a merge verdict yet: current-head CI is still running and the exact head has no approval. Any push resets this review/check binding.

skyl:
Current-base handoff: exact head `b0bbfe1cc5eeb4107e97786dc46877e0f18df404` had an independent exact-head review with no unresolved P0/P1/P2 findings, and both required gates were green on its previous base. `main` has since advanced through #802 to `7bf653b16dcecedb5768f616d47a703aa14bbd38`, so that CI/review binding is no longer strict-current-base evidence. The branch owner must refresh/rebase it onto current `main`; the resulting exact-head delta must be independently rebound and fresh required gates must pass before merge.

skyl:
Independent exact-head adversarial review is bound to `42947715c19d549385aec1b18350baeeb73c06ce` on current `origin/main` `a4478fb1e920bc022a9ab49518d4f26264442837`.

No P0/P1/P2 findings.

Evidence reviewed: the complete `origin/main...HEAD` diff, every participant-count producer and consumer, the merge delta, and the focused mutation controls. Missing, failed, partial, malformed, fractional, negative, and unsafe counts remain explicit `null`; known zero remains distinct; complete sums retain only authoritative numeric inputs; known counts sort descending ahead of stable unknowns; visible and accessible copy share one formatter; public listing and local host/guest paths do not invent server authority or `+1`; and the browser regressions bind pre/post-join UI behavior. The current-base merge has no combined conflict-resolution delta, `git diff --check` is clean, and the exact-head `zuuli / frontend` and `rs / gate` checks are green.

This is a review comment only: I did not approve or merge. Any new push invalidates this SHA binding.

skyl (commented):
## Adversarial review — code read against `origin/pull/798/head` (4294771), diffed vs `main`

**Verdict: solid, closes #794. Approve to merge after rebase.**

### What I verified

- **`number | null` is threaded end to end, no coercion anywhere.** `LiveStatus.participants` and `Livestream.participants` are both `number | null` (`types.ts`). `normalizeParticipantCount` accepts only safe non-negative integers and returns `null` for everything else (missing, string, negative, float, NaN, Infinity — all covered by tests). `sumParticipantCounts` returns `null` if the input array is empty *or* any member fails to normalize — a partial/malformed multi-room hydration doesn't silently drop the bad member and sum the rest, it invalidates the whole total. `mapLivestream` sets `participants: null` unconditionally (the public listing endpoint never carried a count — previously hardcoded to `0`, which was the actual bug). `live.status`'s catch block now returns `participants: null` instead of `0`.
- **No fabricated `+1`/self-counting anywhere.** Grepped the full `wallet/zuuli/src` tree post-patch for `.participants` usage: every remaining site is `participantCountCopy(...)`, `compareParticipantCountsDescending(...)`, or a bare pass-through (`ParticipantStrip count={stream.participants}`). The three previous synthesis sites are gone: `Room.tsx`'s local "just started" stream literal (`participants: 1` → `null`), `ConnectedDetails`'s and `HostControls`'s `stream.participants + 1` → `stream.participants`, and the room-badge's `stream.participants + (ticket ? 1 : 0)` → `participants.watching`.
- **Unknown never sorts as zero.** `compareParticipantCountsDescending`: known values (including real `0`) always sort ahead of `null`; `null` vs `null` is stable-order (`0`). Unit-tested directly, and Discovery's sort call site was updated to use it instead of raw subtraction (which would have produced `NaN`/undefined ordering once `participants` could be `null`, likely silently reordering the whole list — worth noting this is also a latent-bug fix, not just a copy fix).
- **Accessibility text matches visible text.** Both `StreamCard` and `LiveRail` build their `aria-label` from the same `participantCountCopy(...).watching` string rendered visibly, verified by a StreamCard test that checks both the visible pill text and the `aria-label` agree, for both the unavailable and the real-zero case (real zero and unknown are asserted as visibly and accessibly *distinct* — `"0 people watching"` vs `"Participant count unavailable"`).
- **Mock path agrees with the real path.** `mockLivestreams` now has one entry with `participants: null` alongside real numeric entries (214, 91, 0); mock `live.status` runs the same `normalizeParticipantCount` as the real path rather than a separate mock-only shortcut. A dedicated mock-contract test confirms a missing mock creator resolves to `null`, not `0`.
- **UI copy never truncates.** New/changed strings: `"Unavailable"`, `"Participant count unavailable"`, `"0 people watching"`, `"12,345 people watching"`. None of the touched containers (`bottom-3 right-3` pill, `ConnectedDetails`/`HostControls` rows, `ParticipantStrip`) apply `truncate`/`overflow-hidden`/fixed width to this text — they're flex rows that grow, so nothing ellipsizes; confirmed by reading the containing JSX in each of the four call sites.
- Playwright coverage (`participant-counts.pw.ts`) exercises the real DOM for known vs. unavailable counts on Home and Discovery, confirms a locally-started host/guest room shows "unavailable" (not "1 person watching"), and confirms joining a known-count public room leaves the *known* number unchanged (not `+1`) — i.e. it directly re-proves the bug this PR fixes end-to-end, not just at the unit level.
- CI is green across the board on this PR as of review time.

### Findings

1. **Minor, cosmetic only:** `ParticipantStrip` now wraps the *entire* phrase (`"12,345 people watching"` or `"Participant count unavailable"`) in a `tabular-nums` span, where previously `tabular-nums` applied only to the numeric portion. `tabular-nums` is a no-op on non-digit text, so this has no visible effect and isn't worth a revision — flagging only because it reads as slightly sloppy, not because it's wrong.
2. No fake-zero, no coercion, no truncation issues found. Nothing here touches the live-join/ticket path, so it has no bearing on #813.

### Composition with #819 on shared files

Same conclusion as posted on #819: `Room.tsx`, `free2z.ts`, `types.ts` overlaps with #819 are on disjoint fields/concerns (this PR: `participants` typing/display; #819: `meetingId`/ticket binding/retry/creator guard). The only literal-adjacency is inside `mapLivestream`'s return object and a couple of nearby lines in `Room.tsx` — a mechanical rebase merge, not a design conflict. Neither PR's logic reads the field the other one changes.

skyl:
Rebased onto `origin/main` at f629de4 (new head: `61c34dd`).

Conflicts resolved (both concerns kept, nothing dropped):

- `src/features/live/Room.tsx` — kept both #819's `entryState` (viewer-safety gating) and this PR's `participants = participantCountCopy(stream.participants)`; both are used independently later in the file.
- `src/lib/api/free2z.ts` `live.status()` — kept #817/#819's strict per-entry kind-contract check (`streamKindFromType` on both `key` and `entry.meeting_type`, throwing `LivestreamKindContractError` on mismatch) *and* this PR's `number | null` return type + `sumParticipantCounts` aggregation (so a missing/malformed member makes the total `null` instead of silently summing as 0), including returning `participants: null` (not `0`) on the catch path.

Follow-up fixes needed after the rebase (pushed as a second commit, `61c34dd`):

- `participant-count-contract.test.ts` — updated mocked `live-status` responses to include a `meeting_type` matching each key, satisfying the stricter contract check now merged into `live.status()`; count-related assertions are otherwise unchanged.
- `tests/participant-counts.pw.ts` — the host flow now goes through `installMockCapture` + "Set up camera and microphone" → "Preview ready" → "Confirm and start" (media-preflight gate merged since this PR was opened, button copy changed from "Start stream"), matching the pattern in `tests/private-live.pw.ts`.

Local verification (not just vitest):
- `npm run typecheck` — clean
- `npx vitest run` — 91 files / 1156 tests passed, 5 skipped
- `node scripts/ui-copy-truncation.node-test.mjs` — 4/4 passed
- `npx playwright test tests/participant-counts.pw.ts` — 3/3 passed

No coercion introduced: participant counts remain `number | null` end to end, unknown never renders as `0` or a fabricated `+1`, and unknown sorts last via `compareParticipantCountsDescending`.

Ready to merge once `gate` and `rs / gate` are green.

Issue #790, opened 2026-08-30
ZUULI tipping: make ZEC a first-class Tip the creator option


skyl:
## Problem

**Tip the creator** does not offer ZEC even though ZUULI is a Zcash wallet. The primary creator-support flow must support native ZEC, not only 2Z.

## Product direction

Offer ZEC and 2Z as clear, first-class choices. The ZEC path should use the creator's verified payment destination and the wallet's existing safe-send machinery.

## Acceptance criteria

- [ ] **Tip the creator** offers ZEC wherever creator tipping is available.
- [ ] The UI clearly distinguishes ZEC from 2Z balances, amounts, and outcomes.
- [ ] Recipient resolution cannot silently substitute or mutate the destination.
- [ ] The confirmation binds creator, destination, amount, asset, fee, and memo before signing.
- [ ] Prefer shielded ZEC and clearly disclose any privacy downgrade before confirmation.
- [ ] Insufficient funds, sync state, fee calculation, cancellation, failure, and retry are handled without duplicate payment.
- [ ] Success links to durable wallet transaction state and creator attribution.
- [ ] Android and iOS tests cover the complete path.

skyl:
## Note: splits across two apps under #904 — intent moves, execution stays

ZUULI is being split into three surfaces. Tipping straddles the boundary:

- **The intent moves to `free2z`** — the creator page, the amount entry, the "Tip the creator" affordance.
- **Execution stays in ZUULI** — address validation, proposal creation, native confirmation, signing, broadcast. Spending authority never leaves the wallet app.

The handoff is a payment intent over a **verified** deep link, with ZUULI showing its own native confirmation before executing.

**The existing design already models this correctly.** `lib/wallet/creator-tip.ts` is a frozen snapshot keyed by nonce, re-validated field-by-field on the wallet side and failing closed on any mismatch — that is exactly the shape a cross-app intent needs, it just currently travels via React Router state instead of a deep link. #528's one-use native confirmation token is the other half.

**Blocked on #461** (verified Universal Links / App Links) before it can ship cross-app — custom schemes are not an authenticated channel and a hostile app could forge or intercept an intent.

Note the 2Z tip path is unaffected: it is a server-side balance operation over HTTP and never touches the wallet.

Context: #904.

PR #799, opened 2026-08-30
feat(zuuli): lock creator ZEC tip intents

### LINKED ISSUE #795: ZUULI ZEC tips: carry a locked creator payment intent into Wallet Send
Parent: #790. This is the first unblocked client slice and must not close the parent.

## Scope

Carry an immutable creator ZEC-tip intent from the creator tip dialog into the existing safe Wallet Send pipeline. Reuse the current proposal, native confirmation, execution, recovery, balance refresh, and history flow; do not build a second sender.

## Acceptance criteria

- [ ] The creator tip dialog offers distinct ZEC and 2Z choices with correct units; existing 2Z behavior is unchanged.
- [ ] A strict, non-persisted route-state contract carries canonical username, display label, and the exact loaded `CreatorDetail.p2paddr`.
- [ ] Missing, malformed, altered, or wrong-network intent fails closed without falling back to editable recipient text.
- [ ] Creator-tip recipient is read-only; stale validation, creator navigation, or proposal work cannot substitute it.
- [ ] Native `wallet.validateAddress` and the existing propose → native confirm → execute/recover path remain authoritative.
- [ ] Review shows creator context adjacent to—never instead of—the full native proposal destination, amount, fee, memo, network, and change truth.
- [ ] Transparent destinations get an explicit privacy/no-memo warning before proposal; no attribution memo is invented.
- [ ] Cancel, insufficient funds, proposal failure, native-confirm cancellation, unknown broadcast, retry, and accepted-to-history behavior reuse existing safe-send handling.
- [ ] Tests at 320px prove route/creator/destination mutation cannot alter a pending proposal and no intent or proposal secret enters web storage.

Likely surfaces: `wallet/zuuli/src/features/creator/index.tsx`, `src/features/wallet/Send.tsx`, a new `src/lib/wallet/creator-tip.ts`, and focused tests. No Rust change is expected in this slice.

Durable creator attribution and cryptographic payout-address ownership remain follow-ups under #790.

skyl:
## Summary
- offer distinct ZEC and 2Z creator-tip choices while preserving the existing 2Z flow
- carry an alteration-detecting, in-memory creator intent into the existing Wallet Send proposal/confirmation/execution path
- lock the creator recipient, fail closed on missing/reloaded/changed/wrong-network state, and disclose transparent-address privacy and memo limits
- show creator context beside the complete native review and retire accepted intents so browser history cannot reuse them

## Verification
- `npm run typecheck`
- `npm run typecheck:tests`
- `npx vitest run --reporter=dot` (858 passed, 5 skipped)
- `ZUULI_PW_PORT=1479 npx playwright test tests/creator-zec-tip.pw.ts tests/send-review.pw.ts`
- `ZUULI_PW_PORT=1479 npx playwright test tests/paid-actions.pw.ts tests/viewport.pw.ts`
- `node --test scripts/send-review-boundary.node-test.mjs scripts/ui-copy-truncation.node-test.mjs`
- `npm run build`

Closes #795
Part of #790


skyl:
Exact-head review is bound to `0a369939312b185faea152279c93f9c59da0d6fb`.

Blocking P2 — an unrelated recovered payment can be presented and retried as this creator tip. `Send` unconditionally loads any pre-existing native unknown broadcast, then the creator-tip route renders the newly selected creator/address above that recovered transaction and makes `retryPendingSend` the available action. The recovered transaction predates this intent and is not bound to it; the UI exposes only its txid, not its recipient/amount. If retry is accepted, the handler also retires the new creator intent even though no proposal for that creator was made.

Please make the creator-tip route fail/redirect into a clearly separate recovery surface before displaying creator attribution, or prove that the recovered transaction is bound to this exact creator proposal (the current persisted status appears not to carry that binding). Do not retire the new creator intent for an unrelated recovery. Add a focused regression that seeds a pre-existing pending/unknown send before entering the creator-tip route.

No other P0/P1/P2 finding from the current review so far. CI is still running; this verdict must be rebound after any new push.

skyl:
A second exact-head P2 is confirmed on `0a369939312b185faea152279c93f9c59da0d6fb`: the creator-tip privacy warning is keyed only to `validation.addressType === "transparent"`. Native validation also accepts a Unified Address containing only a supported transparent receiver and returns `address_type = "unified"`, `can_receive_memo = false`. That payment is transparent-only/no-memo, but the current UI omits the explicit public-on-chain warning. The mock marks every `u1` memo-capable, so the focused tests cannot exercise this case.

Please key the disclosure to authoritative capability/privacy information (at minimum, creator-tip validation that is valid but cannot receive a memo), or expose an authoritative privacy classification. Add a transparent-only Unified Address case. Do not infer privacy solely from the display address type.

The earlier unrelated-pending-send attribution finding also remains open. Any new push resets the head/check/review binding.

skyl:
Independent exact-head adversarial review is bound to `8457bfb7f8aca6efca9607fd3f9a538dc3f0b5d5` on current `origin/main` `a4478fb1e920bc022a9ab49518d4f26264442837`.

No unresolved P0/P1/P2 findings.

I inspected the full current-base diff and the remediation delta from merge head `b324792c9b04acc80a0566a7969b4b3a87a54260`. The two prior P2s are closed: a pre-existing native unknown broadcast now blocks creator attribution and replaces the route with neutral Wallet Send before the creator form can render; it is neither shown nor retired as the new tip. Transparent disclosure now follows the native capability (`valid && !canReceiveMemo`), so a valid Unified Address with only a supported transparent receiver warns and disables memos even though its display type is `unified`.

Native-validator provenance is consistent with the shipping implementation: `validate_parsed_recipient` returns `can_receive_memo = false` for a Unified Address with no supported shielded receiver and a supported transparent receiver, while rejecting unknown-only UAs; its existing Rust regression constructs the same P2PKH-plus-unknown receiver shape used by the new realistic browser fixture. The mock regression fixes the exact encoded fixture to `{ valid: true, addressType: "unified", canReceiveMemo: false }` instead of deriving capability from the `u1` prefix.

Safe-send authority remains unchanged and immutable: creator route state resolves through its in-memory nonce snapshot; the proposal recipient comes from that snapshot rather than the DOM; the native proposal must exactly match recipient, amount, memo, totals, policies, network, digest, and proposal token; native confirmation freshness and confirmation-token execution remain mandatory. The creator label stays adjacent to, never in place of, the full native proposal review.

Evidence: `git diff --check origin/main...8457bfb7f8aca6efca9607fd3f9a538dc3f0b5d5` is clean; focused `creator-tip.test.ts` plus `mock.test.ts` passed 26/26. Fresh exact-head CI is still running, so this is a review finding verdict, not a CI or merge verdict. Any push invalidates this binding.

skyl (commented):
**Verdict: APPROVE** (adversarial review; posted as a comment rather than an approval because the authenticated \`gh\` identity is also the PR author — GitHub rejects self-approval on this repo).

Focused on whether the "locked" creator ZEC tip intent is genuinely immutable between the creator page and Wallet Send, whether Send.tsx validates against what it displays, and ZEC amount/precision handling.

**Locking mechanism — real, not just a frozen object.** \`src/lib/wallet/creator-tip.ts\` doesn't trust \`location.state\` as the source of truth: \`createCreatorTipRouteState\` stores the canonical username/label/recipient in a module-private \`Map\` keyed by a random nonce, and \`readCreatorTipRouteState\` re-validates every field of the passed-in route state against that stored record, returning \`null\` (fail closed) on any mismatch, unknown nonce, wrong shape, or extra/missing keys. So even if the route \`state\` object is mutated in place (history.replaceState, or a compromised renderer patching the DOM), the value actually used to build the send request (\`creatorTip.recipient\` in \`Send.tsx\`) still comes from the verified module-memory snapshot, not from the tamperable route object. This is validated by an actual adversarial test, not just unit coverage: \`tests/creator-zec-tip.pw.ts\` ("locks the issued destination through native review") mutates both \`window.history.state\` and the readonly \`#to\` DOM input mid-flow via \`page.evaluate\`, and then asserts the confirm dialog still shows the original, unmutated recipient. That's the right test for exactly the concern raised.

**Display-matches-what's-signed.** The proposal review (\`send-review-recipient\` in \`Send.tsx\`) renders \`proposal.review.payments[0]?.recipient\`, i.e. the native/mocked engine's echo of the proposal it actually built — not a locally-held string — so the user cannot be shown one address while a different one is what gets signed, short of the native engine itself lying about its own proposal (out of scope for this PR).

**No amount to tamper with.** The ZEC tip flow intentionally locks only the recipient; the amount field is the ordinary user-entered Send amount (unchanged parsing/zatoshis logic), so there's no creator-supplied amount to round or mutate.

**House rule on client-side "locking" claims:** flagging per doctrine that this is enforced entirely in the renderer (module memory + route-state verification), so a genuinely modified/patched client binary could still call the native \`propose_send\`/\`execute_send\` commands directly with an arbitrary address — that's true of any client-side control and isn't specific to this PR. The PR's own language ("the creator destination is locked from the profile you just loaded") is scoped to normal in-app navigation/tampering, which is the realistic threat model here (confirmed via the Playwright test above), and it doesn't overclaim beyond that. Calling this out for the record per the house rule, not as a defect.

**Fails closed on every edge case checked:** unissued/never-created route state, reload (in-memory map is lost by design), reuse after acceptance (\`retireCreatorTipIntent\` removes the nonce so browser Back can't replay it), and a pre-existing unresolved pending-send is explicitly *not* attributed to a new creator tip (routed to the neutral \`/wallet/send\` recovery surface instead). All four are covered by both \`creator-tip.test.ts\` and the Playwright suite.

**Compose safety with #805:** confirmed by materializing the actual merge of both PRs (git merge-tree, no conflicts) and running \`npm run typecheck\` + the full Vitest suite on the combined tree — clean typecheck, 885 tests passed, 0 failed. Shared files \`src/lib/wallet/mock.ts\` and \`src/features/wallet/index.tsx\` touch disjoint regions (wallet-identity mock additions vs. \`validateAddress\`/\`getPendingSend\` mock additions; onboarding-gate line vs. new route/section-title entries).

No blocking defects found.

PR #800, opened 2026-08-30
fix(f2zmsg): fail closed on relay error contexts

skyl:
Refs #550

## Outcome

This focused correction closes the public-contract/shipping-runtime error and durable-bind divergences from #550 without claiming the production-only handle acceptance item:

- maps every relay-returned `ERR_NO_ACCESS` (10) to `relay-protocol-violation`; state-dependent send refusals remain `ERR_UNAVAILABLE` (15) → `send-unavailable`;
- carries exact command context through the shipping response mapper and fails closed for known codes in invalid contexts;
- persists `Fresh` / `OutcomeUnknown` / `Confirmed` before irreversible bind work and reconciles unknown outcomes through same-key `APPEND` without rewriting relay acceptance;
- makes fresh-address theft loud, queue-scoped, and historically attributable through nullable `Alarm.conversationId` and `Alarm.relayUrl`, without claiming the relay operator was necessarily the actor;
- exposes the active compromised queue relay directly as nullable `Conversation.compromiseRelayUrl`, so current UI attribution never guesses from historical alarm ordering;
- retains failed-persistence alarms as queue-identity-scoped volatile evidence, includes them in public alarm lists and status, and atomically flushes evidence on acknowledgement, replacement, leave, unenrollment, graceful stop, or shutdown before clearing the blocker;
- gives alarm evidence collision-resistant IDs so distinct same-millisecond thefts remain separately listable and acknowledgeable;
- preserves identical adverts while giving every genuinely distinct replacement a fresh CSPRNG send-key seed and retaining historical alarms;
- truthfully reserves the unimplemented exporter rotation schedule and tracks its implementation or retirement in #827;
- binds WIRE, CLIENT-CONTRACT, architecture/ADR/threat wording, Rust models/runtime, shipping TypeScript schema/UI, persistence ordering, replacement key generation, active attribution, and recovery paths with mutation-sensitive checks;
- normally merged `origin/main` at `73b14ff4b7922b7577ac0ac5ed0129aaa5fcabb7`; no rebase or force push was used.

## Scope split

This remains error/bind-recovery-only. The removed work is preserved in focused public issues:

- #820 — raw-ASCII handle candidate semantics, terminology, and Rust/mock parity;
- #821 — bounded PoW work, signed capability policy, and calibration-safe wording;
- #827 — implement or retire the reserved queue rotation schedule.

#550 remains open for its authorized backend census/collision cleanup. No private production data is included here.

## Verification

- required relay-harness/plugin gate: 67 library + 2 containment + 10 lifecycle tests passed;
- relay protocol: 96 unit + 6 property + 8 redaction tests passed;
- messaging MLS: 17 unit + 11 credential + 7 redaction + 18 two-instance tests and the doc test passed;
- shipping `cargo build --locked` and plugin permission generation/self-test passed;
- `cargo +1.97.1 clippy --locked --all-targets ... -- -D warnings` passed;
- `node wallet/zuuli/scripts/messaging-contract.node-test.mjs` — 9/9 passed with injected contract/schema/runtime/order/key/alarm/active-attribution mutations;
- frontend production typecheck, WASM verification, and Vite build passed after the normal main merge;
- Android 32-bit policy — 44 mutation controls plus live selector passed;
- workflow/pinned-action self-tests and live checks, Rust formatting, public-repository boundary, and `git diff --check` passed;
- CI-race hardening: the restart fixture drops its pre-restart WebSocket and reconnects after reopening SQLite, reports the exact failure code on recurrence, passed 200/200 serial exact runs before the fixture repair, 200/200 and then 100/100 concurrent full-suite runs locally, and 50/50 fully-qualified exact runs in a clean Ubuntu 24.04 amd64 CI-image build on Rust 1.97.1.

Adversarial review is requested against exact pushed head `1b849ecb23c71c613f7ec32bc8b0e5d4bc845a04`.


skyl:
Exact-head review is bound to `f984cfb185e8a55d5341c1296b73d485cad0ec8d`.

Two P2/process blockers remain:

1. The PR says every public contract restatement is updated, but exact head still publishes the old ambiguous rule in `rs/crates/f2z-authority/src/types.rs` (`lowercase(username) ...`), `wallet/plugins/tauri-plugin-f2zmsg/src/handle.rs` (generic lowercase/folding and “after lowercasing”), `wallet/plugins/tauri-plugin-f2zmsg/src/models.rs`, and CLIENT-CONTRACT's `too-long` comment. The implementations are currently safe (`is_ascii()` before ASCII-only lowering), but these standalone public comments preserve the divergent-implementation instruction #550 is meant to close. Update every restatement to the raw-ASCII-first, ASCII-only `ascii_lower` rule and remove the remaining generic-lowercase wording.

2. The current diff combines two independently shippable/revertible #550 corrections: the PoW algorithm/difficulty/TTL contract, and the username-eligibility/census contract plus Zuuli mock change. Under the one-logical-change queue rule, please split those into focused PRs (or otherwise narrow this PR to one correction family). Sharing one issue number does not create a technical dependency between them.

`Refs #550` is the correct non-closing linkage because authorized backend census/collision cleanup remains. Current CI is still pending, and any push resets review/check binding. #800 semantically overlaps #769’s WIRE/client-contract work, so the surviving focused correction(s) should land before #769 is refreshed against main.

skyl:
Rebound to new exact head `767df2cb4724c1582685c61606749dffdab8361a`. The previous blockers remain:

- The new commit does not update the stale standalone generic-`lowercase(username)` restatements in `f2z-authority` and the messaging plugin.
- The PR is still not one logical change. It continues to combine independent PoW and handle/census correction families, and the new commit now adds a third independent item-1 error-code mapping correction.
- The PR body still requests review against the obsolete `f984cfb…` head.

Please split/narrow the independent correction families and update every surviving restatement/body reference. Current-head CI is pending; prior checks/reviews are obsolete.

skyl:
New exact-head P1 on `767df2cb4724c1582685c61606749dffdab8361a`: the added CLIENT-CONTRACT mapping now requires send-side/state-dependent wire code 10 (`ERR_NO_ACCESS`) to surface as `relay-protocol-violation`, but shipping `wallet/plugins/tauri-plugin-f2zmsg/src/wire_codes.rs` still maps `WireCode::NoAccess + CommandSide::Send` to `SendUnavailable`, with tests explicitly asserting the old “whichever way #550 lands” hedge. This commit therefore creates fresh public-contract/runtime divergence in the same surface #550 is closing.

Please either include the coherent runtime mapping/test change in the focused item-1 PR after splitting the current multi-family diff, or do not publish the new mapping yet. The stale lowercase restatements, multi-logical-change scope, and obsolete head reference also remain unresolved.

skyl:
Rebound to exact head `93bfe58fd5d684d9bb84f64cf55f087f04af5402` after #782. This is only a rebase onto `4b939558…`; none of the blockers is fixed: stale generic-lowercase restatements remain, CLIENT-CONTRACT still contradicts shipping code/tests for send-side wire code 10, the diff still combines PoW + handle/census + error-code families, and the body still names an obsolete review head. Fresh CI is starting; do not treat prior checks as current.

skyl:
Exact-head review is rebound to `0a0ecfb38b292d93efa8b714913038e3a8e3de4a`, whose merge base is current main `7bf653b16dcecedb5768f616d47a703aa14bbd38`. This push is a pure rebase over merged #802: the prior-head → current-head delta contains only #802's relay-testkit changes, and #800's own eight-file diff is unchanged. `git diff --check` is clean.

The existing blockers therefore remain:

- **P1 — published contract still contradicts shipping behavior.** `docs/e2ee/CLIENT-CONTRACT.md:1686` and `:1738-1754` require send-side/state-dependent wire code 10 (`ERR_NO_ACCESS`) to map to `relay-protocol-violation`. Shipping `wallet/plugins/tauri-plugin-f2zmsg/src/wire_codes.rs:89-90` still maps `WireCode::NoAccess + CommandSide::Send` to `SendUnavailable`, and its assertions at `:229-234` explicitly preserve that mapping. This exact head still creates the public-contract/runtime divergence #550 is intended to close.
- **P2 — stale generic-lowercase public restatements remain.** Current head still says `lowercase(username)` or “after lowercasing” in `rs/crates/f2z-authority/src/types.rs:163-164`, `wallet/plugins/tauri-plugin-f2zmsg/src/handle.rs:26,33`, `wallet/plugins/tauri-plugin-f2zmsg/src/models.rs:275`, and `docs/e2ee/CLIENT-CONTRACT.md:383`. These must state raw-ASCII rejection first and ASCII-only A-Z → a-z lowering.
- **P2/process — not one logical change.** The diff still combines independently shippable/revertible PoW defaults, handle/census semantics plus Zuuli mock behavior, and relay error-code mapping. Split or narrow to one correction family.
- **P2/process — stale review identity.** The PR body still requests adversarial review against obsolete head `f984cfb…`, not this head.

`Refs #550` remains the correct non-closing linkage because #550 is still open, `blocked`, and `needs-backend`; its production census/uniqueness work remains. Fresh exact-head CI is only beginning, prior-head checks are stale, and there is no approval. Do not merge this head.

skyl (commented):

**1. Command-context error mapping (`wire_codes.rs`).** `from_relay(code, command, attempt)` is a `match` over `#[non_exhaustive] f2z_codec::ErrorCode`, ending in `_ => ErrorCode::RelayProtocolViolation` — the only legal default for an external non-exhaustive enum, and it's the fail-closed one, never `Internal` or a permissive pass-through. Every non-default arm is additionally guarded by `matches!(command, ...)`/`attempt` context, so a structurally "known" code in a context the table doesn't name still falls through to the same default. This is backed by a brute-force exhaustiveness test (`every_command_code_and_attempt_matches_the_allowed_context_matrix`) that cross-products `WireCode::ALL × Command::ALL × BindAttempt::{...}` and asserts every cell not explicitly whitelisted equals `RelayProtocolViolation` — about the strongest test shape available for this claim. `from_directory` and `from_proto` (incl. `ProtoError::UnknownStatus` and any future non-exhaustive `ProtoError` variant) follow the identical default rule.

**2. `ERR_NO_ACCESS` (10) vs `ERR_UNAVAILABLE` (15).** Confirmed as claimed: `WireCode::NoAccess` maps unconditionally to `RelayProtocolViolation` for every `Command` (test `no_access_is_never_ordinary_send_unavailability` iterates `Command::ALL`), while `WireCode::Unavailable` maps to `SendUnavailable` only for `BindSend | Append | ContactAppend`. `docs/e2ee/CLIENT-CONTRACT.md`'s new §8.1 table row for code 10 and its worked-example note match this exactly.

**3. Context threading in `relay.rs`.** Every `proto()`/`from_proto` call site passes the real in-flight command: `Command::BindSend`, `Command::Hello`, or the generic `C::COMMAND` (the signed-command type parameter's associated const) for `call_signed`. No call site passes a stale/generic placeholder. `bind_send_once` correctly narrows `ERR_ALREADY_BOUND` to `BindSendOutcome::AlreadyBoundAfterUnknown` only under `BindAttempt::OutcomeUnknown`, else it flows to the public mapper (theft on `FirstForFreshAdvert`, protocol-violation on `Later`).

**4. `Fresh`/`OutcomeUnknown`/`Confirmed` ordering.** Verified the durable transition precedes the irreversible network call in `ensure_bound` (wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs): `persist_bind_state(..., OutcomeUnknown)` is committed *before* `connection.bind_send(...)` is issued, in both the `Fresh` and `OutcomeUnknown` arms — comment states the invariant explicitly. `Confirmed` is persisted only after a successful `BIND_SEND`, or (for the `AlreadyBoundAfterUnknown` reconciliation path) only after the same-key `APPEND` actually succeeds (`append_and_confirm_binding`). No path performs the irreversible action first and persists after.

**5. Fail-open shortcuts:** none found. The two places that log-and-continue on a *secondary* write failure never weaken the value returned to the caller: a failed `Confirmed`-state write after an already-accepted `APPEND` is cosmetic (the wire fact already happened; `OutcomeUnknown` reconciles safely next pass); a failed durable write of theft evidence still returns the `SendAddressStolen` error to the caller and is tracked in `volatile_compromise_alarms` (also checked by `send_address_is_stolen`) until flush.

**6. Docs-vs-code spot checks.** (a) `ERR_NO_ACCESS` table row — matches code, see #2. (b) Queue-rotation exporter (`free2z/queue/v1`): `WIRE.md`/`exporter.rs` now say it's "reserved... not shipping" and the shipping path never invokes it — confirmed by grep, `ExportLabel::Queue` has zero non-doc references in the tree. This *retracts* a previous over-claim (the doc used to imply it actively derived rotation schedule + signing keys), which is exactly the right direction for a PR closing a spec-divergence issue — issue #827 tracks implementing or removing it for real.

**7. TS side.** `mock.ts`/`types.ts` changes are purely additive schema/mock fields (`compromiseRelayUrl`, `Alarm.conversationId`/`relayUrl`) mirroring new Rust `Conversation`/`Alarm` fields; the mock sets them to `null` rather than fabricating a compromised/active state, so it isn't more permissive than production. The `ErrorCodeSchema` enum itself (incl. `relay-protocol-violation`, `send-unavailable`) was already present pre-PR and is unchanged here — no drift introduced.

**5 new unwrap/expect in production code:** none — all new `.expect()`/`.unwrap()` in the `engine.rs` diff are inside `#[cfg(test)] mod tests`.

No blockers found.

skyl (commented):
Adversarial review — PR #800 (`docs/550-wire-divergence`)

## Verdict: APPROVE. Fails closed; docs and code agree; TS mirrors Rust.

**1. Fail-closed mapping** (`wallet/plugins/tauri-plugin-f2zmsg/src/wire_codes.rs:
from_relay`, ~line 47-140): every context-sensitive arm is a match GUARD
(`if matches!(command, ...)`); a guard miss falls through the match, not into
a permissive branch — final arm is `_ => ErrorCode::RelayProtocolViolation`
(wire_codes.rs ~line 143). Confirmed exhaustive and total by the crate's own
`every_command_code_and_attempt_matches_the_allowed_context_matrix` test, which
iterates the full cross-product of `WireCode::ALL × Command::ALL ×
BindAttempt::ALL` and asserts every cell not explicitly allow-listed is
`RelayProtocolViolation`. Ran it: `cargo test --locked wire_codes` — 9/9 pass,
including this matrix test. `ErrorCode::Internal` is reachable ONLY from
`WireCode::Internal` for every command (asserted by
`only_an_explicit_peer_internal_fault_maps_to_internal`) — an unknown/unparseable
code can never present as "internal" (which the frontend contract treats as a
component fault) or as success.

**2. No fail-open coalescing found.** No `.unwrap_or(<success>)`, no `_ =>` that
resolves to anything but `RelayProtocolViolation`/`DirectoryProtocolViolation`.
The one place a wire error is deliberately *not* surfaced as a public code is
`ERR_ALREADY_BOUND` under `BindAttempt::OutcomeUnknown`
(`relay.rs::bind_send_once`, ~line 462-478): this converts to
`BindSendOutcome::AlreadyBoundAfterUnknown`, not to success. The caller
(`engine.rs::ensure_bound`, ~line 2660-2665) then requires a same-key `APPEND`
to actually confirm the bind (`append_and_confirm_binding`, ~line 2722-2748);
the `?` on that append propagates failure, so `BindState::Confirmed` is only
persisted after the relay accepts a signed write from *this* key — a relay that
actually gave the capability to someone else fails that append instead of being
waved through.

**3. Known-code-wrong-context fails closed.** `ERR_NO_ACCESS` unconditionally
maps to `RelayProtocolViolation` regardless of command (wire_codes.rs, `NoAccess`
in the unconditional violation arm) — this closes the old #550 ambiguity
correctly: state-dependent send refusals must use code 15 (`Unavailable`), and
code 10 in that position is now itself flagged as a violation rather than
collapsed into `send-unavailable`. Test
`no_access_is_never_ordinary_send_unavailability` and
`state_revealing_codes_fail_closed_outside_their_commands` cover this and pass.

**4. No panics found.** No new `unwrap`/`expect`/`indexing` on wire-derived
values in wire_codes.rs or relay.rs; `cargo check --locked` on the whole plugin
crate compiles clean.

**Docs vs code — spot-checked 5 load-bearing claims, all backed:**
- OutcomeUnknown reconciliation via same-key APPEND: matches
  `ensure_bound`/`append_and_confirm_binding` (engine.rs ~2634-2748).
- Fresh CSPRNG seed on a genuinely distinct replacement advert, identical replay
  preserved: `set_peer_advert` (engine.rs ~3320-3350) — early-returns on
  identical advert, else `rand::rng().fill_bytes(&mut send_key_seed)`.
- `free2z/queue/v1` exporter "reserved, not shipping": `exporter.rs` diff matches
  exactly — doc explicitly narrows an earlier over-claim rather than widening one.
- `Conversation.compromiseRelayUrl` / `Alarm.conversationId`/`relayUrl`: present
  in `models.rs` structs, camelCase-serialized, matching CLIENT-CONTRACT.md's
  new interface fields exactly.
- Evidence flushed atomically on stop/shutdown/leave/unenroll/acknowledge: each
  has its own commit-then-clear code path (`flush_volatile_compromise_alarms`,
  `unenroll`, `leave_conversation`, `acknowledge_alarm` in engine.rs), backed by
  dedicated tests (`recovered_storage_flushes_volatile_theft_before_stop_and_shutdown`,
  `unenroll_clears_in_memory_compromise_only_after_durable_commit`,
  `leave_clears_queue_compromise_only_after_durable_removal`).
No doc claim found that over-states what the code does; if anything the docs
correct a prior over-claim (queue rotation exporter) to be more honest, which is
exactly the direction #550/#552 wanted.

**TS/mock parity**: `types.ts`'s `ErrorCodeSchema` and `AlarmSchema`/
`ConversationSchema` add the same fields the Rust `models.rs` enums/structs add
(`compromiseRelayUrl`, `conversationId`, `relayUrl` — all `nullable()`, matching
Rust's `Option<String>` with unconditional serialization). `mock.ts` sets these
to `null` in its fixtures, which is the safe/strict direction (a mock that always
reports "not compromised" cannot mask a bug that requires ATTACKER-controlled
state, since compromise UI is exercised via the mock's other scripted alarm
fixtures, not via this field). Rust and TS retryable-code sets are asserted
equal by the codebase's own `tests/error_contract.rs` (pre-existing mechanism,
not added by this PR, but still enforced). No case where the mock accepts
something production would reject.

No findings that would block merge.

skyl (commented):
## Adversarial review — verdict: implementation matches docs, fail-closed confirmed

**1. Fail-closed mapping (`wallet/plugins/tauri-plugin-f2zmsg/src/wire_codes.rs`).** This is the core fix and it holds up. `from_relay` is a `match` over the closed, `#[non_exhaustive]` `WireCode` enum with a catch-all `_ => ErrorCode::RelayProtocolViolation` (never `Internal`, never a benign/retryable default). Confirmed via `git diff` against the merge-base that the *previous* code mapped several codes (`Quota`, `Unavailable`, `TooManyInflight`, `StaleTimestamp`, `RateLimited`, `Backpressure`, `PowRequired`, `PowInvalid`) with **no command-context guard at all** — any command producing that code got mapped unconditionally. The new code guards every one of them with `if matches!(command, ...)` and falls through to `RelayProtocolViolation` otherwise — i.e. a known code appearing in a context the wire spec doesn't give it now fails closed instead of being silently accepted as its "normal" meaning. `from_directory` and `from_proto` have the identical catch-all pattern. A dedicated test (`every_command_code_and_attempt_matches_the_allowed_context_matrix`) enumerates every *allowed* cell and asserts everything else is a protocol violation — I ran it (`cargo test --lib`, tauri-plugin-f2zmsg): all 9 `wire_codes` tests pass, including `no_access_is_never_ordinary_send_unavailability` and `state_revealing_codes_fail_closed_outside_their_commands`, which are exactly the properties in question. No `_ => <permissive>` or `.unwrap_or(<ok variant>)` pattern anywhere in the mapping.

The headline claim — "every `ERR_NO_ACCESS` (10) maps to `relay-protocol-violation`" — is real: the old code branched `WireCode::NoAccess` on `CommandSide::Send` vs `Receive`, mapping the send side to `SendUnavailable` (an existence-oracle leak this PR is explicitly closing); the new code maps `NoAccess` unconditionally to `RelayProtocolViolation` with no side distinction at all.

**2. Docs vs code.** I checked every substantive claim in the `WIRE.md`/`CLIENT-CONTRACT.md` diff against its implementation rather than taking the PR description at its word:
- `Fresh`/`OutcomeUnknown`/`Confirmed` bind-state persistence: real `BindState` enum (`store.rs:263-269`), threaded through `engine.rs` and `relay.rs`, with `effective_bind_state()` conservatively treating a legacy/missing record as `OutcomeUnknown` (never `Fresh`) — matches the "restarts can never manufacture theft evidence" claim (`engine.rs:2637`).
- `compromiseRelayUrl` (current, not historical): field exists on both sides (`models.rs:419-420` `Option<String>` camelCase-serialized ↔ `types.ts:245` `z.string().nullable()`), only set from the live queue, historical alarms retained separately.
- Collision-resistant alarm IDs: `send_address_stolen_alarm_at` appends a 16-byte CSPRNG nonce to the id (`engine.rs:3024-3056`); test `same_millisecond_volatile_alarms_remain_distinct_and_acknowledge_independently` exercises exactly this.
- Fresh CSPRNG send-key seed per genuinely distinct replacement, identical adverts preserved: `set_peer_advert` early-returns on an identical (relay, address) pair before touching key material, and only generates via `rand::rng().fill_bytes(...)` for a real change (`engine.rs:3320-3345`).
- Volatile evidence flushed atomically on all six named triggers: verified call sites in `unenroll`, `leave_conversation`, `acknowledge_alarm`, `set_peer_advert` (replacement), `stop`, and `shutdown` — all touch `volatile_compromise_alarms` and commit durably; `unenroll_clears_in_memory_compromise_only_after_durable_commit` and `recovered_storage_flushes_volatile_theft_before_stop_and_shutdown` both pass.
- Exporter rotation schedule "truthfully reserved": `ExportLabel::Queue` doc-comment now says "reserved; not shipping" instead of claiming a derivation scheme, and I confirmed `ExportLabel::Queue` is **never constructed anywhere** in `rs/` or `wallet/` — it's genuinely inert, not a partially-wired feature hiding behind honest docs.

I did not find a doc claim unmatched by code, which is the specific defect class #550/#552 is about.

**3. TS side (`types.ts`, `mock.ts`).** Field/shape parity with Rust confirmed (`compromiseRelayUrl`, `Alarm.conversationId`/`relayUrl`); `ErrorCodeSchema` itself was untouched (`relay-protocol-violation`/`send-unavailable` already existed). `mock.ts` doesn't implement any relay-error-code mapping at all — before or after this PR — so it can't be "more permissive" than production on the fail-closed path; the only mock/test changes are the new required-but-null fields needed to keep the widened schema compiling. Not a source of hidden-bug risk.

**Verification.** Ran `cargo test --lib` for `tauri-plugin-f2zmsg`: 67/67 pass, matching the PR's own reported count.

No blocking findings. Approve pending green CI.

skyl (commented):
Fast targeted re-review at 9e07ef7 (post-rebase onto #769), scoped only to the rebase-touched invariants. Verdict: **SAFE TO MERGE**, and the three self-updated digest pins (`deliver`, `ensure_bound`, `set_peer_advert`) are justified.

**1/2. `ensure_bound`/`deliver` ordering (engine.rs).** The #769 insertion is purely additive at the top of each function: `ensure_outbound_connection(&outbound).await?` (ensure_bound:2803, deliver:2725). Its only effect is an early return via `?`/`mark_delivery("pending", ...)` on a failed connect — nothing irreversible has happened yet at that point (no persist, no BIND_SEND, no APPEND), which matches the "acceptable" case in the review brief. Downstream of that insertion, the pre-existing ordering is untouched: `persist_bind_state(..., OutcomeUnknown)` (engine.rs:2822/2831) still commits unconditionally before `connection.bind_send(...)` (engine.rs:2841) in every branch (`Fresh`, `OutcomeUnknown`); `deliver` still calls `ensure_bound` before the single `append_and_confirm_binding` call (engine.rs:2732, 2753-2762). No reorder, no short-circuit around the persist, no early return after an irreversible send.

**3. `Confirmed` persistence order.** `append_and_confirm_binding` (engine.rs:2925-2952) sends the same-key `APPEND` first via `?` — a failure there never reaches the persist — and only then conditionally persists `Confirmed`, and only when `bind == NeedsAppendConfirmation`. In `ensure_bound`, `Confirmed` is persisted immediately after `bind_send` returns `Ok(Bound)` (engine.rs:2843), i.e. after relay success. Both paths match the required "confirmed only after relay success or a successful same-key APPEND" property.

**4. `send_address_is_stolen`.** Still checked at the top of both `deliver` (engine.rs:2712) and `ensure_bound` (engine.rs:2791), both fail-closed (`Err(SendAddressStolen)`, no bypass).

**5. `wire_codes.rs` exhaustive test derivation.** The new `signed` list (`Command::ALL.filter(|c| c.auth() != Auth::None)`) is not an independent redefinition — it's the literal same predicate production's `from_relay` already uses at wire_codes.rs:92 and :143, and `auth()` itself is the canonical protocol-level source of truth in `f2z-codec/src/commands.rs`, separately pinned as reviewed literals against WIRE.md §6 in `auth_and_transcript_address_match_section_6` (confirms `ContactAppend => Auth::None`, `PublishKeyPackages => Auth::RecvKey`, matching the comment about the newly-inserted unsigned `ContactAppend` before the new signed `PublishKeyPackages`). The matrix test's outer loop (wire_codes.rs:302-325) is still the full `WireCode::ALL × Command::ALL × 3 attempts` cartesian product with default-deny (`RelayProtocolViolation`) for anything not explicitly allow-listed — unchanged and not narrowed.

**6. Fail-closed default / `ERR_NO_ACCESS`.** `WireCode::NoAccess` (value 10 per `f2z-codec/src/error.rs`) still sits in the unconditional match arm at wire_codes.rs:77 that maps to `RelayProtocolViolation` for every command, no guard clause, plus a dedicated regression test (`no_access_is_never_ordinary_send_unavailability`, wire_codes.rs:343-354) enforcing it across `Command::ALL`.

**On the self-updated digests specifically:** I don't just trust the hash literals or the rebasing agent's inline comments explaining them. `wallet/zuuli/scripts/messaging-contract.node-test.mjs` has separate, live string-position assertions against the actual `engine.rs` source (not gated behind the digest) that independently re-verify: persist-before-`bind_send` in `ensure_bound` (lines ~442-457), compromise-check-before-`ensure_bound`-before-single-`append_and_confirm_binding` in `deliver`/`send_control` (lines ~465-498), and append-before-Confirmed-persist ordering (lines ~552-564). I ran the actual suite against this PR's real head in a disposable worktree (`git worktree add --detach ... 9e07ef7`, cleaned up after) rather than trusting the PR description: all 9 subtests passed, including the ~10s mutation-testing subtest that applies structural mutations (early-return before `ensure_bound` in `deliver`, reordering `persist_bind_state` vs `bind_send`, reordering the `Confirmed` write vs `.append(...)`) and asserts each one is caught. That gives independent corroboration beyond the three pinned hashes — if the rebase had actually broken the ordering, these positional checks would have failed regardless of whether the digest was "self-updated" correctly.

Did not re-review anything outside the rebase-touched surface (deliver/ensure_bound/ensure_outbound_connection/wire_codes signed-derivation), per the ask.

skyl:

Issue #779, opened 2026-08-30
release(zuuli): deliver internal build 0.1.0+17 to TestFlight and Play


skyl:
## Goal

Ship ZUULI `0.1.0+17` to the internal production channels: Apple TestFlight and Google Play internal testing, from one reviewed immutable `main` commit.

## Why

- TestFlight is current at `0.1.0+16`.
- Play internal is still at build 14; builds 15 and 16 failed before upload in release-only Android steps that are now fixed on `main`.
- `main` contains newer ZUULI messaging UI and reliability work that has not reached either tester cohort.

## Ordering

1. Resolve and merge #769 after adversarial review and both required gates are green.
2. Re-derive `wallet/zuuli/STATUS.md` against the final release-impacting `origin/main` SHA in a docs-only PR; do not soften any disposition or invent physical-device evidence.
3. From the resulting clean `origin/main`, create a separate release worktree and run:
   ```bash
   cd wallet/zuuli
   npm ci
   npm run release:bump -- --version=0.1.0 --build=17
   npm run release:verify
   ```
4. Merge the bump PR only with required `gate` and `ZUULI / packaging smoke` green.
5. Monitor the automatic protected release. Rerun failed jobs only when the exact artifact remains valid; if bytes or source must change, advance to build 18 and never reuse 17.
6. Run exact-build TestFlight read-only recovery and the authenticated both-provider store audit.

## Acceptance

- [ ] Release-impacting prerequisites are merged before the STATUS audit anchor is recorded.
- [ ] STATUS freshness guard passes without overstating runtime/product readiness.
- [ ] `0.1.0+17` is introduced by a focused reviewed PR.
- [ ] TestFlight evidence says uploaded, processed, and available to `ZUULI Internal Testers`.
- [ ] Play audit says exact release 17 is present on the internal track.
- [ ] Signed artifacts, checksums, SBOMs, provenance, and sanitized store evidence are retained.
- [ ] No public-store promotion is attempted; `publicationReady` remains false until its separate blockers close.
- [ ] Release result and any follow-up defect are recorded here without credentials or tester identities.

## Known risk

#754 tracks protected release steps that have not yet completed in a live run. It is not a formal internal-release blocker, but it should be worked in parallel and Android should be watched closely.

skyl:
[...]

Issue #374, opened 2026-08-09
security(reader): prevent creator articles from beaconing wallet-reader activity


skyl:
## Summary

Creator-controlled Markdown can automatically load remote images and media from arbitrary HTTPS origins. Opening an article therefore sends the reader's IP address, timing, User-Agent/network metadata, and an attacker-chosen unique URL token to the creator or any third-party host they select.

This is particularly sensitive in a wallet: reading activity can be correlated with account actions, donation timing, or other identities. The comment renderer already recognizes this risk and blocks remote media; the main article path does not apply the same policy.

## Evidence

- `wallet/zuuli/src/services/http.ts:46-65` preserves arbitrary absolute media hosts.
- `wallet/zuuli/src/components/common/Markdown.tsx:347-355` renders remote images automatically.
- `Markdown.tsx:295-305` renders arbitrary HTTPS audio/video sources.
- Reader content uses this path at `wallet/zuuli/src/features/reader/Reader.tsx:149-151`.
- `wallet/zuuli/src-tauri/tauri.conf.json:22-24` permits `img-src https:` and `media-src https:`.
- The comment path has an explicit remote-media block at `Markdown.tsx:313-325` and `:465-471`, demonstrating the intended safer pattern is already available.

## Required remediation

- Do not automatically fetch creator-selected third-party media in a wallet view.
- Default to a placeholder with explicit click-to-load consent showing the destination hostname.
- Prefer a privacy-preserving owned proxy with cache normalization and request-metadata stripping for supported content.
- Apply the policy consistently to inline images, cover images, audio/video, and provider embeds.
- Set an explicit no-referrer policy as defense in depth; it does not hide IP/timing and is not a complete fix.

## Acceptance criteria

- Opening an article makes no request to a creator-controlled host without an explicit user action.
- Tests cover Markdown images, cover images, audio, video, YouTube/Vimeo embeds, redirects, and protocol-relative/encoded URLs.
- Click-to-load clearly identifies the remote origin and remembers no broader consent by default.
- Comment and article renderers share one remote-media policy.

skyl:
Implementation is ready in #440. It provides one per-item consent policy for article/comment inline media, covers, direct audio/video, and YouTube/Vimeo; proves zero matching request and zero URL-bearing media DOM before consent; and is locally green on the exact Node 24 head (378 Vitest, 13 Node contracts, 79 Playwright). Awaiting required hosted gates/packages; no merge performed.

skyl:
[...]

PR #801, opened 2026-08-30
ZUULI reader: auto-load trusted Free2Z images

### LINKED ISSUE #791: ZUULI reader: auto-load trusted first-party Free2Z images
## Problem

ZUULI currently makes readers tap **Load image** even when the image is served by Free2Z. That adds friction without meaningfully improving privacy: the app is already communicating with Free2Z to render the page.

## Product direction

Auto-load trusted first-party Free2Z images by default. Keep consent or a privacy-preserving proxy for third-party media. An optional strict privacy mode may block first-party images too, but it must not be the default.

## Acceptance criteria

- [ ] Images from approved Free2Z-owned origins load automatically in articles.
- [ ] Redirects are revalidated before loading; a first-party URL cannot redirect into automatic third-party loading.
- [ ] Arbitrary third-party images retain the existing consent/proxy protection.
- [ ] A clearly named strict privacy setting can disable first-party auto-load for users who want it.
- [ ] The default reader has no **Load image** interstitial for `free2z.cash` media.
- [ ] Tests cover canonical Free2Z URLs, subdomains, redirects, lookalike hosts, and strict mode.

Follow-up to #374: preserve its protection against creator-controlled beacons while removing friction for genuinely first-party media.

skyl:
## Summary

- auto-load validated raster images from `free2z.cash` and its HTTPS subdomains
- inspect every redirect before requesting the next hop and render only local blob URLs
- retain per-item consent for third-party media and add persistent Strict image privacy in the reader
- bind native transport/CSP scope to the reviewed first-party image path, including zero-follow native requests
- give the Strict image privacy switch a stable visible-label accessible name while retaining its help description
- preserve current-main RealtimeKit HTTPS/WSS CSP endpoints alongside the exact first-party image transport scope

## Verification

- `npm run typecheck`
- `npm run typecheck:tests`
- `npm run build`
- focused Vitest: 61/61
- native media-authority contract: 10/10 plus live detector
- combined first-party image, third-party consent/redirect, and RealtimeKit CSP Playwright: 18/18
- prior full frontend gate at the superseded head: Vitest 871 passed / 5 skipped; Node contracts 372 passed / 5 platform skips; Playwright 136/136
- prior Zuuallet sensitive-entry companion: 3/3

## Integration

Merged latest `origin/main` normally in `33ed51014c2a5b05ac879718e5a2cf26bedba04c` (base `73b14ff4b7922b7577ac0ac5ed0129aaa5fcabb7`); no rebase or force push. The single CSP conflict retained both #801's `blob:` / exact Free2Z transport boundary and main's RealtimeKit endpoint allowlist, with the exact CSP contract updated accordingly.

## Review

Fresh adversarial exact-head review is in progress for `33ed51014c2a5b05ac879718e5a2cf26bedba04c`. The prior `b03556b868698033aa12feb66b4a8f82b996c327` verdict and CI are superseded.

Closes #791


skyl:
Exact-head review is bound to `ee48f9483d38a9c34b98077b9aac13f5c4faa6fa` and found a blocking P1 in the central redirect boundary.

At this submitted head, `fetchFirstPartyImage` passes browser `RequestInit { redirect: "manual" }` directly to `@tauri-apps/plugin-http`, but the native transport does not forward that browser field to its Rust client. A packaged mobile/desktop build can therefore auto-follow a trusted `free2z.cash` URL into a third-party redirect before `downloadTrustedFirstPartyImage` receives and validates the response. That defeats #791's required per-hop revalidation on the shipping native path.

Please disable native redirects through the plugin's explicit transport option (for example `maxRedirections: 0`) and bind that invariant with a source/contract test, then push and obtain fresh exact-head CI/review. I am reviewing only the submitted Git head; uncommitted owner worktree state is not a verdict. Issue linkage and one-change scope are otherwise coherent.

skyl:
Rebound to new exact head `90cb5787ab1238830d8bf3079bff54d3eceb1e93`. The prior native redirect P1 is fixed: packaged Tauri fetch now uses explicit `maxRedirections: 0`, returning every 3xx to the existing per-hop trusted-host validator, and a source-bound test holds that shipping call shape.

I found no remaining P0/P1/P2 in the current diff's URL canonicalization, dot-boundary host trust, manual redirect loop, opaque/unobserved redirect refusal, raster MIME/20 MiB limits, blob-only auto-rendering, strict/per-item consent behavior, or CSP/capability scope. `Closes #791` and the logical scope are coherent.

This is not a merge verdict yet: all previous-head checks are obsolete and current-head CI/review are pending. Any push resets this binding. It is release-impacting and must precede #779's final STATUS anchor if merged.

skyl:
Exact-head P1/QA blocker on `90cb5787ab1238830d8bf3079bff54d3eceb1e93`: the required frontend lane failed deterministically in job https://github.com/free2z/zuu/actions/runs/33294236766/job/99211510308.

All four touch/a11y matrix cases failed, including their retries (signed out/in at 320px and 360px), with `/articles/why-shielded-by-default: button has no accessible name`; the run finished with 132 passing tests and these 4 failures.

The failure is introduced by this PR's new `ImagePrivacySetting` at `Reader.tsx:134`. Its `Switch` at `ImagePrivacySetting.tsx:26-32` renders a Radix `<button>` but supplies only `id="strict-image-privacy"` and `aria-describedby`; it has no `aria-label` or `aria-labelledby`. The required audit's accessible-name contract therefore sees the new button as unnamed.

Please give the switch an explicit stable accessible name (for example, bind `aria-labelledby` to an id on the visible “Strict image privacy” label, preserving the help text as the description) and add focused rendered/accessibility coverage. Push a new head and obtain fresh exact-head CI/review. The previous native redirect P1 remains fixed in this head, but CI is never bypassable.

skyl:
Adversarial exact-head review for `b03556b868698033aa12feb66b4a8f82b996c327`

Verdict: **no P0, P1, or P2 findings**.

I reviewed the complete diff against current `origin/main`, its focused tests, and the exact-head hosted checks. Local focused evidence was green: media-authority Node tests 10/10 plus the live detector, relevant Vitest 61/61, and focused Playwright 16/16. Hosted `zuuli / frontend`, `gate`, `rs / gate`, native Apple lanes, and `Desktop / macOS universal app` are all successful at this exact SHA.

Validated invariants:

- only first-party images may auto-load; third-party media remains inert until explicit, item-scoped consent
- third-party beacon and redirect chains fail closed, including final-origin validation
- requests omit credentials and referrer data; abort and object-URL cleanup remain deterministic
- MIME, SVG, and 20 MiB limits remain enforced
- the real wallet gate fails closed instead of exposing the gated action
- consent and failure states retain accessible names, keyboard behavior, status communication, and mobile/RTL usability

This verdict is bound only to `b03556b868698033aa12feb66b4a8f82b996c327`.

skyl (commented):
Adversarial review, privacy-focused, on the owner's behalf. Verdict: APPROVE (posted as a comment review, not a formal approval — GitHub blocks self-approval since this PR and my CLI session share the `skyl` account; treat this as the reviewer-agent's real sign-off).

**1. What counts as "trusted first-party"?** `wallet/zuuli/src/lib/media/remote-media-policy.ts`, `isTrustedFirstPartyImageTarget` (new): requires `protocol === "https:"` AND (`hostname === "free2z.cash"` OR `hostname.endsWith(".free2z.cash")`), with an explicit dot-boundary — this is NOT a `startsWith`/substring check, so it correctly rejects the exact attack you asked me to check: `free2z.cash.evil.example` and `notfree2z.cash` both fail (covered by `remote-media-policy.test.ts`'s "does not approve lookalikes" cases, and by the mock-data fixture `https://free2z.cash.evil.example/pixel.svg` plus the Playwright e2e in `tests/first-party-images.pw.ts`, which asserts that host is never even requested). Ports are also pinned: `if (parsed.port) return false` after WHATWG URL parsing strips the default `:443`, so `:8443` is rejected but `:443` (which normalizes to no port) passes — verified this is correct in the URL spec, and it's tested (`free2z.cash:443` approved, `free2z.cash:8443` rejected). Userinfo (`user:pass@host`) is also rejected. IDNA lookalikes (`xn--free2z-9za.cash`) are tested and rejected. Only exception: an exact-origin `http://` loopback (the Vite dev proxy), gated to `isLoopback()` + same-origin — not reachable in production. This is a solid allowlist, not a naive `includes()`/`startsWith()` check.

**2. Does auto-loading leak IP/timing to anyone but Free2Z?** No, and the implementation goes further than a plain `` would: `downloadTrustedFirstPartyImage` in the same file re-validates `isTrustedFirstPartyImageTarget` on *every* redirect hop before following it (`redirect: "manual"`, inspect `Location`, reject if the next hop isn't in the trusted zone) — this specifically closes the TOCTOU hole where a first-party URL 302s to a third-party tracker. Tested directly: `stops before requesting an unapproved redirect` in `remote-media-policy.test.ts`, and end-to-end in `first-party-images.pw.ts` (`/uploadz/redirect-out` → `tracker.example` is set up and asserted never contacted). Requests are also sent with `credentials: "omit"` and `referrerPolicy: "no-referrer"`, so even the trusted first-party fetch doesn't leak cookies/referrer it doesn't need to. Bytes are pulled into a `Blob`/local `blob:` object URL rather than handing the DOM a live network URL, so there's no live `` beaconing at all for trusted images. Non-first-party images are unaffected — they still go through the pre-existing manual-consent `RemoteMedia` path.

**3. CSP / capabilities — does widening img-src also widen script/connect/frame, and does this worsen #367 or regress #818's RealtimeKit CSP?**
- `tauri.conf.json` diff is exactly two directives: `img-src` gains `blob:` (needed because the new code renders `URL.createObjectURL(blob)` results, previously CSP only had `'self' data: https:'`), and `connect-src` gains `https://free2z.cash` + `https://*.free2z.cash` (needed for the `window.fetch` fallback path used on non-Tauri/web builds — packaged Tauri builds instead use the native `@tauri-apps/plugin-http` fetch, which isn't subject to page CSP but is gated by the `http:default` capability allowlist instead). **`script-src` (`'self' 'wasm-unsafe-eval'`) and `frame-src` (`'none'`) are byte-for-byte unchanged.** So this cannot re-open #367 (privileged IPC via remote iframes) — #367 is about `frame-src`/iframe embeds and the `zcash:default`/`opener:default` capability grants, none of which this PR touches.
- `tests/realtimekit-csp.pw.ts` diff only *inserts* the two new `free2z.cash` connect-src entries into `expectedConnectSources`; every RealtimeKit endpoint from #818 (`api.realtime.cloudflare.com`, `*.dyte.io`, `socket-edge.realtime.cloudflare.com`, etc.) stays in the same array, unremoved and still asserted. **This is additive, not a regression of #818.**
- `capabilities/default.json` and `capabilities/mobile.json` gain `https://*.free2z.cash/*` in the `http:default` allow-list (native fetch plugin scope) — a real but narrow widening (any free2z.cash subdomain can now be hit by native fetch, vs. only the apex + `stage.` before). This is consistent with `env.ts`'s existing `DEFAULT_API = "https://free2z.cash"` and is enforced/pinned by `scripts/mobile-webview-authority.mjs`'s updated `REVIEWED_HTTP_URLS` contract test, so a future silent widening would fail CI. Worth noting for the record, not a blocker: it does not touch `zcash:*`/`f2zmsg:*` wallet-command permissions or `opener:default`, so it doesn't expand the IPC/command surface #367 is about — only the native HTTP fetch surface, scoped to the same DNS zone the app already trusts for its API.

**4. Is the default safe, and is the setting honored?** Default is auto-load-first-party (matches issue #791's acceptance criteria verbatim: "must not be the default" for strict/blocking mode). Third-party images are unaffected by the default either way — they still require a tap, in both modes. The `Strict image privacy` toggle is real, not cosmetic: `RemoteMedia.tsx`'s branch is `trustedFirstPartyImage && (!strictImagePrivacy || consentedUrl === target.url)` — when strict is on and the image hasn't been individually consented, it falls through to the same manual `Load image` button as third-party media (tested in `Markdown.test.tsx` and the strict-mode Playwright test, which asserts zero network requests fire until the button is clicked). Preference persists via `localStorage` (`STRICT_IMAGE_PRIVACY_KEY`), read/write wrapped in try/catch so private-browsing/storage-blocked contexts fail closed to the in-memory default rather than throwing.

**House-rule check (no theatrical controls):** the strict-mode toggle is a client-side self-protection setting — it changes what the *user's own, unmodified* app does by default, not a boundary meant to resist a hostile client. That's the right frame for it (nobody is attacking themselves by disabling their own consent prompt), so I'm not counting it as a security control that a modified client could "bypass." The actual privacy-relevant boundary — which network origins get contacted automatically at all — is enforced by `isTrustedFirstPartyImageTarget`/`downloadTrustedFirstPartyImage`, which are real runtime checks (not just UI copy) with redirect revalidation, content-type sniffing (rejects SVG and non-raster types, closing the beacon-via-subresource vector `#374` was about), and a size cap. That's a substantive control, not an advisory one.

No blocking issues found. Nicely defensive implementation (redirect revalidation, credential/referrer stripping, SVG rejection, size cap, exact-zone matching) — this clears a materially higher bar than "just add the domain to an allowlist and call it a day."

PR #802, opened 2026-08-30
fix(relay-testkit): close established connections on shutdown

### LINKED ISSUE #766: f2z-relay-testkit: the double has no server-shutdown path at all, so nothing can hold it in step with production's
Follow-up to #678 / PR #764, deliberately left out of that PR and reported rather than done.

## The correction

#678's divergence table has a "shutdown" row claiming production sends `Outbound::close(CLOSE_GOING_AWAY)` while the testkit `break`s with no close frame. **That row compares two different arms**, and PR #764's author caught it:

- **Production has two watches.** `closed_rx` breaks with no frame — correctly, because the writer has already sent the close. `shutdown` sends `CLOSE_GOING_AWAY` and then breaks.
- **The testkit has one**, `shutdown_rx`, and it is the *former* — the writer-finished signal, not server shutdown.

So the testkit does not diverge on server shutdown. **It has no server-shutdown path at all.** `RelayServer::shutdown` only stops accepting new connections; nothing tells an established connection's loop to send a close and wind down.

The other two rows of that table were real and are fixed in #764: `CLOSE_GOING_AWAY` now exists in `f2z-relay-testkit` (it was not defined in the crate at all) and the keepalive-failure close is 1001 in both implementations, with a unit test in each crate pinning all four codes independently.

## Why this is worth its own issue

The testkit is the second implementation the conformance vectors exist to hold in step with `f2z-relay`. A behaviour production has and the double **cannot express** is a gap in the double — that was the argument for adding the missing constant in #764, and it applies here with more force: a missing close-code constant is one value, a missing shutdown path is a whole arm of the state machine.

Concretely, nothing today can hold the two implementations in step on: does an established connection get a close frame when the server goes down, with which status, and does it get it before the socket dies?

## Why #764 did not do it

Adding it needs a new channel plumbed into the testkit's `drive`, which changes that function's signature. That is a larger change than the issue it came from was about, and it does not belong bundled with a test-coverage PR. Doing it separately also means it gets its own falsification rather than riding along on someone else's.

## Wanted

1. Give `f2z-relay-testkit` a real server-shutdown signal, distinct from the existing writer-finished `shutdown_rx`, so an established connection can be told to wind down.
2. Make an established connection send `CLOSE_GOING_AWAY` on that signal, matching production.
3. **Prove the arm is exercised.** Break it and show a test turns red — a shutdown path nobody has watched fail is not a shutdown path. A positive control is needed too: a connection that is *not* being shut down must not receive a close frame.
4. Decide whether a conformance vector should assert this, or whether it stays a crate-level test. WIRE.md names no status for shutdown, so **this is alignment with production, not conformance** — do not write it up as a spec requirement.

## Scope

`rs/` only. Do not change production `f2z-relay` behaviour to match the testkit; production is the reference here, and its two-watch structure is correct as it stands.

skyl:
Closes #766

## Summary
- give `connection::drive` a server-shutdown watch distinct from its writer-finished watch
- send WebSocket status 1001 before an established testkit connection winds down, matching production
- track listener-owned connection tasks so `RelayServer::shutdown().await` waits until their close path finishes
- keep in-process connections independent of the listener shutdown lifecycle

## Test rationale
`WIRE.md` names no shutdown close status, so this stays a crate-level production-alignment test rather than a conformance vector.

The new real-WebSocket test was first run against the pre-fix implementation and failed red by timing out while waiting for the shutdown frame. With the implementation present it asserts the target server sends `CLOSE_GOING_AWAY`, and its positive control proves a connection to an independently running server receives no frame.

## Verification
- `cargo +1.97.1 test --locked -p f2z-relay-testkit --test server_shutdown`
- `cargo +1.97.1 test --locked --all-targets`
- `cargo +1.97.1 test --locked --doc`
- `scripts/check-rust-clippy.sh --root rs`
- `scripts/check-rust-deny.sh --root rs --config rs/deny.toml`
- `git diff --check`

### DIFF
--- a/rs/crates/f2z-relay-testkit/src/connection.rs
+++ b/rs/crates/f2z-relay-testkit/src/connection.rs
@@ -33,15 +33,19 @@ use crate::faults::{Effect, FaultInjector};
 use crate::outbound::{CLOSE_GOING_AWAY, CLOSE_NORMAL, CLOSE_PROTOCOL_ERROR, OutKind, Outbound};
 use crate::transport::{Transport, TransportSink, WireMessage};
 
-/// Serve one connection until either end closes.
-pub async fn drive(relay: Arc<Relay>, transport: Transport) {
+/// Serve one connection until either end closes or the server shuts down.
+pub async fn drive(
+    relay: Arc<Relay>,
+    transport: Transport,
+    mut server_shutdown: watch::Receiver<bool>,
+) {
     let (sink, mut stream) = transport.split();
     let (outbound_tx, outbound_rx) = mpsc::unbounded_channel::<Outbound>();
-    let (shutdown_tx, mut shutdown_rx) = watch::channel(false);
+    let (writer_finished_tx, mut writer_finished_rx) = watch::channel(false);
 
     let mut connection = relay.open_connection(outbound_tx.clone());
     let faults = relay.faults().clone();
-    let writer = tokio::spawn(writer_task(sink, outbound_rx, faults, shutdown_tx));
+    let writer = tokio::spawn(writer_task(sink, outbound_rx, faults, writer_finished_tx));
 
     let handshake_timeout = relay.config().handshake_timeout;
     let ping_interval = relay.config().ping_interval;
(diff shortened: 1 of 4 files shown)


skyl:
Exact-head review is bound to `351cdd0f72cfb50fe68db29061b200e8f76c8a31`. I found no unresolved P0/P1/P2. `Closes #766`, the four-file testkit diff is one logical server-shutdown change, and `git diff --check` is clean.

Adversarial trace: the listener owns its connection `JoinSet`; the server-shutdown watch is distinct from writer-finished; an established `drive` enqueues close 1001, drops both outbound senders, and awaits the writer; the writer closes the sink and finishes before the listener's JoinSet drain lets `shutdown().await` return. In-process `FakeRelay` retains an independent false sender for exactly the lifetime of `drive`. The real-socket regression covers target-server 1001 and an independently live server as a no-frame control, matching the issue's crate-level/non-WIRE scope.

This is not merge-ready yet: exact-head CI is still running and the head has no approval. Any push resets this binding. It can land independently before #769; it does not affect the store release surface.

skyl:
Rebound to exact head `c298d1714e4ec01eb6cb93e49fc0a0ea10155258` after #782. The delta is a pure rebase onto `4b939558a21dc38164ff2b2fb56263ebc855acf2`: range-diff is unchanged and the patch-id remains `28e44f04d67f4617d303e6cedc36b4932754712e`. The prior adversarial verdict therefore rebinds with no unresolved P0/P1/P2; scope/linkage/release impact are unchanged and `git diff --check` is clean.

Fresh strict-base CI is only starting, so this is not a merge verdict yet. Any push resets this binding.

PR #797, opened 2026-08-30
fix(zuuli): keep search copy concise

skyl:
## Summary
- remove algorithm and implementation explainers from global and article search
- shorten visible search placeholders while preserving detailed accessible names
- replace exposed server diagnostics with concise catalog-backed recovery guidance
- add a focused source regression plus 320/360px, de-DE, 200% text, and Spanish runtime coverage
- integrate finalized i18n kernel commit `08630946b746a58decc210c1ac6832ecfa9af149` and route every changed string through exact-parity English, Spanish, and French catalogs

## Integration
- merged current `origin/main` at `a4478fb1e920bc022a9ab49518d4f26264442837` normally; no rebase or force push
- normally merged the finalized #325 kernel commit; this imports 40 files and about 4,040 added lines before this PRs focused catalog wiring
- removed the redundant no-creator description instead of translating a sentence that repeated the empty-state title

## Gates
- `npm ci`
- `npm run typecheck`
- `npm run typecheck:tests`
- `npm run build`
- `npx vitest run src/i18n src/app-bootstrap.test.tsx` — 117 passed
- `ZUULI_PW_PORT=1797 npx playwright test tests/ui-copy.pw.ts tests/i18n-kernel.pw.ts tests/search.pw.ts` — 9 passed
- `ZUULI_PW_PORT=1797 npm test` — 978 Vitest passed / 5 expected skipped; 372 Node policy passed / 5 expected skipped; 139 Playwright passed
- GitHub Actions pin policy plus self-test
- Rust toolchain policy plus self-test
- `git diff --check`

## Ordering
Land this before #803 because both update Search and Articles copy-adjacent UI.

Closes #786

Issue #786, opened 2026-08-30
ZUULI UI: remove implementation-explainer copy and enforce concise translatable microcopy


skyl:
## QA finding

ZUULI leaks implementation/explainer prose into the interface. Example from the Search surface:

> Semantic search — results ranked by meaning, not just keywords.

This reads like a code comment or product pitch, not elegant interface copy. Remove it.

## Product rule

UI copy must be concise, intuitive, and necessary. Do not explain technical implementation details, ranking algorithms, architecture, transport, crypto, storage, or other internals in normal product chrome. The interface should communicate the user action, state, or decision—not narrate how the software works.

This is especially important for the planned 50+ language rollout: every unnecessary sentence multiplies translation cost, layout risk, review burden, and inconsistency.

## Acceptance

- [ ] Remove the quoted Semantic Search explainer from the shown ZUULI surface; do not replace it with another sentence describing the algorithm.
- [ ] Audit the same Search feature for adjacent implementation-explainer copy and reduce it to essential labels, states, and actions.
- [ ] Audit ZUULI product chrome for similar code-comment leakage: technical descriptions that do not help the user make a decision or recover from a problem.
- [ ] Keep user-facing errors actionable but short; put diagnostic/engineering detail in logs or an intentional support/details affordance, not the default UI.
- [ ] Preserve necessary security, money, consent, destructive-action, and recovery warnings; concision must not remove information required for a safe decision.
- [ ] Prefer established icons, labels, direct verbs, and stateful controls over explanatory paragraphs.
- [ ] New/changed copy uses the i18n system and avoids concatenated sentence fragments that become untranslatable.
- [ ] Mobile viewport tests prove copy does not overflow, truncate, or create excess vertical noise at 320/360px and with text expansion.
- [ ] Add a focused regression for the reported Search surface and a lightweight source audit/baseline that prevents the removed phrase or equivalent technical explainer from returning.

## Evidence

Founder QA screenshot on 2026-08-30 shows the line directly beneath Search. The screenshot itself may remain in the private QA session; do not add unrelated private/local files to the public repo.

## Related

- #322 — ZUULI i18n epic
- #325 — i18n kernel/string extraction
- #395 — mobile polish audit

PR #803, opened 2026-08-30
feat(zuuli): make discovery autocomplete-first

### LINKED ISSUE #788: ZUULI search: replace the topic-pill wall with autocomplete-first discovery
QA finding (2026-08-30): the current ZUULI search screen is still an unbounded wall of topic pills. This issue must deliver autocomplete-first discovery, not merely a collapsed version of the same wall.

Updated product direction:

- Search-as-you-type is the primary interaction.
- Rank concise suggestions across relevant topics/tags, creators, and content.
- Show recent/trending suggestions only when useful; never dump the vocabulary on screen.
- Deduplicate language variants and make ranking locale-aware for 50+ languages.
- Keep selected filters compact and removable.
- Cover keyboard, screen reader, touch, loading, empty, error, cancellation, and stale-response behavior.
- The default mobile view must remain visually quiet before and during search.

Screenshot supplied during QA; the current pill wall is the failure case.

skyl:
Closes #788

## Summary
- replace the Articles topic-pill wall with a quiet topic autocomplete and compact removable selections
- add mixed topic, creator, and page suggestions to global Search with locale-aware ranking and stable-identity dedupe
- preserve keyboard-first selection, screen-reader status, touch access, cancellation, stale-response rejection, explicit retryable errors, and paginated result safety
- keep legacy comma-bearing tags displayable but never selectable as ambiguous feed filters

## Remediation
- expose article-topic transport/schema failures so the filter can distinguish unavailable data from an empty corpus
- keep case-variant creators and same-title pages distinct by immutable backend identity
- keep autocomplete panels in normal flow so loading/empty states cannot intercept mobile tabs
- defer click-away collapse through the complete pointer transaction so controls below the in-flow panel do not shift before their click; globally settle pointer releases/cancellation
- use concise Search and Articles copy without reintroducing a topic pill wall

## Exact-head verification
Verified at `00c583aa552a92313d7e24a2f23ee5830ef12c2c`:
- `npm run typecheck`
- `npm run typecheck:tests`
- `npm run build` with fresh 116-byte WASM verification
- focused discovery Vitest: 6/6
- copy policy Node: 2/2
- full Vitest: 853 passed, 5 skipped
- full Node policy: 370 passed, 5 skipped
- full Playwright: 145/145, including all autocomplete, pagination, remote-media consent, mobile tab, locale, pointer-release, and current-main RealtimeKit CSP cases
- prior repair stress remains green: pagination 15/15, media consent 5/5, lost pointer release 5/5

## Integration
This branch is self-contained and normally merged current `origin/main` at `73b14ff4b7922b7577ac0ac5ed0129aaa5fcabb7` in merge commit `00c583aa552a92313d7e24a2f23ee5830ef12c2c`. It does not depend on #786 or #797 and requires no rebase.

## Review
Fresh adversarial exact-head review is in progress for `00c583aa552a92313d7e24a2f23ee5830ef12c2c`; the earlier `6011bb` verdict is superseded.

### DIFF
--- a/wallet/zuuli/src/components/layout/TopBar.tsx
+++ b/wallet/zuuli/src/components/layout/TopBar.tsx
@@ -19,11 +19,7 @@ import {
   TooltipContent,
   TooltipTrigger,
 } from "@/components/ui/tooltip";
-import {
-  Avatar,
-  AvatarFallback,
-  AvatarImage,
-} from "@/components/ui/avatar";
+import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -57,8 +53,7 @@ export function TopBar() {
   // link starts at 0). Reading it during render keyed on `location` keeps it
   // reactive. Hidden on the root/home route, where "back" is meaningless, so
   // it never dead-ends the user out of the app.
-  const historyIdx =
-    (window.history.state?.idx as number | undefined) ?? 0;
+  const historyIdx = (window.history.state?.idx as number | undefined) ?? 0;
   const canGoBack = historyIdx > 0 && location.pathname !== "/";
 
   return (
--- a/wallet/zuuli/src/features/articles/components/TopicFilterAutocomplete.tsx
+++ b/wallet/zuuli/src/features/articles/components/TopicFilterAutocomplete.tsx
@@ -172,7 +172,7 @@ export function TopicFilterAutocomplete({
             onClick={() => onChange([])}
             className="min-tap inline-flex items-center rounded-full px-2.5 py-1 text-xs font-medium text-muted-foreground hover:text-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
           >
-            Clear tags
+            Clear topics
           </button>
         </div>
       ) : null}
--- a/wallet/zuuli/src/features/articles/components/TopicFilterAutocomplete.tsx
+++ b/wallet/zuuli/src/features/articles/components/TopicFilterAutocomplete.tsx
@@ -1,0 +1,279 @@
+import { useEffect, useId, useMemo, useRef, useState } from "react";
+import { Loader2, Search, X } from "lucide-react";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { articles } from "@/lib/api/free2z";
+import { rankTopicSuggestions } from "@/lib/discovery-autocomplete";
+import type { ArticleTagSuggestion } from "@/lib/api/types";
+import { MAX_ARTICLE_TAGS } from "@/lib/article-tags";
+
+interface TopicFilterAutocompleteProps {
+  selected: string[];
+  onChange: (topics: string[]) => void;
+}
+
+const DEBOUNCE_MS = 180;
+
+export function TopicFilterAutocomplete({
+  selected,
+  onChange,
+}: TopicFilterAutocompleteProps) {
+  const listId = useId();
+  const statusId = useId();
+  const requestRef = useRef(0);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [draft, setDraft] = useState("");
+  const [focused, setFocused] = useState(false);
+  const [dismissed, setDismissed] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [initialized, setInitialized] = useState(false);
+  const [error, setError] = useState(false);
+  const [retryNonce, setRetryNonce] = useState(0);
+  const [activeIndex, setActiveIndex] = useState(-1);
+  const [rawSuggestions, setRawSuggestions] = useState<ArticleTagSuggestion[]>(
+    [],
+  );
+  const selectedKey = selected.join("\u0000");
+  const locale =
+    typeof navigator === "undefined" ? undefined : navigator.language;
+  const suggestions = useMemo(
+    () => rankTopicSuggestions(rawSuggestions, draft, selected, locale),
+    // selectedKey intentionally tracks semantic selection without treating a
+    // newly allocated equivalent array as a changed request.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [draft, locale, rawSuggestions, selectedKey],
+  );
+  const hasDraft = draft.trim().length > 0;
+  const full = selected.length >= MAX_ARTICLE_TAGS;
+  const open = focused && !dismissed && hasDraft && !full;
+
+  useEffect(() => {
+    const request = ++requestRef.current;
+    setActiveIndex(-1);
+    if (!focused || !hasDraft || full) {
+      setLoading(false);
+      setInitialized(false);
+      setRawSuggestions([]);
+      return;
+    }
+
+    setLoading(true);
+    setInitialized(false);
+    setError(false);
+    const timer = window.setTimeout(() => {
+      void articles
+        .suggestTags(draft, selected)
+        .then((next) => {
+          if (requestRef.current !== request) return;
+          setRawSuggestions(next);
+          setInitialized(true);
+        })
+        .catch(() => {
+          if (requestRef.current !== request) return;
+          setRawSuggestions([]);
+          setError(true);
+          setInitialized(true);
+        })
+        .finally(() => {
+          if (requestRef.current === request) setLoading(false);
+        });
+    }, DEBOUNCE_MS);
+
+    return () => {
+      requestRef.current += 1;
+      window.clearTimeout(timer);
+    };
+    // selectedKey captures selected without firing for an equivalent array.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [draft, focused, full, hasDraft, retryNonce, selectedKey]);
+
+  useEffect(() => {
+    setActiveIndex((current) =>
+      current >= suggestions.length ? suggestions.length - 1 : current,
+    );
+  }, [suggestions.length]);
+
+  function addTopic(name: string) {
+    if (full) return;
+    onChange([...selected, name]);
+    setDraft("");
+    setDismissed(false);
+    setRawSuggestions([]);
+    inputRef.current?.focus();
+  }
+
+  function removeTopic(name: string) {
+    onChange(selected.filter((topic) => topic !== name));
+    inputRef.current?.focus();
+  }
+
+  function onKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
+    if (event.key === "ArrowDown") {
+      event.preventDefault();
+      setDismissed(false);
+      setActiveIndex((current) =>
+        suggestions.length === 0
+          ? -1
+          : Math.min(current + 1, suggestions.length - 1),
+      );
+    } else if (event.key === "ArrowUp") {
+      event.preventDefault();
+      setActiveIndex((current) =>
+        suggestions.length === 0
+          ? -1
+          : current <= 0
+            ? suggestions.length - 1
+            : current - 1,
+      );
+    } else if (
+      event.key === "Enter" &&
+      activeIndex >= 0 &&
+      activeIndex < suggestions.length
+    ) {
+      event.preventDefault();
+      addTopic(suggestions[activeIndex].name);
+    } else if (event.key === "Escape") {
+      event.preventDefault();
+      setDismissed(true);
+      setActiveIndex(-1);
+    } else if (event.key === "Backspace" && !draft && selected.length > 0) {
+      removeTopic(selected[selected.length - 1]);
+    }
+  }
+
+  const status = loading
+    ? "Loading topics"
+    : error
+      ? "Topics unavailable"
+      : initialized && suggestions.length === 0
+        ? "No topics"
+        : suggestions.length > 0
+          ? `${suggestions.length} topics`
+          : "";
+
+  return (
+    <div className="mb-6 max-w-md" data-topic-filter-autocomplete>
+      {selected.length > 0 ? (
+        <div className="mb-2 flex flex-wrap gap-2" aria-label="Selected topics">
+          {selected.map((topic) => (
+            <button
+              key={topic}
+              type="button"
+              onClick={() => removeTopic(topic)}
+              aria-label={`Remove topic ${topic}`}
+              className="min-tap inline-flex max-w-full items-center gap-1.5 rounded-full border border-primary/40 bg-primary/10 px-3 py-1 text-left text-xs font-medium text-primary focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+            >
+              <span className="min-w-0 break-words">#{topic}</span>
+              <X className="h-3.5 w-3.5 shrink-0" aria-hidden />
+            </button>
+          ))}
+          <button
+            type="button"
+            onClick={() => onChange([])}
+            className="min-tap inline-flex items-center rounded-full px-2.5 py-1 text-xs font-medium text-muted-foreground hover:text-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+          >
+            Clear tags
+          </button>
+        </div>
+      ) : null}
+
+      <div
+        className="relative"
+        onFocus={() => setFocused(true)}
+        onBlur={(event) => {
+          if (!event.currentTarget.contains(event.relatedTarget)) {
+            setFocused(false);
+            setActiveIndex(-1);
+          }
+        }}
+      >
+        <Search
+          className="pointer-events-none absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground"
+          aria-hidden
+        />
+        <Input
+          ref={inputRef}
+          type="text"
+          value={draft}
+          onChange={(event) => {
+            setDraft(event.target.value);
+            setDismissed(false);
+          }}
+          onKeyDown={onKeyDown}
+          placeholder={full ? "Maximum topics selected" : "Filter by topic"}
+          aria-label="Filter by topic"
+          role="combobox"
+          aria-autocomplete="list"
+          aria-expanded={open && suggestions.length > 0}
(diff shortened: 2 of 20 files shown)


skyl:
Exact-head review is bound to `1616e7aadbf57738904070e61cde1d37882ddf0d`. Four blocking P1 findings remain:

1. The PR explicitly depends on #786/#797 landing first and then rebasing. #797 is still blocked on #786's i18n acceptance, and the two PRs overlap Search/Feed. This intentionally pre-dependency head cannot receive an exact-base merge verdict.
2. Articles retry/error UI is unreachable: `TopicFilterAutocomplete` expects `articles.suggestTags(...).catch(...)`, but `suggestTags` catches every network/schema/validation failure internally and resolves `[]`. Production can render only “No topics,” never “Topics unavailable”/retry, contrary to #788.
3. Comma-bearing stored tags regress: `sanitizeArticleTags` deliberately preserves them, while the old feed guarded them with `isArticleTagFilterable`. The new autocomplete suggests/selects them and serializes with `canonical.join(",")`, turning one stored tag into multiple filters on reload.
4. Creator/page dedupe uses locale-normalized username/title rather than entity identity. Real distinct case-variant accounts are documented in this repo, so one creator can be hidden/routed as another; distinct pages with the same normalized title are also collapsed despite different IDs/authors. Dedupe topics by locale label, but creators/pages by their stable entity identity.

Add focused regressions for each behavior, land/fix #797 first, then update this branch to resulting main and rerun exact-head review/CI. `Closes #788` and the overall feature scope are otherwise coherent.

skyl:
Exact-head QA adds another P1 blocker on `1616e7aadbf57738904070e61cde1d37882ddf0d`: required `zuuli / frontend` failed in run `33294358962`, job `99211680513`. Four 320/360 signed-in/out touch-target cases time out because the autocomplete loading/“No matches” overlay intercepts clicks on the Pages tab; retries fail the same way. This is deterministic UI behavior, not a gate to rerun as flaky.

Please ensure the suggestion/status overlay closes or cannot intercept unrelated tab navigation, and add a focused regression that changes Search tabs while loading/empty at 320px and 360px. The dependency and four earlier correctness P1s also remain open; this head is behind current main.

skyl:
Adversarial exact-head review is bound to `6011bb125872433ad42d2b3aa99fa241a5c8bc56`, diffed against current `origin/main` `a4478fb1e920bc022a9ab49518d4f26264442837`. The branch integrated that base by normal merge `ad1c2ebdf411f01707df8e9f083f95d8d0bb61ea`; the PR body now correctly marks the former #786/#797 dependency and old-head verdict superseded.

I found no remaining P0/P1/P2.

The exact diff repairs every documented blocker: article-topic transport/schema failures reach the retryable UI; comma-bearing legacy tags remain displayable but cannot become ambiguous comma-delimited filters; creators and pages dedupe by immutable entity identity; and the bounded in-flow loading/empty autocomplete no longer intercepts 320/360px tab controls. I also reviewed the follow-up pointer transaction repair: result-card media consent and other controls retain a stable target through click, window-level pointer release/cancel clears drag-out state, keyboard/outside blur still dismisses, autocomplete-contained choices retain their own handler ordering, and unmount clears listeners/timers.

Preservation review covered autocomplete-first behavior, locale-aware normalization/ranking, the quiet initial and bounded loading UI, compact removable selections with no vocabulary pill wall, debounce/request-generation cancellation, stale-response rejection, partial/error retry behavior, pagination/cache identity, keyboard/active-descendant/screen-reader semantics, touch selection, and the existing third-party media-consent boundary.

Independent exact-head evidence:
- focused Vitest: 22/22
- copy-policy Node tests: 2/2
- focused Playwright: 4/4 (mixed keyboard/screen-reader suggestions, deterministic stale generation + retry, lost pointer release, and the formerly failing card consent)
- a real `hasTouch` Chromium tap selected `#zcash` and produced `?topic=zcash`
- owner full evidence reviewed coherently: complete Playwright 143/143; pagination stress 15/15; consent stress 5/5; lost-release stress 5/5
- exact-head `zuuli / frontend` and `rs / gate` are green

Other Rust/native/package lanes are still running or queued with no failures, so aggregate merge readiness still awaits their required gate. Any push invalidates this verdict.

skyl:
Fresh adversarial exact-head review is bound to `00c583aa552a92313d7e24a2f23ee5830ef12c2c`, diffed against current `origin/main` `73b14ff4b7922b7577ac0ac5ed0129aaa5fcabb7`. This supersedes my prior `6011bb125872433ad42d2b3aa99fa241a5c8bc56` comment.

Verdict: **no P0, P1, or P2 findings**.

The full discovery diff remains coherent after the normal, conflict-free main merge. Current main adds the RealtimeKit package/CSP contract only; it does not alter the autocomplete, pagination, article-topic, or pointer transaction code. Both RealtimeKit positive and negative CSP controls pass alongside the existing result-card media-consent boundary.

I revalidated every prior repair and preservation invariant: retryable topic transport/schema errors, refusal of ambiguous comma-bearing filters, stable creator/page identity, mobile tab operability, autocomplete-first bounded mixed suggestions, locale-aware ranking/dedupe, quiet initial/loading UI, compact selected filters with no vocabulary pill wall, debounce/request-generation cancellation, stale-response rejection, keyboard/active-descendant/screen-reader behavior, touch selection, pointer release/cancel cleanup, and stable target ordering for card consent.

Independent exact-head evidence:
- focused Vitest: 22/22
- copy-policy Node tests: 2/2
- focused Playwright: 6/6, covering both RealtimeKit CSP controls, mixed keyboard/screen-reader discovery, deterministic stale generation + retry, lost pointer release, and the formerly failing card consent
- owner complete evidence reviewed: Vitest 853 passed / 5 skipped; Node 370 passed / 5 skipped; Playwright 145/145
- exact-head `zuuli / frontend` and `rs / gate` are green

Remaining native/package lanes are still running or queued without failures, so aggregate merge readiness continues to await the required gate. Any push invalidates this verdict.

skyl (commented):
## Adversarial review — verified at head `00c583aa5`

Re-ran independently rather than trusting the PR description: `npm run typecheck` clean, `vitest run` on the three touched/added unit-test files (19/19 pass), and `node --test scripts/ui-copy-truncation.node-test.mjs` (2/2 pass, both `line-clamp-2` sites in `search/index.tsx` carry `data-user-content` and clamp article title/body — user content, not UI chrome — so no truncation-policy violation).

### #797 rebase risk — the one thing you asked about

**No.** `TopicFilterAutocomplete.tsx` is not, and cannot currently be, a translation-sink problem for `MESSAGE_KEYS`. It never imports `i18n`, never calls `t()`/`useTranslation`, and never receives a translated string as a prop — every string in it ("Filter by topic", "No topics", "Topics unavailable", "Clear topics", "Remove topic …", the `${n} topics` status line, etc.) is a hardcoded English literal. The sink-detection mechanism you described only matters when a `t(SOME_KEY)` result flows into a component the static analyzer must recognize as a legitimate render target — since #803 never produces a `t()` call, there's nothing for that analyzer to trace into this component. Confirmed no `i18n`/`MESSAGE_KEYS`/`NavigationItem`/`NAVIGATION_GROUP_LABEL` references anywhere in the 8 touched source files, and no `src/i18n` directory even exists on this branch (pre-#797, as expected).

**Adjacent risk worth flagging anyway:** #803 lands roughly two dozen brand-new hardcoded English strings at the exact moment #797 stands up the catalog as the source of truth — "No topics", "Topics unavailable", "Search unavailable", "No matches", "Start typing.", "Try again", "Clear topics", "Selected topics", "Filter by topic", "Maximum topics selected", etc., none of them catalogued. None of your four listed invariants technically bans this (they govern dead-key detection, catalog parity, direct JSON imports, and truncation — not "new copy must be translated"), so it won't fail the stated CI gates. But it's in tension with the stated purpose of #797, and `Feed.tsx`'s `<PageHeader title="Articles">` / `<EmptyState title="No matching articles">` already pass raw literals into components on your own `REVIEWED_IMPORTED_SINKS` allowlist (pre-existing, unchanged by #803) — so the moment someone *does* wire i18n through those sinks, #803's copy is exactly the kind of thing that'll need a follow-up pass. Not a blocker; flagging so it's a tracked follow-up rather than a surprise.

**Second discrepancy**: #803's own PR body claims "does not depend on #786 or #797 and requires no rebase," which conflicts with the plan that it will be rebased onto #797 after #797 merges. Worth reconciling before merge — if the branch truly predates #797's `main`, the merge conflict surface (new `TopicFilterAutocomplete.tsx`, rewritten `search/index.tsx`, `Feed.tsx`, `free2z.ts`) is nontrivial busywork even though no *logical* invariant is violated.

### Does it deliver #788 — pill wall gone, not hidden

Yes, cleanly. `Feed.tsx` deletes the `knownTags` accumulator, `toggleTag`, and the entire pill-wall JSX block outright (not feature-flagged/hidden) and replaces it with `<TopicFilterAutocomplete>`. A new Playwright assertion (`search-autocomplete.pw.ts:134`) explicitly checks `[aria-label="Filter by tag"]` has count 0 on `/articles` — a real regression guard against the wall coming back, not just an omission.

### `discovery-autocomplete.ts` / autocomplete race & debounce

Sound. `TopicFilterAutocomplete` uses a monotonic `requestRef` counter: bumped at effect start, bumped again in the cleanup, checked before every `setState` inside `.then/.catch/.finally`. Traced the interleaving by hand — a slow earlier response is always discarded correctly, including the case where debounce-timer cancellation (via `clearTimeout` in cleanup) happens before the network call is even issued, vs. the case where the call is already in flight (no `AbortController`, so the request itself isn't cancelled, but its result is provably ignored — matches the new Playwright test "rejects an older response that resolves last"). Global `Search` reuses the pre-existing `usePaginatedSearch` request-id guard (unchanged logic, just the identity function changed) for the same property.

Short-circuit: `!hasDraft` (empty/whitespace query) returns before scheduling the debounce timer at all — confirmed no network call fires for empty input, and `buildDiscoverySuggestions` independently returns `[]` for a blank/whitespace query without touching the network.

### `pagination.ts`

The only change here is `creatorIdentity` (`username.toLowerCase()`) → `creatorSearchIdentity` (`free2zaddr`), fixing a real bug: two creators differing only by username case previously collapsed into one during page-merge dedup. New test proves it. The actual pagination/merge logic (`mergeSearchPage`, off-by-one cursor validation, shrinking-result-set handling via the count-drift `console.warn` path) is untouched by this PR — nothing new to find there.

### Accessibility

Both autocompletes implement the ARIA combobox pattern correctly: `role="combobox"` + `aria-autocomplete="list"` + `aria-expanded`/`aria-controls`/`aria-activedescendant`, arrow-key navigation with wraparound, Enter-to-select, Escape closes the popup without clearing typed text (matches the accompanying Playwright assertions), Backspace-on-empty removes the last chip. `onPointerDown preventDefault` on option buttons correctly keeps focus on the input during selection.

Global Search's blur/pointer handling (`pointerInteractionRef` + capture-phase `onPointerDownCapture`/`onClickCapture` on the route root + window-level `pointerup`/`pointercancel` fallback) is a deliberately intricate state machine to solve a real problem: the suggestion panel renders in normal document flow (not an absolutely-positioned overlay), so collapsing it on blur *before* a click on a control below it completes would shift that control out from under the pointer. I traced the browser event ordering (pointerdown → any default-driven blur → mouseup → click) by hand and it holds together — capture-phase handlers run before the blur-driven state check, so an in-subtree click never collapses the panel before its own click handler fires, while a click entirely outside the subtree (e.g., TopBar) still resolves correctly via the ordinary blur path since it isn't subject to the same layout-shift problem. It works, but it's genuinely fragile: correctness depends on unstated browser event-ordering guarantees, and a future refactor could silently reintroduce the bug it fixes without any type error to catch it. Worth a code comment pointing at this review or a Playwright regression test that fails loudly if the ordering assumption ever breaks (one such test already exists — "clears a pointer transaction released outside the route" — so this may already be covered).

### Behavior change worth a second look (not a blocker)

`articles.suggestTags` used to fail-soft per-malformed-item (`.flatMap` silently dropped bad entries, kept the good ones). It now throws on the *first* malformed suggestion, discarding the entire batch — intentional, per the PR's stated goal of distinguishing "unavailable" from "empty," and it's well tested. But it means one bad record from the backend (e.g., a legacy tag with a non-numeric, non-coercible count) now nukes an otherwise-good suggestion list and shows "Topics unavailable" instead of just quietly omitting the one bad entry. Confirmed this doesn't regress the *other* existing consumer, `ArticleTagInput.tsx` — it already wraps `suggestTags` in its own `.catch()` with a comment anticipating exactly this API-contract tightening, so no unhandled rejection there.

### Copy / truncation

No UI-copy truncation found; the two `line-clamp-2` usages in `search/index.tsx` are pre-existing, clamp user-authored article title/body, and carry `data-user-content` — compliant with the house rule and #797's stated allowance.

### Verdict

Approve on the merits, modulo the owner's admin override for the human-review requirement (author can't self-approve). No correctness bugs found beyond the one behavioral trade-off above (flagged, not blocking). Rebase-risk-wise: the specific concern about `TopicFilterAutocomplete` being an undetected translation sink does not materialize under the four stated invariants — but reconcile the PR body's "no rebase needed" claim against the stated plan to rebase onto #797, and expect a real (if mechanical) rebase given the size of the diff to `search/index.tsx` and `Feed.tsx`.

skyl:
## Deferred from this session's drain — same diagnosed cause as #822

Approved on the merits (adversarial review confirmed the topic-pill wall is genuinely deleted, the autocomplete race-guard is correct, and there are no truncation-policy violations). Deferred only because it cannot be fixed and re-gated before this session's release freeze.

**What's failing:** `zuuli / frontend`. Every other gate dependency is green or still running clean — no Rust failures.

**Root cause (same as #822):** this branch was rebased onto #797 (the i18n kernel). The rebase correctly reconciled the source — including keeping `t(MESSAGE_KEYS.navSearchPlaceholder)`, `navSearch`, and `articlesSearchAccessible` that a later commit had tried to replace with hardcoded strings (which would have tripped #797's "message keys lack executable consumers" check) — and **vitest passes 1111/1111**. But `tests/*.pw.ts` is a *separate* Playwright suite that was not run locally, and those specs still assert pre-i18n copy/markup.

Likely affected specs, given this PR's own file list: `tests/search.pw.ts`, `tests/search-autocomplete.pw.ts`, `tests/search-pagination.pw.ts`, `tests/article-tags.pw.ts`, `tests/touch-targets.pw.ts`.

**To land it next session:**
1. `npx playwright test tests/search.pw.ts tests/search-autocomplete.pw.ts tests/article-tags.pw.ts tests/touch-targets.pw.ts` against this branch.
2. Update assertions to the copy the i18n catalog now renders.
3. Push **exactly once** — do not `gh run rerun`; competing runs cancel each other via the workflow concurrency group.

**Process note for whoever picks this up:** the rebase instructions for this batch asked for typecheck + vitest + the truncation node-test, but did not ask for the Playwright suite. That omission is why both this PR and #822 look green locally and fail in CI. Any future rebase across the i18n kernel should run `npx playwright test` too.

skyl:
Rebased onto current `origin/main` and fixed both gaps so the `zuuli / frontend` gate goes green. New head: `a26ec11`.

## Gap 2 (the real i18n regression) — fixed

The dedicated `/search` page's combobox took its accessible name and placeholder from a hardcoded `SEARCH_INPUT_LABEL` constant in `search-route.ts` and a literal `placeholder="Search"`, bypassing `t()` entirely — confirmed the loss under `es`.

Fix: route both through the catalog, reusing existing keys rather than adding new ones:
- `aria-label={t(MESSAGE_KEYS.navSearch)}`
- `placeholder={t(MESSAGE_KEYS.navSearchAction)}` (the exact pattern `Feed.tsx` already uses)

`navigation.search` (`navSearch`)'s catalog value is promoted from `"Search creators and pages"` to `"Search creators, pages, and topics"` in all three locales (en/es/fr), because that's the name `search.pw.ts` (this PR's own edit), `search-autocomplete.pw.ts`, `search-pagination.pw.ts`, and `touch-targets.pw.ts` already expect on **both** the TopBar's home-page quick search and the dedicated combobox — they share the key now, so they can't drift again. Removed the now-dead `SEARCH_INPUT_LABEL` export.

Translations added:
- es: `"Buscar creadores, páginas y temas"`
- fr: `"Rechercher des créateurs, des pages et des sujets"`

Fixing the localization surfaced two test files with selectors that had gone stale relative to this same PR's earlier (correct) rename of the combobox's ARIA role from `searchbox` to `combobox` (required by the WAI-ARIA combobox pattern given `aria-expanded`/`aria-controls`/`aria-activedescendant`):
- `tests/ui-copy.pw.ts` (untouched by this PR until now) still queried `role="searchbox"` with the pre-topics name, so its locator could never find the element — regardless of translation. Updated the two English assertions and the Spanish assertion to `role="combobox"` + the new name. Every substantive check is untouched: viewport fit, translated heading/placeholder/empty-state text, and the removed-explainer-copy check.
- `tests/search-autocomplete.pw.ts`'s "locale-aware Search" test asserted the combobox under `es-ES` by its literal *English* name — that only ever worked because the label was unlocalized (a bug masking a bug). Now that it's translated, that test looks up the Spanish name instead.

## Gap 1 (copy mismatch) — resolved by keeping the existing copy

`search.pw.ts`'s own diff changed the expected empty-state text from `"Search all of free2z"` to `"Start typing."` without ever touching a locale file. Per the task's own tiebreaker: `tests/ui-copy.pw.ts`'s Spanish test — untouched by this PR — independently pins `"Buscar en todo free2z"` (the *existing* translated copy) as a passing assertion. That's proof `"Search all of free2z"` is the accepted, existing design, not a stale value. Reverted `search.pw.ts`'s assertion back to `"Search all of free2z"` rather than inventing new copy across three catalogs on a guess.

## Verification (from `wallet/zuuli`, `npm ci`)

- `npm run typecheck` — clean
- `npx vitest run` — 92 files / 1163 passed, 5 skipped (unchanged), including `i18n/build-boundary.test.ts`'s consumer/sink checks
- `npx playwright test` — full suite, **171/171 pass**, including all originally-failing tests (`search.pw.ts` x3 viewports, `ui-copy.pw.ts` x3)

Worktree removed locally; not merging — leaving that to CI + review per the usual gate.

Issue #371, opened 2026-08-09
security(release): keep Apple credentials out of dependency-controlled builds


skyl:
## Summary

The iOS and macOS release jobs materialize Apple credentials before running dependency-controlled build steps. A malicious or compromised Rust/npm/CocoaPods/build dependency can read those credentials during its build script, even if the later upload tooling is the only component that needs them.

Android's analogous build/sign/publish split is already tracked in #293. This issue covers the unaddressed Apple credential lifetime.

## Evidence

- The iOS job writes the App Store Connect private key to a global filesystem location before the build; the build runs later in `.github/workflows/zuuli-release.yml` around lines 380-478.
- `wallet/zuuli/scripts/mobile-release.sh:138-143` copies the App Store Connect key and then runs the application build.
- The script only needs that API key for later `altool` upload operations around lines 161-169.
- The macOS release job imports signing/App Store Connect material and then runs the release build around `zuuli-release.yml:704`.
- `wallet/zuuli/docs/releasing.md:68` describes an Admin App Store Connect key, increasing the blast radius.

## Impact

Arbitrary code executed by a transitive build dependency can exfiltrate long-lived Apple API keys or signing material, enabling unauthorized uploads or other actions allowed by the credential's role.

## Required remediation

- Build unsigned application artifacts in a credential-free job.
- Produce and attest immutable artifact digests.
- Pass only those artifacts into a minimal signing/notarization/upload job with no package-manager install, dependency build, or project-controlled build hooks.
- Materialize credentials immediately before the single operation that requires them and remove them immediately afterward.
- Replace Admin credentials with the minimum App Store Connect role and scope that supports the required upload.
- Apply the same invariant to macOS signing identities and notarization credentials.

## Acceptance criteria

- npm, Cargo, CocoaPods, Tauri, Xcode project build phases, and third-party build scripts never run while Apple private keys or signing secrets are available.
- The credential-bearing job accepts only a digest-verified artifact produced by the credential-free build.
- Store credentials have minimum privileges and are short-lived where supported.
- A canary build script proves Apple secret paths/environment variables are absent during all dependency-controlled steps.
- Android work remains tracked by #293.

PR #804, opened 2026-08-30
ZUULI Live: add camera and microphone preflight

### LINKED ISSUE #792: ZUULI Live preflight: camera, microphone, and device preview before provisioning
Parent: #784.

## Scope

Add a local camera/microphone/device preflight before ZUULI provisions or starts a livestream. This slice does not create backend rooms.

## Acceptance criteria

- [ ] Enumerate and select devices where supported.
- [ ] Request permissions only after explicit host intent.
- [ ] Show a local preview with clear mute defaults.
- [ ] Permission denial, no-device, and removed-device states have concise recovery actions.
- [ ] Cancel and unmount release every acquired track.
- [ ] No `live.start` call occurs before successful confirmation.
- [ ] Keyboard, screen-reader, reduced-motion, touch-target, and 320px layout coverage is included.
- [ ] Native permission manifests remain exact and are tested where applicable.

Likely surfaces: new `wallet/zuuli/src/features/live/media-preflight.ts`, `useMediaPreflight.ts`, preflight UI, `GoLiveDialog.tsx`, and focused tests.

skyl:
## Summary
- add an explicit-intent camera/microphone preflight with device enumeration and exact selection
- fail closed across denied, missing, busy, removed-device, and provisional-device races while releasing every current, provisional, or stale track
- require a distinct confirmed preview before provisioning, with accessible mute truth and mobile-safe layout
- lock exact Android/iOS/macOS capture permission manifests and signed-artifact policy with mutation-sensitive tests
- merge current `origin/main` normally and preserve its reviewed RealtimeKit CSP endpoints and browser regressions

## Exact-head evidence

Verified at `16737c2ea500b16bfcbf62a050bcba08f21bde8c`:

- `npm run typecheck`
- `npm run typecheck:tests`
- `npm run build`
- full Vitest: 882 passed, 5 skipped
- full Node policy suite: 379 passed, 5 skipped
- focused media Vitest: 36/36
- focused media Playwright: 9/9
- standalone full Playwright: 141/141, including both RealtimeKit CSP regressions
- action-pin and Rust-toolchain mutation self-tests plus live checks
- Apple credential, macOS entitlement, and permission-manifest mutation/live checks: 160/160
- unsigned universal Apple Tauri app build (`arm64` + `x86_64`); emitted bundle has both architectures and exact camera/microphone usage strings

Closes #792


skyl:
Exact-head CI blocker on `17e5f4560f3c5cccdfc7fc2667abff7fd84edcf0`: both change detectors and required gates fail because this PR changes the canonical `wallet/zuuli/package.json#test` literal to add `media-permission-manifests.node-test.mjs` without updating the compiler-bound contract in `scripts/check-github-actions-pins.mjs`. The rs log reports the expected canonical script verbatim.

Please preserve the canonical `test` script by folding this audit into an already-bound Node test, or deliberately update the checker and its mutation controls as part of the same authority change. A new push requires fresh exact-head code review and CI. The media/security review is still in progress; this comment is not a complete correctness verdict.

skyl:
A second exact-head blocker is confirmed on `17e5f4560f3c5cccdfc7fc2667abff7fd84edcf0`:

P2 — removed-device recovery is currently a dead end. The `devicechange` path detects a missing selected camera/microphone and sets `removed`, but leaves `selectedRef`/`selected` holding the removed exact device IDs. The failure-state **Try again** calls `requestPreview`, which reuses those IDs and asks `getUserMedia` for the gone device again. The failure UI also hides the selectors while telling the user to “choose another device.” The focused test proves only the transition to `removed`, not recovery.

Please clear/fallback the vanished selection and/or keep remaining-device selectors available in recovery, then add a test that actually recovers using a surviving camera/microphone. The hard-red canonical-test-script gate blocker from the earlier comment also remains at this head. Any new push resets the review and CI binding.

skyl:
A third P2 is confirmed on exact head `17e5f4560f3c5cccdfc7fc2667abff7fd84edcf0`: the Android “exact permission set” test is fail-open. Its regex recognizes only the exact single-attribute, self-closing `<uses-permission android:name="…" />` spelling. A new permission with another attribute, a non-self-closing element, or a `<uses-permission-sdk-23>` element is invisible, so authority can expand while the claimed exact/non-duplicated test stays green.

Please parse or conservatively enumerate every Android permission element, fail closed on unrecognized permission-element forms, and add negative mutations for alternate valid spellings/permission element types. The checked-in manifests currently retain the expected set; this finding is about the guard's claimed ability to hold that set.

The canonical test-script gate failure and removed-device recovery finding also remain open at this head.

skyl:
A release-level P1 is confirmed on exact head `17e5f4560f3c5cccdfc7fc2667abff7fd84edcf0`: the signed/notarized macOS product cannot satisfy this capture feature under the current authority contract.

The release workflow signs `ZUULI.app` with Hardened Runtime (`codesign --options runtime`) while `src-tauri/Entitlements.plist`, the source checker, signing step, and signed-artifact verifier all require the exact three-key Keychain-only entitlement set. Apple documents that Hardened Runtime camera/audio capture requires `com.apple.security.device.camera = true` and `com.apple.security.device.audio-input = true`. The desktop app also has no macOS Info.plist carrying `NSCameraUsageDescription` / `NSMicrophoneUsageDescription`; those descriptions currently exist only on the iOS path, while Apple requires the keys for apps using camera/microphone APIs.

Primary references:
- https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.device.camera
- https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.device.audio-input
- https://developer.apple.com/documentation/bundleresources/information-property-list/nscamerausagedescription
- https://developer.apple.com/documentation/bundleresources/information-property-list/nsmicrophoneusagedescription

Please add the narrowly scoped camera/audio-input entitlements and macOS usage descriptions, then update the source, workflow, credential-boundary, and signed-artifact verification policies so the exact new authority is compiler-bound. Add packaged macOS artifact verification for both entitlements and both Info.plist strings. Packaging smoke cannot prove runtime capture without this authority. The prior red gates and P2 findings also remain open.

skyl:
Rebound to exact head `872fba2e6d377b872b25ffae381d7375f1a59fc7`. The submitted delta fixes the canonical-test-script gate defect and removed-device recovery with generation-bound enumeration/fallback tests. Those prior findings are closed.

Two blockers remain unchanged: the Android exact-permission guard is still fail-open to alternate valid permission-element spellings/types (P2), and the signed Hardened Runtime macOS product still lacks camera/audio-input entitlements plus macOS usage descriptions and exact release/artifact verification (P1). This head is also behind current main and fresh CI is pending. Any push/base update resets the binding.

skyl:
One additional P2 remains on exact head `872fba2e6d377b872b25ffae381d7375f1a59fc7`: stale exact device selections survive close/reopen and acquisition errors. `release()` does not clear selection; reopening enumeration only updates the device list, and `devicechange` returns whenever there is no current stream. If a selected device disappears while the dialog is closed—or between selection and acquisition—`getUserMedia` receives the stale exact ID, the `OverconstrainedError` path leaves it intact, and **Try again** repeats the same dead request despite a valid replacement device.

Please sanitize stored selection against each successful enumeration even with no current stream and/or clear invalid exact IDs on `no-device`. Add close → unplug → reopen recovery and selection-race recovery regressions. The Android permission-test P2 and signed macOS authority P1 also remain at this head.

skyl:
## Adversarial review — release-workflow and Apple entitlement surface

I focused on the protected-release/entitlement changes and verified them by running the actual boundary scripts against the PR's real files (not just reading the diff):

- `node scripts/apple-credential-boundary.mjs` → **passes** against the PR's `zuuli-release.yml`. This confirms the updated `macos-sign` job-body SHA-256 pin (`1373f642...`) is the correct hash of the *new* job content (extra `plutil -lint`/`NSCameraUsageDescription`/`NSMicrophoneUsageDescription`/entitlement-key checks), not a blindly-bumped value. No new secrets, no new job, no relocation of credential material — the added lines are read-only `plutil`/`test` assertions.
- `node scripts/macos-keychain-entitlements.mjs` → **passes**, confirms `tauri.conf.json`'s new `bundle.macOS.infoPlist: "Info.macos.plist"` and the two new entitlement keys are exactly the reviewed/pinned values.
- `node --test` on `apple-credential-boundary.node-test.mjs`, `macos-keychain-entitlements.node-test.mjs`, `media-permission-manifests.node-test.mjs` → **160/160 pass**.
- `vitest run` on `MediaPreflight.test.tsx`, `media-preflight.test.ts`, `useMediaPreflight.test.tsx` → **36/36 pass**.

### 1. Protected-release workflow (`zuuli-release.yml`, `zuuli-packaging.yml`)
No weakening of the credential boundary. Every change in the diff is an *additional* verification step (`plutil -lint`, exact-string checks on `NSCameraUsageDescription`/`NSMicrophoneUsageDescription`, and two new required entitlement keys added to the existing exact-keyset assertion `keys | sort | join(",")`). `Info.macos.plist` is now shipped through the same checksum-manifest/attestation path as `Entitlements.plist` (added to `CHECKSUMS.sha256` inputs and the expected-member inventory in both the unsigned-builder and signer jobs) — so it's covered by the same integrity chain, not a side channel. `apple-credential-boundary.mjs`'s per-job SHA-256 pin plus its new `requireText` assertions on `macos-sign`/`macos-finalize` mean any future attempt to touch these checks without updating the reviewed hash fails CI. This directly satisfies the #371 concern — nothing here creates a path for a dependency-controlled build to reach Apple credentials.

### 2. Entitlements/Info.plist
Minimal and matched to actual use: `com.apple.security.device.camera` + `com.apple.security.device.audio-input` (both `true`), and `NSCameraUsageDescription`/`NSMicrophoneUsageDescription` describing exactly the live-broadcast/join use case. No broader entitlement (no photo library, no location, no network-server, etc.) was added. `media-permission-manifests.node-test.mjs` further cross-checks that the Android manifest permission set is exact (`INTERNET`, `CAMERA`, `RECORD_AUDIO`, `MODIFY_AUDIO_SETTINGS`, no more) and that the iOS `Info.ios.plist`/generated Info.plist/`project.yml` carry byte-identical copy to the new macOS strings.

### 3. `tauri.conf.json` vs. #818 RealtimeKit CSP
The only change in this file is `bundle.macOS.infoPlist`. **No CSP or `capabilities` key is touched at all** — there is zero overlap with #818's RealtimeKit CSP-endpoint allowlist, so no regression risk to the #813 livestream path.

### 4. `check-github-actions-pins.mjs`
The only change is adding `media-permission-manifests.node-test.mjs` to the required frontend `test` script contents list — this is a stricter requirement (CI now also enforces this script is wired into `npm test`), not a relaxation. All actions remain pinned by SHA elsewhere in the workflows (unchanged by this PR).

### 5. Preflight UX — fail-closed and cleanup
- `GoLiveDialog.tsx`: `canStart` requires `preflight.status === "ready"`; `handleStart` re-checks `preflight.status !== "ready"` before starting. Denial/no-device/busy/removed/unsupported/failed states all block the Start button and show recovery copy (`MediaPreflight.tsx` `FAILURE_COPY`).
- Track cleanup is thorough: `stopMediaStream` wraps each `track.stop()` in try/catch so one bad track can't block releasing the rest; `useMediaPreflight`'s effect teardown calls `invalidateAndStop()` on unmount/`active` flip; `GoLiveDialog.handleOpenChange` calls `preflight.release()` on any close path (Cancel button, outside click, Escape — all route through Radix's `onOpenChange`); `startAfterMediaConfirmation` explicitly releases the preview stream before handing off to `live.start()` so the local preview stream is never leaked into the provisioning call, even on failure.
- Microphone defaults muted (`track.enabled = false` on install, `microphoneEnabled` initial state `false`) — camera does not silently start hot-mic'd.
- `MediaPreflight.tsx`'s own video-element effect also tears down (`video.pause()`, `srcObject = null`) on stream change/unmount.

### Minor / non-blocking nits
- `useMediaPreflight.ts`'s `onDeviceChange` handler computes `nextSelection = selectedRef.current` and then immediately does `selectedRef.current = nextSelection; setSelected(nextSelection)` — a no-op reassignment (dead code, not a bug).
- Camera preview turns off then back on around `startAfterMediaConfirmation` (preview stream stopped, then RealtimeKit's own `getUserMedia` call re-acquires) — a brief flicker/re-prompt-free re-acquire, not a defect, just worth knowing about if a future bug report mentions a momentary camera-light blink when going live.

### CI status caveat (process, not code)
At review time the PR's **required** `gate` check (workflow "wallet/zuuli") had not yet reported — `zuuli / frontend` (runs exactly the `npm test` suite that includes the two boundary scripts) was still `IN_PROGRESS`, and `Desktop / macOS universal app` (runs the new `plutil`/entitlement assertions against a real built `.app`) was still `QUEUED`. Only `rs / gate` (the Rust-side required check) and non-required jobs (`Release identity`, `Rust / formatting`, `changes`) had completed. Branch protection requires both `gate` and `rs / gate`, so this PR cannot actually merge until the frontend/packaging gate finishes — this is not a defect in the PR, just a note that "CI already green" wasn't fully accurate as of this review; I verified the equivalent checks manually instead (see above).

## Verdict: APPROVE

No security or correctness defect found in the release-workflow, entitlements, or CSP surfaces. The change is additive and self-verifying (new SHA-pinned job bodies, new exact-value assertions at every stage of the macOS signing/build/finalize/DMG chain, cross-platform copy-consistency test). The media preflight hook fails closed and cleans up capture tracks on every exit path I could find. Confirmed independently by running the actual boundary/entitlement scripts and full preflight test suites against the PR's real files, not just reading the diff.

**Release-workflow and `tauri.conf.json` changes are safe to have on `main` while builds 18/19 ship**: the protected-environment boundary, the `zuuli-mobile-store` concurrency lane it doesn't touch, and the status-freshness/release-identity checks are all unaffected; `tauri.conf.json` makes no CSP change at all, so there's no interaction with #818's RealtimeKit CSP fix. The only remaining gate is CI finishing (`gate` required check), which is a merge-timing matter, not a code-review one.

Issue #813, opened 2026-08-30
P0 ZUULI Live: Join Free fails with RealtimeKit ERR0001 on an active stream


skyl:
## Founder QA failure

ZUULI correctly shows a creator as live, but tapping **Join Free** fails before the viewer initializes:

```text
[ERR0001]: {Client} Failed to initialize. Unable to connect to the server. Check the internet connection
```

The same livestream is active on Free2Z and has real viewers, so this is not adequately explained by generic offline copy. This was observed in the Play-distributed ZUULI build; exact build identity is not yet exposed in-app (#810).

## Required diagnosis

Trace the exact ZUULI path from listing → membership/join response → RealtimeKit initialization. Determine whether the failure is caused by room lifecycle, environment/configuration, meeting ID, participant token, token expiry/audience, API shape parsing, SDK mismatch, transport policy, or another client boundary. Compare the same live room and credentials contract with the working Free2Z web viewer.

## Acceptance criteria

- [ ] A real active Free2Z livestream can be joined from Android and iOS ZUULI.
- [ ] ZUULI validates the join response strictly before initializing RealtimeKit and never substitutes fabricated identifiers/tokens.
- [ ] Meeting ID, participant token, environment, and SDK expectations agree across ZUULI and the authoritative backend contract.
- [ ] Expired, malformed, wrong-room, wrong-environment, lifecycle-race, and connectivity failures produce concise actionable states rather than the misleading generic internet error.
- [ ] Retry obtains/reconciles authoritative state and cannot create, rotate, replace, or mutate the active creator room.
- [ ] Diagnostics identify the failing boundary without logging tokens, meeting secrets, user identity, or other private data.
- [ ] End-to-end coverage uses one authoritative room to prove the web viewer and ZUULI viewer can join concurrently.
- [ ] Device evidence covers background/foreground, network transition, and a freshly installed app.

## Related private backend work

- free2z/tuzi#1514 — room lifecycle/viewer 404 P0 and creator-safe public URL behavior
- free2z/tuzi#1515 — RealtimeKit modernization audit
- free2z/tuzi#1516 — public web viewer inline stage collapsing to 0px
- free2z/tuzi#1507 — authoritative ZUULI host/viewer lifecycle contract

The web CSS collapse in #1516 is a distinct rendering bug; this issue owns the ZUULI client initialization failure while keeping shared lifecycle/configuration dependencies explicit.

skyl:
[...]

PR #819, opened 2026-08-30
fix(zuuli): bind live join tickets to meetings

### LINKED ISSUE #815: ZUULI Live: bind join tickets to rooms and obtain a fresh ticket on retry
Parent: #813.

## Scope

Strictly validate ZUULI Live join tickets, bind them to the intended room, refresh them on retry, and classify client-safe failures without exposing credentials.

## Acceptance criteria

- [ ] Strictly parse the join response and require nonempty `auth_token` and `meeting_id` where the contract requires them.
- [ ] Decode only the minimal safe participant-token claims needed to prove response/token meeting identity agreement; fail closed on mismatch.
- [ ] Retry obtains a fresh authoritative join ticket instead of reusing the same token.
- [ ] Retry cannot create, rotate, replace, or mutate the creator's active room.
- [ ] Distinguish policy/CSP, offline/transport, timeout, malformed ticket, token rejection, ended room, and lifecycle race with concise actionable UI.
- [ ] Diagnostics identify the failing boundary without tokens, full claims, meeting secrets, identity, or private URLs.
- [ ] Tests cover expiry, revocation, wrong room/environment, rotation, replay, background/foreground, and network changes.

skyl:
Closes #815
Part of #813

## Summary
- strictly parse RealtimeKit participant JWTs and bind the token, authoritative join response, discovery meeting, provider environment, role, and expiry before SDK initialization
- replace same-token SDK retry with a fresh participant-only authoritative join, rejecting replay, room rotation, environment drift, and host-role refresh
- prevent an unresolved or creator-owned public viewer page from invoking the overloaded host/join route
- classify policy/CSP, offline, transport, timeout, malformed/expired/replayed tickets, provider rejection, ended rooms, and lifecycle races into safe actionable copy
- emit allowlisted boundary/category/provider-code diagnostics only; never raw SDK errors, token claims, meeting IDs, identities, or private URLs

## Verification
- npm run typecheck:tests
- npm run build
- npx vitest run (68 files; 887 passed, 5 skipped)
- focused live suite (58 passed)
- gitleaks git --staged --redact --no-banner .

## Backend boundary
The client blocks the known creator-viewer trigger and its retry path never calls start or private room management. The backend must still enforce the second-layer creator-safe public-route invariant tracked in free2z/tuzi#1514; this PR does not claim that server guard or signed-device end-to-end evidence.

### DIFF
--- a/wallet/zuuli/src/features/live/Room.tsx
+++ b/wallet/zuuli/src/features/live/Room.tsx
@@ -55,6 +55,8 @@ import {
   runSingleFlight,
 } from "./membership";
 import { Stage } from "./Stage";
+import { viewerEntryState } from "./viewer-safety";
+import { classifyLiveFailure, safeLiveDiagnostic } from "./connection-failure";
 
 interface JustStarted {
   ticket: DyteJoinTicket;
@@ -76,6 +78,7 @@ export function Room() {
   const tuzis = useSession((s) => s.tuzis);
   const adjustTuzis = useSession((s) => s.adjustTuzis);
   const user = useSession((s) => s.user);
+  const sessionLoading = useSession((s) => s.loading);
 
   // A host who just went live already has their stream + ticket in navigation
   // state, so we can render the room immediately and skip the public listing
@@ -146,6 +149,12 @@ export function Room() {
   const [ticket, setTicket] = useState<DyteJoinTicket | null>(
     justStarted?.ticket ?? null,
   );
+  // Private invites remain memory-only. Keeping the active secret here lets a
+  // failed viewer request obtain a fresh ticket without putting it in storage,
+  // diagnostics, query parameters, or the ticket itself.
+  const [activeJoinSecret, setActiveJoinSecret] = useState<string | undefined>(
+    inviteSecret ?? undefined,
+  );
 
   useEffect(() => {
     if (!data?.inviteTicket) return;
@@ -213,6 +222,12 @@ export function Room() {
   const kind = KIND_META[stream.kind] ?? KIND_META.broadcast;
   const creatorName = stream.creator.display_name ?? stream.creator.username;
   const seed = stream.username + stream.title;
+  const entryState = viewerEntryState({
+    sessionLoading,
+    viewerUsername: user?.username,
+    creatorUsername: stream.username,
+    kind: stream.kind,
+  });
 
   return (
     <div className="space-y-6 animate-slide-up">
@@ -278,7 +293,21 @@ export function Room() {
 
             {/* Stage center: either the "off-air" identity or the real meeting */}
             {ticket ? (
-              <Stage ticket={ticket} />
+              <Stage
+                ticket={ticket}
+                {...(ticket.as === "participant"
+                  ? {
+                      refreshTicket: (previous: DyteJoinTicket) =>
+                        live.refreshParticipant(
+                          stream.username,
+                          stream.kind,
+                          previous,
+                          activeJoinSecret,
+                        ),
+                      onTicketRefreshed: setTicket,
+                    }
+                  : {})}
+              />
             ) : (
               <div className="absolute inset-0 grid place-items-center">
                 <div className="flex flex-col items-center gap-3 text-center">
@@ -349,16 +378,23 @@ export function Room() {
               />
             )
           ) : (
-            <JoinPanel
-              stream={stream}
-              tuzis={tuzis}
-              onJoined={(t) => {
-                setTicket(t);
-                if (stream.kind === "ppv" && stream.price_tuzis > 0) {
-                  adjustTuzis(-stream.price_tuzis);
-                }
-              }}
-            />
+            entryState === "checking-session" ? (
+              <ViewerIdentityCheck />
+            ) : entryState === "creator-blocked" ? (
+              <CreatorViewerGuard />
+            ) : (
+              <JoinPanel
+                stream={stream}
+                tuzis={tuzis}
+                onJoined={(t, secret) => {
+                  setActiveJoinSecret(secret);
+                  setTicket(t);
+                  if (stream.kind === "ppv" && stream.price_tuzis > 0) {
+                    adjustTuzis(-stream.price_tuzis);
+                  }
+                }}
+              />
+            )
           )}
         </div>
       </div>
@@ -395,7 +431,7 @@ function JoinPanel({
 }: {
   stream: Livestream;
   tuzis: number;
-  onJoined: (ticket: DyteJoinTicket) => void;
+  onJoined: (ticket: DyteJoinTicket, secret?: string) => void;
 }) {
   const navigate = useNavigate();
   const location = useLocation();
@@ -503,14 +539,22 @@ function JoinPanel({
     try {
       const ticket = await runExclusive(async () => {
         if (action) await action();
-        return live.join(stream.username, stream.kind, joinSecret);
+        return live.join(
+          stream.username,
+          stream.kind,
+          joinSecret,
+          "participant",
+          { expectedMeetingId: stream.meetingId },
+        );
       });
       if (!ticket) return false;
-      onJoined(ticket);
+      onJoined(ticket, joinSecret);
       return true;
     } catch (e) {
+      const failure = classifyLiveFailure(e, "ticket-request");
+      console.warn("Livestream join failed", safeLiveDiagnostic(failure));
       toast.error("Could not join", {
-        description: e instanceof Error ? e.message : "Please try again.",
+        description: failure.message,
       });
       return false;
     }
@@ -568,7 +612,10 @@ function JoinPanel({
             // does not expose while replacing all authoritative account data.
             setUser({ ...useSession.getState().user, ...authoritative });
           },
-          join: () => live.join(stream.username, stream.kind),
+          join: () =>
+            live.join(stream.username, stream.kind, undefined, "participant", {
+              expectedMeetingId: stream.meetingId,
+            }),
         }),
       );
       if (!result) return;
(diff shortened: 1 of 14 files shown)


skyl:
Requesting changes at exact reviewed head `36c921e49a9e6b904c30ab75df93e4f217fc3836`.

## P1 — refresh is not participant-only and can still rotate the creator room

`live.refreshParticipant()` checks only the client-authored `previous.as`, then calls `live.join()` (`wallet/zuuli/src/lib/api/free2z.ts:1915-1928`). For every public kind, `live.join()` sends the process-authenticated POST to `/api/dyte/{username}/{type}` (`wallet/zuuli/src/lib/api/free2z.ts:1884-1897`). This is the exact overloaded host/join route that free2z/tuzi#1514 proves chooses host behavior from the authenticated username and can create/rotate/overwrite the creator room. The test at `src/lib/api/live-join-contract.test.ts` mocks this POST and calls it non-management, but path inspection cannot prove participant intent because this path is itself the management boundary.

There is a concrete app path around the new creator guard: on any non-401/403 `auth.me()` bootstrap failure, the session store sets `user: null, loading: false` while deliberately retaining the Knox token (`wallet/zuuli/src/store/session.ts:50-59`). `viewerEntryState()` therefore allows the creator as an apparent guest (`wallet/zuuli/src/features/live/Room.tsx:225-230`), while `request()` still attaches the retained token. The backend sees the creator, takes the host/rotation branch, and returns a host-preset token; this PR then labels it `participant` from the caller and enables another mutating refresh.

This recreates the linked P0 incident under a transient session-hydration failure. The fix needs a server-enforced participant intent or dedicated participant endpoint from free2z/tuzi#1514, plus a fail-closed client session state. `previous.as` cannot be the authorization boundary.

## P1 — environment, role, and expiry are not strictly bound

The parser accepts a participant token with no `exp`: expiry is validated only when present (`wallet/zuuli/src/lib/api/live-ticket.ts:124-140`). I ran a dedicated negative control requiring a missing-exp token to be rejected; it failed because the parser accepted it.

Initial start/join also supplies no trusted `expectedEnvironmentId`; environment is compared only when a previous ticket provides one during refresh (`wallet/zuuli/src/lib/api/live-ticket.ts:179-184`). Any syntactically valid initial `orgId` therefore passes. Role/preset is not decoded or returned by the authoritative response at all: `as` is copied directly from `constraints.as` into the accepted ticket (`wallet/zuuli/src/lib/api/live-ticket.ts:193-202`). A host credential can consequently be mislabeled participant.

Current tests cover an already-present expired `exp` and environment drift after the first ticket, but not missing expiry, wrong initial environment, or authoritative role disagreement. The join contract needs an authoritative role/preset and environment binding, and expiry must be required if strict expiry is claimed.

## P1 — lifecycle handling and background/network evidence are incomplete

After `initMeeting()` resolves with any truthy client, `Stage` has no bounded wait for `roomJoined` (`wallet/zuuli/src/features/live/Stage.tsx:55-93`). Its `roomLeft` handler recognizes only `ended` and `failed` (`wallet/zuuli/src/features/live/Stage.tsx:95-124`), while bundled RealtimeKit 1.5.1 declares states including `rejected`, `kicked`, `disconnected`, `left`, and others. A rejected/terminal non-`failed` exit never reaches the app's actionable fresh-ticket retry, and a client that initializes but never joins has no app timeout.

The claimed background/foreground test never provides a meeting or emits any SDK lifecycle event. It dispatches `visibilitychange`, but `Stage` has no visibility listener; the assertion only proves that no refresh happened. It does not exercise suspension/resume, expiry while backgrounded, an established session's network transition, reconnection, or terminal disconnect. Add mutation-sensitive meeting-event tests and a bounded join lifecycle.

## P2 — the exact CSP `ERR0001` remains classified as transport

The known production CSP failure reaches the app as `[ERR0001] ... Unable to connect`, without `SecurityError` or CSP prose. The classifier's policy branch only recognizes those synthetic signals, then the exact error falls through to `transport` (`wallet/zuuli/src/features/live/connection-failure.ts:155-189`). The test explicitly expects `transport` for that exact `ERR0001`, so the shipped copy would again tell the user to switch networks rather than identify the policy boundary. Classify from a trustworthy preflight or `securitypolicyviolation` signal; the provider code alone is not enough to distinguish CSP from transport.

## Verification performed

- Focused live suite: 46/46 passed.
- `npm run typecheck:tests`: passed after installing both app dependency trees.
- `npm run build`: passed, including fresh WASM verification.
- Controlled mutations of response/token mismatch, refresh room binding, hydration guard, and fresh-token handoff were all killed by the focused tests.
- Missing-expiry negative control failed, exposing the coverage/contract gap above.
- `gitleaks` scanned the exact commit range with no leaks.
- No raw token, full claim, identity, meeting ID, or private URL exposure was found in the new app diagnostics.

The positive meeting/response/listing binding, replay rejection, safe diagnostic allowlist, and core mutation controls are useful, but the P1 items above must be resolved before merge.

skyl:
CTO handoff: DO NOT MERGE. Independent review found blocking P1s: participant retry reuses the overloaded host/join route and can rotate the creator room (tuzi#1514), retained Knox token after auth.me failure, missing authoritative exp/environment/role binding, and incomplete lifecycle handling. Successor must obtain the server-enforced participant-intent contract and re-review a new exact head before any release inclusion.

skyl (commented):
## Adversarial review — code read against `origin/pull/819/head` (36c921e), diffed vs `main`

**Verdict: solid, closes #815, does not endanger the #813 join path or #818's CSP fix. Approve to merge after rebase, with one follow-up worth filing (not a blocker).**

### What I verified

- **Binding is real, not cosmetic.** `parseJoinTicketResponse` (`live-ticket.ts`) decodes the participant JWT's `meetingId`/`orgId`/`participantId`/`exp` claims and cross-checks them against the join response's `meeting_id`, the caller-supplied `expectedMeetingId` (from the listing the user actually clicked, via the new `Livestream.meetingId` ← `RawDyteMeeting.meeting_id`, which is a required wire field), and `expectedEnvironmentId`. Any mismatch throws a typed `LiveTicketError` and fails closed before RealtimeKit ever sees the token. A room-A-ticket-for-room-B substitution is rejected by the `meeting-mismatch` check.
- **Retry gets a fresh ticket, not a replay.** `live.refreshParticipant` re-issues one authoritative `live.join` call (never `start`, never a host-management route) and passes `previousAuthToken`; `parseJoinTicketResponse` explicitly rejects the response if `auth_token === previousAuthToken` (`replayed-ticket`). `Stage.tsx`'s retry effect keys off `ticket.authToken`, so a genuinely fresh token re-triggers `initMeeting`; the old same-token `attempt`-counter retry is gone entirely. Confirmed with a live test (`live-join-contract.test.ts`) that mocks a server replaying the same token and asserts rejection.
- **Fails closed, never fakes "connected."** Both the falsy-client-resolution path and the catch path in `Stage.tsx` route through `classifyLiveFailure` and set `status: "failed"`; `"connected"` is only ever set by RealtimeKit's own `joined` event, unchanged from before. No swallowed errors.
- **Diagnostics are actually scrubbed.** `safeLiveDiagnostic` returns only `{boundary, category, providerCode}` — verified by a test that stuffs a fake token and private URL into the raw error and asserts neither leaks into the logged JSON. The old code path (`toast.error(..., { description: e.message })`) that echoed raw SDK text is replaced with static per-category copy.
- **CSP/endpoint behavior from #818 is untouched** — no CSP or endpoint config file appears in this diff; `connection-failure.ts` only pattern-matches error text to choose UI copy, it doesn't touch `connect-src` or transport config.
- **Creator-viewer guard (`viewer-safety.ts`) is a real fix for a second bug**, not scope creep: `live.start` and `live.join` for one's own public stream hit the *same* backend URL shape (`/api/dyte/{username}/{kind}`), so an authenticated creator viewing their own public stream page previously could have their "Join Free" click ambiguous with host-start. The new gate blocks exactly that case (case-insensitive username match, non-private kind) and nothing else — guests, other viewers, and the dedicated private-invite route are all still `"allowed"` (unit-tested). `sessionLoading` resolves near-instantly for anonymous users (`isAuthed()` false → immediate `loading:false`), so this doesn't stall real viewers in practice.
- CI is green across the board on this PR as of review time.

### Findings

1. **(Minor, worth a follow-up issue, not a blocker)** Hosts lose their retry button on `Stage` failure. `Room.tsx` only passes `refreshTicket`/`onTicketRefreshed` to `<Stage>` when `ticket.as === "participant"`; for a host ticket there's no equivalent, so `failure.retryable && refreshTicket && onTicketRefreshed` is always false and no "Try again" renders at all — a host whose own stage fails to connect has no in-app recovery (previously they at least got a same-token re-init button, which was itself the bug being fixed). This is a defensible deliberate trade-off (never touch/rotate the creator's own room from this code path) but it's a regression in host UX that should get its own follow-up ticket for a safe host-side reconnect (e.g., re-fetch host status/ticket via the dashboard, not a raw retry).
2. The JWT decode is explicitly *not* a signature check (documented in a comment) — trust is anchored on TLS + the authoritative HTTP response, which is reasonable for a client-side binding check but worth keeping in mind if #813's root cause turns out to be a spoofable claim rather than a stale/rotated ticket.

### Composition with #798 on shared files

`Room.tsx`, `free2z.ts`, and `types.ts` are touched by both PRs but on **disjoint concerns**: #819 adds `meetingId?`/`environmentId?`/`expiresAt?` to `Livestream`/`DyteJoinTicket` and reworks join/retry/guard logic; #798 changes `participants: number` → `number | null` and reworks display copy. The only true textual overlaps are adjacent lines inside the same object literals/functions (`mapLivestream`'s return, and a couple of spots in `Room.tsx` within ~10 lines of each other) — mechanically resolvable rebase conflicts with no semantic collision; neither PR reads or depends on the field the other one changes.

skyl:

PR #822, opened 2026-08-30
ZUULI About: bind exact build identity

### LINKED ISSUE #810: ZUULI About: show exact version, build, channel, platform, and commit
Parent: #809

## Scope

Add an **About** surface under Settings/More that identifies the exact ZUULI build currently running.

## Acceptance criteria

- [ ] Show product name, semantic version, build number, release channel, platform, and short source commit from build-time immutable metadata.
- [ ] Never derive identity from mutable runtime state or display placeholder/fabricated values.
- [ ] Provide one concise **Copy build info** action with a stable plain-text format.
- [ ] Make full commit/provenance details available through a compact disclosure or link when the build supports it.
- [ ] Work offline and fail honestly if an optional field is unavailable.
- [ ] Build metadata agrees with release artifacts and existing release verification.
- [ ] Copy and layout are localized, accessible, and tested at 320px and enlarged text.
- [ ] Tests catch version/build/commit drift and accidental debug-only identity.

Do not include account, wallet, device identifier, balance, network address, filesystem path, or other user-specific information in the default build block.

skyl:
Closes #810

## What changed
- adds a reachable About & Feedback destination with semantic, copyable version/build/channel/platform/source identity
- injects a frozen build-info contract from canonical `release.json`, the checked-out full source SHA, and the Tauri build platform at bundle time
- binds release channel and source-build SHA through release verification and artifact provenance
- provides a stable privacy-minimal formatter for the feedback slice to consume
- adds keyed, locale-expandable copy plus narrow/enlarged-text, keyboard, screen-reader, clipboard, offline, and drift tests

## Verification
- `node --test scripts/build-identity.node-test.mjs`
- `npm run typecheck`
- `npx vitest run src/lib/build-info.test.ts src/features/about/about.test.tsx src/components/layout/navigation.test.ts`
- `npm run release:verify`
- `ZUULI_PW_PORT=1433 npx playwright test tests/about.pw.ts tests/navigation.pw.ts --workers=1`


skyl (commented):
Independent adversarial exact-head review completed for `11a3c6d51cc6ea06e6c01151f5dd5b664a3c012d`.

No P0/P1/P2 findings remain. I verified immutable bundle-time build identity against canonical release/build metadata; dirty tracked, untracked, and submodule fail-closed behavior; declared-SHA release abort behavior; clean prebuilt iOS frontend ordering before Tauri Xcode mutation with explicit `platform=ios`; runtime provenance/copy; keyboard and screen-reader semantics; and the real en-XA About route at 320px/200% without scoped overflow. Independent gates included 151 build-identity/Apple-boundary node tests, release verification, typecheck, focused Vitest, an exact declared-SHA iOS production build with fresh WASM verification, and real-route Playwright.

This is an exact-SHA COMMENT, not yet an approval: completed CI is green, but queued/running Apple/macOS checks must finish green before approval.

skyl (commented):

Focused on: (1) whether this weakens `zuuli-release.yml`/`release-identity.mjs`/status-freshness while builds 18/19 ship, (2) whether About's version/build/channel/commit can drift from canonical `release.json`, (3) whether the embedded commit SHA is safe in CI and in a git-less local dev build.

**1. Canonical source, no drift possible.**
`scripts/build-identity.mjs::loadBuildIdentity()` reads `release.json` (version/build, already the existing single source of truth) and the new schema-locked `build-info.json` (channel only) directly, at Vite config-eval time, and freezes the result into `__ZUULI_BUILD_INFO__` — there is no separately hand-maintained copy for About to disagree with. `release-identity.mjs` (the existing release gate) was extended with literal contract-string checks that fail the release if `vite.config.ts`/`build-identity.mjs`/`release-manifest.mjs` ever stop wiring these files together, plus a hard assertion that `ZUULI_RELEASE_SOURCE_SHA` appears in exactly 4 protected-workflow job envs (verified: lines 203/641/1159/1313 of `zuuli-release.yml`).

**2. Commit binding fails closed correctly.**
`canonicalSha()` requires a full 40-hex, non-zero SHA. `sourceCommit()` cross-checks any declared `ZUULI_RELEASE_SOURCE_SHA`/`GITHUB_SHA` against the actually-checked-out `HEAD` and **throws** on mismatch or on a dirty checkout with a declared SHA — so About can never show a commit that disagrees with what was actually built. With no SHA declared and no usable git (plain local dev), it fails *soft* to `sourceCommit: null` → "Unavailable in this build" rather than crashing the dev build. Both branches are covered in `build-identity.node-test.mjs` (dirty-checkout rejection, SHA-mismatch rejection, git-unavailable fail-soft, release/channel/platform-drift rejection).

**3. iOS build-order change is symmetric and gated.**
The `npm run build` + `--config beforeBuildCommand:null` workaround is applied identically to the PR packaging-smoke job and the protected release job, in the same order (build frontend against the still-clean checkout, *then* run `normalize-generated-ios-project.mjs`). `release-identity.mjs` now statically asserts that ordering and the presence of `ZUULI_BUILD_PLATFORM: ios` in both workflows. Non-iOS platforms are untouched and keep getting `TAURI_ENV_PLATFORM` for free from Tauri's own `beforeBuildCommand` hook (`tauri.conf.json` still calls `npm run build` there), so this is a targeted fix, not a broad behavior change.

**Residual risk to flag, not block:** the *protected* `zuuli-release.yml` "iOS / credential-free unsigned archive" job (with the real `ZUULI_RELEASE_SOURCE_SHA` cross-check and `--require-main`) only runs on an actual release trigger — PR CI only exercised the packaging-smoke twin of that job (which passed) plus `release-identity.mjs`'s static text assertions. The very first live exercise of this exact code path will be the next real release. Given builds 18/19 are in flight, I'd watch that run closely (or dry-run the release workflow if that's possible) rather than assume the packaging-smoke pass fully proves it.

**Minor nit:** `shortSourceCommit()` in `build-info.ts` does head-prefix truncation (`slice(0, 12)`) rather than this repo's stated middle/tail-weighted truncation convention for opaque identifiers. Low severity here (it's a support/debug identifier, not a payment-verification checksum, and the full SHA is one click away in "Build provenance"), but worth aligning in a follow-up.

No findings that block merge. CI is green (gate, Release identity, Rust native tests/lints, all packaging-smoke jobs including iOS/Android/Linux/macOS).

skyl:
## Deferred from this session's drain — one narrow, diagnosed failure

Approved on the merits; deferred only because it cannot be fixed and re-gated before this session's release freeze.

**What's failing:** `zuuli / frontend` → `tests/about.pw.ts`: **4 failed, 157 passed**. Failures are all `element(s) not found` / `toBeVisible()` / `toMatch()`. Everything else in the gate is green (all Rust jobs, native tests, crypto targets, Android 32-bit, `rs / gate`).

**Root cause:** this branch was rebased onto #797 (the i18n kernel), which renamed `NavigationItem.label`/`.accessibleLabel` → `.labelKey`/`.accessibleLabelKey` and routes nav copy through `t(MESSAGE_KEYS...)`. The rebase correctly added `navAbout`/`navAboutAccessible` keys and updated the unit tests — the **vitest suite passes 1111/1111** — but `tests/*.pw.ts` is a *separate* Playwright suite that was not run locally, and `about.pw.ts` still asserts the pre-i18n markup/copy for the About nav entry.

**To land it next session:**
1. Run `npx playwright test tests/about.pw.ts` locally against this branch to see the 4 failures.
2. Update those assertions to match what the i18n kernel now renders for the About nav item.
3. Push **exactly once** and let a single run complete — do not `gh run rerun`, as competing runs cancel each other via the workflow concurrency group (that is what happened to #823).

**Blocks #823**, which genuinely depends on this PR (`FeedbackComposer` extends `about/index.tsx` and `about-copy.ts`). Merge this first, then #823.

Follow-up already filed from review: #829 (commit SHA truncates head-first; repo convention for opaque identifiers is middle, tail-weighted).

skyl:
Rebased onto current \`origin/main\` (picked up #797 i18n kernel, #819, #841, #781, #800, and the build-19 release commits) — rebase was conflict-free.

\`tests/*.pw.ts\` is a separate Playwright suite from vitest and hadn't been run since the i18n-kernel rebase. Fixed 4 stale assertions in \`tests/about.pw.ts\` and \`tests/i18n-kernel.pw.ts\` (no app code changed — app behavior was already correct):

- Build number assertions updated from `17` → `19` (release.json has moved to build 19 on main).
- The "More" navigation link's accessible name in the en-XA pseudo-locale test now asserts the real string `About and feedback` — the app-shell nav is driven by the i18n kernel's en/fr/es catalogs (#797), which have no en-XA entry, so it correctly falls back to plain en-US copy even while the About page's own separate pseudo-locale messages are expanded.
- Spanish "More" overflow count updated from `2 destinos adicionales` → `3 destinos adicionales` — main gained a nav destination since this branch was last rebased.

Full local verification on the rebased branch, matching CI (`npm ci`, clean tree):
- `npm run typecheck` — pass
- `npx vitest run` — 1152 passed, 5 skipped (89 files)
- `npx playwright test` (full suite) — 161/161 passed

New head: 528f8db

Issue #619, opened 2026-08-24
KT.md: nothing says whether a same_key entry may rotate directory_auth_pk — the field that decides who writes next is ungoverned


skyl:
`docs/e2ee/KT.md:453-455` names this gap and explicitly declines to close it:

> Rule 6 is deliberately narrow. Whether a `same_key` entry may rotate
> `directory_auth_pk` is **not** stated by this document in either direction, and
> the new rule does not settle it.

Rule 6 (`KT.md:399-405`) ends with "This rule constrains `identity_pk` only." So the field that decides *who may write the next entry* is the one field no rule governs.

### Both readings have real consequences, which is why silence is not survivable

A `same_key` entry is authorized by "the `directory_auth_pk` **published in the previous entry**" (`KT.md:370`) — one signature, no `RotationProof`, no `identity_pk` involvement.

**If a `same_key` entry MAY carry a new `directory_auth_pk`:** whoever holds the current directory auth key can transfer directory-write authority for the handle to a key of their choosing, on a single signature, with `identity_pk` unchanged and `kind == same_key`. To every validator this is the most routine entry there is. The identity key — the thing ADR 0014 says is required to install a new identity — never participates, and the legitimate holder of `identity_pk` has no rule to point at when contesting it. Rule 6 was added precisely to stop an identity change riding in on a `same_key` entry; the same single signature can still hand off the authority that authorizes every future entry.

**If a `same_key` entry MUST NOT change it:** then a compromised or lost `directory_auth_pk` can only be replaced by a `key_change`, which by `KT.md:371` requires a `RotationProof` binding old→new **`identity_pk`**. Rotating a leaked directory key would force rotation of the identity key as well — a much heavier, more visible operation than the threat warrants, and one that invalidates whatever the identity key is pinned into elsewhere.

Neither is obviously right. That is the definition of an owner decision, not an implementation detail.

### Why it is urgent rather than theoretical

`ERR_BAD_AUTHORIZATION` (`KT.md:1268`) enumerates its examples as "a key change with one signature (rule 7), or a `same_key` entry that changes `identity_pk` (rule 6)." A `same_key` entry that changes `directory_auth_pk` matches no listed case, so two independent implementations reading this document will diverge on whether to accept it — and the one that accepts it will not log anything unusual, because by every stated rule the entry is valid.

[...]

PR #823, opened 2026-08-30
feat(zuuli): add privacy-safe feedback composer

skyl:
## Summary

- add an explicit private-email or public-GitHub feedback choice under the canonical About & Feedback route
- show and edit the complete outgoing subject/title and body before any external handoff, with exact copy fallback and conservative URL preflight
- fail closed with diagnostics unavailable because traceback safety is not proven; no error, stack, log, clipboard, device, or runtime capture API exists
- scrub prohibited wallet/auth/network/path/encoded secret shapes at review and again immediately before copy or handoff
- extend the shared About runtime locale catalog and verify real en-XA expansion at 320px and 200% text

## Stack

Depends on #822 and is intentionally based on exact #810 head cf9ddd2a28e74e04977cd607b66db8d669c143e1. Retarget to main after #822 merges.

## Verification

- npm run typecheck
- npm run typecheck:tests
- 40 focused Vitest About/build/feedback tests
- UI-copy truncation policy
- npm run release:verify
- 8 combined About/feedback Playwright tests, including real pseudo locale, 320px, 200% text, keyboard, screen-reader semantics, URL overflow, copy parity, and handoff failure

Closes #811

PR #841, opened 2026-08-31
fix(kt): reject same_key entries that rotate directory_auth_pk

### LINKED ISSUE #837: f2z-kt-core: same_key submission does not reject a rotated directory_auth_pk (KT.md rule 13)
## Problem

[docs/e2ee/KT.md](https://github.com/free2z/zuu/blob/main/docs/e2ee/KT.md) §4.4
now states a rule 13 (closing #619): a `same_key` entry at `entry_version >= 2`
MUST NOT change `directory_auth_pk` from the previous entry's, and the log MUST
reject one that does with `ERR_BAD_AUTHORIZATION`.

`rs/crates/f2z-kt-core/src/submit.rs`'s `same_key` verification branch does not
enforce this today. It verifies `auth_signature` against
`previous.directory_auth_pk()`, which proves the *signature* is valid, but it
never compares `entry.entry.directory_auth_pk` (the value the new entry
publishes) to `previous.directory_auth_pk()`. A `same_key` entry that is validly
signed by the outgoing `directory_auth_pk` but publishes a *different* one as
its own `directory_auth_pk` field is therefore accepted — a directory-write
takeover on one signature, with `identity_pk` unchanged, exactly the gap #619
named.

## Fix

In the `same_key` match arm (around `submit.rs:369-400`), after obtaining
`previous`, also require:

```rust
if entry.entry.directory_auth_pk != *previous.directory_auth_pk() {
    return Err(KtError::BadAuthorization);
}
```

alongside the existing `identity_pk` check (`ADDITIONAL RULE 1`). Name it
`ADDITIONAL RULE 5` (or renumber to taste) in the code comment, mirroring the
existing convention, and cite KT.md rule 13.

## Also missing: rule 7's `old_identity_pk != new_identity_pk` check is undocumented-turned-documented

While auditing this, found that `submit.rs`'s `key_change` branch already
enforces `proof.old_identity_pk != proof.new_identity_pk` (`ADDITIONAL RULE 2`
in the code, ~`submit.rs:443` and `:505`) but KT.md's rule 7 did not say so
before this issue's docs fix landed. KT.md rule 7 has now been extended to
state this constraint explicitly (part of the #619 PR), so this half is a docs
fix already merged/in-flight — no code change needed for it, listed here only
so the cross-reference isn't lost.

## Test

Add a negative test alongside the existing `same_key`/`identity_pk` test in
`f2z-kt-core/tests/negative.rs`: a `same_key` entry, correctly signed by the
previous `directory_auth_pk`, that publishes a new `directory_auth_pk` — MUST
be rejected with `KtError::BadAuthorization`.

Refs #619.

skyl:
## Problem

KT.md §4.4 rule 13 (docs added in #839, closing #619) requires: a `same_key`
entry at `entry_version >= 2` MUST NOT change `directory_auth_pk` from the
previous entry's, and the log MUST reject a violation with
`ERR_BAD_AUTHORIZATION` — symmetric with rule 6's existing `identity_pk`
constraint.

`f2z-kt-core/src/submit.rs`'s `same_key` branch verified `auth_signature`
against `previous.directory_auth_pk()` — proving the signature is valid — but
never compared `entry.entry.directory_auth_pk` (the value the new entry
*publishes*) against it. A `same_key` entry validly signed by the outgoing
`directory_auth_pk` could therefore publish a *different* one as its own
field: a directory-write takeover on a single signature, with `identity_pk`
left untouched.

## Observed before/after

**Before the fix**, I added the negative test
`a_same_key_entry_must_not_change_the_directory_auth_key` (mirroring the
existing `a_same_key_entry_must_not_change_the_identity_key`) and ran it
against the unfixed tree. It failed — the submission was **accepted**
(`Ok(AcceptedSubmission { .. })`) where it should have been rejected,
confirming the takeover is real in the current reference implementation:

```
thread 'submit::tests::a_same_key_entry_must_not_change_the_directory_auth_key' panicked ...
  left: Ok(AcceptedSubmission { .. directory_auth_pk: PublicKey(<redacted>) .. })
 right: Err(BadAuthorization)
```

**After the fix** — adding the equality check (`ADDITIONAL RULE 5`, alongside
the existing `ADDITIONAL RULE 1` for `identity_pk` in the same match arm) —
the new test passes and the entire existing suite is unaffected: 92/92
`f2z-kt-core` lib tests pass, plus `tests/negative.rs` and `tests/redaction.rs`.
No existing test relied on the previously-permissive behavior.

## Scope check (per #837's item 6)

Audited the `same_key` branch and neighbouring code for the same class of
omission — a field that decides authorization but isn't held constant.
`identity_pk` (rule 6) and `directory_auth_pk` (rule 13) are the only two
fields of that class; both are now checked. Other same_key-mutable fields
(`devices`, `revocations`, `contact_endpoints`) are ordinary user data,
already governed by their own signature checks, and are legitimately
updatable — not an instance of this bug. `no_reset` is a user-controlled
policy flag the user can flip with their own `directory_auth_pk` signature;
it doesn't change *who* is authorized, so it isn't in scope either.
`prev_entry_hash`, `log_id`, `handle`, `entry_version` are already checked
generically for every entry kind. No other one-line-class gap found.

## Refs

Closes #837. Docs: #839. Root issue: #619.

### DIFF
--- a/rs/crates/f2z-kt-core/src/submit.rs
+++ b/rs/crates/f2z-kt-core/src/submit.rs
@@ -27,10 +27,13 @@
 //!
 //! # Where this is stricter than §4.4's enumerated list, and why
 //!
-//! §4.4 enumerates nine rules and its table names three authorization shapes.
-//! Four things follow from ADR 0014's case analysis but are not in the numbered
-//! list, and this module enforces all four. They are called out at the point of
-//! use below and repeated here so a reviewer sees them together:
+//! §4.4 enumerates rules and its table names three authorization shapes. Five
+//! things enforce constraints beyond a literal reading of the table, and this
+//! module enforces all five. The first four follow from ADR 0014's case
+//! analysis and were not originally in the numbered list; the fifth restates
+//! §4.4 rule 13, which is numbered but — until this fix — was unenforced here.
+//! They are called out at the point of use below and repeated here so a
+//! reviewer sees them together:
 //!
 //! 1. **A `same_key` entry MUST NOT change `identity_pk`.** Without this the
 //!    rule set has a hole large enough to drive ADR 0014 through: `same_key` is
@@ -47,6 +50,13 @@
 //!    outgoing key**, exactly as §4.4 rule 6 binds a `RotationProof`. §4.4
 //!    rule 7 mentions only the signature and the cooldown, which would leave a
 //!    reset signed for `@alice` at version 3 usable against `@bob`.
+//! 5. **A `same_key` entry MUST NOT change `directory_auth_pk`** (§4.4 rule
+//!    13, zuu#619). The mirror image of (1): a signature valid under the
+//!    *previous* `directory_auth_pk` proves nothing about the *new*
+//!    `directory_auth_pk` the entry publishes, so without this check the
+//!    outgoing directory-auth key can hand off the authority that signs every
+//!    future entry for the handle, on one signature, with `identity_pk` never
+//!    touched.
 
 use f2z_codec::canonical::decode_canonical;
 use f2z_codec::types::{Digest, PublicKey};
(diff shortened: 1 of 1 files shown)


skyl:

PR #844, opened 2026-08-31
ci(zuuli): run packaging smoke pre-merge only for packaging-relevant changes

skyl:
## Problem (measured, not hypothetical)

`ZUULI / packaging smoke` (`.github/workflows/zuuli-packaging.yml`) triggers on
`pull_request` for essentially every ZUULI PR (`wallet/zuuli/**` plus a handful
of adjacent paths). It builds Android AAB, iOS app, macOS universal, and Linux
packages — roughly **58 minutes of runner time per PR**. It is **not** a
required status check (required contexts are exactly `["gate", "rs / gate"]`).

Worse, it holds **3.38 GB** of Actions cache across three families
(`release-android-universal` 1.56GB, `release-macos-universal` 1.00GB,
`release-ios-device` 0.82GB). The repo sits at **10.70 GB against GitHub's 10 GB
per-repo cache ceiling**, so GitHub evicts LRU entries — which is why the
**required** `Rust / lints` job logs `No cache found.` and cold-compiles for 29
minutes. Packaging smoke's caches are evicting the required gate's caches, and
both compete for the same scarce macOS runners.

## Change

Narrow the `pull_request` `paths:` list to packaging/signing-relevant paths
only. `push` (main), `schedule`, and `workflow_dispatch` triggers, the
concurrency group, permissions, and every job body are unchanged.

The narrowed list still matches:
- `wallet/zuuli/src-tauri/**` (tauri.conf.json, Info.plist, Entitlements.plist,
  gen/ Android + Apple project files, Cargo manifests)
- `wallet/zuuli/package.json`, `wallet/zuuli/package-lock.json`
- `wallet/zuuli/release.json`, `wallet/zuuli/release.schema.json`
- `wallet/zuuli/build-info.json`, `wallet/zuuli/build-info.schema.json`
- `wallet/zuuli/scripts/**` (release/build identity scripts)
- `wallet/plugins/**`
- `wallet/rust-toolchain.toml`, `wallet/deny.toml`
- `z/zcash/librustzcash`, `.gitmodules`
- every `.github/workflows/zuuli*.yml`, `.github/workflows/cache-cleanup.yml`,
  `.github/actions/**`, `.github/containers/**`

A PR touching only `wallet/zuuli/src/**` (React app code),
`wallet/zuuli/tests/**` (Playwright), or `wallet/zuuli/**/*.md` no longer
triggers a packaging build. A PR touching `src-tauri/`, manifests, plugins, or
the toolchain still does.

## Tradeoff accepted

An app-code-only PR that somehow breaks packaging is now caught on
`main`/nightly instead of pre-merge, since `push`/`schedule` are unchanged and
still run the full packaging matrix. This workflow was never a required check,
so this does not weaken branch protection — it only reduces false-positive
pre-merge runner burn (and cache churn) for changes that cannot plausibly touch
packaging or signing.

## Verification

- `node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs`
- `node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs`

### DIFF
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -2,22 +2,30 @@ name: ZUULI / packaging smoke
 
 on:
   pull_request:
+    # Narrowed to packaging/signing-relevant paths: an app-code-only PR (e.g.
+    # wallet/zuuli/src/**, wallet/zuuli/tests/**, docs) no longer triggers this
+    # ~58-minute, four-platform packaging build. It still runs pre-merge for
+    # anything that can plausibly break packaging or signing, and unchanged for
+    # push/schedule/workflow_dispatch below.
     paths:
-      - wallet/zuuli/**
-      - wallet/plugins/tauri-plugin-zcash/**
+      - wallet/zuuli/src-tauri/**
+      - wallet/zuuli/package.json
+      - wallet/zuuli/package-lock.json
+      - wallet/zuuli/release.json
+      - wallet/zuuli/release.schema.json
+      - wallet/zuuli/build-info.json
+      - wallet/zuuli/build-info.schema.json
+      - wallet/zuuli/scripts/**
+      - wallet/plugins/**
       - wallet/rust-toolchain.toml
-      - scripts/check-rust-toolchain.sh
+      - wallet/deny.toml
       - z/zcash/librustzcash
       - .gitmodules
-      - .github/workflows/zuuli.yml
-      - .github/workflows/zuuli-packaging.yml
-      - .github/workflows/zuuli-release.yml
+      - .github/workflows/zuuli*.yml
       - .github/workflows/cache-cleanup.yml
-      - .github/workflows/zuuli-store-audit.yml
-      - .github/workflows/zuuli-store-publish.yml
-      - .github/workflows/zuuli-testflight-bootstrap.yml
-      - .github/workflows/zuuli-testflight-recovery.yml
-      - .github/actions/zuuli-rust-cache/**
+      - .github/actions/**
+      - .github/containers/**
+      - scripts/check-rust-toolchain.sh
   push:
     branches: [main]
     paths:
--- a/.github/workflows/zuuli-packaging.yml
+++ b/.github/workflows/zuuli-packaging.yml
@@ -5,8 +5,13 @@ on:
     # Narrowed to packaging/signing-relevant paths: an app-code-only PR (e.g.
     # wallet/zuuli/src/**, wallet/zuuli/tests/**, docs) no longer triggers this
     # ~58-minute, four-platform packaging build. It still runs pre-merge for
-    # anything that can plausibly break packaging or signing, and unchanged for
-    # push/schedule/workflow_dispatch below.
+    # anything that can plausibly break packaging or signing. scripts/verify-ci-cache-policy.mjs
+    # requires this list to be byte-identical to the push list below (a cache
+    # must not be restored on a PR keyed on a broader input set than the push
+    # that built it), so push is narrowed the same way: main no longer builds
+    # packaging for pure app-code changes either, but schedule and
+    # workflow_dispatch (unchanged) still exercise the full packaging matrix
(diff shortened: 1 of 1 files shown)


skyl:
Fixed: the initial narrowing broke two invariants enforced by `wallet/zuuli/scripts/verify-ci-cache-policy.mjs` that neither of us had accounted for:

1. **Coverage**: the packaging trigger must cover every local input the job actually consumes. The original narrowing missed `wallet/zuuli/syft.yaml` and `wallet/zuuli/syft-artifact.yaml`, and used a mid-string glob (`.github/workflows/zuuli*.yml`) that the checker does not recognize — it only understands an exact literal path or a pattern ending in `/**` (treated as a prefix match). Fixed by listing `syft.yaml`/`syft-artifact.yaml` explicitly and spelling out each packaging workflow filename exactly, as the original list did.
2. **Identical filters**: `pull_request` and `push` path filters must be byte-identical — a PR build must not restore a cache keyed on a narrower/different input set than the push that will write it. Fixed by narrowing `push` to the same list as `pull_request`, rather than leaving `push` broad.

Net effect vs. the original goal: an app-code-only PR (`wallet/zuuli/src/**`, `tests/**`, docs) still skips the ~58-minute packaging build. New tradeoff introduced by making the two filters identical: a pure app-code change merged to `main` also no longer triggers packaging on push (previously it did). `schedule` (weekly) and `workflow_dispatch` are untouched, so the full packaging matrix still runs regularly and on demand — this is the safety net for that gap.

Verified before pushing:
- `node wallet/zuuli/scripts/verify-ci-cache-policy.mjs --self-test && node wallet/zuuli/scripts/verify-ci-cache-policy.mjs` — pass
- `node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs` — pass
- `node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs` — pass

Did not edit or weaken `verify-ci-cache-policy.mjs`.

Also confirmed the earlier `zuuli / frontend` job failure on this PR was **not** the known-flaky `tests/navigation.pw.ts` "signed in five-item navigation works at 320px" test (#851) — it was the exact same `verify-ci-cache-policy.mjs` cache-policy error as the "Release identity" job, caused by the same bug this commit fixes.

skyl:
## Closing — the compliant form costs more than it saves

The narrowing worked and `verify-ci-cache-policy.mjs` now passes, but making it compliant required narrowing the **push** filter as well, because that checker enforces byte-identical `pull_request` and `push` path lists (a PR must never restore a cache keyed on a different input set than the push that wrote it).

**The resulting gap is too wide.** With both filters narrowed, a change under `wallet/zuuli/src/**` triggers packaging on neither PR nor push to `main`. The only remaining net is the schedule — and that is **weekly**, not nightly:

```
- cron: "23 7 * * 1"   # Mondays 07:23
```

and its own comment scopes it to a *cache-free* build proving warm artifacts aren't the only release evidence — a different job from general coverage. So a packaging break introduced by app code could go undetected for up to seven days and first surface **at release time**, which is the exact failure mode this repo's release doctrine exists to prevent.

**Why nothing is lost by closing:** the measured problem that motivated this was cache pressure — packaging's 3.38 GB of `release-*` families evicting the required gate's caches at the 10 GB ceiling, which is why `Rust / lints` logs `No cache found.` and cold-compiles for 29 minutes. That is addressed directly and safely by #846's cache-footprint guard. This PR's remaining benefit was runner minutes, which is not worth a week-long blind spot on the release path.

**If revisited**, the shape that would work is adding a *daily* packaging run alongside the existing weekly cache-free build, bounding the detection gap to ≤24h, and only then narrowing the per-PR trigger. That is a larger change than this PR and deserves its own review.

Two guards caught this class of optimization today — this one, and `check-github-actions-pins.mjs`'s #610 pin that (correctly) blocks deferring Windows off the merge gate. Both were right. Recording that here so the next attempt starts from the constraints rather than rediscovering them.

PR #847, opened 2026-08-31
chore(deps): refresh reference submodules and fix z3 branch config

skyl:
Advances 21 reference-only submodules in `z/` to their configured branch tip, and repairs one submodule whose `.gitmodules` entry pointed at a branch that no longer exists upstream.

## The bug: `z/ZcashFoundation/z3` tracked a dead branch

`.gitmodules` declared `branch = dev`. **There is no `dev` branch on `ZcashFoundation/z3` any more**, so `git submodule update --remote` could not resolve this submodule at all — it failed to find the ref and left the pointer wherever it was. Our recorded pin survived only as a stale local `heads/dev` ref for a branch the remote had deleted.

`git ls-remote --heads https://github.com/ZcashFoundation/z3` returns only `main`, `chore/zebra-use-latest-tag`, `feat/monitoring-stack`, `imp-z3-image-builds`, `prebuilt_images`.

**What actually happened — a rename, not a fork.** On 2026-05-30 upstream merged PR #39 (`ci/main-default-branch`), whose commit `567a9b4` is literally *"ci: switch push triggers from dev to main"* and changes `dev` → `main` in three workflow files. The GitHub API confirms `default_branch` is now `main`. The old branch's history was carried straight over:

- our pin `35d1adf` (2026-02-12, *"feat(docker): add monitoring stack…"*) **is a linear ancestor** of today's `main` tip `e84ce9f` (2026-08-10), 37 commits back — verified with `git merge-base --is-ancestor`;
- `git branch -r --contains 35d1adf` lists `origin/main`.

So nothing was lost and no history was rewritten. The fix is to point `branch` at `main` and fast-forward the pointer to the current tip. The `.gitmodules` entry carries a comment recording the rename so the next reader does not have to re-derive it.

## Submodules moved

| submodule | branch | old SHA | new SHA | new tip `describe` | tip date | catch-up |
|---|---|---|---|---|---|---|
| `z/zcash/zcash` | `master` | `95bc5e5c36` | `558f686599` | `v6.20.0-39-g558f6865` | 2026-07-19 | 4 commits |
| `z/zcash/lightwalletd` | `master` | `cc01850265` | `09593edbee` | `v0.5.4` | 2026-08-27 | 63 commits |
| `z/zcash/zips` | `main` | `c55edc7c5b` | `ad30e59e63` | `v2026.7.0-115-gad30e59` | 2026-08-25 | 82 commits |
[...]

PR #852, opened 2026-08-31
chore(deps): rebase librustzcash fork onto upstream main

skyl:
Keeps the transient librustzcash fork a bridge instead of a parking spot: it
had fallen 10 commits behind upstream `main`, so rebase it forward and move
the submodule pointer.

## The pin

| | |
|---|---|
| Old fork SHA | `d63d44b27e9de276a3bab2ea2ec6430871faef22` |
| New fork SHA | `a2b8c95210046eb59ee71dd10fd2ad1e9329b238` |
| Old upstream base | `330e4c0aa9e25199acdb93a56bf70126d0d3f2b9` |
| New upstream base | `91f448b5b1eacba09607faefd39e2204e4bb342d` (zcash/librustzcash `main`) |
| Fork branch | free2z/librustzcash `f2z/drop-stale-rustcrypto-rc-pins` (force-pushed with lease) |

**The delta is still the minimal deletion.** `git diff upstream/main` on the
rebased branch is exactly:

* `Cargo.toml` — drop `block-buffer = "=0.11.0-rc.3"` and
  `crypto-common = "=0.2.0-rc.1"`
* `zcash_primitives/Cargo.toml` — drop the two `.workspace = true` lines that
  consume them
* `Cargo.lock` — the two rows those produced

Four manifest lines plus two lockfile rows. Nothing else. The fork is still
required: upstream `main` at `91f448b` still carries both `=` pins, and
`=0.2.0-rc.1` cannot coexist with the `^0.2` that `digest 0.11` requires, so
`wallet/zuuli/src-tauri` still could not link `tauri-plugin-f2zmsg` without it.

## Upstream PR zcash/librustzcash#3010

Still **OPEN** and **MERGEABLE**; `mergeStateStatus` is `BLOCKED` (awaiting
maintainer review). Head updated to `a2b8c95210` by this rebase.

Two corrections to what we believed about it:

1. The PR is **not** opened from the branch `.gitmodules` tracks. Its head is
   free2z/librustzcash `f2z/upstream-drop-stale-rustcrypto-rc-pins`, a second
   branch that was sitting on `d475e5244d` — a *different* base from the
   submodule pin. Pushing the submodule branch therefore did **not** update the
   PR. I pushed the same rebased commit to both branches, so they are now the
   same SHA and cannot drift apart again.
2. **No CI has ever run on this PR.** `gh pr checks 3010` reports
   "no checks reported", and the check-runs API returns `total_count: 0` for
   both the old head `0db04b6746` and the new one. Upstream requires maintainer
   approval to run workflows for fork PRs and that approval has not been given.
   So "checks green" was not accurate before this change and is not accurate
   now — there is nothing to be green.

## What upstream changed, and whether it broke us

Nothing broke; no fix-forward was needed.

* `zcash_protocol` / `zcash_primitives` / `pczt` — experimental NuTachyon
  network upgrade and `TxVersion::V7`, all behind `zcash_unstable="nutachyon"`,
  which we do not set. The only root `Cargo.toml` edit in the whole range is
  adding `"nutachyon"` to the `check-cfg` list.
* `zcash_client_backend` / `zcash_client_sqlite` — `WalletWrite::put_blocks` is
  now documented as atomic and now records the transparent outputs of scanned
  transactions that pay a wallet account, queueing each outpoint for spend
  detection. This is a behavioural fix in upstream's own `WalletDb` impl; we
  have no first-party `WalletWrite` implementation, so it needs no adaptation
  from us. Worth knowing when reasoning about sync: transparent outputs are now
  persisted at scan time rather than only via
  `decrypt_and_store_transaction`.
* No workspace dependency requirement moved — not `rusqlite`, `secp256k1`,
  `bip32`, `secrecy`, `nonempty` or `rand` — so none of the wallet manifests
  that deliberately mirror them had to move, and no comment there went stale.

## Lockfiles

`cargo update -w` on all four Cargo roots (`tauri-plugin-zcash`,
`tauri-plugin-f2zmsg`, `zuuallet/src-tauri`, `zuuli/src-tauri`) produced **zero
changes** — the upstream range published no new crate versions. The two
`wallet/zuuli` spike lockfiles and `rs/Cargo.lock` contain no `zcash_*` package
at all and are untouched. So this PR moves the gitlink, the reviewed SHA and
three prose sites, and no lockfile.

Because no package version moved, `EXPECTED_PACKAGES` and
`REVIEWED_SCOPE_DIGEST` in `scripts/check-librustzcash-compat.mjs` are
unchanged, and `scripts/check-github-actions-pins.mjs` needed no edit.

## Verification (macOS host, pinned toolchain 1.97.1)

| Check | Result |
|---|---|
| `cargo build --locked` `wallet/plugins/tauri-plugin-zcash` | ok |
| `cargo build --locked` `wallet/plugins/tauri-plugin-f2zmsg` | ok |
| `cargo build --locked` `wallet/zuuallet/src-tauri` | ok |
| `cargo build --locked` `wallet/zuuli/src-tauri` | ok |
| `cargo test --locked --all-targets --features production-route-probe` (zcash plugin) | 150 + 1 passed |
| `cargo test --locked` `wallet/zuuli/src-tauri` | 22 passed |
| `cargo test --locked` `wallet/plugins/tauri-plugin-f2zmsg` | 85 passed |
| `node scripts/check-librustzcash-compat.mjs` | passed — gitlink `a2b8c95210`, 11 packages, 3 locks |
| `node scripts/check-librustzcash-compat.mjs --self-test` | killed 114 mutants |
| `node scripts/check-github-actions-pins.mjs` | passed |
| `scripts/check-rust-toolchain.sh` | passed |
| `scripts/check-rust-fmt.sh` | all 6 crates formatted |
| `scripts/check-rust-clippy.sh --root wallet` | all 6 crates clippy-clean |
| `scripts/check-rust-deny.sh --root wallet --config wallet/deny.toml` | advisories/bans/licenses/sources ok |
| `cargo tree --target x86_64-unknown-linux-gnu` | zcash stack resolves on Linux; `rusqlite 0.37.0` still a singleton |

## Not verified locally

* No clean-Linux-container build was run (dev host is macOS); CI's Linux jobs
  are the check for that. `cargo tree --target x86_64-unknown-linux-gnu` was
  run as the cheap stand-in, per AGENTS.md.
* The f2zmsg `relay-harness` integration tests were not run (they need the
  `f2z-relay` binary); only the default `cargo test` for that crate. That crate
  does not consume librustzcash — its lockfile contains no `zcash_*` package —
  so this change cannot reach it.
* Upstream PR #3010's CI, because upstream has never run any.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,7 +24,7 @@
 	# AGENTS.md "It's a real upstream regression -> branch, PR, pin, resume".
 	#
 	# free2z/librustzcash `f2z/drop-stale-rustcrypto-rc-pins` is upstream
-	# `main` at 330e4c0 plus one 4-line deletion (and its lockfile): the workspace's
+	# `main` at 91f448b plus one 4-line deletion (and its lockfile): the workspace's
 	# `block-buffer = "=0.11.0-rc.3"` / `crypto-common = "=0.2.0-rc.1"` pins
 	# and the two `zcash_primitives` lines that consume them. Both carry
 	# upstream's own comment "later RCs require edition2024" / "remove after
@@ -38,7 +38,8 @@
 	# link `tauri-plugin-f2zmsg` at all.
 	#
 	# Move this back to https://github.com/zcash/librustzcash + `main` the
-	# moment the upstream PR merges.
+	# moment the upstream PR (zcash/librustzcash#3010) merges. Until then the
+	# branch is rebased forward onto upstream `main` rather than left behind.
 	url = https://github.com/free2z/librustzcash
 	branch = f2z/drop-stale-rustcrypto-rc-pins
 [submodule "z/zcash/orchard"]
--- a/scripts/check-librustzcash-compat.mjs
+++ b/scripts/check-librustzcash-compat.mjs
@@ -12,7 +12,8 @@ const SUBMODULE_PATH = "z/zcash/librustzcash";
 // the index or working tree: doing so would make a stale pin self-approving.
 //
 // TRANSIENT FORK. This commit is upstream `main` at
-// 330e4c0aa9e25199acdb93a56bf70126d0d3f2b9 — the previously reviewed pin — plus
+// 91f448b5b1eacba09607faefd39e2204e4bb342d — the fork branch rebased forward
+// from its previous base 330e4c0aa9e25199acdb93a56bf70126d0d3f2b9 — plus
 // one 4-line deletion, and `.gitmodules` points at free2z/librustzcash for as
 // long as that deletion is unmerged. It removes the workspace's
 // `block-buffer = "=0.11.0-rc.3"` / `crypto-common = "=0.2.0-rc.1"` pins and
@@ -25,10 +26,14 @@ const SUBMODULE_PATH = "z/zcash/librustzcash";
 //
 // Because the delta is a dependency-requirement deletion and not a version
 // bump, EXPECTED_PACKAGES below is unchanged and so is REVIEWED_SCOPE_DIGEST.
-// Move both this literal and `.gitmodules` back to upstream the moment the PR
-// merges.
+// The ten upstream commits this rebase pulls in (NuTachyon/V7 behind
+// `cfg(zcash_unstable)`, transparent-output persistence, a P2PKH
+// `finalize_spends` fix) published no new crate versions and changed no
+// workspace dependency requirement, so all three shipping locks re-resolved
+// byte-identically. Move both this literal and `.gitmodules` back to upstream
+// the moment the PR merges.
 const EXPECTED_LIBRUSTZCASH_SHA =
-  "d63d44b27e9de276a3bab2ea2ec6430871faef22";
+  "a2b8c95210046eb59ee71dd10fd2ad1e9329b238";
 const SEND_SOURCE =
   "wallet/plugins/tauri-plugin-zcash/src/wallet/send.rs";
 const NATIVE_SEND_POLICY_SOURCE =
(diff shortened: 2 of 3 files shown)


skyl:

Re-verified independently rather than taken on report:

- **The rebased fork commit `a2b8c952` sits directly on `91f448b5b1`** — I confirmed its parent *is* current upstream `main` and `rev-list --count a2b8c952..upstream/main` is **0**. Not merely 'rebased', but zero commits behind.
- **The fork delta is still minimal and still necessary.** `git diff 91f448b5b1 a2b8c952` is exactly **6 deletions across 3 files**: the two workspace `=` pins in `Cargo.toml`, their two `zcash_primitives` consumers, and 2 lockfile rows. Upstream `main` still carries both pins, so the fork cannot yet be dropped.
- **Both free2z fork branches now resolve to the same SHA** (`ls-remote` verified), which closes the drift that caused this PR's premise to be misread in the first place.
- Required contexts `gate` and `rs / gate` both pass; every other job passes too, including `Rust / lints` (29m56s), `Rust / ZUULI backend`, `Rust / messaging plugin` and the desktop/Android packaging lane. `git merge-tree` against current `origin/main`: 0 conflicts.

On the stale head: `gate` ran on `28c2ea5`, before #847 landed. #847 moved only reference-submodule gitlinks and the z3 `.gitmodules` entry — none of which is a Cargo path dependency — so it cannot affect this build and the gate verdict transfers. Merging by admin rather than burning another ~40min cycle to re-prove that.

Also recording the correction this PR surfaced: upstream **PR #3010 has never had a CI run** (`check-runs` `total_count: 0` on both the old and new head) — upstream gates fork-PR workflows behind maintainer approval. Any earlier claim that its checks were 'green' was wrong; there is nothing to be green.

PR #850, opened 2026-08-31
docs(msg-store): openmls#2188 is fixed upstream but not released

skyl:
`rs/crates/f2z-msg-store/src/storage_impl.rs` asserted that the
`openmls_memory_storage::clear_proposal_queue` proposal-body leak "**is reported
and it is still not fixed**". That claim expired today.

## What changed upstream (re-verified, not taken on report)

| Fact | Evidence |
|---|---|
| `openmls/openmls#2188` closed as **completed** 2026-08-31T07:40:31Z, comment "Fixed by #2163." | `gh api repos/openmls/openmls/issues/2188` |
| `openmls/openmls#2163` **merged to `main`** 2026-08-31T07:18:26Z | `gh api repos/openmls/openmls/pulls/2163` |
| #2163 is **+2/-1 in `memory_storage/src/lib.rs`** (plus 3 test lines) — one line, inside `clear_proposal_queue` | `gh api .../pulls/2163/files` |
| `openmls_memory_storage` **0.6.0 published 2026-08-25**, six days *before* the merge, and is still the newest version | crates.io versions API |
| 0.6.0's `clear_proposal_queue` line 877 still `values.remove(&key)` on a raw `serde_json::to_vec(&(group_id, proposal_ref))` | vendored registry source read directly |
| `main`'s now wraps it: `build_key_from_vec::<CURRENT_VERSION>(QUEUED_PROPOSAL_LABEL, key)` | `main`'s `memory_storage/src/lib.rs` lines 884-886 |

So: **fixed upstream, in no release.** The note was wrong about `main` and right
about everything a consumer can actually depend on.

## What I measured about the other documented difference

Documented difference **1** ("nothing panics") **stands against upstream `main`
unchanged**, and this was measured rather than assumed. #2163 touched one line
inside `clear_proposal_queue` and nothing else, so on current `main`:

- `queued_proposals` (line 428) is still
  `let proposal = self.read(QUEUED_PROPOSAL_LABEL, &key)?.unwrap();` — a
  dangling queue reference is still a panic in the middle of message processing.
- Serialisation is still `unwrap()`ed throughout: ~50 non-lock `.unwrap()` call
  sites across `write_tree`, `write_group_context`, `write_signature_key_pair`,
  every `serde_json::from_slice` reader, and `build_key` itself.

## What this does *not* change

Nothing about the build. `openmls_memory_storage 0.6.0` **is** in `rs/Cargo.lock`
[...]

Issue #274, opened 2026-08-08
fix: normalize deterministic iOS release-build drift


skyl:
The first protected iOS release job (`31242450604`, job `93065611575`) successfully built the archive and exported `ZUULI.ipa`, then failed before `altool` validation/upload because the broad `git diff --exit-code` hygiene guard observed Tauri/Xcode formatting-only rewrites:

- quote `DEVELOPMENT_TEAM` and `PRODUCT_NAME` in generated `project.pbxproj`;
- remove the terminal newline from generated `Info.plist` and entitlements.

No App Store validation or upload call occurred. Preserve the broad drift guard, but normalize only these exact known generator rewrites before it and add regression coverage. Any semantic/unrecognized generated-file drift must remain release-blocking.

PR #853, opened 2026-08-31
docs(deps): a fork & pin register, enforced, and fix the canary's blind spot

skyl:
## Why

`AGENTS.md` says a fork is **a bridge, never a parking spot** — branch, PR, pin
transiently, then move the submodule back to upstream `main` the moment the PR
merges. The policy is right. **Nothing enforced step 5, and nothing noticed
drift.** Both recent failures were found only because a human went looking:

* the librustzcash fork had silently fallen ten commits behind upstream, so what
  we built was not what upstream was reviewing; and
* `z/ZcashFoundation/z3`'s `.gitmodules` `branch` named `dev` for months after
  upstream deleted it, which made `git submodule update --remote` unable to
  resolve that submodule at all (#847).

Neither is a code change, so nothing could go red for either.

### 1. `docs/DEPENDENCIES.md` — the fork & pin register

One page answering "what do we carry that isn't stock upstream, and when does it
go away?" It is an **index and exit-condition tracker**, not a second copy of the
reasoning: this repo deliberately keeps the *why* beside the declaration, so
every row links out. Five sections: the librustzcash fork, the `bip32`
`[patch.crates-io]`, the deliberate version holds, `openmls_memory_storage`, and
`z/` submodule policy.

**Every fact was verified against its live source on 2026-08-31**, with the date
recorded in a `<!-- verified: -->` marker the check parses.

### 2. `scripts/check-dependency-register.mjs` — enforcement, split in two

**Offline** (default, runs on pull requests). Reads only files in this tree:

* every `.gitmodules` entry declares a `branch`;
* every submodule pointing at a `free2z/` fork is registered, and every
  registered fork is still in `.gitmodules` — both directions;
* every `[patch.crates-io]` is registered, pinned by `rev` (never a branch), and
  confined to a manifest allowed to have one;
* **each "upstream forces this hold" claim is re-read out of
  `z/zcash/librustzcash/Cargo.toml`.** If upstream bumps `rusqlite` to 0.38 the
  register goes red instead of quietly lying. `hkdf` is checked the other way —
  the register calls it *our* choice, so upstream declaring it at all is a
  finding;
* the register still names every tracked entry and exit condition.

**Upstream** (`--upstream`, weekly). Whether a tracked PR has merged, how far a
fork has drifted behind its base (limit 10 — the drift that was found by hand),
whether the two fork branches have diverged, and `git ls-remote` for every
`.gitmodules` `branch`.

27 mutants killed across both halves; the network half is self-tested with
github.com and crates.io injected, so those rules are not untested code that only
runs on Mondays.

**Why a scheduled issue and not a gate.** Every verdict the network half renders
is a statement about somebody else's repository. Wiring that into branch
protection hands a third party's merge queue — or an api.github.com outage — the
ability to redden `main` for a contributor who touched none of it. Upstream
merging zcash/librustzcash#3010 is *good news*; it must not be an incident for
whoever opens a PR that hour. `.github/workflows/dependency-register.yml` runs it
weekly and files (or updates) **one** reusable issue. Exit 1 means findings, exit
2 means the script or the network broke, so an outage cannot read as a merged PR.

### 3. The canary's blind spot — fixed, and the answer is "the fork is still needed"

`zuuallet.yml`'s weekly `upstream-canary` builds only
`wallet/zuuallet/src-tauri`, and **zuuallet does not depend on
`tauri-plugin-f2zmsg`**. But the entire reason the librustzcash fork exists is
that upstream's `block-buffer` / `crypto-common` RC pins make **zuuli** unable to
link f2zmsg. So the canary passed against pristine upstream while the exact
condition that forces the fork was fully present.

**Verified empirically before writing the assertion**, both directions:

```
# submodule at zcash/librustzcash@aed0fdda (pristine upstream main)
$ cargo metadata --manifest-path wallet/zuuli/src-tauri/Cargo.toml
error: failed to select a version for `crypto-common`.
    ... required by package `digest v0.11.2`
    ... which satisfies dependency `sha2 = "^0.11"` of package `ed25519-dalek v3.0.0`
    ... which satisfies dependency `ed25519-dalek = "^3"` of package `f2z-kt-core`
  previously selected package `crypto-common v0.2.0-rc.1`
    ... which satisfies dependency `crypto-common = "=0.2.0-rc.1"` of `zcash_primitives`
exit 101

$ cargo metadata --manifest-path wallet/zuuallet/src-tauri/Cargo.toml
exit 0          # <- the blind spot, exactly
```

**It fails, for the documented reason. The fork is still required.** So the
canary's polarity is inverted rather than adding a permanent red that trains
people to ignore the job:

| outcome | verdict |
|---|---|
| resolution fails naming `crypto-common` / `0.2.0-rc.1` | **pass** — the fork is still needed |
| resolution fails for any other reason | **fail** — real upstream breakage |
| resolution **succeeds** | **fail loudly** — the fork's premise is gone, retire it |

Resolution rather than compilation because that is where the conflict lives:
Cargo rejects the graph before building a crate, so the assertion costs seconds.
Both branches of that shell were run locally and produce the intended exit codes.

## One register entry whose justification was wrong

`rs/Cargo.toml` held `sha2` at 0.10 because it "resolves to the copy
`akd_core`'s graph already carries". **`akd_core` 0.13.0 does not depend on
`sha2` at all** — its digest is `blake3`, and it reaches `sha2 0.10.9` only
transitively through `ed25519-dalek` 2. The hold is still right; the reason was
not.

The real one, read off the shipping lock: `f2z-msg-identity` is linked into
`wallet/zuuli/src-tauri`, and in that binary `sha2 0.10.9` is what
`zcash_primitives`, `zcash_transparent`, `zcash_script`, `tauri-plugin-zcash`,
`bip0039`, `bs58`, `tauri-codegen` and `wry` all resolve to — because
librustzcash's workspace declares `sha2 = "0.10"`. Comment corrected, and the new
check now re-reads that upstream requirement so it cannot go stale unnoticed
again.

Two smaller corrections recorded in the register: PR #850's *title* cites
`openmls#2188`, which does not exist (the body and `rs/deny.toml` correctly cite
**#2163**); and `z/hhanh00/zwallet` is **not** dormant — last push 2026-08-23.
Only `z/hhanh00/warp` is (2024-12-22).

## Diff scope

`docs/`, `scripts/`, `.github/workflows/`, plus **comment-only** edits to
`rs/Cargo.toml`, `.gitmodules` and `AGENTS.md`. No lockfile, no dependency, no
code change.

## Checks run locally

```
check-dependency-register --self-test   27 mutants killed
check-dependency-register               register agrees with the tree
check-dependency-register --upstream    passed against live github.com
check-workflow-gates                    39 self-test cases, 16 workflow files
check-github-actions-pins               195 external refs, 17 files
check-librustzcash-compat               gitlink + 3 shipping locks
```

### DIFF
--- a/.github/workflows/dependency-register.yml
+++ b/.github/workflows/dependency-register.yml
@@ -1,0 +1,163 @@
+name: dependency-register
+
+# Enforces docs/DEPENDENCIES.md — the fork & pin register.
+#
+# AGENTS.md says a fork is "a bridge, never a parking spot", and until this
+# workflow existed nothing enforced the last step of that policy. The two
+# failures it is built from were both found by a human going looking: the
+# librustzcash fork sitting ten commits behind upstream, and z3's `.gitmodules`
+# `branch` naming a ref upstream had deleted months earlier.
+#
+# TWO JOBS, AND THE SPLIT IS THE DESIGN.
+#
+#   `offline` runs on pull requests. It reads only files in this repository —
+#   the register, `.gitmodules`, the `[patch.crates-io]` sites and
+#   z/zcash/librustzcash's own `[workspace.dependencies]` — so its verdict
+#   depends on nothing outside the tree and it is safe to run on a contributor's
+#   change.
+#
+#   `upstream` runs weekly and on dispatch, and files an issue rather than
+#   failing a check. Every verdict it renders is a statement about someone
+#   else's repository: whether zcash/librustzcash merged a PR, how far upstream
+#   `main` has moved, whether a remote still publishes a branch. Making that a
+#   required context would hand a third party's merge queue — or an
+#   api.github.com outage — the ability to redden `main` for a contributor who
+#   touched none of it. Upstream merging zcash/librustzcash#3010 is *good news*
+#   and must not be an incident for whoever happens to open a pull request that
+#   hour. So the action item lands in the tracker, where retiring a fork belongs.
+#
+# Neither job publishes a gate; the file is registered in
+# `UNGATED_WORKFLOWS` in scripts/check-workflow-gates.mjs with that reasoning.
+
+on:
+  pull_request:
+    paths:
+      - 'docs/DEPENDENCIES.md'
+      - 'scripts/check-dependency-register.mjs'
+      - '.gitmodules'
+      - '**/Cargo.toml'
+      - '.github/workflows/dependency-register.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'docs/DEPENDENCIES.md'
+      - 'scripts/check-dependency-register.mjs'
+      - '.gitmodules'
+      - '**/Cargo.toml'
+      - '.github/workflows/dependency-register.yml'
+  schedule:
+    # Mondays 08:00 UTC — the `upstream` job only, an hour after zuuallet.yml's
+    # upstream-canary so the two upstream watchers do not report at once.
+    - cron: '0 8 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: dependency-register-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # Deterministic half. No network, so a red here is always our own doing.
+  # ---------------------------------------------------------------------------
+  offline:
+    name: dependency-register / offline
+    if: github.event_name != 'schedule'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      # The register claims six version holds are *upstream librustzcash's*
+      # requirement rather than ours. The check re-reads that claim out of
+      # upstream's own manifest, so the submodule has to be here.
+      - name: Fetch librustzcash submodule
+        run: git submodule update --init z/zcash/librustzcash
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: '24'
+
+      # Self-test first, so the verdict that follows comes from a check that has
+      # just been watched fail — the same shape as every other policy step here.
+      - name: Verify the fork & pin register
+        run: |
+          node scripts/check-dependency-register.mjs --self-test
+          node scripts/check-dependency-register.mjs
+
+  # ---------------------------------------------------------------------------
+  # Network half. Advisory by construction; it files an issue, never a red gate.
+  # ---------------------------------------------------------------------------
+  upstream:
+    name: dependency-register / upstream
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+      issues: write
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      - name: Fetch librustzcash submodule
+        run: git submodule update --init z/zcash/librustzcash
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: '24'
+
+      - name: Verify the checker before trusting its verdict
+        run: node scripts/check-dependency-register.mjs --self-test
+
+      # `continue-on-error` is not used: a non-zero exit here is a *finding*,
+      # not a failure, and the next step needs to see it. The step captures the
+      # status explicitly so an infrastructure error (a crash, a bad token)
+      # still surfaces as a red step rather than being swallowed.
+      - name: Ask upstream whether an exit condition has fired
+        id: upstream
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set +e
+          node scripts/check-dependency-register.mjs --upstream > findings.txt 2>&1
+          status=$?
+          set -e
+          cat findings.txt
+          echo "status=$status" >> "$GITHUB_OUTPUT"
+          if [ "$status" -gt 1 ]; then
+            echo "::error::dependency-register --upstream crashed (exit $status)"
+            exit "$status"
+          fi
+
+      # One issue, reused. A weekly job that opens a fresh issue every Monday
+      # trains people to close them unread, which is the failure mode this
+      # whole workflow exists to fix.
+      - name: File or update the tracking issue
+        if: steps.upstream.outputs.status != '0'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TITLE: 'chore(deps): fork & pin register needs attention'
+        run: |
+          set -euo pipefail
+          {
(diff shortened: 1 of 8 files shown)


skyl:
## Part 3 verified in real CI, not just locally

`zuuallet.yml` was dispatched on this branch — [run 33431743698](https://github.com/free2z/zuu/actions/runs/33431743698) — so the extended `upstream-canary` actually ran against `zcash/librustzcash@main`. **Job conclusion: success.** Both halves of the blind spot are visible in one log:

* **`Build against latest upstream`: success.** Zuuallet compiles fine against pristine upstream. That is the blind spot — this step is all the canary used to do.
* **`Assert the librustzcash fork is still required (ZUULI vs pristine upstream)`: success**, via the *expected-failure* branch:

```
error: failed to select a version for `crypto-common`.
    ... required by package `digest v0.11.2`
    ... which satisfies dependency `digest = "^0.11"` of package `curve25519-dalek v5.0.0`
    ... which satisfies dependency `curve25519-dalek = "^5.0.0"` of package `ed25519-dalek v3.0.0`
    ... which satisfies dependency `ed25519-dalek = "^3"` of package `f2z-kt-core v0.1.0`
    ... which satisfies path dependency `f2z-kt-core` (locked to 0.1.0) of package `zuuli v0.1.0`
versions that meet the requirements `^0.2` are: 0.2.2, 0.2.1

all possible versions conflict with previously selected packages

  previously selected package `crypto-common v0.2.0-rc.1`
    ... which satisfies dependency `crypto-common = "=0.2.0-rc.1"` of package `zcash_primitives v0.30.1`
    ... which satisfies path dependency `zcash_primitives` (locked to 0.30.1) of `tauri-plugin-zcash`
    ... which satisfies path dependency `tauri-plugin-zcash` (locked to 0.1.0) of `zuuli v0.1.0`

failed to select a version for `crypto-common` which could resolve this conflict
Expected: ZUULI cannot resolve against pristine upstream librustzcash.
This is the conflict free2z/librustzcash's one 4-line deletion removes,
so the fork is still required. Nothing to do.
```

**Answer to the question the task posed: it fails, for exactly the documented reason. The fork is still needed.** Because the step's only success path is "resolution failed AND the log names `crypto-common` AND `0.2.0-rc.1`", a green here *is* the assertion — it cannot go green by accident, and the two other outcomes (unexpected failure reason, or unexpected success) both exit 1.

The inverse control was run locally before the assertion was written: with the submodule back on the fork (`a2b8c952`), the same shell exits 1 with `::error::ZUULI now RESOLVES against pristine upstream librustzcash.` So both branches are exercised, not just the one CI happens to hit.

`dependency-register / offline` also passed on this PR with the submodule present — `27 mutants killed across both halves`, then `docs/DEPENDENCIES.md agrees with .gitmodules, 1 patch site(s), and 6 upstream-forced version holds`, which it can only say by having read the real `z/zcash/librustzcash/Cargo.toml`.

skyl:
## Both review points fixed in c6c55a0

**1. The `#2188` claim was wrong, and thank you for catching it.** I tested existence with `repos/openmls/openmls/pulls/2188`, which 404s **because 2188 is an issue, not a pull request** — exactly your diagnosis. Verified:

| ref | `/issues/{n}` state | has `pull_request` key |
|---|---|---|
| `openmls/openmls#2188` | closed, `state_reason: completed` | **no** — plain issue |
| `openmls/openmls#2163` | closed | **yes** — pull request |
| `zcash/librustzcash#3010` | open | yes, `merged_at: null` |

The note is deleted. In its place the register states the real relationship: **#2188 is the issue** (ours, closed as completed 2026-08-31 with "Fixed by #2163"), **#2163 is the PR** that fixed it, PR #850's title is therefore correct, and the exit condition is keyed to #2163 because *a release either contains that commit or does not, whereas a closed issue says nothing about what was published*. Both numbers are now required tokens in the checker, so the register cannot lose the half that makes the other unambiguous.

**The audit you asked for found the same defect live in the script, so this was not only a docs fix.** `upstreamFindings` resolved every tracked exit reference through `/pulls/{n}`. Register an issue number and the 404 throws → exit 2 → `dependency-register.yml` reports *"the script or the network broke"*. A wrong registration would have been indistinguishable from an api.github.com outage. Now:

* resolve `/issues/{n}` (resolves both shapes) and branch on the `pull_request` key;
* `merged` comes from `pull_request.merged_at`;
* `/pulls/{n}` is only called once we know there *is* a head SHA to ask about;
* an issue where a PR belongs is a **named finding** telling you to register the PR that closes it.

New self-test mutant for it, and its fake **throws on `/pulls/{n}`** so the test fails if that call is ever reached for a non-PR. **28 mutants, up from 27.** The fakes were re-checked against the live shapes in the table above rather than invented.

**2. `zwallet` tightened.** You're right that "not dormant" kept the recency and dropped the signal. Its last commit is `Update README to reflect Zcash deprecation (#274)` (2026-08-23); the one before is 2026-06-04, a release chore. The register now says it is *recently touched, but the touch was a deprecation notice* — the fact a reader needs when deciding whether to keep vendoring it — and notes that neither "dormant" nor "actively maintained" is the useful word. `warp` (2024-12-22) is unambiguous and unchanged.

Re-ran green: `--self-test` (28 mutants), offline verdict, and **live `--upstream` against real github.com**, which now exercises the `/issues/` path end to end.

skyl:

Re-verified independently:

- **The `#2188` correction is right, and the same defect was live in the checker.** `/issues/2188` resolves (closed, `state_reason: completed`, **no** `pull_request` key); `/pulls/2188` 404s because it is an issue. The register now states the real relationship — #2188 is the issue we filed, #2163 is the PR that fixed it, PR #850's title is correct — and keys the exit condition to **#2163** because a release either contains that commit or does not. `scripts/check-dependency-register.mjs` now resolves `/issues/{n}`, branches on the `pull_request` key, takes merge state from `pull_request.merged_at`, and only reaches `/pulls/{n}` once a head SHA is known. The new mutant whose fake **throws** on `/pulls/{n}` is the right way to pin it — it fails if that call is ever reached for a non-PR. 28 mutants, up from 27.
- **The `sha2` correction is the most valuable thing in this PR**, and it survives the #854 merge cleanly. `akd_core 0.13.0` genuinely has no `sha2` dependency — its digest is blake3, reaching `sha2 0.10.9` only transitively via `ed25519-dalek 2`. The real constraint, correctly identified, is that `f2z-msg-identity` links into `wallet/zuuli/src-tauri` where librustzcash's `sha2 = "0.10"` governs; moving to 0.11 would put a second SHA-256 in the shipped wallet. Right hold, wrong reason, now fixed.
- **The canary fix closes a blind spot that made a green meaningless.** Inverting the polarity — resolution failing on `crypto-common`/`block-buffer` is PASS, failing otherwise is FAIL, and **succeeding is FAIL loudly because the fork's premise is gone** — is better than the plain `cargo build` I asked for: it turns the fork's justification into something CI asserts every Monday instead of something a comment claims. Using `cargo metadata` rather than a build is the right call, since the conflict is a resolution error and costs seconds.
- **Scheduled issue over a required gate is the correct choice.** Every network verdict is a statement about a third party's repository; upstream merging #3010 is good news and must not redden `main` for a contributor who touched none of it.
- `gate` and `rs / gate` both pass; 0 conflicts against current `origin/main`; the `AGENTS.md` edit correctly wires the register into the existing branch-PR-pin-resume doctrine so step 5 finally has an enforcer.

One follow-up already assigned elsewhere, noted here so the trail is complete: the `tls_codec` comment merged in #854 cites "(#853)" for its byte-compatibility measurement, which is this PR's number rather than its own. Being corrected in the freshness pass; not this PR's defect.

PR #854, opened 2026-08-31
chore(deps): unify tls_codec on 0.5, closing the two-major split

skyl:
## The finding

`rs/Cargo.lock` and both wallet lockfiles carried **two majors of `tls_codec` at once**:

| version | who |
|---|---|
| `0.4.2` | every first-party crate — `f2z-authority`, `f2z-codec`, `f2z-kt`, `f2z-kt-core`, `f2z-msg-dag`, `f2z-msg-identity`, `f2z-relay`, `f2z-relay-proto`, `f2z-relay-testkit`, `f2z-witness`, and `tauri-plugin-f2zmsg` |
| `0.5.0` | `openmls 0.9.0`, `openmls_traits 0.6.0`, `openmls_libcrux_crypto 0.4.0`, `hpke-rs 0.7.0`, `libcrux-ecdh`, `libcrux-ml-kem` |

This PR unifies on `0.5.0`. The three questions were answered in order, and the answers are below.

## 1. Does any `tls_codec` type cross the boundary?

**No — and it was already deliberate.** The only `tls_codec` use in `f2z-msg-mls` is two traits imported as `use openmls::prelude::tls_codec::{Deserialize as _, Serialize as _}`, applied to `MlsMessageOut`/`MlsMessageIn` at `engine.rs:956,962`, plus `DeserializeBytes` on an OpenMLS `KeyPackage` in `keypackage.rs`. `tls_codec` is not a dependency of that crate at all. A `DeviceCredential` reaches MLS as the opaque identity bytes of a `BasicCredential`, encoded by `f2z-codec`'s canonical encoding — so no `tls_codec` *type* appears in an interface between the two halves.

So the split was **cosmetic in effect** — but survivable by accident, one `use tls_codec::…` away from a trait-not-implemented error whose cause is invisible. It also quietly falsified [`WIRE.md` §3.1](docs/e2ee/WIRE.md)'s own second reason for choosing this codec: *"it is already in the client's dependency graph… the crypto core already contains this decoder, already ships it to WASM"*. Two majors are two decoders in the WASM bundle and two in the audit scope, which is exactly the cost §3.2 refused to pay for CBOR.

## 2. Is the encoding byte-compatible? — **Yes, and it was measured, not read**

**Direct statement: the encoding is byte-identical between 0.4.2 and 0.5.0 for every construct this tree uses.**

Read from the actual crate sources side by side (both are vendored locally), not only the changelog:

- **`src/arrays.rs`, `src/primitives.rs`, `src/tls_vec.rs`** — every encode-side change is `writer.write(..)` → `writer.write_all(..)` (the `Serialize`/`Write` half, which this tree does not use) and the `SerializeBytes::tls_serialize` → `tls_serialize_bytes` rename. Emitted bytes are unchanged: integers still `to_be_bytes`, `[u8; N]` still `to_vec`, `TlsByteVecU*` still length prefix + bytes, `Option<T>` still tag + value.
- **`tls_codec_derive/src/lib.rs`** — the derived `SerializeBytes` still appends fields in declaration order and still writes the discriminant first for enums. The only additions are `#[cfg(not(target_pointer_width = "64"))]` overflow guards that trip solely at `usize::MAX`.
- Every other change in 0.5.0 is **decode-side and strictly narrowing**: element vectors must now consume their declared length exactly, a zero-length element is refused, `checked_add` replaces an overflowing sum, and read-based deserialization no longer pre-allocates from an untrusted length. None of these can widen what is accepted. (This tree does not use `TlsVecU*`/`VLBytes` at all — only `TlsByteVecU8/16/24` and its own `f2z_codec::vec::VecU*`.)
- MSRV 1.85, edition 2024, licence `Apache-2.0 OR MIT` — all fine against the pinned 1.97.1.

### The empirical proof, in the right order

Reading a changelog is not evidence, so:

1. **Under 0.4.2**, `f2z_codec::canonical::encode`/`decode_canonical` — the single chokepoint every crate in the tree encodes through — was temporarily instrumented to append `type_name::<T>()` + hex to a file, and the whole `rs/` suite was run. Result: **30,315 distinct (type, bytes) vectors over 81 types in seven crates** (`f2z-codec`, `f2z-kt-core`, `f2z-authority`, `f2z-msg-dag`, `f2z-kt`, `f2z-witness`, `f2z-relay-proto`).
2. **Then the bump**, and the same corpus was replayed through `decode_canonical` under **0.5.0** — which decodes, re-encodes and byte-compares.

```
REPLAYED   206 vectors for f2z_authority::
REPLAYED  7593 vectors for f2z_codec::
REPLAYED    66 vectors for f2z_kt::
REPLAYED   504 vectors for f2z_kt_core::
REPLAYED 21925 vectors for f2z_msg_dag::
REPLAYED    20 vectors for f2z_witness::
                = 30,314, all identical, zero failures
```

The one vector not replayed is `f2z-relay-proto`'s private `Empty`, whose encoding is zero bytes and cannot differ.

Generating the vectors *after* the bump would have proved only that 0.5 agrees with itself, which is why it was done this way round. The instrumentation and the replay harness were scratch-only and are **not** in this diff.

**The independent half:** `rs/crates/f2z-codec/tests/wire_vectors.rs` already exists — 98 tests whose expected bytes are derived **by hand from the `WIRE.md` spec text**, explicitly not from the implementation (#564). It passes unchanged.

## 3. What breaks at the source level?

One thing: `SerializeBytes::tls_serialize` → `tls_serialize_bytes`. Mechanically renamed at ~45 sites across 11 files. No `#[tls_codec(with = ...)]` anywhere, no use of the deprecated `Error::InvalidWriteLength`, no `Size` trait change, no derive-attribute change.

## Result

`rs/Cargo.lock` and both wallet lockfiles now resolve exactly one `tls_codec`; `0.4.2` and `tls_codec_derive 0.4.2` are gone with no new package added.

The reasoning, the measurement, and the obligation on whoever bumps this next are recorded **at the declaration** in `rs/Cargo.toml` — including the warning that `TlsSerializeBytes` and `TlsDeserializeBytes` move together, so a round-trip test stays green straight through a format change. The stale notes in `f2z-msg-mls`'s manifest and `engine.rs` are corrected; the indirection through `openmls::prelude` **stays**, because it must not become load-bearing on the two happening to resolve to one crate today.

## Verification

- `scripts/check-rust-fmt.sh --root rs` / `--root wallet`
- `scripts/check-rust-clippy.sh --root rs`; `cargo clippy --locked --all-targets -- -D warnings` in `tauri-plugin-f2zmsg`
- `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` and `--root wallet` — advisories, bans, licences, sources all ok, the `[[bans.deny]]` floors untouched
- `cargo test --locked --all-targets` in `rs/` — **1150 passed, 0 failed**; `cargo test --locked --doc`
- Every wasm32 build `rs.yml` runs: the five client crates, `f2z-msg-store --no-default-features`, `f2z-msg-mls --features js`, `f2z-kt-core`/`f2z-kt-client --features verifier`
- `cargo build --locked` of `wallet/plugins/tauri-plugin-f2zmsg` and `wallet/zuuli/src-tauri`

### DIFF
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -772,7 +772,7 @@ version = "0.1.0"
 dependencies = [
  "ed25519-dalek 3.0.0",
  "f2z-codec",
- "tls_codec 0.4.2",
+ "tls_codec",
 ]
 
 [[package]]
@@ -782,7 +782,7 @@ dependencies = [
  "blake2",
  "proptest",
  "syn 2.0.119",
- "tls_codec 0.4.2",
+ "tls_codec",
 ]
 
 [[package]]
@@ -800,7 +800,7 @@ dependencies = [
  "f2z-kt-core",
  "log",
  "protobuf",
- "tls_codec 0.4.2",
+ "tls_codec",
  "tokio",
  "tower",
 ]
@@ -834,7 +834,7 @@ dependencies = [
  "f2z-codec",
  "proptest",
  "protobuf",
- "tls_codec 0.4.2",
+ "tls_codec",
 ]
 
 [[package]]
@@ -848,7 +848,7 @@ dependencies = [
  "f2z-msg-mls",
  "f2z-msg-store",
  "proptest",
- "tls_codec 0.4.2",
+ "tls_codec",
 ]
 
 [[package]]
@@ -863,7 +863,7 @@ dependencies = [
  "proptest",
  "rand_core 0.10.1",
  "sha2 0.10.9",
- "tls_codec 0.4.2",
+ "tls_codec",
  "zeroize",
 ]
 
@@ -909,7 +909,7 @@ dependencies = [
  "futures-util",
  "rand 0.9.5",
  "serde",
- "tls_codec 0.4.2",
+ "tls_codec",
  "tokio",
  "tokio-rustls",
  "tokio-tungstenite",
@@ -924,7 +924,7 @@ dependencies = [
  "f2z-codec",
  "proptest",
  "subtle",
- "tls_codec 0.4.2",
+ "tls_codec",
 ]
 
 [[package]]
@@ -945,7 +945,7 @@ dependencies = [
  "f2z-codec",
  "f2z-relay-proto",
  "futures-util",
- "tls_codec 0.4.2",
+ "tls_codec",
  "tokio",
  "tokio-tungstenite",
 ]
@@ -961,7 +961,7 @@ dependencies = [
  "f2z-kt",
  "f2z-kt-core",
  "log",
- "tls_codec 0.4.2",
+ "tls_codec",
  "tokio",
  "ureq",
 ]
@@ -1288,7 +1288,7 @@ dependencies = [
  "log",
  "serde",
  "subtle",
- "tls_codec 0.5.0",
+ "tls_codec",
  "zeroize",
 ]
 
@@ -1726,7 +1726,7 @@ dependencies = [
  "libcrux-curve25519",
  "libcrux-p256",
  "rand 0.10.2",
- "tls_codec 0.5.0",
+ "tls_codec",
 ]
 
 [[package]]
@@ -1831,7 +1831,7 @@ dependencies = [
  "libcrux-sha3",
  "libcrux-traits",
  "rand 0.10.2",
- "tls_codec 0.5.0",
+ "tls_codec",
 ]
 
 [[package]]
@@ -2117,7 +2117,7 @@ dependencies = [
  "serde",
  "serde_bytes",
  "thiserror 2.0.20",
- "tls_codec 0.5.0",
+ "tls_codec",
  "web-time",
  "zeroize",
 ]
@@ -2135,7 +2135,7 @@ dependencies = [
  "p384",
  "rand_core 0.6.4",
  "serde",
- "tls_codec 0.5.0",
+ "tls_codec",
  "zeroize",
 ]
 
@@ -2158,7 +2158,7 @@ dependencies = [
  "openmls_memory_storage",
  "openmls_traits",
  "rand 0.10.2",
- "tls_codec 0.5.0",
+ "tls_codec",
 ]
 
 [[package]]
@@ -2199,7 +2199,7 @@ dependencies = [
  "rand_core 0.6.4",
  "sha2 0.11.0",
  "thiserror 2.0.20",
- "tls_codec 0.5.0",
+ "tls_codec",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


skyl:

Re-verified independently:

- **The unification is real and complete.** `tls_codec` resolves to a single **0.5.0** in `rs/Cargo.lock`, `wallet/plugins/tauri-plugin-f2zmsg/Cargo.lock` and `wallet/zuuli/src-tauri/Cargo.lock`; `tls_codec 0.4.2` and `tls_codec_derive 0.4.2` are gone from all three, and the only remaining `0.4.2` packages are unrelated crates (`constant_time_eq`, `ctutils`, `num-rational`, `windows-strings`). No package was added.
- **The proof method is the right one, and the ordering is what makes it evidence.** Capturing 30,315 (type, bytes) vectors through `f2z_codec::canonical`'s single chokepoint **under 0.4.2 first**, then replaying decode -> re-encode -> byte-compare under 0.5.0, is the honest direction; generating them after the bump would have proven nothing. 30,314 replayed identical, and the one omission (`f2z-relay-proto`'s private `Empty`) encodes to zero bytes. Instrumentation correctly kept out of the diff.
- **The independent half is the stronger half.** `f2z-codec/tests/wire_vectors.rs` predates this PR and derives its expected bytes from `WIRE.md`'s spec text rather than from the implementation (#564), so it cannot rubber-stamp an implementation change. It passes unchanged.
- **The narrowing argument checks out.** Every 0.5.0 decode-side change tightens (exact length consumption, zero-length elements refused, `checked_add`, no eager allocation from untrusted lengths); encode-side is `write` -> `write_all`, which this tree does not use.
- Diff scope is exactly right for a rename plus a version bump — 18 files, no unrelated edits. `gate` and `rs / gate` both pass; 0 conflicts against current `origin/main`.

Two judgement calls I want to endorse explicitly, because a less careful change would have got both wrong:

1. **Keeping the `openmls::prelude::tls_codec` indirection** rather than importing `tls_codec` directly in `f2z-msg-mls`. The two resolving to one crate today must not become load-bearing.
2. **Documenting that `TlsSerializeBytes`/`TlsDeserializeBytes` move together**, so a round-trip test stays green straight through a format change. That is precisely why a round-trip test was not sufficient evidence here, and it tells the next bumper they owe the same corpus measurement.

One number I could not reproduce: the report describes `wire_vectors.rs` as 98 tests; the file has 42 `#[test]` functions and no generating macros. Immaterial to the conclusion — the corpus replay is the substantive proof — but recording it rather than passing it through.

Also worth noting for the record: this closes a real gap in `WIRE.md` §3.1's own reasoning, which justifies the codec partly as 'already in the client's dependency graph'. Two majors meant two decoders in the WASM bundle and two in audit scope — the exact cost §3.2 declined to pay for CBOR.

PR #855, opened 2026-08-31
chore(deps): refresh two stale lockfiles, and collapse the rand 0.9 island

skyl:
## 1. Two stale lockfiles, refreshed

`wallet/plugins/tauri-plugin-zcash/Cargo.lock` and
`wallet/zuuallet/src-tauri/Cargo.lock` were months behind their siblings.
Refreshed with `cargo update` per manifest — never hand-edited:

| | before | after |
|---|---|---|
| `tauri` | 2.10.2 | **2.11.5** |
| `tonic` | 0.14.4 | **0.14.6** |
| `rustls` | 0.23.36 | **0.23.43** |
| `zeroize` | 1.8.2 | **1.9.0** |

### Why they drifted apart

Nothing regenerates a lockfile on a schedule — a lock moves only when a PR
touches the graph that owns it. Every dependency-moving change of the last six
weeks has been **messaging** work, and messaging links neither the Zcash plugin
nor the reference wallet:

* `wallet/zuuli/src-tauri` and `wallet/plugins/tauri-plugin-f2zmsg` — last moved
  by #854 (2026-08-31), before that #759 and #756.
* `wallet/plugins/tauri-plugin-zcash` and `wallet/zuuallet/src-tauri` — last
  moved by **#750 (2026-08-25)**, and nothing since.

So this is not neglect of a rule; it is the absence of one. The four locks
converge only when a change happens to touch all four, and for six weeks none
did.

### `orchard` is deliberately held at 0.15.3

`cargo update` moved it to 0.15.5 and `scripts/check-librustzcash-compat.mjs`
correctly went red: `orchard` is one of its 11 reviewed packages across 3
shipping locks. Bumping that inventory is its own procedure (regenerate all
three locks, re-derive `--print-scope-digest`, copy it into two checkers) and
does not belong in a lockfile-freshness PR. Pinned back with
`cargo update -p orchard --precise 0.15.3` in both locks; the compat checker
passes. **`orchard` was the only one of the 11 that moved.**

### The regenerated Tauri schemas are not optional

`zuuallet.yml` regenerates `wallet/zuuallet/src-tauri/gen/schemas` and runs
`git diff --exit-code` over the whole directory, and tauri 2.11 adds a
`supports-multiple-windows` permission. Tauri's build script only refreshes the
schema for the *host* target family, so `linux-schema.json` was produced in a
clean `ubuntu:24.04` container on the pinned 1.97.1 toolchain — the file Linux CI
owns — and `desktop-schema.json` / `macOS-schema.json` / `acl-manifests.json`
locally.

---

## 2. Version drift — two bumps, seven holds

Every row was measured on the real graph, not inferred. "Δ" is the change in
duplicate crate copies.

| crate | from | to | Δ | measured reason |
|---|---|---|---|---|
| **`rand`** (`tauri-plugin-f2zmsg`) | 0.9 | **0.10** | **−3 in two locks** | Not on the original list; found by measurement and the biggest win here. `rand 0.9.5` had **exactly one** consumer in both the f2zmsg lock and `wallet/zuuli/src-tauri`'s — this crate — and the whole island hung off that single edge. Bumping it deletes `rand 0.9.5`, `rand_chacha 0.9.0` and `rand_core 0.9.5` outright: excess copies **75 → 72** in the f2zmsg lock and **105 → 102** in `wallet/zuuli/src-tauri`'s, where `rand`, `rand_chacha` and `rand_core` each drop from *three* resolved versions to two (the remaining 0.8/0.3/0.6 line is librustzcash's documented hold). Two-line source change (`RngCore` → `Rng`; `rand::rng()` unchanged). |
| **`rand`** (`f2z-relay`, `rs/`) | 0.9 | **0.10** | **−3 shipped, 0 in the lock** | The island does **not** collapse in the lockfile, and the brief's premise that it would is wrong: `proptest 1.11.0` (dev-only) pulls `rand 0.9` + `rand_chacha 0.9` + `rand_xorshift` → `rand_core 0.9` → `getrandom 0.3`, so `rs/Cargo.lock` stays at 39 excess copies. What *does* change is what ships: `cargo tree --edges normal` now finds no `rand 0.9`, `rand_core 0.9` **or `getrandom 0.3`** anywhere in the relay binary, where before this crate was their sole non-dev root. |
| `ripemd` | 0.1 | **held** | 0 | **The brief's premise is inverted.** `zcash_script 0.4.5` declares `ripemd = "0.1"`, and librustzcash's `[workspace.dependencies]` declares `ripemd = { version = "0.1" }` — which is what `zcash_transparent` and `pczt` inherit. We are on 0.1.3 *with the majority*: 3 of the 4 consumers. The lone `ripemd 0.2.0` consumer is `bip32 0.6.0-pre.1`, the `=`-pinned RustCrypto pre-release island. Bumping cannot remove 0.1.3 (two other consumers hold it), so Δ is 0 — and it would move our HASH160 onto `digest 0.11` while the `Sha256` composed with it *in the same function* stays on `digest 0.10`. |
| `chacha20poly1305` | 0.10 | **held** | **+1 in all four wallet locks** | `zcash_note_encryption 0.4.2` declares `chacha20poly1305 = "0.10"`; `hpke-rs-rust-crypto` and `openmls_rust_crypto` pin 0.10 as well. Ours is currently the *only* copy — bumping makes it the second. |
| `base64` | 0.22 | **held** | **+1 in two locks** | librustzcash pins `base64 = "0.22"` and 11 crates resolve 0.22.1 (`zcash_client_backend`, `zip321`, `tonic`, `wry`, `reqwest`, …). `tauri-plugin-zcash` and `zuuallet` carry no 0.23 at all, so bumping ours takes them from two base64s to three. |
| `nonempty` | 0.11 | **held** | **+1, and it would not compile** | librustzcash pins `nonempty = "0.11"` and 8 of its crates use it — including the `NonEmpty<Payment>` handed to `propose_transfer`. Two majors would not merely duplicate, they would stop type-checking. |
| `getrandom` (zuuli) | 0.3 | **held** | 0 | `tauri 2.11.5` itself declares `getrandom = "0.3"`, so 0.3.4 stays regardless. All three of 0.2.17 / 0.3.4 / 0.4.3 remain either way. |
| `bip0039` | 0.12 | **held** | 0 | A singleton — `tauri-plugin-zcash` is the only consumer, since librustzcash's use sits behind the `zcashd-compat` feature we do not enable. Zero graph benefit, and it is `mnemonic_to_seed`, the root of every wallet's key material, where a byte-level change is unrecoverable funds. No advisory forces it. |
| `blake2` (`rs/`) | 0.10 | **held** | 0 | A singleton, and `digest 0.10.7` stays regardless because `sha2 0.10` and `hkdf 0.12` are held. `f2z-msg-identity/src/blake.rs` deliberately drives `Blake2bVarCore` through `digest::core_api` (`VariableOutputCore`, `Buffer`, `UpdateCore`) *because* the high-level `Blake2bMac512` was silently wrong — and `digest 0.11` reshapes exactly that API. Re-porting the most delicate crypto in the tree, against outputs pinned by `tests/derivation_vectors.rs`, to remove zero duplicates. |
| `keyring` | 3.6.3 | **held** | 0 | A singleton. A major bump of the crate holding wallet seeds in the macOS/iOS Keychain, Linux keyutils + secret-service and Windows Credential Manager needs per-platform testing a lockfile PR cannot honestly provide. Recommend a dedicated PR. |

---

## 3. One citation fix

`rs/Cargo.toml`'s `tls_codec` note credited the byte-compatibility measurement to
#853. Verified with `gh pr view`: **#854** is "unify tls_codec on 0.5" — where the
30,315-vector replay was done — and #853 is the unrelated dependency-register PR.

---

## Verification

* `cargo build --locked` green for all four wallet roots and `rs/`.
* `cargo test` green: `tauri-plugin-zcash` 151 (`--all-targets --features
  production-route-probe`), `tauri-plugin-f2zmsg`, `zuuli/src-tauri`, and the
  full `rs/` workspace including doctests.
* All five wasm32 targets from `rs.yml` build.
* `scripts/check-rust-{fmt,clippy,deny,toolchain}.sh` clean on both trees;
  `check-librustzcash-compat.mjs` and `check-dependency-register.mjs` pass.
* `rusqlite` is still the repo-wide singleton at 0.37.0 / `libsqlite3-sys` 0.35.0
  in every lock, verified for `x86_64-unknown-linux-gnu`.
* Linux verified in a clean `ubuntu:24.04` container on pinned 1.97.1, not just
  on the macOS dev host.


skyl:

The most valuable thing in this PR is the seven bumps it **declined**, each with a measured duplicate-count delta rather than a version comparison. That is the right shape for a freshness pass: fewer duplicate crates and fewer unexamined versions, not higher numbers.

Re-verified independently:

- **The `ripemd` rejection is correct and my brief had the premise inverted.** `wallet/zuuli/src-tauri/Cargo.lock` shows `ripemd 0.1.3` consumed by `tauri-plugin-zcash`, `zcash_script 0.4.5` **and** `zcash_transparent 0.10.0`, while `ripemd 0.2.0` has exactly one consumer: `bip32 0.6.0-pre.1`, the `=`-pinned pre-release island. `z/zcash/librustzcash/Cargo.toml:85` declares `ripemd = { version = "0.1" }`. We were already with the 3-of-4 majority, and bumping would have stranded HASH160's RIPEMD on `digest 0.11` beside a `digest 0.10` `Sha256` in the same function. Good catch; the brief was wrong.
- **The other upstream-forced rejections check out** against librustzcash's : `nonempty = "0.11"` (line 94), `base64 = "0.22"` (113), `bip0039 = "0.12"` (192); and `chacha20poly1305 = "0.10"` in `zcash_note_encryption/Cargo.toml:24`.
- **`orchard` correctly held at 0.15.3** in both wallet locks. `cargo update` moved it to 0.15.5, `check-librustzcash-compat.mjs` went red, and the right response was to respect the gate rather than re-baseline the reviewed inventory. That restraint is worth more than the bump would have been.
- **The `rand` bumps are the real win**, and the `tauri-plugin-f2zmsg` one was found by measurement rather than handed over in the brief. Both declarations are at 0.10 on the branch.
- Lockfile refresh lands tauri **2.11.5** in both previously-stale roots. The regenerated `gen/schemas/*.json` are a genuine consequence of tauri 2.11's new `supports-multiple-windows` permission, and `Zuuallet / generated schema freshness` passing confirms the container output is byte-identical to CI's.
- The `(#853)` -> `(#854)` citation fix is in, and was verified against the PRs rather than taken on my word.
- `gate` and `rs / gate` both pass; zero failing checks; 0 conflicts against `origin/main`.

Also correct to flag rather than fix: `docs/DEPENDENCIES.md` §3's `rand` row goes stale the moment this merges, and the holds table is missing the upstream-forced entries this PR measured. Staying out of a file another PR owned was the right call — I am handling the reconciliation separately.

PR #856, opened 2026-08-31
docs(deps): reconcile the hold register with #855, and give every row a reader

skyl:
Follows #853 (the register) and #855 (the graph measurement). Documentation and
check-script work only — **no dependency version, lockfile, or manifest
requirement is changed** in this PR.

## 1. §3's `rand` row was factually wrong

It said the shipping wallet lock "legitimately carries 0.8, 0.9 and 0.10".
After #855 collapsed the `rand 0.9` island, `wallet/zuuli/src-tauri/Cargo.lock`
carries **0.8.7 and 0.10.2 only** (`rand_core` 0.6.4 and 0.10.1). Verified at
lines 5319/5330 and 5361/5370 of that lock. Corrected.

The check passed anyway, because it validates holds against librustzcash's
manifest — the right source for *who forces a hold*, and no source at all for
*what the graph contains*. That is what §3 below fixes.

## 2. A second row had rotted the same way, and nobody had named it

The `sha2` correction still ended "Moving `f2z-msg-identity` to 0.11 would add
a second SHA-256 to the shipped wallet." It would not. The wallet lock already
carries **0.10.9 and 0.11.0** — the MLS half of the graph brought 0.11 in
through `bip32`, `ed25519-dalek`, `hpke-rs-rust-crypto` and
`openmls_rust_crypto`. **14** packages resolve `sha2` on 0.10, **4** on 0.11.
The cost of moving is *defection* from the copy every Zcash crate in the binary
shares, not a new copy. Corrected, and both counts are now re-read.

## 3. Six holds added, each re-verified with line numbers

| Crate | Held | Constraint source | Verified |
|---|---|---|---|
| `ripemd` | 0.1 | `z/zcash/librustzcash/Cargo.toml:85` | ✅ |
| `nonempty` | 0.11 | `z/zcash/librustzcash/Cargo.toml:94` | ✅ |
| `base64` | 0.22 | `z/zcash/librustzcash/Cargo.toml:113` | ✅ |
| `bip0039` | 0.12 | `z/zcash/librustzcash/Cargo.toml:192` | ✅ |
| `chacha20poly1305` | 0.10 | `z/zcash/zcash_note_encryption/Cargo.toml:24`, plain `[dependencies]` — **not** librustzcash | ✅ |
| `getrandom` | 0.3 | `tauri 2.11.5` | ✅ **confirmed** |

**`getrandom` is confirmed forced by tauri.** `tauri-2.11.5`'s published
manifest declares `[dependencies.getrandom] version = "0.3"` — unconditional,
not optional, not target-gated — and `wallet/zuuli/src-tauri/Cargo.lock:6825`
records the resolved edge `tauri 2.11.5 → getrandom 0.3.4`. So 0.3 is in the
shipping graph whatever we choose. Our single call site is `oauth.rs:179`'s
PKCE randomness; bumping to 0.4 would not retire tauri's copy (the lock already
carries 0.2.17, 0.3.4 and 0.4.3) — it would only move us off the copy the
framework links.

Measured consequences, each re-read from a lockfile:

* **`chacha20poly1305`** — every wallet lock carries exactly one copy (0.10.1),
  shared by 5 packages: ours twice plus `zcash_note_encryption`,
  `hpke-rs-rust-crypto` and `openmls_rust_crypto`. Bumping ours to 0.11 strands
  the other three and adds a second AEAD to all four wallet locks.
* **`ripemd`** — bumping adds no copy (`bip32` already carries 0.2.0), but 0.2
  is on `digest 0.11`, so `hash160`'s `Ripemd160` and its `Sha256` would sit on
  two different `digest` traits in the same six-line function.
* **`nonempty`** — crosses an API boundary rather than merely duplicating:
  `send/native.rs:182` returns `nonempty::NonEmpty<TxId>` out of a librustzcash
  call, so a second copy is a type mismatch.
* **`bip0039`** — deliberately labelled the weakest row: upstream declares it
  only `optional`, so today exactly 1 package resolves it, and it is ours.

## 4. The rot risk, which is the real deliverable

A hold justified by a source the checker does not read is exactly the claim that
rotted before, wearing a checked-looking badge. Three changes close it:

**Every hold names its own constraint source, and the checker reads whichever is
named.** `source.kind` is `manifest` (a named dependency table in a named TOML
file), `lockfile` (a resolved graph, where the forcing crate is a registry
dependency with no manifest in this tree), or `ours` (the negative claim,
against *every* registered source rather than just librustzcash's).

**The table and the registry must agree row for row, in both directions.** The
forward direction stops a hold being dropped while the version is still pinned.
The reverse stops a row being *added* to the page with no source anything
re-reads — and each row must name, in its own text, the source that verifies it.

**Every number the page states is re-read, and the sentence must carry it.**
32 `evidence` entries across the four wallet locks: version sets, resolved
edges, consumer counts. Verifying the lock while letting the prose say something
else is precisely how the `rand` row stayed wrong through a green check, so
`claimText` derives the required phrase from the measurement — the two cannot
drift.

`getrandom` is labelled as the weaker kind it is: a lockfile is a resolution,
not a requirement, and tauri has no manifest here to re-read. Pinning the check
to **2.11.5** is what makes that visible — the next tauri bump turns the row red,
which is not the check crying wolf but the check saying nobody has re-read
tauri's requirement since.

## Verification

* `--self-test`: **50 mutants killed** (41 offline + 9 upstream), up from 28.
  The 22 new ones cover the non-librustzcash manifest source, the lockfile
  source, all three evidence kinds, the row-for-row agreement in both
  directions, and prose drifting from a verified measurement — including a
  mutant that reproduces the exact `rand` bug this PR fixes.
* Offline verdict: green — 13 holds and 32 measured consequences across 6 files.
* `--upstream` against real github.com: green.
* `check-workflow-gates.mjs`, `check-rust-fmt.sh`, `check-rust-toolchain.sh`,
  `check-rust-deny.sh`: all green.

`dependency-register.yml` now inits `z/zcash/zcash_note_encryption` alongside
librustzcash and triggers on `**/Cargo.lock`.


skyl:

Re-verified independently:

- **The second rotted claim is real, and catching it is the most valuable thing here.** I confirmed it against the shipped lock directly: `wallet/zuuli/src-tauri/Cargo.lock` already carries **`sha2 0.10.9` (14 consumers: `bip0039`, `bs58`, `dbus-secret-service`, `ed25519-dalek`, `f2z-msg-identity`, `p256`, `p384`, `secret-service`, …) and `sha2 0.11.0` (4: `bip32`, `ed25519-dalek`, `hpke-rs-rust-crypto`, `openmls_rust_crypto`)**. So #853's "would add a second SHA-256 to the shipped wallet" was false — there are already two — and the real cost of moving `f2z-msg-identity` is defection from the 14-consumer shared copy to the 4-consumer one. A correction to a correction, found because the new checker re-reads the number instead of trusting the sentence. That is the mechanism working exactly as intended, one PR after it was built.
- **The `getrandom` row is properly resolved and honestly weakened.** I flagged it as unverified; you verified it (`tauri-2.11.5` manifest, unconditional `getrandom = "0.3"` above the first `[target…]` block, so not platform-gated) *and* then correctly noted the row is weaker than #855 implied — the lock already carries 0.2.17, 0.3.4 and 0.4.3, so bumping ours moves us off the framework's copy rather than adding one. Stating that plainly is better than banking the win.
- **Task 3 is answered at the right altitude.** Per-row `source.kind` (`manifest` / `lockfile` / `ours`) means a hold justified by `zcash_note_encryption` or a registry crate is machine-checked by whatever actually constrains it, not by a manifest that happens to be vendored. The bidirectional table-registry agreement is the part I would have missed: it makes it impossible to *add* a prose row that nothing re-reads, which is the failure mode both rotted claims came from.
- **Deriving `claimText` from the measurement** is the right fix for the specific way the `rand` row stayed wrong — verifying a lock while letting the prose say something else is not verification.
- Ran the branch's checker against the real tree myself: **50 mutants killed** (up from 28), offline verdict clean and now naming every source it reads. `gate`, `rs / gate` and `dependency-register / offline` all pass; 0 conflicts; diff confined to 3 files with no dependency, lockfile or manifest change, as instructed.

Agreed on the item you flagged rather than fixed: `sha2 0.11.0` reaching the shipping wallet through the MLS crates is a real graph fact that predates this work and was never written down. Documenting it and leaving the decision alone was the correct call — whether that second SHA-256 is acceptable is a dependency question for the owner, not a docs change.

PR #857, opened 2026-09-01
chore(zuuli): publish store-capture toolchain image (Phase A, #387)

skyl:
## Summary

Publishes a digest-pinned, immutable Docker image combining Playwright 1.62.1's Chromium with the exact Rust 1.97.1 `wasm32-unknown-unknown` toolchain ZUULI's production WASM build requires, plus the CI wiring and source-policy validator (`scripts/check-zuuli-store-capture-image.mjs`) that keep the Dockerfile/workflow honest.

This directly answers the blocker recorded in the #387 comment thread: the current pinned Playwright worker only mounts `wallet/zuuli` and has no Cargo/Rust/rustup, so it cannot run the mandatory WASM build during a deterministic store-screenshot capture. That comment explicitly calls for "a separately reviewed digest-pinned Playwright + Rust 1.97.1/wasm32 capture image plus wallet-root mount" as the next tooling prerequisite before the real two-pass 20-image capture.

This is Phase A only: it builds and inspects the image on PRs, and publishes to GHCR (with SBOM/provenance/attestation) only on `main`/schedule/dispatch. It does **not** wire the new image into the actual capture path yet — `docs/ZUULI-STORE-CAPTURE-IMAGE.md` spells out that a follow-up Phase B PR must review the published digest, pin it into the capture config, add the wallet-root mount, and then run the real two-pass capture. No store publication and no screenshot replacement happens here.

## Judgment call

This is a rebase-and-verify salvage of a stranded branch (`chore/387-store-capture-image`, 1 commit) that was never opened as a PR. `main` has moved substantially since (i18n kernel #797, build 19, etc.), but none of that touched this area — the rebase onto current `origin/main` produced exactly one conflict, in `.github/workflows/zuuli.yml`'s file-classification `case` pattern (both this branch and a merged main commit added entries to the same glob list). I merged the two pattern lists so all entries from both sides are preserved: the store-capture paths this PR adds (`scripts/check-zuuli-store-capture-image.mjs`, `.github/containers/zuuli-store-capture/*`, `.github/workflows/zuuli-store-capture-image.yml`, `docs/ZUULI-STORE-CAPTURE-IMAGE.md`) plus the `wallet/zuuli/*.md`-skip guard and `docs/e2ee/*` entries that landed on main afterward.

I confirmed via `gh issue view 387` and its comment history that this is a live, correctly-sequenced next step toward the issue (not a dead end) — the issue's own directory schema/validators/metadata/locales/screenshots/read-only-audit/publish-gate acceptance criteria are already substantially implemented on `main` (#417, #420, #425, #444, #536, #707); the only missing piece is exactly the deterministic-capture toolchain this PR adds.

## Verified

From `wallet/zuuli` after `npm ci` (not `npm install`), rebased onto current `origin/main`:

- `node scripts/check-zuuli-store-capture-image.mjs --self-test` and (no flag) — pass (28 mutants killed)
- `node scripts/check-workflow-gates.mjs --self-test` and (no flag) — pass (39 self-test cases; production check passes: 2 gates, 15 registered-ungated, 66 jobs, 17 workflow files)
- `node scripts/check-github-actions-pins.mjs` — pass (204 external refs, 18 files, including this PR's 5 new pinned actions)
- `npm run typecheck` — clean
- `npx vitest run` — 96 test files / 1408 passed, 5 skipped
- `npx playwright test` — 180 passed, 2 failed (`tests/navigation.pw.ts` "five-item navigation works at 320px/360px" tap-size measurement). Re-ran both failures in isolation with a fresh dev server: both pass in ~2s each. This branch touches zero UI/component source (CI/Docker/docs/scripts only), so the failure is pre-existing full-suite flakiness under parallel-worker CPU contention, not a regression from this change.

No merge performed; this is ready for the QA gate + adversarial review per repo doctrine.


skyl:
Update: CI failed on \`check-zuuli-android-gate.mjs\`'s pinned change-detector digest — this PR edits the \`case "$file" in ...\` pattern list inside \`.github/workflows/zuuli.yml\`'s "Detect release-impacting ZUULI changes" step (adding the store-capture paths this PR introduces to the same \`zuuli=true\` arm), and that script digest-pins the entire step so an edit there requires a reviewed re-derivation.

Confirmed the edit is additive only — it does not touch the fail-open/fail-closed logic (invalid/missing base SHA, diff failure) above the \`case\` block, it only adds new file globs to the existing \`zuuli=true\` arm so this PR's own new files correctly trigger the gate. Re-derived the digest with the script's own \`--print-change-detector-digest\` flag (per its inline guidance) and updated the constant — did not edit or weaken the checker itself.

Re-verified before push:
- \`node scripts/check-zuuli-android-gate.mjs --self-test && node scripts/check-zuuli-android-gate.mjs\` — pass (44 mutations)
- \`node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs\` — pass (39 self-test cases; 2 gates/15 registered-ungated/66 jobs/17 files)
- \`node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs\` — pass (340 mutation cases; 204 external refs/18 files)
- \`node wallet/zuuli/scripts/verify-ci-cache-policy.mjs\` — pass
- \`npm run typecheck\` — clean

Pushed once with \`--force-with-lease\`.

Issue #691, opened 2026-08-25
wallet: apps reach into each other and into shared code by counting ../, with no module boundary and no guard


skyl:
Raised from #690, where `import { RestoreWallet } from "../../../../zuuallet/src/pages/RestoreWallet"` produced **five of that PR's six CI failures**. The import is the symptom; the absence of a module boundary between the two wallet apps is the cause. Swept `wallet/` and `ts/` for the whole family.

## Why it is not a style question

`wallet/zuuli/src/lib/wallet/zuuallet-sensitive-entry-routes.test.tsx:56` is a **static** import, so zuuli's TypeScript program follows it into zuuallet's source and typechecks that source under **zuuli's** `tsconfig` and **zuuli's** `node_modules`. `zustand`'s generics do not resolve there, so on `e21f4fe5`:

```
../zuuallet/src/store/wallet.ts(45,52): error TS7006: Parameter 'set' implicitly has an 'any' type.
…13 more TS7006, 6 more TS18046…
```

`zuuli / frontend` fails at **Typecheck** and all four `ZUULI / packaging smoke` jobs fail at **Build packages**, on one line in one test file.

**The five `vi.mock` calls above it do not help**, and that is the part worth internalising. That file already mocks `../../../../zuuallet/src/store/wallet`, `.../sensitive-entry`, `.../useWallet`, `.../tauri` and `.../sensitive-seed`. Those are runtime substitutions; `tsc` still walks the static graph through `RestoreWallet` into the real modules. A test can be fully mocked at runtime and still drag another application's entire dependency graph into this project's build.

## The sweep

**1. Cross-project source imports — the damaging kind.** One reaches an *application page*:

| file | reaches |
|---|---|
| `wallet/zuuli/src/lib/wallet/zuuallet-sensitive-entry-routes.test.tsx:56` | `../../../../zuuallet/src/pages/RestoreWallet` |
| `wallet/zuuli/src/lib/wallet/zuuallet-created-seed.test.ts:5` | `../../../../zuuallet/src/lib/sensitive-seed-session` |

The second is **already on `main`** and has not bitten only because it lands on a leaf module with no store, hooks or third-party generics behind it. It is the same crossing, one import away from the same failure.

**2. Shared code reached by counting dots.** `wallet/shared/sensitive-entry-session.ts` is a genuine shared surface, and every consumer reaches it by relative path — at a *different depth from each side*, from **production** files, not only tests:

```
[...]

PR #862, opened 2026-09-01
fix(wallet): enforce project boundaries between zuuli, zuuallet, and shared

skyl:
## Summary

Salvages a 12-commit branch that had been stranded since 2026-08-29 (never opened as a PR) and rebases its intent onto current `main`, which had moved 42 commits since this branch's base (i18n kernel restructuring under `wallet/zuuli/src`, media preflight, About/build-identity, search changes, build 19).

Delivers what #691 asked for:

- `wallet/shared/` becomes a real package (`@free2z/wallet-shared`, `wallet/package.json` workspace root) instead of a path both apps reach by counting `../`.
- `wallet/zuuli/scripts/project-boundary.mjs` — a static + build-time scanner that walks both apps' source graphs (imports, path aliases, tsconfig `rootDirs`/`include`, Vite aliases/plugins/Rollup input, even a real constrained Vite build to catch a plugin that reads outside its project at runtime) and fails on any crossing into a sibling app's source, with the shared package as the one permitted crossing.
- `wallet/zuuli/scripts/project-boundary.node-test.mjs` — an 85-case self-test with fixtures that inject a real crossing, per the repo's "a check nobody has watched fail is not a check" convention.
- Wired into CI as a required step (`Verify wallet project boundaries` in `.github/workflows/zuuli.yml`), and `scripts/check-zuuli-android-gate.mjs` was extended to assert the workflow still selects the ZUULI gate for wallet-boundary and shared-package changes.
- The two damaging crossings named in #691 (`zuuallet-sensitive-entry-routes.test.tsx` reaching into `zuuallet/src/pages/RestoreWallet`, and `zuuallet-created-seed.test.ts` reaching `sensitive-seed-session`) are gone; both apps now import the shared session module by package name.

## Assessment: rebased, not reimplemented

This was case (a): the work was still valuable and cleanly rebaseable. Only 3 files actually conflicted against 42 intervening `main` commits — `.github/workflows/zuuli.yml`, `scripts/check-zuuli-android-gate.mjs`, and `wallet/zuuli/package-lock.json` — because this branch and `main` mostly touched disjoint files (the i18n/media/About/search work never touched `wallet/shared/*` or the two production files this branch also edited, `wallet/zuuli/src/lib/wallet/bridge.ts` and `mock.ts`, both of which auto-merged cleanly).

Conflict resolution (merge commit `ab4aabf`, done by hand, not blind pick):
- `zuuli.yml`: merged this branch's broadened `wallet/zuuallet/*` + `wallet/shared/*` + `wallet/package.json` selectors with main's independently-added `docs/e2ee/CLIENT-CONTRACT.md`/`WIRE.md` selector.
- `check-zuuli-android-gate.mjs`: kept this branch's `requiredWalletBoundarySelectors` (package-glob checks, which supersede the old per-file `classicSeedBoundaryInputs` list now that `wallet/shared/sensitive-entry-session.ts` moved to `wallet/shared/src/...`) alongside main's independent `messagingContractInputs` check; recomputed the workflow-detector self-check digest (`--print-change-detector-digest`) for the merged file.
- `package-lock.json`: regenerated via `npm install --package-lock-only` rather than hand-splicing JSON.
- Also fixed a **pre-existing** bug this exposed: `wallet/zuuallet/package-lock.json` still recorded `@free2z/wallet-shared` as a plain `file:` dependency instead of a workspace link, which made `npm ci` fail post-merge (`Missing: @free2z/wallet-shared@0.1.0 from lock file`). Regenerated it to match the link-style entry already present in `wallet/zuuli`'s lockfile (separate commit `809cd87`).

## Verified (from `wallet/zuuli`, Node 24 to match CI)

- `npm ci` in `wallet/zuuli`, `wallet/zuuallet`, and `wallet/` (root workspace) — all clean.
- `npm run typecheck` and `npm run typecheck:tests` — clean. (This is the exact failure mode #691 documents: zuuallet's store leaking into zuuli's `tsc` program. Confirmed fixed.)
- `npx vitest run` — 95 files / 1396 tests / 5 skipped, all pass.
- `node scripts/ui-copy-truncation.node-test.mjs` — 4/4 pass.
- `npx playwright test` — **182/182 pass** on a clean run. (An earlier run under contention from unrelated concurrent Playwright processes on the same machine showed 4 flaky failures; isolated reruns of each proved they were resource contention, not regressions. A separate run also intermittently showed `about.pw.ts`'s build-identity test failing with "Source commit: Unavailable in this build" — traced to my own `npm run build` in `zuuallet` leaving a dirty `dist/` checkout, which `build-identity.mjs` correctly treats as unavailable by design (`never presents a dirty checkout as the exact source commit`, confirmed in its own self-test). Not a bug; passes on a clean tree.)
- `node scripts/project-boundary.mjs` (the live scanner, not just its self-test) — passes against the real merged repo: "3 discovered projects, 324 source files, 1629 parsed module references, 5 production shared-package consumers, 2 constrained production Vite builds."
- `node --test scripts/project-boundary.node-test.mjs` — **81/85 pass.** See Risk below.
- `node --test scripts/seed-capture-boundary.node-test.mjs` + `node scripts/seed-capture-boundary.mjs`, `npm run test:seed-capture`, `node --test scripts/status-freshness.node-test.mjs`, `node --test scripts/wasm-boundary.node-test.mjs scripts/messaging-contract.node-test.mjs` — all pass.
- `node scripts/check-github-actions-pins.mjs` and `node scripts/check-zuuli-android-gate.mjs` (both self-test and live) — pass.
- Packaging path: `npm run build` in both `wallet/zuuli` (feeds `src-tauri`, includes the WASM dist-verify step) and `wallet/zuuallet` complete successfully end to end. This branch touches zero Rust/`src-tauri` files — confirmed via diff against its merge-base — so the only packaging exposure is the Vite build sandbox itself, and both production builds prove it doesn't break the build output.

## Risk being flagged

`scripts/project-boundary.node-test.mjs` has **4 pre-existing failing sub-tests** (confirmed present on the original unmerged branch tip too, not introduced by this merge/rebase):
- `rejects a temp output symlink redirected into owner source`
- `rejects a Vite transform that reads sibling-app source after a real build proves injection`
- `production boundary rejects a Vite load hook that reads sibling-app source`
- `rejects a CSS import of sibling-app source after a real build proves bundling`

In every case I inspected, **the actual boundary enforcement still fails closed** — the constrained Vite build correctly aborts with "constrained production Vite graph build failed; a plugin or module may have read outside wallet/zuuli and wallet/shared" in all 4 cases. The failures are the test's regex additionally expecting the specific offending path (e.g. `zuuallet/src/index.css`, `transform-secret.ts`) to appear in the error text; instead Node's `--permission` model intercepts the read at the OS layer before the custom `auditedPath` wrapper in `project-boundary-vite-build.mjs` gets to format its own message with that path, so the surfaced text is Vite/Rollup/postcss's generic "Access to this API has been restricted" wording. This looks like drift in Node's experimental permission-model error text between whatever Node 24.x the author tested against and this environment's 24.19.0 — not a hole in the guard itself.

This step is wired into CI as a required gate (`Verify wallet project boundaries` in `zuuli.yml`, `node --test scripts/project-boundary.node-test.mjs` first), so **I expect this PR's own CI run to go red on that step** until those 4 assertions are relaxed to match current Node's actual permission-denial wording (or the wrapper is changed to reliably re-attach the attempted path via `error.resource`, which `formatPermissionError` already has a code path for). I did not attempt that fix myself — it touches unfamiliar Node permission-model/Vite-internals interaction and didn't feel safe to guess at under this task's scope. Flagging for review/follow-up rather than merging past it.

## Not merging

Per instructions, this PR is left for review — CI gate + adversarial review should run, and the risk above should be resolved (or explicitly accepted) before squash-merge.


skyl:
## Fixed the 4 pre-existing `project-boundary.node-test.mjs` failures

Reproduced locally on Node **24.18.0** (matching `ZUULI_NODE_VERSION` in CI) in an isolated worktree. Confirmed these 4 sub-tests failed on the branch tip *before* this fix, and that they are message-text failures, not enforcement failures:

- `rejects a temp output symlink redirected into owner source`
- `rejects a Vite transform that reads sibling-app source after a real build proves injection`
- `production boundary rejects a Vite load hook that reads sibling-app source`
- `rejects a CSS import of sibling-app source after a real build proves bundling`

### What was actually failing

All 4 assertions used `assert.rejects(fn, /constrained production Vite graph build failed[\s\S]*<something node-version-specific>/)`, where `<something>` was either a literal Node permission-model token (`FileSystemWrite`/`allow-fs-write`) or the specific leaked filename (e.g. `transform-secret.ts`).

Under Node 24.18.0:
- The symlink-write case now throws `fs.symlink API requires full fs.read and fs.write permissions` (no `FileSystemWrite`/`allow-fs-write` substring).
- The transform/load/CSS cases now sometimes trip Node's OS-level `--permission` check on an **earlier, unrelated read** (e.g. `index.html`) before the deliberately-injected cross-app read is ever reached, producing the generic `Access to this API has been restricted. Use --allow-fs-read to manage permissions.` instead of a message naming the injected file.

### Verified enforcement still fails closed

In every one of the 4 cases, `assert.rejects()` on the promise itself still passed — only the message-content `assert.match` failed. I checked each case individually:

- The symlink test's second assertion (`readFile(marker)` → `ENOENT`) already independently proves the malicious write never landed — that assertion was passing.
- `assertProjectBoundaries`/`main()` reject **only** when `scanProjectBoundaries` reports `violations.length > 0` (`project-boundary.mjs:1050`), and the constrained Vite `build()` call is wrapped so any thrown error (regardless of *why* Node/Vite/Rollup denied the read) becomes a violation and `buildVerified` stays `false` (`project-boundary-vite-build.mjs:674-678`). So a denial anywhere in the read path — whether it's our own JS-level `auditedPath` check or Node's OS-level `--permission` model — still aborts the whole build and gets surfaced as a violation.
- Each of the 4 tests still contains its own "unrestricted build proves the leak is real" preamble (a plain, unsandboxed Vite build that does successfully read/bundle the forbidden sibling content), so the tests still prove there's a real vulnerability that the sandboxed build then blocks.

No case showed the sandboxed build succeeding, or the injected/leaked value slipping into any produced artifact. This is a message-wording drift, not a fail-open defect.

### The fix

Replaced the 4 brittle regexes with a single shared constant, `ZUULI_CONSTRAINED_BUILD_FAILURE`, matching only the part of the failure message this codebase fully controls — the literal wrapping prefix emitted by `project-boundary-vite-build.mjs` itself:

```
zuuli/vite.config.ts: constrained production Vite graph build failed; a plugin or module may have read outside wallet/zuuli and wallet/shared:
```

This is generated by our own code (not by Node or Vite), so it's stable across Node/Vite versions. It still proves the specific project's production build was blocked for a boundary reason, while no longer depending on whichever internal Node/Vite/Rollup wording happens to trail it. I deliberately did **not** touch the other, currently-passing tests that use similarly-shaped regexes (e.g. lines matching `FileSystemWrite|allow-fs-write` elsewhere) — those still pass today and are out of scope for this fix; if they start flaking on a future Node bump the same pattern applies.

### Local verification (Node 24.18.0, in an isolated worktree — nothing landed on local `main`)

- `node --test scripts/project-boundary.node-test.mjs` → **85/85 passing**
- `node scripts/project-boundary.mjs` (live scan) → passing: "3 discovered projects, 324 source files, 1629 parsed module references, 5 production shared-package consumers, 2 constrained production Vite builds"
- `node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs` → passing
- `node scripts/check-zuuli-android-gate.mjs --self-test && node scripts/check-zuuli-android-gate.mjs` → passing
- `wallet/zuuli`: `npm ci && npm run typecheck` → clean; `npx vitest run` → 95 files, 1396 passed / 5 skipped (pre-existing skips)

Rebased/merged onto current `origin/main` (two unrelated commits had landed in `rs/crates/`, no overlap) and pushed once with `--force-with-lease`. Not merging — leaving that to the QA gate + review per repo doctrine.

skyl:
## Fixed the CI-only failure at `7666bd9` — the `build-identity` / boundary-sandbox collision

Confirmed the diagnosis: reproducing locally requires `GITHUB_ACTIONS=true` + `GITHUB_SHA=<40-hex>` (unset outside CI, which is exactly why my first local run never hit it).

### Root cause

`wallet/zuuli/scripts/project-boundary-vite-build.mjs` (the constrained Vite build helper) is spawned by `constrainedViteConfigurationAudit()` in `project-boundary.mjs` under Node's `--permission` model with only `--allow-addons` and a handful of `--allow-fs-read`/`--allow-fs-write` roots — **no `--allow-child-process`**, deliberately, since that flag is broad enough to defeat the read-boundary audit the sandbox exists to enforce (Node itself flags it as dangerous enough to "invalidate the permission model").

`wallet/zuuli/vite.config.ts` calls `build-identity.mjs`'s `loadBuildIdentity()` at config-load time, which shells out to `git status`/`git rev-parse` to confirm the checkout is clean and matches any declared source SHA. Inside the sandbox that spawn is denied (verified directly: `execFileSync("git", ...)` under the sandbox's exact permission flags throws `ERR_ACCESS_DENIED` for the `ChildProcess` permission). `checkoutIsDirty()`'s catch-all then falls back to "assume dirty". Because the subprocess inherited the runner's real `GITHUB_ACTIONS=true`/`GITHUB_SHA` (the spawn used `env: { ...process.env, ... }`), `build-identity` correctly refuses to vouch for a declared source SHA it cannot verify, and throws `"declared source SHA cannot identify a dirty checkout"` — failing the whole boundary scan on an unrelated concern.

Locally this never reproduced because `GITHUB_ACTIONS`/`GITHUB_SHA` aren't set outside CI, so `build-identity` took its documented no-declared-SHA branch (`sourceCommit: null`) instead of ever exercising the dirty-checkout throw.

**Both guards are independently correct** — I did not weaken either:
- `build-identity.mjs`'s dirty-checkout check is untouched. It still runs with the *real* environment for every actual build (dev, CI's real `npm run build`, etc.) and still fails closed exactly as before.
- The boundary sandbox still runs under the same restrictive `--permission` flags, auditing the exact same production Vite config/build — no new read/write access was granted.

### The fix

In `constrainedViteConfigurationAudit()` (`wallet/zuuli/scripts/project-boundary.mjs`), strip `GITHUB_ACTIONS`, `GITHUB_SHA`, and `ZUULI_RELEASE_SOURCE_SHA` from the sandboxed subprocess's environment (setting them to `undefined` in the `env` object, which Node's `child_process` correctly drops rather than stringifying — verified directly). This makes the sandbox run **undeclared**, which is the documented local-dev shape `build-identity` already supports: no declared SHA means a dirty (or unverifiable) checkout is fine and `sourceCommit` is simply `null`. The sandbox's job is to audit reads, not to mint a releasable, provenance-stamped build, so it shouldn't be claiming a CI provenance it structurally cannot verify (since verifying it requires `git`, which the sandbox correctly withholds).

### Verification (Node 24.18.0, CI-like env: `GITHUB_ACTIONS=true`, `GITHUB_SHA=<HEAD>`, `CI=true`)

- Reproduced the failure first on the pre-fix code with `GITHUB_ACTIONS=true GITHUB_SHA=$(git rev-parse HEAD)` — got the exact same `declared source SHA cannot identify a dirty checkout` error.
- Also confirmed it fails identically with a genuinely dirty tree (untracked file) + `GITHUB_SHA` set — same error, same layer.
- After the fix, same CI-like env, clean committed tree:
  - `node --test scripts/project-boundary.node-test.mjs` → **85/85 passing**
  - `node scripts/project-boundary.mjs` (live scan) → passing
  - Also re-verified with a genuinely dirty tree + `GITHUB_SHA` set → still passes (sandbox no longer cares, as intended)
  - `node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs` → passing
  - `node scripts/check-zuuli-android-gate.mjs --self-test && node scripts/check-zuuli-android-gate.mjs` → passing
  - `node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs` → passing (345 self-test cases)
  - `wallet/zuuli`: `npm ci && npm run typecheck` → clean; `npx vitest run` → 98 files, 1486 passed / 5 skipped; `npx playwright test` → **184/184 passing**

### Also merged the newly-landed `origin/main` (#859, #861)

`origin/main` had moved again (RTL layout mirroring, #861) since my last push, with real conflicts this time in `.github/workflows/zuuli.yml` and `scripts/check-github-actions-pins.mjs` — both were the same shape: this branch's "Verify wallet project boundaries" step/self-tests landing adjacent to #861's new "Verify RTL source policy" step/self-tests. Resolved by keeping both step blocks and both sets of self-test cases (verified no `<<<<<<<`/`=======`/`>>>>>>>` markers remain, YAML parses, `node --check` passes on the pins script, and its own 345-case self-test suite passes). `wallet/zuuli/package.json`/`package-lock.json` auto-merged cleanly (verified with a fresh `npm ci`).

Note: before this merge, `npx playwright test` showed 181/182 passing with one unrelated failure (`tests/navigation.pw.ts` touch-target sizing at 320px) — that was against a pre-#861 tree and disappeared entirely after merging (184/184 now), so it was a stale-tree artifact, not a real regression.

Pushed once with `--force-with-lease` (`7666bd9..2f7d395`). Not merging — leaving that to the QA gate + review per repo doctrine.

skyl:
## Rebased once more onto `origin/main` (#857, #863) — pure conflict resolution

`origin/main` had moved again since my last push: `4bc05fc` (#857, store capture image) and `355d648` (#863, snackbar insets). As expected, #857 was the third contributor to the same two files reconciled last time — it also edited the "Detect release-impacting ZUULI changes" change-detector step and re-derived the Android-gate digest.

### Conflicts (2 files, as anticipated)

**`.github/workflows/zuuli.yml`** — one conflict, in the `zuuli=true` case-pattern line of the change-detector step. Diffed both sides item-by-item before touching anything:
- Items only on `origin/main` (genuinely new from #857): `scripts/check-zuuli-store-capture-image.mjs`, `.github/containers/zuuli-store-capture/*`, `.github/workflows/zuuli-store-capture-image.yml`, `docs/ZUULI-STORE-CAPTURE-IMAGE.md`, and a narrow `wallet/shared/sensitive-entry-session.ts`.
- Items only on this branch (from #862's own earlier broadening of the selector for the new `wallet/shared` npm package): `wallet/package.json`, `wallet/package-lock.json`, `wallet/shared/*`, `wallet/zuuallet/*`.
- `wallet/shared/sensitive-entry-session.ts` is already subsumed by this branch's broader `wallet/shared/*`, so it was dropped as redundant (no coverage lost — confirmed by the selector self-tests below). Everything else from both sides was kept, verbatim.
- Also added #857's new "Verify the store-capture build image policy" step, alongside this branch's "Verify wallet project boundaries" and #861's "Verify RTL source policy" — all three now sit side by side, none dropped.
- **First pass had a bug I caught before deriving the digest:** my merged glob line was missing the closing `)` that bash's `case` syntax requires directly after the last pattern (both parents' originals had it; my line-splice dropped it). `check-zuuli-android-gate.mjs --self-test` caught this immediately — the live `check()` call failed with a wall of "must select the full ZUULI suite" / "must retain full wallet/zuuli/* selector" errors, because the malformed line made every pattern-set membership check come up empty. Confirmed the fix by isolating `job()`/`namedStep()` output in a standalone script before re-deriving anything, restored the trailing `)`, and only then re-derived the digest — so I didn't paper over that bug with a hash update.

Confirmed the merged change-detector step, once fixed, is **purely additive**: `git diff` against my previous merge commit shows only the one new step and the appended glob entries — the fail-open logic (`No usable base commit`, `Diff failed`), the `.md` prose skip-guard, and the `zuuallet_schema` case block are byte-for-byte untouched.

**`scripts/check-zuuli-android-gate.mjs`** — the `changeDetectorDigest` constant, same pattern as last time. Re-derived with `node scripts/check-zuuli-android-gate.mjs --print-change-detector-digest` only after confirming the step above was correctly reassembled (including catching and fixing the missing `)`).

### Verification (Node 24.18.0, CI-like env: `GITHUB_ACTIONS=true`, `GITHUB_SHA=<HEAD>`, `CI=true`, clean committed tree)

- `node --test wallet/zuuli/scripts/project-boundary.node-test.mjs` → **85/85 passing**
- `node wallet/zuuli/scripts/project-boundary.mjs` (live scan) → passing
- `node scripts/check-zuuli-android-gate.mjs --self-test && node scripts/check-zuuli-android-gate.mjs` → passing (34 mutations)
- `node scripts/check-workflow-gates.mjs --self-test && node scripts/check-workflow-gates.mjs` → passing (39 cases)
- `node scripts/check-github-actions-pins.mjs --self-test && node scripts/check-github-actions-pins.mjs` → passing (345 cases)
- `wallet/zuuli`: `npm ci && npm run typecheck` → clean; `npx vitest run` → 99 files, 1487 passed / 5 skipped; `npx playwright test` → **184/184 passing** (run twice; one run showed `tests/navigation.pw.ts` — the same touch-target-sizing test I flagged as environment-flaky last round — fail-then-recover as "1 flaky" under Playwright's own `retries: 1` CI policy, both runs finished exit 0. Diffed #863's CSS changes: they only touch Sonner/toast positioning variables, nothing in the `.app-mobile-more-dialog` row layout this test measures, so it's unrelated to anything merged here.)

Pushed once with `--force-with-lease` (`2f7d395..c2eb699`). Not merging — leaving that to the QA gate + review per repo doctrine.

PR #871, opened 2026-09-01
release(zuuli): prepare internal build 0.1.0+20

skyl:
Bumps ZUULI to internal build `0.1.0+20`. Version stays `0.1.0`.

**Stacked on #870** (the build-20 STATUS re-derive), so this PR's diff against
`main` currently shows both commits. Once #870 squash-merges, the STATUS half
drops to a no-op and this reduces to `release.json` plus the four generated
identity surfaces. Stacking is deliberate: a docs-only STATUS change still
triggers the full ~40-minute ZUULI gate, so both gates run concurrently. Same
shape as builds 18 and 19.

## What changed

Produced mechanically, nothing hand-edited:

```
node wallet/zuuli/scripts/release-bump.mjs --version=0.1.0 --build=20
```

| File | Change |
|---|---|
| `wallet/zuuli/release.json` | `build` 19 → 20 |
| `src-tauri/gen/android/app/build.gradle.kts` | `versionCode` default 19 → 20 |
| `src-tauri/gen/apple/project.yml` | `CFBundleVersion` 19 → 20 |
| `src-tauri/gen/apple/zuuli_iOS/Info.plist` | `CFBundleVersion` 19 → 20 |
| `src-tauri/tauri.conf.json` | iOS `bundleVersion` and Android `versionCode` 19 → 20 |

## What build 20 carries

Thirty-one PRs landed after build 19's source commit `cafa488` and reach a
build for the first time here. Notably it is the **first build that fixes the
mobile More sheet rendering fully off-screen at rest** on narrow viewports
(#869) — a user-facing navigation defect that shipped in builds 18 and 19,
measured at `left` of -160 (320px) and -180 (360px). It is also the first build
without the KT `same_key` / `directory_auth_pk` spec-code divergence (#841,
closing #837). See #870 for the full re-derived disposition.

Nothing here changes what is *proven*: #813 remains open and unproven, no store
readback exists at build 18 or 19, and messaging still cannot start a
conversation on the shipping `NoDirectory` default.

## Verification

- `npm run release:verify` → `{"identity":"0.1.0+20","tag":"zuuli-v0.1.0+20", ...}`
- `node wallet/zuuli/scripts/status-freshness.mjs --source-sha=<this commit>` → **passes**: `{"auditSha":"9c78e90fda7aa4c886411164b4f490c2e522f46c","auditDate":"2026-09-01"}`
- Diff confirmed to be exactly the five files above

### DIFF
--- a/wallet/zuuli/STATUS.md
+++ b/wallet/zuuli/STATUS.md
@@ -8,74 +8,96 @@ back from the named store. Authenticated, money-moving, wallet, KYC, and media
 operations are not called working without recorded evidence from that path.
 
 Last re-derived from `origin/main` at
-`e2be49c8e7b319c383b449d4ea1747c9a162c726` on 2026-08-31. Before a release,
+`9c78e90fda7aa4c886411164b4f490c2e522f46c` on 2026-09-01. Before a release,
 update the evidence and disposition for every non-ready row; do not carry this
 commit or date forward mechanically.
 
-This anchor is the first one since build 18 to carry application changes.
-Thirteen PRs landed between `73b14ff` (the build-18 anchor) and this commit:
-the i18n kernel ([#797](https://github.com/free2z/zuu/pull/797)), the wallet
-identity store foundation ([#805](https://github.com/free2z/zuu/pull/805)),
-locked creator ZEC tip intents ([#799](https://github.com/free2z/zuu/pull/799)),
-the authoritative `ppv` stream kind
-([#817](https://github.com/free2z/zuu/pull/817)), trusted first-party image
-auto-load ([#801](https://github.com/free2z/zuu/pull/801)), camera/microphone
-preflight ([#804](https://github.com/free2z/zuu/pull/804)), MLS `KeyPackage`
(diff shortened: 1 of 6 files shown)


skyl:
Closing: this bump cannot pass as-is, and the cause is a real defect rather than anything wrong with the bump.

`zuuli / frontend` failed on `tests/about.pw.ts`, which pins the release identity as literals (`getByText("19")` and a regex containing `Build: 19`). Bumping to 20 makes it fail by construction — so **every** ZUULI release bump would fail the required gate until that is fixed. This is the first bump since #822 added the About build-identity binding (#822 landed after build 19 was cut), so build 20 is the first release to hit it.

The fix cannot ride along in the release source commit: `status-freshness.mjs` correctly rejects a non-ceremony release-impacting change there. So it is #872, landing ahead of the bump.

Once #872 merges, main's tip moves and `wallet/zuuli/tests/about.pw.ts` is a release-impacting path, so STATUS.md needs re-deriving at the new anchor before the bump. Build 20 will be re-cut on that new base. No bytes were published for build 20, so the number is still free to reuse.

PR #878, opened 2026-09-03
fix(rs): align relay and KT runtime with corrected specs

### LINKED ISSUE #646: rs: align f2z-relay-proto / f2z-relay / f2z-kt with the 2026-08-24 spec corrections (#630, #633, #634)
Follow-up to the docs-only reconciliation in #630 / #633 / #634. Those corrections
turned three implementation workarounds into **specified behaviour**, and turned
one client policy knob into a non-conforming one. Docs moved first, deliberately;
this is the code half.

## 1. `f2z-relay-proto`: a token-gated relay is now a refusal, not a warning

`crates/f2z-relay-proto/src/capabilities.rs` has
`ClientPolicy::allow_token_queue_creation`, which lets a client accept a relay
advertising `queue_creation_mode == 2` if the user opts in.

`WIRE.md` §13.1's correction makes that non-conforming: **value 2 is reserved and
MUST NOT be published**, and a client that reads it **MUST treat the relay as
unusable and MUST NOT offer an override** — there is nothing to override, because
no consent makes a token presentable when the wire has no field for one. The knob
and its test (`a_token_gated_relay_is_refused_unless_the_user_accepts_the_linkability`)
should become an unconditional refusal.

`ErrorCode::QueueCreationTokenGated` stays — a client still has to name why it
refused, and error codes are not retired.

## 2. Comments that call these specification defects

Each of these was correct when written and is now stale, because the specification
moved to meet the implementation:

- `crates/f2z-relay/src/config.rs` — the startup refusal for `token` mode is now
  `WIRE.md` §13.1's specified behaviour, not "the right fail-closed call, reported
  as a spec defect."
- `crates/f2z-kt/src/log.rs` — the heartbeat module note's "Reported as a
  specification defect rather than coded around silently". `KT.md` §5.1's
  correction now specifies the label, the value derivation, the handle-label-space
  exclusion and the `tree_size` effect; the note should cite it instead of
  claiming an invention.
- `crates/f2z-kt-core/src/api.rs` — `Presence`'s "This field is an admission, and
  it should not have to exist". It is now required: `KT.md` §8.1's correction
  makes the unproved labelling **normative**, and the type is the conforming
  shape rather than a stopgap. It should still say the proof does not exist and
  why (`KT.md` §11.5), because that part has not changed.

## 3. Worth checking while in there

- `f2z-relay-testkit` and the conformance tests for anything that still treats
  `token` as a producible configuration.
- Whether `f2z-kt`'s lookup response rejects a populated `proof` alongside the
  absent discriminant in both directions — `KT.md` §9.2 now states it as a
  malformedness rule rather than leaving it to `LookupResponse::validate`.

None of this changes behaviour a client can observe except item 1, which removes
a way to reach a relay no conforming client can use anyway.

skyl:
Closes #646

## Summary

- remove the client policy override for reserved `queue_creation_mode = 2` and reject it as `QueueCreationTokenGated` during capability validation
- align relay, testkit, heartbeat, and `Presence` documentation with the 2026-08-24 WIRE/KT corrections
- cover the complete present/absent lookup-response shape matrix and prove the client rejects an absent discriminant carrying a real populated proof

## Verification

- `cargo test --locked -p f2z-relay-proto -p f2z-relay-testkit -p f2z-relay -p f2z-kt-core -p f2z-kt-client -p f2z-kt` (from `rs/`, Rust 1.97.1)
- pinned Rust 1.97.1 workspace `cargo clippy --locked --all-targets -- -D warnings` pass (via `scripts/check-rust-clippy.sh --root rs`; per-crate repetitions left to CI after the workspace verdict)
- `git diff --check`

### DIFF
--- a/rs/crates/f2z-kt-client/tests/acceptance.rs
+++ b/rs/crates/f2z-kt-client/tests/acceptance.rs
@@ -91,6 +91,7 @@ struct DirectTransport {
     entry_override: Mutex<Option<Vec<u8>>>,
     drop_cosignatures: Mutex<bool>,
     claim_absent: Mutex<bool>,
+    claim_absent_with_proof: Mutex<bool>,
     /// Serve a history with the entry at this index removed — the **truthful
     /// subset** of `KT.md` §8.2. Everything still served is real.
     omit_history_entry: Mutex<Option<usize>>,
@@ -111,6 +112,7 @@ impl DirectTransport {
             entry_override: Mutex::new(None),
             drop_cosignatures: Mutex::new(false),
             claim_absent: Mutex::new(false),
+            claim_absent_with_proof: Mutex::new(false),
             omit_history_entry: Mutex::new(None),
             swap_history_entry: Mutex::new(None),
         }
@@ -145,6 +147,10 @@ impl DirectTransport {
 
     fn assert_absent(&self, value: bool) {
         *self.claim_absent.lock().unwrap() = value;
+    }
+
+    fn assert_absent_but_retain_proof(&self) {
+        *self.claim_absent_with_proof.lock().unwrap() = true;
     }
 
     fn bundle_bytes(&self, mut bundle: f2z_kt_core::api::TreeHeadBundle) -> Vec<u8> {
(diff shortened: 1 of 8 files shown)


skyl:
Adversarial review of exact head `1e8bd0b5988a9092a83c9879be8827d2dafb57ad`: **approved pending the required gate**.

I checked the complete diff and searched the head for surviving `allow_token_queue_creation` / stale stopgap language. `ClientPolicy::accept` still calls `validate` first, so reserved wire value 2 is refused even by an otherwise-lenient policy. The present/absent shape matrix independently mutates both payload fields, and the end-to-end client test retains a genuine proof while falsifying the discriminant, so it covers the smuggling direction rather than only garbage decoding. Relay and FakeRelay both retain startup refusals.

No blocking finding. The public field removal is intentional and correctly fail-closed; CI must prove there are no unported callers before merge.

skyl:
Rebase review of exact head `cfd89d0ea37f6afd2139c3ef899db3ab70b3bcc7`: the branch rebased cleanly onto security-fix main `b86116d`; its stable patch ID remains exactly `2cd5d1e986bfeb97c440587d122fe03fcf118c80`, identical to the previously adversarial-reviewed head. Approval therefore carries forward, pending this exact head’s aggregate gate.

PR #879, opened 2026-09-03
fix(msg-identity): remove alternate MLS signing seam

### LINKED ISSUE #700: DeviceSignatureKey::sign_mls_content signs MLS framing with ed25519-dalek, has no callers, and is the ADR 0001 seam #697 says it calls out but does not
## The seam

`rs/crates/f2z-msg-mls/tests/common/mod.rs:13-26` states it plainly:

> `f2z-msg-identity::DeviceSignatureKey` signs MLS framing with **`ed25519-dalek`** (`sign_mls_content`), and this engine signs it with **libcrux** (`f2z_msg_mls::DeviceSigner`) — which is what closes #693 and what ADR 0001's single crypto core requires. […] That is a real seam and it needs reconciling: either `DeviceSignatureKey` signs through libcrux, or it stops offering `sign_mls_content`. **It is called out in the pull request rather than papered over here.**

It is not called out in the pull request. #697's body has a "Closes #693" section arguing ADR 0001's one-crypto-core is "now literally true for this graph", and nothing about `sign_mls_content`. The note points at a discussion that does not exist, so the decision it defers would have been lost at merge. That is why this is filed.

## What is actually there

```
$ grep -rn 'sign_mls_content' rs/crates/
rs/crates/f2z-msg-identity/src/device.rs:137:    pub fn sign_mls_content(&self, framed_content_tbs: &[u8]) -> Signature {
rs/crates/f2z-msg-identity/src/device.rs:257:            key.sign_mls_content(b"one message"),
rs/crates/f2z-msg-identity/src/device.rs:258:            key.sign_mls_content(b"another message")
```

A public method named for MLS `FramedContentTBS`, with **zero production callers** — its only two uses are in its own unit test. The engine that actually signs MLS framing uses a different key type on a different crypto library.

## Why it matters rather than being tidy-up

1. **It is the ADR 0001 violation #693 is about, still in the tree, just not on the path.** #697's claim is scoped to "this graph" and is true as scoped. But the API that would violate it is public, named for the job, and is what any future caller reaches for first. Closing #693 while leaving `sign_mls_content` present makes the next violation a one-line mistake with no gate against it: `deny.toml` bans the *crate* `openmls_basic_credential`, not `ed25519-dalek`, which `f2z-msg-identity` uses directly.
2. **The two keys are not interchangeable and cannot be bridged.** `DeviceSignatureKey` correctly does not expose its private bytes, so a caller who picks it for MLS gets signatures the engine's verifier — libcrux via `openmls_libcrux_crypto` — has no reason to agree with. Both are Ed25519, so the failure would not be a type error; it would be a signature that verifies or does not depending on which path checked it.
3. **It is a general signing oracle over caller-chosen bytes.** #694's `lib.rs` lists as property 4 that "no key type exposes a general `sign(&[u8])`". `sign_mls_content(framed_content_tbs: &[u8])` takes arbitrary bytes, as do `sign_ceremony_payload` and `sign_queue_command`. I raised the missing test for property 4 when reviewing #694; this is the same gap seen from the other side, and `sign_mls_content` is the member of that set with no caller at all to justify it.

## The decision

Exactly the two the harness note names:

- **Remove `sign_mls_content`.** Nothing calls it, `DeviceSigner` is the engine's signer, and removing it makes ADR 0001 enforced by absence rather than by convention. This is my recommendation — it is a deletion plus its unit test, and it is the only option that cannot rot.
- **Or make `DeviceSignatureKey` sign through libcrux**, so the two paths agree and the method has a defensible meaning. Larger, and it buys nothing until something needs the account-side key to sign MLS framing — which today nothing does.

Either way, a test asserting property 4 — that the crate exposes no general-purpose signature over caller-chosen bytes — would keep the answer from drifting. That test does not exist today.

Found while reviewing #697; relates to #693 and #694.

skyl:
## Summary

- remove the unused `DeviceSignatureKey::sign_mls_content` ed25519-dalek path so MLS framing uses the libcrux-backed engine signer
- replace the stale MLS coordination-test seam with the resolved ownership boundary
- reject restoration of that exact public API from an external consumer using a pinned `trybuild` E0599 compile-fail fixture
- remove the source-text public-surface census and its broad assurance: pinned stable Rust 1.97.1 does not expose rustdoc JSON, and the unstable format represents only each invocation's active cfg/feature expansion

The compile-fail test deliberately makes no claim that it discovers renamed methods, wrappers, re-exports, traits, macros, or raw-key exposure. A follow-up compiler-derived API policy should wait for stable rustdoc JSON, snapshot every supported target/feature combination, normalize resolved type identities and re-exports, and mutation-test cfg- and macro-generated surfaces.

## Verification

- `cargo +1.97.1 test --locked -p f2z-msg-identity --all-features --all-targets`
- `cargo +1.97.1 test --locked -p f2z-msg-identity --doc`
- `cargo +1.97.1 test --workspace --all-targets --locked`
- `cargo +1.97.1 test --workspace --doc --locked`
- `cargo +1.97.1 clippy --workspace --all-targets --locked -- -D warnings`
- `scripts/check-rust-fmt.sh --root rs` and self-test
- `scripts/check-rust-clippy.sh --root rs` and macOS target-native self-test
- `scripts/check-rust-deny.sh --root rs --config rs/deny.toml` and self-test
- `scripts/check-rust-toolchain.sh` and self-test
- `cargo +1.97.1 build --locked -p f2z-msg-identity --target wasm32-unknown-unknown`
- exact restoration mutation: adding back `sign_mls_content(&[u8])` makes the compile-fail fixture compile and the test fail

Closes #700

### DIFF
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -864,6 +864,7 @@ dependencies = [
  "rand_core 0.10.1",
  "sha2 0.10.9",
  "tls_codec",
+ "trybuild",
  "zeroize",
 ]
 
@@ -3255,6 +3256,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "target-triple"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3a6bfce3d99adfa72d24750a61f782f3036a81e7f86d8841ee1326deaebd171"
+
+[[package]]
 name = "tempfile"
 version = "3.27.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3265,6 +3272,15 @@ dependencies = [
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "termcolor"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06794f8f6c5c898b3275aebefa6b8a1cb24cd2c6c79397ab15774837a0bc5755"
+dependencies = [
+ "winapi-util",
 ]
 
 [[package]]
@@ -3506,6 +3522,21 @@ name = "tower-service"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8df9b6e13f2d32c91b9bd719c00d1958837bc7dec474d94952798cc8e69eeec3"
+
+[[package]]
+name = "trybuild"
+version = "1.0.120"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e605bf6b39357663d8ba4e984f8be8da8df6bb32e81031d6889024ea8fd68e4"
+dependencies = [
+ "glob",
+ "serde",
+ "serde_derive",
+ "serde_json",
+ "target-triple",
+ "termcolor",
+ "toml 1.1.4+spec-1.1.0",
+]
 
 [[package]]
 name = "tungstenite"
--- a/rs/crates/f2z-msg-identity/Cargo.toml
+++ b/rs/crates/f2z-msg-identity/Cargo.toml
@@ -58,3 +58,6 @@ tls_codec = { workspace = true }
 
 [dev-dependencies]
 proptest = { workspace = true }
+# Compile the removed public method from an external consumer's perspective
+# and pin the expected E0599 diagnostic on the repository's Rust toolchain.
+trybuild = "1"
--- a/rs/crates/f2z-msg-identity/src/device.rs
+++ b/rs/crates/f2z-msg-identity/src/device.rs
@@ -108,10 +108,11 @@ device_key!(
     /// `DeviceSignatureKey` (DSK) — the MLS leaf `signature_key`
     /// (RFC 9420 §7.2), bound to the account by a `DeviceCredential`.
     ///
-    /// It signs MLS handshake and application framing, which is why it is
-    /// **not** the [`crate::account::IdentitySigningKey`]: §4.1 requires that
-    /// the key which signs message content and the key a peer pins as an
-    /// identity are different keys.
+    /// It is **not** the [`crate::account::IdentitySigningKey`]: §4.1 requires
+    /// that the key which signs message content and the key a peer pins as an
+    /// identity are different keys. The former ed25519-dalek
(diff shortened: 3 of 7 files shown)


skyl:
Adversarial review of exact head `9cf72bf544498619ffeaaa4557d3845cec31cd3c`: **approved pending the required gate**.

I checked the full diff and the retained head. The unused ed25519-dalek MLS method and its stale coordination note are removed; no production `sign_mls_content` caller survives. The new guard derives its population from every crate source file, parses ordinary inherent methods, separately inspects macro bodies (necessary because these keys are macro-declared), distinguishes restricted visibility and trait implementations, and includes positive/negative controls plus a demonstrated live mutation failure. `syn` is dev-only and already present in the locked graph.

No blocking finding. CI still needs to prove the intentional public API removal has no unported consumer.

skyl:
Rebase review of exact head `acb7219ae1bf4b1d0f6ba228d2dc2ea36dec6b47`: the branch rebased cleanly onto security-fix main `b86116d`; its stable patch ID remains exactly `54f726b99944a470664b823c036c3373704fd56c`, identical to the previously adversarial-reviewed head. Approval therefore carries forward, pending this exact head’s aggregate gate.

skyl:
Exact-head adversarial re-review: 4f9ea8c5047f8048633ad3a825e2e4e2f0532d6c is a conflict-free, patch-identical rebase of the unused MLS signing-seam removal already reviewed. Full Rust all-target and doc tests, identity/MLS wasm, fmt/clippy/deny/toolchain checks pass; the API guard rejects a temporary reintroduction of the exact public sign seam and returns green after restoration. Approved at this exact SHA, subject to the required aggregate CI gate.

skyl:
Independent adversarial review refreshed at exact head `8413b0e64e687573a24adafa7cd86d9e88ebbb1e`: **APPROVED — no findings.**

The defensibly narrowed patch is identical by range-diff after rebasing over merged #896/#887 (`d4b404b = 8413b0e`). The consumer trybuild fixture still pins the exact removed `DeviceSignatureKey::sign_mls_content(&[u8])` API to E0599; the exact restoration mutation makes the fixture compile and therefore fail. No repo callers exist, production MLS continues through libcrux, and neither docs nor PR claim a general public-surface census. Pinned identity all-feature/all-target tests and the combined 356-case Actions policy suite pass. Merge remains contingent on exact-head CI.

PR #882, opened 2026-09-03
fix(e2ee): bound signed PoW work policy

### LINKED ISSUE #821: fix(e2ee): bound PoW work and align signed capability policy
Refs #550

Split from PR #800 to preserve one logical change per PR. This issue owns only the proof-of-work default/calibration/runtime family.

## Problem

The proposed public contract allows any nonzero `difficulty_bits`, `f2z-codec` accepts 1..=255, the relay rejects only values above 64, and the shipping client synchronously runs an unbounded fixed-salt wrapping-counter search inside async code. A signed extreme capability can pin a runtime worker indefinitely. The public docs also make unsupported phone-duration claims despite calibration being unmeasured.

## Acceptance

- Publish one bounded client work policy and align WIRE, CLIENT-CONTRACT, codec validation, relay configuration, and shipping client behavior.
- Reject signed capabilities outside that policy before starting work.
- Run solving off the async executor and bind it to a finite work/deadline/cancellation rule; terminate on counter exhaustion without wrapping the same fixed search space.
- Keep the existing algorithm-1 / 20-bit / 60,000 ms interoperability default explicitly provisional and do not invent phone or attacker calibration evidence.
- Add required mutation-sensitive spec↔runtime default/policy checks with a self-test.
- Test maximum and maximum+1 capability validation plus prompt refusal/cancellation for hostile work, while keeping deterministic low-cost solve coverage.

skyl:
## Summary
- define one provisional v1 PoW policy across WIRE, CLIENT-CONTRACT, codec, relay configuration, and the shipping client
- refuse hostile signed capabilities before work and solve on a bounded blocking worker with deadline, cancellation, attempt, and counter-exhaustion limits
- add boundary/refusal/cancellation/low-cost tests plus a mutation-sensitive spec/runtime policy checker wired into the required workflow

## Verification
- `cargo +1.97.1 test --locked -p f2z-codec -p f2z-relay-proto -p f2z-relay`
- `cargo +1.97.1 test --locked --lib` (tauri-plugin-f2zmsg)
- `cargo +1.97.1 fmt --all -- --check` (rs)
- `cargo +1.97.1 fmt --manifest-path Cargo.toml -- --check` (tauri-plugin-f2zmsg)
- `node scripts/check-pow-policy.mjs --self-test`
- `node scripts/check-pow-policy.mjs`

Closes #821

### DIFF
--- a/.github/workflows/zuuli.yml
+++ b/.github/workflows/zuuli.yml
@@ -52,6 +52,8 @@ jobs:
           node scripts/check-workflow-gates.mjs
           node scripts/check-librustzcash-compat.mjs --self-test
           node scripts/check-librustzcash-compat.mjs
+          node scripts/check-pow-policy.mjs --self-test
+          node scripts/check-pow-policy.mjs
 
       # Runs on every event, including changes this job would otherwise skip,
       # because a toolchain pin can drift from any workflow or manifest. The
--- a/docs/e2ee/CLIENT-CONTRACT.md
+++ b/docs/e2ee/CLIENT-CONTRACT.md
@@ -467,12 +467,14 @@ against that entry**, build the group, obtain a second challenge, compute a
 second proof-of-work stamp, and `CONTACT_APPEND` the `Welcome`. Four
 consequences for the UI:
 
-- **It can take seconds, and the delay is proof-of-work on the user's device —
-  now twice.** §12.6 gives the claim its own challenge purpose, so first contact
-  pays two stamps rather than one. On a cheap phone that is meaningfully slow and
-  the reason is unfixable by tuning
-  ([`WIRE.md` §12.4](./WIRE.md#124-the-honest-limits)). Show progress; do not
-  present it as a network wait.
+- **The delay includes proof-of-work on the user's device — now twice.** §12.6
+  gives the claim its own challenge purpose, so first contact pays two stamps
+  rather than one. No supported-phone duration has been measured. Show progress;
+  do not present computation as a network wait. Each search follows the finite
+  [`WIRE.md` §13.1](./WIRE.md#131-the-layers) client policy: reject signed work
+  above 20 bits before starting, run off the async executor, and stop at the
+  16,777,216-candidate, 30,000 ms, challenge-expiry, cancellation, or `uint64`
+  counter boundary without wrapping.
 - **It fails closed when the witness threshold is unmet.** Resolving a *new*
   handle is refused, not degraded
   ([`KT.md` §8.3](./KT.md#83-the-threshold-rule-and-failing-closed)). See rule 5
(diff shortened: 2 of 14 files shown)


skyl:
Blocking adversarial-review finding on exact head `1a30f7793a3d73ceb946487e20744f80ceb958ba`:

`challenge_lifetime(issued)` computes `expires_at_ms - issued.relay_time_ms`. Those are both server timestamps from issuance, so this always grants the solver the original TTL and ignores response transit/queueing time. A delayed response can therefore spend CPU after the relay challenge has already expired, contradicting this PR’s new rule that challenge remaining lifetime is the earlier deadline. `RelayConnection::relay_time_ms()` already maintains the current relay-clock estimate.

Please derive remaining time from `expires_at_ms - self.relay_time_ms()` at the point work starts (for both ordinary and key-package claim challenges), still cap it at `MAX_POW_SOLVE_MS`, and add a regression test proving a response whose issuance timestamp is old but expiry is at/before the current estimated relay time is refused before spawning work. Update the mutation checker if needed so this cannot silently regress. No merge until this is fixed and the exact amended head is re-reviewed.

skyl:
Addressed the blocking finding in `84e5152bf98a2d108d93e5e84c1a3689c9557b82`:

- both ordinary and key-package challenge paths now derive remaining time from `expires_at_ms - self.relay_time_ms()` immediately before entering the bounded solver
- the existing `MAX_POW_SOLVE_MS` cap remains in force
- added a delayed-response regression showing a challenge that was fresh at issuance but expired at current estimated relay time is refused before work
- extended the mutation checker to require current relay time and both shipping call paths

Focused verification is green: pinned 1.97.1 plugin library suite (78 tests), plugin rustfmt check, and PoW policy checker live + self-test. Ready for exact-head re-review; not merged.

skyl:
Re-review of amended exact head `84e5152bf98a2d108d93e5e84c1a3689c9557b82`: **blocking finding resolved; approved pending the required gate and rebase sequence**.

Both solver entry paths now calculate `expires_at_ms - self.relay_time_ms()` immediately before work, so transit/queueing time consumes the challenge lifetime and an expired response fails before `spawn_blocking`. The regression proves the same response is fresh at issuance but rejected against the later relay-clock estimate; the checker now binds current-time subtraction and requires exactly both shipping call sites.

No remaining review finding. This PR overlaps #878 in relay capability/config code, so #878 should land first and this head must then be rebased/resolved and re-reviewed rather than merged stale.

skyl:
Adversarial re-review of exact rebased head `044b29bc0ceb93524a1801a264780da599973f3a`: approved.

I resolved the intentional #878 overlap as the union of both policies: `queue_creation_mode = 2` remains an unconditional document-conformance refusal with no client override, while signed PoW difficulty/TTL over the v1 maxima independently returns `PowWorkPolicyExceeded`. The second commit (expiry from current relay time) range-diffs identically; the first differs only at those #878 token-mode hunks. I also checked that the shared defaults, relay configuration caps, blocking work isolation, cancellation, and transit-time expiry paths remain bound by `check-pow-policy.mjs`.

Fresh post-rebase evidence:
- `cargo +1.97.1 fmt --all --check`
- `cargo +1.97.1 test -p f2z-codec -p f2z-relay-proto` (all unit/integration/doc tests green)
- `node scripts/check-pow-policy.mjs`
- `node scripts/check-pow-policy.mjs --self-test` (all 11 negative controls green)
- `git diff --check`

This approval is bound only to the exact head above; CI must still complete its aggregate gates before merge.

PR #885, opened 2026-09-03
test(crypto): add permanent libcrux KAT suite

### LINKED ISSUE #730: The crypto known-answer tests that justified X-Wing do not exist in the repo — nothing could re-validate the 0.9 migration
Surfaced by the OpenMLS 0.9 migration (#723), which could not re-validate the cryptography it had just bumped.

## The gap

The known-answer tests that justified adopting X-Wing **do not exist in this repository.**

#385 ran NIST ACVP ML-KEM-768 keyGen/encap/decap, RFC 7748 X25519, RFC 8032 Ed25519 (with a tampered-signature negative control) and all three X-Wing draft-06 Appendix C vectors — green, including on a real Android API-29 arm64 device. That evidence is what closed the post-quantum question and let the whole engine be built on libcrux.

**It was a throwaway.** `wallet/zuuli/crypto-target-spike` is a compile-target probe over `libcrux-ml-kem`, not a known-answer suite. So when #723 replaced the entire libcrux family — `libcrux-secrets 0.0.5 → 0.0.6`, `sha3 0.0.8 → 0.0.10`, `ed25519`, `poly1305`, `chacha20poly1305`, `aes` — **there was nothing to re-run.**

The functional suite did exercise real X-Wing, ChaCha20-Poly1305 and Ed25519 natively on aarch64, which is meaningful. It is not a KAT: a functional test proves two of our own components agree with each other, while a KAT proves our component agrees with the *standard*. Two implementations can agree consistently and both be wrong.

## Why this needs fixing now rather than at the next bump

Every future dependency bump in the crypto path has the same hole. The one just merged replaced a constant-time primitive with a **known aarch64 correctness defect** in its predecessor — precisely the change where a KAT is the check that matters, and precisely the one we could not perform.

It also means the strongest claim in `#385` is not reproducible. An auditor asking "show me the vectors passing on the version you ship" currently cannot be answered without rebuilding the harness from scratch.

## Wanted

A **permanent** known-answer suite in `rs/`, run by CI on every change to the crypto graph:

- **NIST ACVP ML-KEM-768** — keyGen, encap, decap.
- **RFC 7748** X25519, **RFC 8032** Ed25519, including negative controls (a tampered signature must fail; an all-zero key must be refused, which is its own RUSTSEC advisory class).
- **X-Wing Appendix C**, all three vectors, asserting the exact combiner output — not merely that encap/decap round-trips. #385 established the vectors are byte-identical between draft-06 and draft-10, so they are stable to pin.
- Vectors **committed** and their provenance recorded, so a reviewer can trace each to its standard rather than to whoever generated it.

Design points worth deciding rather than defaulting:

- **Where it runs.** Host CI catches a bad version bump, which is most of the value and is cheap. But #385's whole finding was that libcrux selects SIMD backends at *compile* time per target — so a KAT that only ever runs on the CI host does not exercise the NEON path that ships to phones. Consider a target-matrix run, or at minimum record that host-only is what we have.
- **Whether it lives beside `f2z-msg-mls` or as its own crate.** Its own crate is probably right: it tests dependencies, not our code, and should not be skipped by a change filter that only watches our crates.

## Acceptance

- [ ] The suite exists in-repo, is committed with its vectors, and passes on the currently shipped versions.
- [ ] CI runs it on any change to the crypto dependency graph — and is **verified to fail** on a deliberately corrupted vector, since a KAT suite that cannot fail is the exact defect class this project has hit repeatedly.
- [ ] The provenance of every vector is recorded.
- [ ] The limits of where it runs are stated plainly, in the manner of the existing docs.

## Related
- **#385** — the spike whose evidence this makes reproducible.
- **#723** — the migration that could not re-run it.

skyl:
## Summary

- add a standalone host-run KAT crate against the exact locked libcrux graph
- cover NIST ACVP ML-KEM-768 keygen, encapsulation, and decapsulation; RFC 7748 X25519; RFC 8032 Ed25519; and all three X-Wing draft-06 Appendix C vectors
- assert exact keys, signatures, ciphertexts, and shared secrets, with tampered-signature, low-order X25519, and zero-only Ed25519 RNG negative controls
- commit standards-publisher vector material with pinned provenance and applicable license/citation notes
- run the suite from the required rs test job and first prove it detects a deliberately corrupted X-Wing combiner output
- bind all three shipping lockfiles to one explicit 34-package OpenMLS/libcrux/HPKE inventory; independently bind the reviewed exact name set by count and SHA-256; reject coordinated inventory/all-lock narrowing, same-size name substitution, version/source/checksum drift, and separator-delimited family additions/removals
- select the rs KAT gate for changes to either wallet consumer manifest as well as either independent wallet lockfile

## Verification

- `node --check scripts/check-crypto-kat-locks.mjs`
- `node scripts/check-crypto-kat-locks.mjs --self-test` (coordinated inventory/all-lock removal, same-size substitution, exact/underscore/hyphen variants, unrelated-prefix controls, and 45 family/lock identity mutations)
- `node scripts/check-crypto-kat-locks.mjs`
- `scripts/check-crypto-kats.sh --self-test`
- `scripts/check-crypto-kats.sh`
- `cargo test --locked --all-targets` in `rs/`
- `scripts/check-rust-clippy.sh --root rs`
- `scripts/check-rust-deny.sh --root rs --config rs/deny.toml`
- `node scripts/check-github-actions-pins.mjs --self-test` (including both wallet manifest selector mutations)
- `node scripts/check-workflow-gates.mjs --self-test`
- `node scripts/check-workflow-gates.mjs`
- `actionlint .github/workflows/rs.yml`
- complete `f2z-msg-mls` tests and wasm build; messaging-plugin 80 unit tests and 10 lifecycle tests

## Target limits

This executes cryptography on the Ubuntu x86_64 CI host. It does not execute the compile-time-selected NEON backend shipped on aarch64 Android and iOS, and the existing wasm job does not run Rust tests in a browser. The crate README records that boundary rather than presenting cross-compilation as runtime KAT evidence.

Closes #730


skyl:
Adversarial review of exact head `814c6bf14a5d7a4bbe19b2f072ba69dd5aadec15`: approved, pending the full QA gate.

I independently verified the committed X-Wing JSON is byte-identical to upstream blob `a179e093a854c224e06680e33230bf82c3cd30f5`, and JSON-normalized all three selected NIST cases against ACVP-Server commit `975de31e` (keyGen tg2/tc26, encap tg2/tc26, decap tg5/tc86); all are exact. The RFC 7748/8032 values match the named published vectors. The requested review fixes are present: libcrux ECDH/KEM now inherit their single workspace declarations, and the all-zero X25519 refusal is accurately described as stricter free2z policy rather than an RFC mandate.

The corruption self-test is mutation-sensitive to an exact X-Wing shared-secret assertion, the negative tests exercise libcrux refusal rather than a wrapper, and the README accurately limits evidence to host execution. No blocking finding remains. I will require this amended head’s aggregate gate before merge.

skyl:
Exact-head adversarial re-review: 5a17484582939ccfcc4abb336f2a86d148b08374 is a conflict-free, patch-identical rebase of the permanent crypto KAT suite already reviewed. The pinned Linux 1.97.1 container rejects a corrupted X-Wing vector and accepts all seven canonical vectors; the full locked Rust suite, formatting, clippy, deny, toolchain, action-pin, and workflow-gate mutation controls pass. Approved at this exact SHA, subject to the required aggregate CI gate.

skyl:
Exact-head adversarial review at `b31a5dfb501d43b654fa8c801a1e8e0380a31acb`: no findings. I independently verified the repaired exact 34-package lock inventory/digest and separator-delimited family census, reproduced the coordinated removal and same-size substitution failures, confirmed unrelated prefix neighbors remain excluded, and ran the corrupted-vector negative control plus all seven live KATs. The committed X-Wing vector file hashes to the documented upstream Git blob `a179e093a854c224e06680e33230bf82c3cd30f5`. Runtime review found all-zero signing material is rejected before fresh-enrollment or restore state mutation. This head is intentionally not merge-ready: obsolete CI was cancelled and it must be patch-identically rebased onto the drained `main`, then pass fresh exact-head CI.

skyl:
Exact-head review update for `5a879c7f8f017d4fddd31d5221102a38b63d48e8`: no findings after rebasing onto `origin/main` at `ba72c7ef5944c9ef9730f7e263c13016a2a6282e`. The only range-diff delta is the deliberate union with merged #898: the rs selector/source-input contract retains `scripts/check-kt-sth-repeat-agreement.mjs` alongside the KAT scripts and four wallet graph inputs, and both repeated-STH credential rules plus both exact KAT steps retain their mutation coverage. Post-integration checks passed: KAT lock self-test/live, corrupted-vector control, all seven KATs, repeated-STH agreement self/live, Actions policy self/live (124 Rust-root cases), workflow-gates self/live, and diff cleanliness. Merge remains gated on the newly restarted exact-head aggregate.

PR #886, opened 2026-09-03
test(zuuli): remove gate timing races

### LINKED ISSUE #865: ZUULI tests: two specs on the required gate are flaky under parallel load
Two separate specs on the required `zuuli / frontend` gate now fail intermittently. Both currently clear the gate via retry, which is exactly why they need tracking — a flaky-but-passing test on a required gate trains everyone to ignore it.

### 1. `tests/navigation.pw.ts` — "signed in five-item navigation works at 320px"
Tracked in detail at #851. Touch-target sizing at the narrowest viewport; observed at ~2/3 and ~3/5 locally, and still flickering against a tree containing #857, #861 and #863. Two candidate causes (genuinely marginal five-item layout after #822, vs a settling race) — #851 calls for **measuring** the rendered sizes rather than inferring from pass rate.

### 2. `src/components/common/BidiIdentifier.test.tsx` — new
Surfaced while working #838. **Times out under full-suite parallel vitest load; passes in isolation.** This is an AST-scanning test introduced by #861 (RTL layout mirroring and bidi identifiers, merged today). AST scans over the source tree are CPU-heavy, so under vitest's default parallelism it can exceed its timeout while a dozen other files compete.

Likely fixes, in order of preference:
- Give the AST scan a realistic timeout, or mark the file to run serially (vitest supports per-file concurrency control) rather than raising global timeouts.
- Or memoize/narrow the scan so it isn't re-walking the whole tree per assertion.

Do **not** fix either by raising retries or loosening what is asserted.

### Acceptance criteria
- [ ] `BidiIdentifier.test.tsx` passes deterministically under a full parallel `npx vitest run`, verified with repeated runs.
- [ ] #851's Playwright flake resolved deterministically (tracked separately there).
- [ ] Neither fix weakens an assertion, raises retry counts, or skips a test.

Context: both were observed during the post-cleanup issue burndown; #861 and #822 are the relevant recent merges.

skyl:
Closes #865

## What changed

- Cache BidiIdentifier policy `SourceFile`s and type checkers by exact file/source identity, so each mutant recompiles only the file it changes instead of rebuilding all eight audited programs.
- Give only the compiler-backed opaque-identifier suite a realistic 30-second budget; global Vitest retries and timeouts are unchanged.
- Run the SQLite theft-fallback test's FakeRelay at the shipping 25-second ping cadence instead of the harness's keepalive-test-oriented 500 ms cadence.
- Prove the established relay transport is still ready with a signed capabilities round trip immediately before the definitive theft assertion. The `SendAddressStolen` assertion and all atomic/no-partial-state checks remain exact.

## Verification

- Full parallel `npx vitest run`: 5/5 stress runs green, 99 files / 1,506 active tests each; post-rebase run also green.
- `BidiIdentifier.test.tsx`: all 76 tests green under concurrent cold Rust clippy load; initial inventory completed in 2.8 s.
- Exact Rust regression: 50/50 stress runs green on Rust 1.97.1; post-rebase focused run green.
- CI-equivalent messaging command green: `cargo test --locked --all-targets --features relay-harness` (74 lib tests plus all integration/two-process harness tests).
- `cargo +1.97.1 clippy --locked --all-targets --all-features -- -D warnings`
- `cargo +1.97.1 fmt -- --check`
- `npm run typecheck:tests`
- `npm run test:rtl` (17 source-policy tests, 88 Vitest tests, 2 Playwright tests)
- Tauri plugin permission self-test and live parity check.

### DIFF
--- a/wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs
+++ b/wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs
@@ -4974,6 +4974,7 @@ fn capabilities_view(capabilities: &f2z_codec::commands::Capabilities) -> RelayC
 mod tests {
     use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+    use std::time::Duration;
 
     use f2z_codec::commands::Command;
     use f2z_codec::types::{Digest, PublicKey, RelayId, ShortBytes};
@@ -4982,6 +4983,7 @@ mod tests {
     use f2z_msg_identity::{AccountKeys, DeviceCredentialRequest};
     use f2z_msg_mls::{DeviceSigner, MlsEngine};
     use f2z_msg_store::{Durability, MemoryBackend, Op, SqliteBackend, StorageBackend, StoreError};
+    use f2z_relay_testkit::config::RelayConfig as FakeRelayConfig;
     use f2z_relay_testkit::fake::FakeRelay;
     use f2z_relay_testkit::faults::{Effect, Fault, Trigger};
     use openmls::prelude::{GroupId, MlsGroup};
@@ -6655,7 +6657,16 @@ mod tests {
 
     #[tokio::test]
     async fn failed_atomic_theft_commit_uses_loud_in_memory_fallback_without_partial_state() {
-        let relay = FakeRelay::with_defaults().expect("fake relay");
+        // This test deliberately parks the client socket while it opens and
+        // primes a real SQLite store. The FakeRelay's 500 ms keepalive is for
+        // keepalive tests, not representative of the shipping relay's 25 s
+        // interval; under a contended gate it could close this otherwise-ready
+        // fixture before the storage-failure assertion even began.
+        let relay_config = FakeRelayConfig {
+            ping_interval: Duration::from_secs(25),
+            ..FakeRelayConfig::default()
+        };
+        let relay = FakeRelay::new(relay_config).expect("fake relay");
         let server = relay.listen_loopback().await.expect("relay listener");
         let url = server.url();
         let root = tempfile::tempdir().expect("tempdir");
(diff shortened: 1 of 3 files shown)


skyl:
Adversarial review of exact head `993ae92f81ee470ea4f404faf2a9c3a221c80481`: approved, pending the full QA gate.

I checked the failure mechanisms rather than only the green-path counts. The TypeScript cache key is the exact `(fileName, source)` pair, so every mutated source receives a distinct AST; the checker cache is keyed by that exact `SourceFile`, and retention is bounded to this test module lifetime. The 30-second allowance is suite-local and does not add global retries or weaken any inventory assertion. On the Rust side, 25 seconds matches the shipping relay default, the CAPABILITIES round trip occurs before the injected storage failure and cannot consume the storage apply counter, and the definitive `SendAddressStolen`, no-partial-state, and in-memory alarm assertions remain unchanged.

The local evidence is proportionate: five full parallel 1,506-test runs, 50/50 exact Rust stress runs, full messaging harness, clippy/fmt/typecheck/RTL/permissions checks. No blocking finding remains. I will still require the exact-head aggregate gate before merge.

skyl:
Exact-head adversarial re-review: 2e3e3c58747893007cb18df7187d95ab66eb0613 is a conflict-free, patch-identical rebase of the two gate-flake corrections already reviewed. The theft fallback regression, full relay-harness tests, all 1,506 active Vitests, RTL mutation/unit/browser suite, permissions policy, clippy, formatting, and high-threshold audit pass. It incorporates no #897 content and has no overlapping hunk. Approved at this exact SHA, subject to the required aggregate CI gate.

skyl:
Exact common-base approval: 9244f06681012ab6584a1ba7339347e84f050f35 is a conflict-free, patch-identical rebase onto 7f5549114d7544ddcd2f733d2b9e4596f32f4fd1. Independent adversarial review confirmed the relay test preserves theft/no-partial-state/fault-consumption invariants under 50 concurrent repetitions and the bidi cache remains exact-source mutation-sensitive. Approved at this SHA, subject to fresh aggregate CI.

PR #898, opened 2026-09-03
fix(kt): compare complete repeated tree heads

skyl:
Summary:
- make KT section 6.3 explicitly branch on epoch: exact complete SignedTreeHead repeats are no-ops, older heads roll back, and only later heads run advancement checks
- retain the last verified SignedTreeHead in LogView and compare the canonical protocol value directly, covering VRF, cadence/merge policy, reset count, successor key, chain metadata, and signature without a parallel encoding
- add independent field mutants plus real witness steady-poll and same-epoch two-handle client acceptance coverage
- add a mutation-tested spec/runtime agreement checker and bind its exact CI ownership

Verification:
- cargo +1.97.1 test --locked -p f2z-kt-core (94 unit + 8 integration tests)
- cargo +1.97.1 test --locked -p f2z-kt-client --test acceptance (19 passed)
- cargo +1.97.1 test --locked -p f2z-witness --test acceptance (5 passed)
- cargo +1.97.1 clippy --locked -p f2z-kt-core --all-targets -- -D warnings
- scripts/check-rust-fmt.sh --root rs --self-test; scripts/check-rust-fmt.sh --root rs
- node scripts/check-kt-sth-repeat-agreement.mjs --self-test; live check
- node scripts/check-github-actions-pins.mjs --self-test; live check
- node scripts/check-workflow-gates.mjs --self-test; live check
- node scripts/check-hash-domain-labels.mjs --self-test; live check

Closes #892

Issue #892, opened 2026-09-03
kt: require repeated tree heads to match every signed field


skyl:
Split from #558.

`KT.md` §6.3 simultaneously requires `epoch > last.epoch` and defines an equal-epoch branch that compares only `root_hash`, `tree_size`, and `published_at_ms`. The runtime accepts that limited equality and returns before checking other signed `SignedTreeHead` fields, including `vrf_public_key`. A same-epoch head with altered signed metadata therefore is not governed by a single byte-identical/no-op rule.

Scope this as a focused spec plus runtime correction, independent of the reset and witness-policy decisions in #551/#770.

Acceptance:

- `KT.md` explicitly branches: a byte-identical re-presentation of the complete signed head is a no-op; a changed field at the same epoch is fork evidence.
- Advancement-only monotonicity and chain checks apply only when `epoch > last.epoch`.
- `f2z-kt-core` compares every consensus-relevant signed field, including the VRF key and cosignature policy representation, without creating an alternate encoding.
- Tests mutate each field independently and prove identical steady-state witness polls and same-epoch multi-handle client lookups remain conforming.
- A mutation-sensitive checker binds the spec/runtime agreement.

Refs #558.

PR #899, opened 2026-09-03
fix(kt): make long-offline catch-up resumable

### LINKED ISSUE #893: kt: make long-offline tree-head catch-up bounded and resumable
Split from #558.

`KT.md` §6.3 requires fetching every intervening head through a one-head endpoint. The client caps one catch-up call at `MAX_EPOCH_CATCHUP = 256`, but a larger gap fails without advancing the trusted checkpoint. Retrying the same lookup repeats the same failure, so a client offline for roughly 43 hours at the proposed 10-minute cadence can be permanently locked out.

Acceptance:

- The protocol defines a bounded batch/range or checkpointed pagination mechanism with canonical ordering and response limits.
- A client advances only after verifying a contiguous prefix and can resume safely after interruption.
- Gaps larger than 256 epochs converge across repeated bounded calls rather than repeating the same failure.
- Fork, rollback, missing-head, duplicate-head, reorder, truncation, and server-stall cases fail closed without skipping an epoch.
- §8.5 states realistic round-trip/byte costs for long dormancy.
- Runtime integration tests prove multi-batch recovery and persisted resume state; mutation controls prevent replacing contiguous verification with latest-head trust.

Refs #558.

skyl:
Closes #893

## Summary

- specify checkpointed pagination over the existing exact-epoch endpoint: an ascending inclusive prefix of at most 256 signed heads, with exact response-epoch binding and an explicit incomplete outcome
- retain the last accepted complete `SignedTreeHead` inside `LogView` as the single canonical durable checkpoint; `checkpoint_bytes()` encodes it and `resume()` canonically decodes, signature-checks, and re-pins it
- integrate #898 without duplicating or weakening its repeated-head semantics: the catch-up target is re-accepted through complete protocol equality, and failed same-epoch alternatives leave the verified exact-epoch checkpoint unchanged
- advance the checkpoint after every consecutively verified head, including prefixes preceding transport or decode failures, while never accepting the target across an unverified gap
- document the honest long-dormancy cost: a year at the ten-minute cadence is about 52,560 exact-epoch round trips and 15.7 MiB without cosignatures / 26.0 MiB with one cosignature per head, before latest requests and HTTP/TLS overhead
- add signed runtime fixtures covering a 600-epoch three-batch recovery, persisted byte-level restart, corrupt-checkpoint refusal, complete repeated-head equality, fork, rollback, chain break, missing, duplicate, reorder, truncation, and server stall

## Compatibility

`KtClient::resume` now accepts canonical signed-head checkpoint bytes and returns `Result<Self>` instead of accepting an in-memory `LogView`. There were no repository callers. This makes the documented restart seam persistable and ensures resumed state is signature-checked.

## Integration with #898

This branch is rebased onto `ba72c7ef5944c9ef9730f7e263c13016a2a6282e`. `LogView::last_head()` exposes the exact complete head already retained by #898; `KtClient` no longer carries a second checkpoint that could drift. The interaction test changes only `reset_count` in a re-signed same-epoch latest head, proves it is fork evidence after the exact page reaches that epoch, proves the canonical page head remains durable, then resumes and accepts the exact canonical repeat as an idempotent no-op.

## Verification

- `cargo +1.97.1 test --locked --manifest-path rs/Cargo.toml -p f2z-kt-client -p f2z-kt-core -p f2z-kt --all-targets`
- `cargo +1.97.1 test --locked --manifest-path rs/Cargo.toml -p f2z-kt-client -p f2z-kt-core -p f2z-kt --doc`
- `cargo +1.97.1 clippy --locked --manifest-path rs/Cargo.toml -p f2z-kt-client -p f2z-kt-core -p f2z-kt --all-targets -- -D warnings`
- `cargo +1.97.1 check --locked --manifest-path rs/Cargo.toml -p f2z-kt-client --target wasm32-unknown-unknown --no-default-features --features verifier`
- Rust fmt, native-clippy, deny, and toolchain self-tests plus live `rs/` checks
- repeated-STH, workflow-gate, hash-domain, bounded-PoW, GitHub Actions, and AKD evidence self-tests plus live checks
- `git diff --check`

## Mutation controls

- replacing the `EpochGap` branch with a fresh `LogView::pin(..., target)` makes `latest_head_trust_mutation_cannot_satisfy_a_catch_up_page` return success instead of the required incomplete checkpoint
- the #898 interaction fixture changes only `reset_count` while keeping the former epoch/root/size/time subset identical; subset equality would accept it, while the complete comparison returns `KtError::Fork` and preserves the exact page checkpoint
- the repeated-STH agreement self-test killed all six spec/runtime/field/coverage mutants

Exact pushed head: `300bc2c2c0f0d1ed0ab166b24bd58c86035ba051`

### DIFF
--- a/docs/e2ee/KT.md
+++ b/docs/e2ee/KT.md
@@ -1704,7 +1704,24 @@ below against the last one it accepted for this `log_id`:
    `prev_sth_hash == H("free2z/kt/v1/tree-head-hash", tls_codec(last.sth))` directly.
    If `epoch > last.epoch + 1`, the verifier MUST fetch every intervening tree
    head and check the chain link by link. **It MUST NOT skip.** A gap accepted on
-   trust is a branch accepted on trust.
+   trust is a branch accepted on trust. Clients paginate that work as follows:
+   the durable checkpoint is the last accepted `SignedTreeHead`; a page starts
+   at `last.epoch + 1`, ends at `min(last.epoch + 256, target.epoch)`, and fetches
+   `GET /kt/v1/sth/{epoch}` once for every epoch in that inclusive range, in
+   ascending order. Each response MUST contain exactly the requested epoch.
+   After each head verifies, the client advances its in-memory checkpoint; if
+   the page ends before `target.epoch`, it returns an explicit incomplete
+   result, the caller persists the advanced checkpoint, and a later call
+   resumes at the next epoch. The caller also persists the verified prefix
+   exposed after a network or decoding failure. A process interruption before
+   that write may repeat part of the page from an older checkpoint, which is
+   safe because every head is reverified; it MUST NOT resume from an unsigned
+   request counter. A missing, duplicate, reordered, truncated, or stalled
+   response fails closed at that response and cannot consume its epoch. The
+   page is therefore bounded to 256 one-head responses (257 round trips
+   including the initial latest-head or lookup response), while repeated calls
+   converge from an arbitrarily old checkpoint without accepting the target
+   across a gap.
 8. For `epoch == last.epoch`, the complete `SignedTreeHead` — every field of
    `SignedTreeHeadTBS` and `signature` — MUST be identical to the last accepted
    value. An identical re-presentation is an idempotent no-op. Any difference is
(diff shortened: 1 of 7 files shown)


skyl:
Adversarial review of exact head `733eae0aefc3670bc78cee5d532a7608184cdd5d`: approved.

The design stays within the existing wire protocol: the signed head itself is the only page checkpoint, exact-epoch GET responses are required to match the requested epoch, and each call accepts at most 256 contiguous heads. A page never trusts the latest target across a gap; partial transport/decoding failures expose only the verified prefix for persistence, and canonical resume re-verifies the checkpoint signature under the configured log identity/key. `CatchUpIncomplete` is retryable progress, not fork evidence.

I reviewed the boundary arithmetic (256, 257, and 600 epochs), exact-target recheck, duplicate/reorder/missing/truncation/stall behavior, failure-state updates, resume corruption rejection, request counts, and executable byte-cost claims. The tests use real signed heads/cosignatures and a process-style byte boundary, not a cloned private `LogView`. The API is intentionally breaking but currently has no downstream resume caller.

The worker passed full client all-target/doc tests, focused `-D warnings` clippy, wasm verifier build, repo-wide fmt/clippy/deny/toolchain checks, and policy self-tests. Approval is bound only to this exact head; #892 will require an integration rebase/re-review because both retain the accepted head state.

skyl:
Re-review of superseding exact head `4f494701231a549a97e40b5d59caeeaf41e8709e`: approved.

The only delta from reviewed `733eae0` is a four-line `Transport::sth_at` contract clarification: the client, not the transport, binds each decoded head to the requested epoch; one operation calls it at most 256 times; the last accepted signed head is the checkpoint. That precisely matches the reviewed implementation and introduces no behavior change. Focused catch-up tests and `clippy -D warnings` reran green. This approval replaces the prior head-bound review; fresh post-integration CI is still required.

Issue #906, opened 2026-09-04
scaffold: stand up cash.free2z.free2z and cash.free2z.e2e2z as buildable Tauri apps


skyl:
Parent: #904. Pairs with #905 (the bridge).

Goal: **working boilerplate for the two new apps before any feature moves**, so surface extraction is a move rather than a build, and so `e2e2z` can be worked on independently.

## The three apps

| App ID | Role | Privileged plugins |
|---|---|---|
| `cash.free2z.zuuli` | wallet authority (exists today) | `zcash` |
| `cash.free2z.free2z` | content — articles, live, AI, creator, search | **none** |
| `cash.free2z.e2e2z` | messaging | `f2zmsg` |

## What each scaffold needs

- `src-tauri/` with its own `tauri.conf.json`: app ID, product name, icons, **its own CSP** (the content app's is permissive by necessity; `e2e2z`'s should be as tight as ZUULI's).
- **Capability files that grant nothing privileged by default.** `free2z` must contain **no** `zcash:*` and **no** `f2zmsg:*` entry — this is the whole point, and it needs a contract test, not a convention. Model it on `wallet/zuuli/scripts/mobile-webview-authority.mjs`, which already enforces exactly this shape for `mobile.json`.
- Deep-link registration for the bridge (#905), per platform.
- Vite + React + the existing i18n kernel; reuse `@free2z/wallet-shared`.
- Registration in `wallet/zuuli/scripts/project-boundary.mjs` so cross-app imports fail CI (#691/#862).
- CI: gate wiring, release identity, packaging. Note `scripts/check-workflow-gates.mjs` requires every job be covered by a gate's `needs:` and its result actually read.
- Its own `release.json` / build identity, following the ZUULI pattern.

## Watch out for

Real hazards found while planning — these will bite whoever does this:

- **The zcash plugin's seed lock is label-blind.** `wallet/plugins/tauri-plugin-zcash/src/lib.rs` clears the in-memory seed on `WindowEvent::Focused(false)` and ignores the event's label. Harmless with one window today; wrong as soon as more surfaces exist.
- **The three app-crate enroll commands** (`f2zmsg_enrollment_status`, `f2zmsg_enroll`, `f2zmsg_unenroll`) are registered via `generate_handler!` and addressed by **no** capability entry — a test in `lib.rs` asserts this. Their ACL story must be re-derived for `e2e2z`, where enrollment becomes a bridge call rather than an in-process one.
[...]

Issue #905, opened 2026-09-04
bridge: versioned intent protocol between cash.free2z.zuuli and every other app


skyl:
Parent: #904.

**This is the keystone of the 3-app split.** Every non-wallet app is a client of it, and more apps are anticipated — `free2z` may itself split into more focused surfaces. So this must be a **reusable, versioned protocol**, not three bespoke integrations, and it must be built **before** any surface moves. Retrofitting a protocol after two apps already talk ad hoc is how you get three subtly different implementations.

## What it is

A request/response protocol carrying **authority delegation** from `cash.free2z.zuuli` (which holds the seed) to callers that hold none.

| Intent | Request carries | ZUULI does | Returns |
|---|---|---|---|
| `sign-challenge` | challenge bytes, purpose, caller identity | native confirmation | signature |
| `issue-device-credential` | device **public** keys, purpose | derives §4.2 account keys, native confirmation | `DeviceCredential` |
| `execute-payment` | recipient, amount, memo, fee | re-derives and shows its **own** payment review | txid or refusal |

The seed never appears in any message. Device private keys never leave their app. Nothing here is a continuous grant — every intent is a one-shot, user-confirmed issuance.

## Non-negotiable properties

Each already has prior art in this repo; the bridge generalises them rather than inventing:

- **Fail closed on any mismatch.** `wallet/zuuli/src/lib/wallet/creator-tip.ts` is the model — a frozen snapshot keyed by nonce, every field re-validated on the wallet side, returning `null` on any mismatch or on reload.
- **One-use, expiring, bound.** #528's native confirmation token binds proposal/recipient/amount/fee and expires on **both** monotonic and wall clocks.
- **Native confirmation is the authority.** The user approves inside ZUULI, seeing *ZUULI's* rendering of the request — never the caller's. A caller must not be able to influence what the confirmation says.
- **Unicode/homoglyph safety in the review.** #528 already renders layout controls visibly rather than letting them spoof the native review; the bridge inherits that requirement.

## The hardest open question — resolve this first

**Caller identity must be established, not asserted.** A deep link does not authenticate its sender: any app can register a custom scheme, and even a verified App Link tells you the *link* was verified, not who opened it.

- **Android:** `getCallingPackage()` on an activity-started-for-result is the honest primitive, and can be checked against a signing-certificate allowlist.
- **iOS:** weaker. Universal Links do not identify the opener. Likely needs a registered-caller allowlist plus per-caller keys established at install/enrollment time.

[...]

Issue #904, opened 2026-09-04
Architecture: split ZUULI into a wallet authority and delegated surfaces (3 apps)


skyl:
## Direction change — read this before continuing work that assumes a single app

ZUULI is today one Tauri app, one window, one WebView, holding the Zcash seed **and** rendering third-party content: article embeds, the Cloudflare RealtimeKit livestream SDK, remote images, AI output. We are splitting it into three surfaces.

### Why

- **#367** (P0, Android-confirmed): Wry injects Tauri's bridge scripts into **every** frame (`setOf("*")`, ignoring the `main_only` flag) and its Android IPC reports the *top-level* URL rather than the requesting frame. A remote subframe therefore resolves as the trusted main window and can invoke privileged commands. Reported upstream separately.
- **#816/#818** (P0): the packaged CSP blocked RealtimeKit's first request and broke Join Free. A CSP is a runtime failure mode that only exists because a privileged WebView renders third-party code.

The generalisation: **any embed policy permissive enough to be useful is too permissive to sit next to seed access.** `frame-src 'none'` works today precisely because the feature was given up.

### Why not sandbox inside one app

This was the preferred option. It does not hold where the threat is:

- [tauri#11528](https://github.com/tauri-apps/tauri/issues/11528) (open since 2024-10-29) — multiple webviews are **desktop-only, behind an unstable feature flag**. [tauri#11794](https://github.com/tauri-apps/tauri/issues/11794) (`add_child` on mobile) also open.
- Multi-*window* on mobile needs **Android 12L / API 32**; our `minSdk = 29`.
- Even at API 32, Android launches a **separate activity** onto the back stack and iPhone **replaces** the UI. The UX cost of another window on a phone already equals another app — so the single-app argument evaporates exactly where the vulnerability lives.

### The principle

**The wallet is a signing authority. Every other surface holds only delegated, scoped, revocable credentials — never the seed.**

This is already our design. `docs/e2ee/ARCHITECTURE.md` §4.2 separates **account keys** (seed-derived, restorable: `CeremonySigningKey`, `DirectoryAuthKey`, `BackupWrapKey`) from **device keys** (OS CSPRNG, *never* seed-derived, never exported: `DeviceSignatureKey`, `DeviceInitKey`, `QueueKey_{q}`), bound by a **device credential** — already implemented at `rs/crates/f2z-msg-identity/src/credential.rs`.

[...]

Issue #816, opened 2026-08-30
P0 ZUULI Live: permit least-privilege RealtimeKit endpoints in packaged CSP


skyl:
Parent: #813.

## Proven failure

ZUULI's packaged CSP allows legacy Dyte origins but blocks the bundled RealtimeKit SDK's first request to `https://api.realtime.cloudflare.com/v2/internals/participant-details`, deterministically producing founder QA's exact `ERR0001`.

## Scope

Add the narrow current RealtimeKit HTTPS/WSS origins required by the bundled SDK and prove the production CSP permits initialization without broadening unrelated network access.

## Acceptance criteria

- [ ] Add only audited RealtimeKit API/socket origins required by a real join; do not use unrestricted `connect-src`, `https:`, or `wss:`.
- [ ] A test loads the actual production CSP and bundled SDK and proves the initial Cloudflare request is permitted.
- [ ] The negative control removes the new source and reproduces the CSP violation and `ERR0001` boundary.
- [ ] Existing Free2Z API, wallet, messaging, media, and legacy-required origins remain exact.
- [ ] No participant token, meeting secret, or private claim enters logs or fixtures.
- [ ] Signed Android and iOS builds join the same active room as the web viewer; background/foreground and network transition are exercised.
- [ ] Record the exact observed WebSocket hostname(s) from credential-safe diagnostics before claiming complete coverage.

This is the immediate P0 restoration slice. Ticket validation/retry hardening remains a separate child of #813.

PR #912, opened 2026-09-04
feat(free2z): port the article content surface out of ZUULI (#904 phase 1)

skyl:
Phase 1 of #904 — **the content surface**. This was written to stack on #909; **#909 merged while it was in flight, so the branch was rebased onto `origin/main` and this PR targets `main` directly.** It also lands after #911, and the creator-tip decision below is stated against that protocol rather than around it.

`wallet/free2z` stops being a placeholder screen: Articles, the Markdown/media pipeline and Creator profiles **run in a browser and under `vite dev`**, in a project whose Tauri backend links **no privileged plugin** and holds **no `zcash:*` and no `f2zmsg:*` capability entry**.

> ⚠️ **Read "The native layer is NOT wired" below before reading anything else in this PR.** This surface is **not** a working packaged Tauri app, and this PR does not make it one. `bundle.active` is `false` so nobody can produce an installer that looks shippable. #918 is the work that earns it back.

## The native layer is NOT wired

An adversarial review confirmed the security property this PR was written for is fully intact — zero `zcash:*` and zero `f2zmsg:*` capability entries, neither privileged plugin linked, and no `invoke_handler` in `wallet/free2z/src-tauri/src/lib.rs` **at all**. Nothing below weakens that; it is the opposite problem.

What the review also found is that the frontend was ported complete, including three code paths that branch on `isTauri() && !DEV` — and the Rust side those paths call into was never written here. **No gate can see this**: vitest, Playwright and `npm run build` all run where `isTauri()` is `false` or `import.meta.env.DEV` is `true`, which is exactly the condition under which none of these branches is taken.

1. **HTTP is unwired.** `wallet/free2z/package.json` adds `@tauri-apps/plugin-http`, but `wallet/free2z/src-tauri/` was never touched: no `tauri-plugin-http` in `Cargo.toml`, no `.plugin(tauri_plugin_http::init())` in `src/lib.rs`, no `http:*` entry in either capability file. `wallet/free2z/src/lib/api/http.ts:129-132` (every API request) and `wallet/free2z/src/components/common/RemoteMedia.tsx:50-51` (every first-party image) both take the native path in a packaged build, invoking `plugin:http|fetch` — a command that does not exist in that binary.
2. **The OAuth commands are unwired.** `wallet/free2z/src/lib/oauth/transport.ts` invokes ten `oauth_*` commands that exist only in `wallet/zuuli/src-tauri/src/oauth.rs`. `nativeTransport()` rethrows rather than falling back to web, and `wallet/free2z/src/App.tsx:58` calls `recoverMobileOAuth()` unconditionally on mount behind a `.catch` that fires `toast.error("Couldn't finish sign-in")` (`wallet/free2z/src/App.tsx:82`) — so a packaged build shows a sign-in error toast on **every launch**.
3. **The deep-link scheme belongs to the wallet.** `wallet/free2z/src/lib/oauth/protocol.ts:3` uses `cash.free2z.zuuli://oauth/callback` and `wallet/free2z/src/lib/checkout/native-return.ts:3` uses `cash.free2z.zuuli://checkout/return` (guard at `:72`). Those are ZUULI's URIs; this app registers only `cash.free2z.free2z://bridge/return`. With both apps installed, an OAuth code or Stripe return initiated by the *content* app is delivered by the OS to the *wallet authority* app — the cross-app URI collision this whole split exists to prevent.

**What this PR does about it, and deliberately does not do:**

* `wallet/free2z/src-tauri/tauri.conf.json` — `bundle.active` is now **`false`**. `tauri build` cannot emit an installer for a binary whose network layer is dead. `cargo check --locked --all-targets` still passes with it (tauri-codegen validates the config at compile time), and no gate reads this key.
* `wallet/free2z/README.md` gains a **"Not yet wired natively"** section stating the three gaps with `file:line`. A silent breakage becomes a stated one.
* **#918** is the follow-up that wires it. It is deliberately not attempted here, and note *why* it is not a one-liner: `wallet/zuuli/scripts/surface-capability-authority.mjs:56-61` holds `REVIEWED_NAMESPACES = {core, deep-link, opener, f2zmsg}`, and `http` is not in it — **adding `http:default` FAILS the required `gate`**. By design. What a content surface's HTTP client may reach is a scope decision that has to be reviewed, not a default anyone can grant in passing.

## What moved

`features/articles/**` · `features/creator/**` · `Markdown` + `Mermaid` (and its worker) + `RemoteMedia` + `ImagePrivacySetting` + `lib/markdown/` + `lib/media/` · the `lib/api` surface those read (`free2z.ts`, `http`, `types`, `mock-data`, `checkout`, `donation`, `social-providers`) · `store/session` · `lib/auth`, `lib/oauth`, `lib/checkout` · the shared `components/ui` set · the i18n kernel · and a shell (`AppShell`/`TopBar`/`Sidebar`/`navigation`) written for **this** app rather than copied.

Routes: `/` → `/articles`, `/articles/*`, `/creator/:username/*`, `/fund`, `/login`, `*` → NotFound.

## What I deliberately left

* **`features/live`** — Cloudflare RealtimeKit is three heavy packages and a `connect-src`/`media-src` host list whose one-entry-short failure mode is literally #816/#818. It needs its own CSP review.
* **`features/ai`**, **`features/search`** — no coupling problem, just volume; they come next with Live.
* **`features/wallet/funding` — decided: it does NOT move in this PR.** The brief said it is "HTTP+Stripe only, needs no `zcash:*`". The card checkout is, but the *feature* is not: `funding/index.tsx`, `BalanceHero.tsx`, `SendTab.tsx` and `zec-top-up-demo.ts` all import `@/store/wallet` — exactly the thing this app's shell must never touch. Long-term it **does** belong with content (2Z is a content credential in #904's own table), but severing the ZEC top-up path from the card checkout is its own reviewable change. `/fund` is a route here today that says where 2Z top-up lives, so the two low-balance CTAs in `TipDialog`/`CommentForm` (and the TopBar 2Z chip) point somewhere real instead of at `/wallet/fund`.
* **Nothing was deleted from ZUULI.** `Markdown`, `RemoteMedia` and `ImagePrivacySetting` are still consumed by ZUULI's `features/ai`, `features/search` and `features/creator`, so removing them now would break the app. Stripping the markdown/embed dependency tree out of the wallet is **#904 phase 4** and is scheduled after every content surface has moved. ZUULI is untouched by this PR apart from one number and one comment in `project-boundary.node-test.mjs`.

## The three couplings

Prior exploration named exactly three. All three still held; all three are now asserted against the running app in `wallet/free2z/tests/surface-separation.pw.ts`.

**1. Creator ZEC tip.** `features/creator/index.tsx:59` importing `createCreatorTipRouteState` was the only social→wallet import. Replaced by `@/lib/bridge/creator-tip.ts` (deliberately **not** `lib/wallet/`, which does not exist on this surface): it applies ZUULI's byte-equivalent validation — so a creator without a usable address still fails in the renderer, not at a signer — records the frozen intent in module memory, bounded and never in web storage, and stops. **No deep link was invented.** The reader is told the tip is signed in ZUULI; the 2Z tip works here.

*Why not `createIntentSession` from `@free2z/wallet-shared`, now that #911 has landed?* Because the intent is incomplete on this screen, and the file says so at length. `encodeExecutePaymentPayload` refuses `amountZatoshis <= 0`, and **no tip dialog collects a ZEC amount** — ZUULI's `creator-tip.ts` carries only `{username, label, recipient}` for the same reason: the amount and memo are entered on the wallet's own Send review, the one surface that can show a payer what they are approving. Minting an amount here to satisfy the encoder would be inventing the number a human is supposed to choose. So this file holds the pre-request destination half only, holds it the way the bridge's session does (frozen, bounded, in memory, lost on reload), and **declares none of the five single-implementation names `project-boundary.mjs` reserves**, mints no `free2z/intent/v1/` label, and re-implements no encoder, version gate or response matcher. `project-boundary.mjs` reports **5 single-implementation intent-bridge guards** with this branch applied, unchanged from `main`.

**2. Zcash login.** Password and linked accounts work. Zcash is *absent*, not stubbed behind a dead button, and the login screen says why. Per #905's correction: signing a challenge is an attestation, not a spend, which is exactly why it looks like the easy first grant — and it is the **weakest** family for native confirmation, because a nonce confirms nothing to the person approving it. A tip names an amount and a recipient a human can read. So it is deliberately not first.

**3. The shell.** `TopBar` has no ZEC chip (the 2Z chip stays — fiat-funded content credit, not a key), `AppShell` renders no `LegacyWalletNotice`, `App` bootstraps a session and nothing else, and `features/auth` no longer calls `useWallet().bootstrap()`. `/wallet/send` renders NotFound.

## The one structural change: free2z now ships a `vite.config.ts`

#906 shipped this app with no Vite config on purpose, and the README explained why. The content extraction ends that, mechanically rather than stylistically — I proved both before deciding:

* `worker.format` defaults to `"iife"` and the Mermaid worker dynamically imports `mermaid`. Rollup refuses to code-split an IIFE: `Invalid value "iife" for option "worker.format"`. The build fails outright.
* `mathjax-full` (via `rehype-mathjax`) reads its version through `eval('require')(...)` unless a global `PACKAGE_VERSION` exists — `require is not defined` the first time an article renders math.

There is a way to dodge this — name the config something `VITE_CONFIG_NAMES` does not match and pass `--config`. That is evasion of a security control, so it is not what I did. Instead the config is real, `project-boundary.mjs` now runs its **real production Rollup build** inside the constrained audit sandbox (**3** verified builds, up from 2), and the protected `frontend` job installs this project's dependencies so that audit can run. That required adding `npm ci --prefix ../free2z` and the lockfile cache path to `.github/workflows/zuuli.yml`, and updating `REQUIRED_FRONTEND_JOB_LINES` in `scripts/check-github-actions-pins.mjs`, which byte-pins that step list. **Both edits are additive; no branch, guard or pattern was removed or weakened**, and the pins checker's own 18 frontend-build mutation cases still pass.

## CSP

Opened `img-src`/`media-src` to `https:` and added the `free2z.com` media host — `RemoteMedia` now renders creator images, audio and video here, every one behind one-item destination consent (`tests/remote-media-consent.pw.ts`, 13 cases).

`frame-src` stays `'none'`. `Markdown` keeps ZUULI's `isTauri()` branch, so a packaged build still opens a YouTube/Vimeo embed in the OS browser. Flipping that is the reviewed product change this app exists to make, with its own test surface — `frame-src` opens with it, not as a side effect of moving a file. RealtimeKit's hosts arrive with `features/live`.

## Non-negotiables

* **No `zcash:*`, no `f2zmsg:*`.** `surface-capability-authority.mjs` green, all **12** mutation controls green. No capability file was touched.
* **`project-boundary.mjs` green, and proven to fail.** Both controls run at real conditions: a relative import of `../../zuuli/src/lib/wallet/creator-tip` → `relative import escapes wallet/free2z`, **exit 1**; a package import `zuuli/src/lib/wallet/creator-tip` → `free2z imports the zuuli application as a package`. Green with them removed.
* **i18n kernel invariants.** New `src/i18n/source-policy.test.ts` enforces both mechanically instead of by description — locale JSON reachable only through `CATALOG_LOADERS` (with the `test-provider.tsx` exemption ZUULI also carries), and every `MESSAGE_KEYS` entry read by real application code. Both mutation controls verified red first: an orphan `navigation.orphan` key and an eager `import en from "@/i18n/locales/en.json"` in `EmptyState.tsx` each fail; green when reverted. `en/es/fr.json` are key-identical (29 keys, checked programmatically).
* **`ui-copy-truncation.node-test.mjs`** now ships in `wallet/free2z/scripts/` and runs in `npm test`. Its chrome-scan anchor moved from ZUULI's Search surface to this app's article `Feed.tsx` — the anchor's purpose (prove the scan reached a real surface rather than passing on an empty selection) is unchanged.
* **Playwright ran, not just vitest.**
* **`lib/bridge/creator-tip.ts` is now tested.** It was ported carrying security-relevant validation and shipped with no unit test at all, while ZUULI's counterpart has `wallet/zuuli/src/lib/wallet/creator-tip.test.ts`. `wallet/free2z/src/lib/bridge/creator-tip.test.ts` ports that suite to this surface's API (no route state, no nonce — there is no `/wallet/send` to hand anything to) and covers what remains: the 150/128/255 **code-point** bounds at the boundary and one past it, trim-equality, control-character and DEL rejection, no-whitespace recipient, non-string/null/undefined fields, `Object.freeze` on both the intent and the `pendingCreatorTipIntents()` copy, and the 32-entry cap. **22 tests, proven load-bearing:** ten separate mutations of `creator-tip.ts` were each verified red before restoring — `150→200`, `128→512`, `255→1024`, dropping `value === value.trim()`, dropping the control-character regex, dropping `!/\s/u.test(recipient)`, `MAX_RECORDED_INTENTS 32→64`, dropping either `Object.freeze`, and counting `value.length` instead of `Array.from(value).length`.

## Test results — all foreground, from each app dir after `npm ci`

> **Caveat, stated because it is the whole point of the section above.** Every measurement in these tables was taken under `isTauri() === false` or `import.meta.env.DEV === true`. That is precisely the condition in which the three native branches (`lib/api/http.ts:129`, `RemoteMedia.tsx:50`, the `oauth_*` invokes) are **not** taken, so none of these numbers says anything about whether they work. `cargo check --locked --all-targets` compiles a backend that is never asked whether the frontend expects commands from it — it proves nothing about the JS→plugin contract. There is currently **no** check in this repo that would catch a `@tauri-apps/plugin-*` dependency with no matching crate, or an `invoke("…")` naming a command the binary does not register; adding one is an acceptance item on #918. These tables are green and honest about what they cover, which is the browser/dev surface only.

`wallet/free2z`

| check | result |
|---|---|
| `npm run typecheck` / `typecheck:tests` | clean |
| `npx vitest run` | **699 passed**, 5 skipped, 51 files |
| `node scripts/ui-copy-truncation.node-test.mjs` | **4 passed** |
| `npx playwright test` | **41 passed** |
| `npm run build` | ok (Mermaid worker + katex + mathjax all emitted) |
| `npm audit --audit-level=high` | exit 0 (4 moderate, same baseline as ZUULI) |
| `cargo check --locked --all-targets` | ok, with `bundle.active: false` |

`wallet/zuuli`

| check | result |
|---|---|
| `node scripts/project-boundary.mjs` | 5 projects, 513 files, 2389 refs, 5 intent-bridge guards, **3 constrained production Vite builds** |
| `node --test scripts/project-boundary.node-test.mjs` | **92 passed** |
| `node scripts/surface-capability-authority.mjs` | green |
| `node --test scripts/surface-capability-authority.node-test.mjs` | **12 passed** |

Policy checkers, `--self-test` **and** live: `check-github-actions-pins` (74 source-policy / 11 gate-verdict / 148 current-workflow / 3 project-discovery / 18 frontend-build / 4 classic-frontend-audit / 124 Rust-root cases), `check-workflow-gates` (39), `check-hash-domain-labels` (17), `check-f2z-images` (19). All pass.

## Specs that moved with their features

`article-drafts` · `article-tags` · `mermaid-security` · `remote-media-consent` · `first-party-images` · `creator-pagination` · `creator-responsive`, plus the new `surface-separation`. `creator-zec-tip.pw.ts` did **not** move — it drives ZUULI's wallet Send review, which is not this app's behaviour. Four specs needed a route the shell no longer carries (Home → Articles, Wallet → Log in, Search → the tag-filtered Feed, and the Watch-live action → a non-navigating live marker, since `/live` is unmounted and a 404 is worse than no destination); each edit carries a comment saying what ZUULI asserts and why this surface differs.

Pushed once on the pre-rebase branch, then force-pushed once after rebasing onto `main` (#909 merged mid-flight), then one follow-up commit after adversarial review: `bundle.active: false`, the README's "Not yet wired natively" section, and `lib/bridge/creator-tip.test.ts`. Then rebased again onto `main` after **#913** merged, with one hand-resolved conflict in `.github/workflows/wallet-surfaces.yml`: both PRs add a `google-chrome --version` step at the same anchor in a matrix job that runs for *both* apps, so the resolution keeps **#913's step name and `run:` byte-identically** and folds free2z's reason into the shared comment rather than adding a duplicate step. The diff of that file against `main` is now comment-only. `main`'s `#913` paths (`wallet/e2e2z/**`, `docs/e2ee/**`, `scripts/check-workflow-gates.mjs`) are untouched by this branch, and the changed-file list is byte-identical before and after the rebase (183 files, 4 of them outside `wallet/free2z/`). CI runs on the final state only; no `gh run rerun`.


skyl:
## Adversarial review — DO-NOT-MERGE (narrow), security property intact

**The load-bearing property holds.** Verified by parsing the capability JSON (not grepping — a `description` field can name a permission it does not grant):

- `wallet/free2z/src-tauri/capabilities/{default,mobile}.json` grant exactly `core:default`, `deep-link:default`, `opener:default`. **Zero `zcash:*`, zero `f2zmsg:*`.**
- Stronger than the capability layer: `src-tauri/Cargo.toml` links neither wallet plugin, and `src-tauri/src/lib.rs:12-18` registers **no `invoke_handler` at all** — there is no command for a confused subframe to reach. That is the #367 property, structurally.
- CSP deltas (`img-src`/`media-src` +`https:`, `connect-src` +free2z hosts) all sit inside ZUULI's existing envelope. `frame-src` stays `'none'`; `script-src` is *stricter* than ZUULI's.
- The "real `vite.config.ts`, not a rename to dodge the scanner" claim is **true** — confirmed against `VITE_CONFIG_NAMES`, and the 2→3 build count is its mechanical consequence.

### Blocker: three native paths are silently dead, and the evidence can't see them

**F1 (critical).** `@tauri-apps/plugin-http` is added to `package.json`, but `src-tauri/` is untouched: no `tauri-plugin-http` in `Cargo.toml`, no registration, no `http:*` capability. `lib/api/http.ts:129-132` and `RemoteMedia.tsx:50-51` both take the native branch when `isTauri() && !isDev`, so a packaged build invokes `plugin:http|fetch`, which does not exist in that binary. Articles, profiles, login, media — all dead.

**Every green row in the test table was measured where this bug cannot appear** (`isTauri()` false or `DEV` true). `cargo check --all-targets` proves nothing about a JS→plugin contract.

**F2 (high).** `lib/oauth/protocol.ts:3` and `lib/checkout/native-return.ts:3` hand out `cash.free2z.zuuli://` URIs. With both apps installed, an OAuth code or Stripe return initiated by the *content* app is delivered by the OS to the *wallet-authority* app — the exact cross-app URI collision this split exists to prevent.

**F3 (high).** `lib/oauth/transport.ts` invokes ten `oauth_*` commands that live only in ZUULI's crate; `nativeTransport()` rethrows rather than falling back to web, and `App.tsx:58` calls `recoverMobileOAuth()` on every mount — so a packaged build shows an error toast on launch.

**F4.** Ported security-relevant validation in `lib/bridge/creator-tip.ts` ships with no unit test, though ZUULI has one.

**F5.** ~35.9k lines are **copied, not moved** — nothing was deleted from `wallet/zuuli/src/`. Disclosed and deferred to phase 4, but the drift window is now open and unguarded.

### Fix in flight

Narrow fix being applied to this branch: bundle `"active": false`, a "Not yet wired natively" section in the README naming all three gaps with `file:line`, a caveat on the test tables, and the missing `creator-tip.test.ts`. The real native wiring gets its own PR — note that `surface-capability-authority.mjs:56-61` `REVIEWED_NAMESPACES` excludes `http`, so wiring it requires a deliberate scope decision, not a one-line capability addition. That is by design.

Gate coverage gap filed as #915.

Issue #461, opened 2026-08-23
zuuli: replace custom-scheme checkout return with verified Universal Links / App Links


skyl:
Split out of #406, which shipped the native checkout return path. Sub-task of #388.

## Why

Checkout returns to the app today over the private-use scheme
`cash.free2z.zuuli://checkout/return`. That scheme is registered by ZUULI but,
like any custom scheme, it is **not uniquely owned** — another installed app can
register the same string.

The credit path is already defended against that: the return is not a bearer of
value. Delivery only produces a short-lived code, and the actual claim is an
authenticated, user-bound, bounded server call, so an app that intercepted the
scheme could not steal a credit. The residual risk is **delivery denial** — an
interceptor can prevent the return from reaching ZUULI, leaving a paying user
stranded on a screen that never resolves.

Verified Universal Links (iOS) and App Links (Android) remove that, because the
OS binds the HTTPS URL to a specific app via a document the domain owner serves.

## Scope

- Move the return target from the custom scheme to an HTTPS URL under a domain
  we control, keeping the existing scheme registered as a fallback during
  migration so older installs keep working.
- Decide which host the app claims, and make sure the association document is
  served on **that exact host** — this is the easiest thing to get subtly wrong,
  and it fails silently: the OS simply declines to open the app and the link
  loads as a web page with no error anywhere.
- Requires an Apple Associated Domains entitlement on the app, and the
  production Play App Signing SHA-256 for Android. Neither is in this repo.
- Keep the existing bounded-claim behavior unchanged — this narrows how the
  return is *delivered*, not what it authorizes.

## Testing note

Association is fetched by the OS and cached, and simulator/emulator behavior
differs from device behavior. Plan for real-device verification on both
platforms, and for a cold-start case (app not running) as well as warm-start —
`#406`'s parser already handles both, so this is about delivery, not parsing.

## Not blocking anything today

Native return is gated off server-side, and `#406` made the client fall back to
web-mode checkout when it is unavailable, so nothing is broken while this is
open and no client release is needed to turn the native path on later.

Related: #388, #406.

skyl:
## Promoted: now a hard prerequisite for #904 Phase 2

This was scoped as a checkout-return correctness issue. Under the 3-app split (#904) it becomes **a security prerequisite for the whole cross-app protocol**.

After the split, these all cross an app boundary over deep links:

- **ZEC tip** — a payment intent handed from free2z to ZUULI for native confirmation and execution
[...]

PR #913, opened 2026-09-04
feat(e2e2z): port the E2EE messaging surface out of ZUULI (#904 phase 3)

skyl:
Phase 3 of #904. Part of #305's implementation track.

**Rebased onto `main` (2026-09-04).** #909 squash-merged while this was in
flight, so the branch was replayed with `git rebase --onto origin/main 82faf66c`
and is now **one commit** on top of `main`. Before: 100 files / +23455,
`CONFLICTING`/`DIRTY` — it was re-proposing the whole already-merged scaffold.
After: **70 files / +3468 / -391, 1 commit, `MERGEABLE`**. The diff touches no
`wallet/free2z/src-tauri/**`, no icon set, no `tauri.conf.json`, and nothing
under `wallet/shared/src/intent/` or `rs/crates/f2z-intent/` — main's merged
intent bridge (#911) is untouched. Full verification was re-run after the
rebase, including both Playwright suites.

## Enrollment: answer in one line

**(a) — enrollment stays in ZUULI, and e2e2z fails closed.** It is not ported,
not stubbed, and e2e2z gets no seed access.

- `f2zmsg_enrollment_status` / `f2zmsg_enroll` / `f2zmsg_unenroll` remain
  app-crate commands in `wallet/zuuli/src-tauri/src/messaging.rs`. `account_keys()`
  is untouched and still the only seed-touching path for chat.
- e2e2z links no `tauri-plugin-zcash`, registers none of the three, and holds no
  capability addressing them. Its crate test asserts the manifest link is absent.
- Its `bridge.ts` keeps the three in the declared command population and
  **refuses** each with a typed `EnrollmentUnavailableError`
  (`reason: "enrollment-requires-wallet-app"`), without reaching Tauri IPC.
- It **never synthesizes an `EnrollmentStatus`.** There is no successful-looking
  enrollment anywhere: the screen renders "Enrollment happens in the wallet app"
  with no claim control, no conversation list, no engine start/stop and no
  "Handle active".
- No cross-app intent was invented — #905's `issue-device-credential` is the way
  out and does not ship before #461 provides an authenticated channel.

The one nuance worth stating: `VITE_MOCK=1` still serves the pre-existing fixture
data layer, as it does for every ZUULI bridge. It is a build-time opt-in that no
packaged build sets, so no shipped e2e2z can reach a mocked enrollment. The
Playwright enrollment-gap spec runs against the **default** build precisely so
the shipped path is what gets asserted.

**Negative control, run:** replacing the refusal with a fabricated
`enrolled: true` turns 3 vitest cases and the Playwright case red. Reverted.

### The gap is repo-wide, not e2e2z-only — state it plainly

On `main` today ZUULI has a **working** claim control:
`wallet/zuuli/src/features/messages/index.tsx` calls
`enrollment.enroll(candidate)`. This PR moves that screen out. Afterwards:

- **e2e2z** refuses (above), and
- **ZUULI** still registers `f2zmsg_enroll` in Rust but has **no frontend caller
  anywhere**.

So **no user can claim a messaging handle in any shipped build** until #905 and
#461 land. That is a real, deliberate regression in reachable functionality, and
it is the honest consequence of the #904 invariant: the surface that runs
conversations must not hold the seed, and the authority that holds the seed has
no channel to hand out a credential yet. Nothing in messaging worked
end-to-end before this PR either — there is no relay, no directory log and no
witness set in a shipping build — so what is lost is a control that could be
clicked, not a capability that functioned.

The in-app copy says so: "Claiming a handle, and issuing this device its
credential, is done in the wallet app. **That handover is not built yet, so
nothing here is enrolled and no conversation can start.**"

## What moved

`wallet/zuuli/src/` → `wallet/e2e2z/src/`, as `git mv` (rename-detected):

- `features/messages/` — the screen, `FirstContact`, `Transcript`,
  `BrowserGuarantee` (§11.1), and their tests.
- `lib/messaging/` — `bridge.ts`, `events.ts`, `types.ts`, `mock.ts`, the
  handle-eligibility fixtures added by #864 and their tests.

Prior exploration held: `features/messages` imports **no** store, **no** wallet
library and **no** session. Re-verified — the only non-messaging imports were
`components/ui/{button,callout,input,skeleton}` and
`components/common/PageHeader`. Those five, plus `lib/utils.ts`, the Tailwind
token layer and the bundled Plex faces, are **copied**: `project-boundary.mjs`
forbids one wallet project reaching into another's tree, and a shared build
config would be exactly that. `button.tsx` loses its `zec` variant — e2e2z
never denominates money.

`wallet/e2e2z` still has **no `vite.config.ts`**, deliberately (#909's README
section, and `viteBuildsVerified` in `project-boundary.node-test.mjs`), so the
moved files use relative imports rather than `@/`. The boundary run still
reports **2** constrained production Vite builds, unchanged.

## The enrollment gap, and how it fails closed

`f2zmsg_enrollment_status` / `f2zmsg_enroll` / `f2zmsg_unenroll` are app-crate
commands in `wallet/zuuli/src-tauri/src/messaging.rs`; `f2zmsg_enroll` calls
`account_keys()`, which reads the wallet seed in process and derives
`ARCHITECTURE.md` §4.2's account keys. **They did not move and cannot.** §4.2 is
what makes the split workable: ongoing messaging needs only device keys (OS
CSPRNG, never seed-derived, never exported) plus a `DeviceCredential`.

e2e2z's `bridge.ts`:

- **keeps all three in its declared population.** `WIRE_COMMANDS`, `RESULTS` and
  `BridgeMethod` are one population that `parity.test.ts` and
  `messaging-contract.node-test.mjs` hold to §3 of `CLIENT-CONTRACT.md`.
  Deleting them would silently shrink the contract rather than record what this
  app cannot do.
- **refuses every call** with a typed `EnrollmentUnavailableError` carrying
  `reason: "enrollment-requires-wallet-app"` and the refused method. It never
  reaches `@tauri-apps/api/core` — letting Tauri answer "command not found"
  would turn a designed boundary into what reads like a packaging bug.
- **synthesizes nothing.** Every field of an `EnrollmentStatus` is a claim about
  the KT directory; a fabricated `enrolled: true` would render a handle nobody
  published and offer first contact that cannot complete.

The screen renders the refusal as a standing state: page header, device notice
and engine summary (the plugin half genuinely works), then "Enrollment happens
in the wallet app" — and **no** claim control, **no** conversation list, **no**
engine start/stop, no "Handle active", no first-contact form. Only the typed
refusal takes that path; any other rejection keeps propagating, so a broken
plugin cannot hide behind a tidy screen that says the wrong thing.

`VITE_MOCK=1` still serves fixtures, as it does for every ZUULI bridge. It is a
build-time opt-in never set in a packaged build, so no shipped e2e2z can reach a
mocked enrollment.

No deep link was invented. #905's `issue-device-credential` intent is the way
out and **does not ship before #461**.

**ACL story for the trio in e2e2z:** unchanged in kind and stricter in fact.
They are app-crate commands, so no capability entry addresses them anywhere —
ZUULI's `lib.rs` assertion that no `f2zmsg-enroll*` identifier appears in either
capability file is untouched and still green (`cargo test --lib
shipping_capabilities` passes). In e2e2z they are additionally not *registered*:
`src-tauri/src/lib.rs` has no `invoke_handler` for them and asserts the manifest
links no `tauri-plugin-zcash`. When #905 lands, the credential arrives over the
intent channel and is installed by the plugin — so the correct end state is
still **no capability entry and no app-crate enroll command in e2e2z**; what
changes is that the refusal is replaced by a call into the wallet authority.

Proved three ways:
- `src/lib/messaging/enrollment-gap.test.ts` — refuses, never invokes IPC (spied
  host), never resolves, keeps the population, recognizes a prototype-less copy.
- `src/features/messages/index.enrollment-gap.test.tsx` — the screen renders the
  gap, offers nothing enrolled, never lists conversations, and does not convert
  an unrelated failure into the gap.
- `tests/enrollment-gap.pw.ts` — a real browser against the **default** build
  with a Tauri IPC host shaped exactly like `src-tauri/src/lib.rs`: the plugin
  answers, the trio does not exist. Asserts the gap renders **and** that no
  `f2zmsg_*` command was ever invoked.

**Negative control, run:** replacing the refusal with a fabricated
`enrolled: true` turns 3 vitest cases and the Playwright case red. Restored.

## `docs/e2ee/WIRE.md` and its pinned prose digest — read this

`rs/crates/f2z-msg-mls/tests/key_packages.rs` pins
`WIRE_RENDERED_PROSE_DIGEST` over **the whole** of WIRE.md's rendered
paragraphs, deliberately, so that any prose drift is reviewed before the pin
moves. This PR trips it. **The change is intentional and the digest is updated
with a stated reason — here it is, so it can be audited from this body alone.**

The complete WIRE.md delta is in §14.1 and is entirely about *where two files
live*:

1. The link to `mock.ts` — `wallet/zuuli/…` → `wallet/e2e2z/…`.
2. The link to `handle-eligibility.fixtures.json` — same move. This one is
   load-bearing: `tauri-plugin-f2zmsg`'s `handle.rs` reads that exact file with
   `include_str!`, and this PR repoints that too. A stale path here would leave
   the normative document naming a file that does not exist.
3. One sentence rewritten: "kept inside `wallet/zuuli` rather than under `docs/`
   so `mock.ts`'s side of it is an ordinary same-project JSON import, not a
   reach outside **the wallet's** own module boundary" → "kept beside `mock.ts`
   … not a reach outside **its own** module boundary, which is also why it moved
   with the messaging surface out of `wallet/zuuli` in #904 phase 3."
   **The reasoning is unchanged; only the project it names moved.**

**No normative claim changed.** §12.6 — the key-package authentication table,
the three claims the test mutates, `last_resort`, exhaustion, privacy,
anti-abuse — is byte-identical, which the same test's structural and
mutation-sensitivity assertions confirm independently of the digest. The new pin
carries a comment recording exactly the above, so the next reader sees why it
moved. `cargo test --locked -p f2z-msg-mls` → **26 passed, 0 failed**; the whole
`rs` workspace (`cargo test --locked --all-targets` + `--doc`) is green.

## ZUULI after the removal

- `lib/routes.ts`, `App.tsx` (lazy import + `<Route>`), `navigation.ts` (entry +
  `MessagesSquare`) — the messages route and nav entry are gone. Nothing else in
  the shell is restructured.
- `i18n/messages.ts` and `en/es/fr.json` drop `navigation.messages` and
  `navigation.messagesAccessible`. The three catalogs stay key-identical, and
  `source-policy.ts`'s "every message key has an executable consumer" rule is
  why the keys had to go with the entry.
- `navigation.test.ts`, `Sidebar.i18n.test.tsx`, `tests/navigation.pw.ts`
  updated to the new census.
- `scripts/rtl-source-policy.mjs`: `CHAT_OWNED_RESIDUALS` is now **empty**, which
  makes ZUULI's policy *stricter* — any residual, in any file, now fails. The
  mechanism is kept, not deleted, for the next surface that earns a counted
  exception. Its node-test's neutral mutation carrier moved from
  `messages/BrowserGuarantee.tsx` to `wallet/Send.tsx`, and the three
  messaging-specific cases became emptiness cases. 17/17 still pass.
  **Known follow-up:** RTL enforcement is *not* ported to e2e2z — that app has
  no document-direction bootstrap or `.bidi-number` layer to hold to, and
  building one is its own change rather than a silent half-measure here.
- `src-tauri/src/messaging.rs`: the `include_str!` wire-name pin now reads
  `../../../e2e2z/src/lib/messaging/bridge.ts`. The pin now spans two packages,
  which is more reason for it, not less.

## Cross-tree pins repointed (all previously read ZUULI's copy)

- `wallet/zuuli/scripts/messaging-contract.node-test.mjs` — stays in
  `wallet/zuuli/scripts` **on purpose**, because that is where the protected
  `gate` runs it; it now reads the client out of `wallet/e2e2z`. 9/9 pass.
- `wallet/plugins/tauri-plugin-f2zmsg/src/handle.rs` — the shared Rust↔TS
  handle-eligibility fixture table (`WIRE.md` §14.1).
- `scripts/check-pow-policy.mjs` — the shipping-UI copy check on
  `FirstContact.tsx`.
- Doc/comment references in `docs/e2ee/{CLIENT-CONTRACT,WIRE}.md`,
  `rs/crates/f2z-msg-dag/*` and the plugin's modules.
- `CLIENT-CONTRACT.md` §2.4 carries a "superseded" note stating the new shape and
  the enrollment gap; the original paragraph is kept as the record.

## New in e2e2z

`tailwind.config.cjs` + `postcss.config.cjs` + a trimmed `index.css` (same token
names and values as ZUULI, only the layers this app renders), `fonts.css`,
`playwright.config.ts` (two dev servers — the fixture layer and the default
build, because this app has two truthful states), `tests/*.pw.ts`, and
`scripts/ui-copy-truncation.node-test.mjs` (the two copy/layout rules; ZUULI's
third test is about Search and stays with it). `App.tsx` renders the surface
directly — one surface, one screen, no router — and the placeholder i18n keys
are replaced by the two the shell actually uses. `.gitignore` gains
`playwright-report` / `test-results`.

`.github/workflows/wallet-surfaces.yml` gains a `google-chrome --version` step,
matching `zuuli.yml`: both Playwright configs select `channel: "chrome"` under
CI, so the runner image's stable Chrome is used and no browser is downloaded.

`scripts/check-workflow-gates.mjs`'s ungated registration for that workflow said
"What remains here is surface-only build coverage, which cannot fail a merge and
is not claimed to." **That justification is no longer true** — as of this PR the
e2e2z lane carries ~162 vitest cases and 4 Playwright cases, including
`enrollment-gap.pw.ts`, the most security-relevant test here, and none of them
can block a merge. The registration text is corrected to say so and to point at
#915, which decides whether the suite moves into the required gate. It is
**not** moved here. The security *property* those tests illustrate is separately
gated: no `zcash:*` for e2e2z and no cross-application import are both decided
by `surface-capability-authority.mjs` and `project-boundary.mjs` inside `gate`.

## Verification (all foreground, all green)

**e2e2z** (`npm ci`): `npm run typecheck` ✓ · `npm run typecheck:tests` ✓ ·
`npx vitest run` — **162 passed / 10 files** ✓ ·
`node --test scripts/ui-copy-truncation.node-test.mjs` — 2/2 ✓ ·
`npx playwright test` — **4 passed** ✓ · `npm run build` ✓

**ZUULI** (`npm ci`): `npm run typecheck` ✓ · `npm run typecheck:tests` ✓ ·
`npx vitest run` — **1366 passed, 5 skipped / 94 files** ✓ ·
all 21 `node --test` script suites — **564 passed, 0 failed** ✓ ·
`npx playwright test` — **185 passed** ✓

> The one Playwright failure seen mid-work (`about.pw.ts`, "Source commit:
> Unavailable in this build") was a dirty-worktree build-identity artifact, not
> a regression: it passes on the committed tree, and the full 185-test rerun
> after committing is clean.

**Policy checkers:**
- `project-boundary.mjs` — 5 projects, 360 files, 1748 references, 2 Vite builds ✓.
  **Both negative controls run:** a relative import out of `wallet/e2e2z` into
  `zuuli/src/lib/wallet/bridge` → `relative import escapes wallet/e2e2z`, exit 1;
  the package form → `e2e2z imports the zuuli application as a package`, exit 1;
  green with the control removed.
- `surface-capability-authority.mjs` ✓ — **negative control run:** adding
  `zcash:allow-get-balance` to `wallet/e2e2z/.../default.json` fails with
  `wallet/e2e2z must never grant zcash:*`, exit 1; green when reverted.
- `check-tauri-plugin-permissions.mjs` `--self-test` + live ✓ (zcash 39/78/3,
  f2zmsg 43/86/4)
- The four gate checkers, `--self-test` **and** live: `check-github-actions-pins`
  ✓ · `check-workflow-gates` ✓ · `check-librustzcash-compat` ✓ ·
  `check-pow-policy` ✓
- Also re-run because this PR touches what they read: `check-akd-doc-evidence` ✓ ·
  `check-hash-domain-labels` ✓ · `check-kt-sth-repeat-agreement` ✓ ·
  `check-zuuli-android-gate` (`--self-test` + live) ✓ · `rtl-source-policy` ✓ ·
  `apple-credential-boundary` ✓ · `macos-keychain-entitlements` ✓

**Rust** (pinned 1.97.1): `cargo check --locked --all-targets` for
`wallet/e2e2z/src-tauri`, `wallet/zuuli/src-tauri` and
`wallet/plugins/tauri-plugin-f2zmsg` ✓ · `cargo clippy --locked --all-targets --
-D warnings` for all three ✓ · `check-rust-fmt.sh --root wallet` — 8 crates ✓ ·
`cargo test`: e2e2z crate ✓, `f2zmsg handle::` 8/8 (including the moved fixture
table) ✓, `zuuli messaging::` 2/2 ✓, `zuuli tests::shipping_capabilities…` ✓,
`f2z-msg-dag` ✓

## Not done here, on purpose

- **No merge.** Stacked on #909.
- **No transport for #905.** Blocked on #461.
- **No `zcash:*` for e2e2z.** Ever.
- **RTL policy for e2e2z** — see above.


skyl:
## Adversarial review — architecture sound; DO-NOT-MERGE on the rebase only

**The central invariant holds: no seed-derived material followed the port.** Confirmed by reading, not trust:

- `wallet/e2e2z/src-tauri/Cargo.toml` links no `tauri-plugin-zcash`; `Cargo.lock` has **zero** `zcash*`, `zip32`, `bip0039` or mnemonic crates.
- `src-tauri/src/lib.rs:18-24` registers **no `invoke_handler` at all** — the enroll trio is not merely un-ACL'd, it does not exist in the process. `manifest_links_no_wallet_plugin` asserts this at the manifest level.
- `src/lib/messaging/bridge.ts:641-661` — all three enrollment methods throw `EnrollmentUnavailableError` **before** reaching `@tauri-apps/api/core`.
- `account_keys()` stays put at `wallet/zuuli/src-tauri/src/messaging.rs:218-241`, still the only seed-touching path.

So the answer is **enrollment left in ZUULI + a typed hard refusal in e2e2z** — not a stub, not a mock, not a success-shaped no-op. Capabilities parse clean: 45 permissions, zero `zcash:*`, no blanket `f2zmsg:default`, `set_relay_trust` deliberately absent.

**The refusal is proved at the real conditions.** `tests/enrollment-gap.pw.ts` runs against a second dev server with `VITE_MOCK=""` and a Tauri IPC host that answers only `plugin:f2zmsg|*` and throws for everything else, then asserts no enroll command was ever invoked. The gap branch renders no claim control and no conversation list. That is the right bar.

Also correctly reasoned: in Tauri v2, `generate_handler!` commands are **not** subject to capability ACL, so ZUULI's existing "no capability names them" test was never an unreachability proof — e2e2z's story (no handler at all) is strictly stronger.

### Blocker (process, not design)

**F1.** Branch merge-base predates both #909 (`bd39b14`) and #911 (`51bfe6f`), so it is `CONFLICTING` and **CI has never run on it**. Every "all green" number was measured on a tree that is not what would land.

**Rebase trap, flagged to the author:** the diff shows #911's intent-bridge guards being *deleted* from `wallet/zuuli/scripts/project-boundary.node-test.mjs` (including `sharedIntentGuardCount === 5`). That is a stale-base artifact — commit `520bf15` does not touch that file — but it *is* in the conflict set, so a careless resolution would silently revert #911's enforcement. Take main's version verbatim.

**F2.** The moved messaging tests left the required gate — `enrollment-gap.pw.ts` could go red without blocking a merge. Filed as #915; the justification text in `check-workflow-gates.mjs` is now false and is being corrected here.

**F3.** The enrollment gap is broader than the body states: ZUULI still registers `f2zmsg_enroll` but now has **no frontend caller anywhere**, so no user can claim a handle in *any* shipped build until #905/#461.

**F4 → #916** (ZUULI keeps blanket `f2zmsg:default` it can no longer use). **F5 → #917** (RTL/bidi policy not ported).

One false claim in the body: "Merge #909 first" — #909 is already merged.

PR #914, opened 2026-09-04
feat(intent-bridge): the ZUULI authority side — execute-payment, natively confirmed

skyl:
Implements the **wallet half** of **#905** — the side that *receives* an intent
inside `cash.free2z.zuuli` and turns it into a native confirmation. Parent
**#904**. Blocked-on tracker: **#461**.

## Rebased onto main — #911 has landed

**#911** was squash-merged as `51bfe6f`, so this branch's copy of that work was
no longer an ancestor of `main` and the PR re-proposed 40 already-merged files.
Rebased with `git rebase --onto origin/main f6f6ed42`, which replays only this
branch's own commit and drops the duplicate. No conflict surfaced and nothing was
hand-resolved: `git diff f6f6ed42 origin/main -- rs/crates/f2z-intent
docs/intent-bridge wallet/shared/src/intent` was empty, so main's merged copy of
#911 is byte-identical to the base this was written against.

**Before:** 40 files, +8258/-23, `CONFLICTING`/`DIRTY`.
**After:** 11 files, +1729/-23, one commit, `MERGEABLE`.

## What it is

`wallet/zuuli/src-tauri/src/intent.rs` — one function, `receive_intent(app,
bytes, attestation)`. It calls `f2z_intent::IntentGate::admit` **once** and
re-implements no guard, then routes an approval through the wallet's existing
path:

```
IntentGate::admit → execute-payment only → propose_send
  → prepare_send_confirmation → review_matches_request
  → native dialog → ConfirmationAuthorization::issue + issue_send_confirmation
  → consume (takes self) → take_send_proposal → ensure_active_seed_loaded → execute_send
```

`docs/intent-bridge/AUTHORITY.md` is the design note, including the parts that
are weaker than they look.

## `execute-payment` only, and that is the point

**#905**'s correction comment, followed literally:

> Strength runs: **strong for `execute-payment`, medium for
> `issue-device-credential`, weak for `sign-challenge`.** … shipping
> `sign-challenge` first because it "only signs" is backwards.

`sign-challenge` and `issue-device-credential` are **admitted by the gate and
refused by this surface** with `INTENT_UNKNOWN_INTENT`, which is not a lie —
this build does not implement them. There is no signer here to gate; the family
restriction is enforced by the absence of code, and §8 of AUTHORITY.md says so
rather than claiming a two-sided proof it does not have.

## What it deliberately does NOT ship

**No transport.** No deep-link parsing, no URL handling, no intent filter, no
scheme registration. `receive_intent` takes bytes somebody else decided to hand
it. #461 is unresolved and a custom scheme would recreate #367 at the OS layer.

**No IPC surface.** `receive_intent` is not a `#[tauri::command]` and appears in
no capability — the privileged WebView must not be able to mint an intent, which
is exactly #367. A test parses `lib.rs`'s `generate_handler!` block and fails if
anything from the module reaches it, with a positive control so an empty parse
cannot pass vacuously.

## The confirmation

Every value on screen is re-derived by `tauri-plugin-zcash` from its own native
proposal. The caller contributes exactly one string — `purpose` — and it is
quoted and escaped with **#528's own** `quote_native_memo`, so a layout control
renders visibly rather than adding a `Total:` line. The requesting app's **name
comes from our registry**, never from the request's `caller` field; `#911`'s
`CallerTrust` has no `Claimed` → `Attested` path and this preserves it, rendering
"Identity: NOT CONFIRMED" for every platform that cannot prove otherwise.

`signing_certs` is empty for both registered callers, which is **fail-closed**:
an Android attestation that matches no registered certificate is a refusal, not a
downgrade. It stays that way until #905's first amended acceptance criterion —
does `startActivityForResult` compose with a verified App Link? — is measured.

Reuse, not a second confirmation: `native_review_lines` is factored out of
`format_native_send_confirmation` so the payment block is literally the same
lines the wallet's own send flow shows.

## One protocol addition: `INTENT_UNAVAILABLE` (12)

The wallet side is the first thing in this protocol that can *try and fail*, and
v1 had no member for it. "No wallet open" / "cannot be funded" / "the broadcast
did not complete" are not `INTENT_INVALID_VALUE` — telling an honest caller its
message was wrong is how it "fixes" a message that was already correct. Added to
both halves (`error.rs`, `error.ts`); carries no detail, because "insufficient
funds" is a balance oracle. An **ambiguous** broadcast is reported as
`UNAVAILABLE`, never as a txid.

## The CI selector, and the digest pin

`wallet/zuuli/src-tauri` now path-depends on `rs/crates/f2z-intent`, and
`scripts/check-zuuli-android-gate.mjs` derives the linked-crate set by reading
every `wallet/**/Cargo.toml` rather than from a list. It went red on the first
CI run, correctly: without a selector entry a change to the intent protocol
would **not** have selected the ZUULI gate, so the wallet authority could be
rebuilt against modified authority-delegation code with the whole native matrix
skipped. Satisfied, not worked around.

One entry — `rs/crates/f2z-intent/*` — alphabetically placed between
`f2z-codec/*` and `f2z-kt-client/*` in the `zuuli=true` arm. Purely additive,
proven mechanically rather than by eye: splitting the old and new arms on `|`
gives **1 insertion, 0 deletions**, and the remaining order is identical. The
`zuuallet_schema` arm is deliberately untouched — `f2z-intent` is linked by
`wallet/zuuli/src-tauri` alone, not by zuuallet or the plugins.

**What `changeDetectorDigest` protects:** it is a reviewed sha256 over the whole
"Detect release-impacting ZUULI changes" step, so that a change to *which* paths
may skip a ~40-minute native matrix cannot land as a quiet diff inside a long
`case` line. Re-deriving it is a deliberate act, and the script provides
`--print-change-detector-digest` precisely so the constant is copied rather than
guessed. It was re-derived **after** confirming the semantic change is exactly
the one entry above:
`12ac7c78…` → `4c97e2b0…`. The self-test's stale-digest negative control still
fires against the previous literal, so the pin is still load-bearing.

- `node scripts/check-zuuli-android-gate.mjs --self-test` — 34 mutations passed
- `node scripts/check-zuuli-android-gate.mjs` — live, exit 0
- `check-workflow-gates`, `check-github-actions-pins`,
  `check-actions-cache-footprint`, `verify-ci-cache-policy`,
  `check-zuuli-linux-image`, `check-zuuli-store-capture-image` — self-test **and**
  live, all exit 0

## Plugin changes are widenings, never weakenings

| Change | Why |
|---|---|
| `format_zec_amount`, `quote_native_memo` → `pub` | one amount format and one escaper for every native confirmation |
| `native_review_lines` split out | shared payment block; the old function reproduces its previous output exactly |
| `ensure_active_seed_loaded` → `crate::wallet` | a second, non-IPC caller; `commands.rs` keeps the name so `send-review-boundary.node-test.mjs`'s "consume before custody" ordering rule still reads out of `execute_send` |

No policy checker was edited. No `#[allow]` was added.

## Conformance — mutation-verified

Every guard deleted or inverted in the source, the suite run, the named test
watched to fail, the guard restored. `AFTER RESTORE: GREEN`.

**30 mutations, 30 failures, 0 survivors** — full table in
`docs/intent-bridge/AUTHORITY.md` §8. Includes the three dual-clock rows,
re-verified *here* because this is where the deadline is actually used.

## Verification, all foreground

- `cargo test --locked -p f2z-intent` — 63 green
- `cargo test --locked --manifest-path wallet/zuuli/src-tauri/Cargo.toml` — 48 green (26 new)
- `cargo test --locked --all-targets --features production-route-probe -p tauri-plugin-zcash` — 150 + 1 green
- `scripts/check-rust-fmt.sh --root rs` and `--root wallet` — 17 + 6 crates formatted
- clippy `-D warnings` on the three touched crates (no full-workspace run)
- `scripts/check-rust-deny.sh --root wallet` — advisories/bans/licenses/sources ok
- `check-hash-domain-labels`, `check-tauri-plugin-permissions`, `check-dependency-register`, `check-rust-toolchain`, `check-public-repo-boundary`, `wasm-boundary` — `--self-test` **and** live
- `node --test send-review-boundary + seed-capture-boundary + messaging-contract` — 120 green
- `project-boundary.node-test.mjs` — 92 green; live: 3 projects / 338 files / 5 intent-bridge guards / 2 Vite builds
- `npx vitest run src/lib/intent-bridge.test.ts` — 24 green; `typecheck` + `typecheck:tests`

## Adversarial review found two survivors; both are closed

Four of the thirty rows exist because the first twenty-six **missed two guards**,
and both were ones this design calls central. Recorded here and in §8 rather than
quietly folded in, because "0 survivors" is only worth something if the times it
was wrong are visible.

- **The native dialog verdict.** Dropping the `!` from
  `if !confirm_natively(…)` — approve on *Cancel* — left every test green: the
  ordering test's step literal matched with or without the negation, and nothing
  else in the suite can reach a dialog. The primary control of the whole design
  had no test that would notice it inverted.
- **The ambiguous-broadcast rule.** Inverting or deleting the
  `BroadcastStatus::Accepted` check left the suite green; nothing referenced
  `BroadcastStatus` at all, while the docs advertised "an ambiguous broadcast is
  reported as `UNAVAILABLE`, never as a txid" as a headline property.

Both are closed by moving the decision into a **pure function a test can drive**,
not by an assertion about the source: `dialog_verdict(Option<bool>)` covers
authorize / cancel / silence, and `payment_outcome(&ExecuteSendResult)` covers
`Accepted` / `Rejected` / `Unknown` plus an accepted-but-malformed identifier.
Each has a positive control, so a mapping that refused everything would fail too.

One half stays a source assertion and is labelled as such: the `!` cannot be
exercised without a real window and a funded wallet, so the ordering test now
pins the literal `"if !confirm_natively(app, message).await {"`.

The campaign was re-run **whole**, not just the four new rows — restructuring
`execute_payment` could have staled an existing anchor, and a silently skipped
row is exactly how the two findings above survived. No row skipped.

## Caller-facing contract that was documented nowhere

**A refusal after admission burns the `request_id`.** Parse, caller and window
refusals do not spend it; the ledger claim and everything after it does —
including a family this build does not implement, a user who cancelled, and
`INTENT_UNAVAILABLE`. So an honest caller retrying after `INTENT_UNAVAILABLE`
**must mint a fresh `request_id`**; re-presenting the same bytes gets
`INTENT_REPLAY`. New `PROTOCOL.md` §4.1 states it with a table.

## Where I think this is weaker than it looks

Stated in AUTHORITY.md §5 rather than buried: on this surface `issue` → `consume`
runs inside one `async` function holding `send_operation`, and the confirmation
token never leaves the process. In #528 that token crosses to a renderer and back
— a real trust boundary. Here it does not. The intent-level authorization is a
**binding** (it is the only thing that can say "the user approved *this
delegation*", which the plugin cannot express) and not a second independent gate,
and its dual-clock deadline bounds an interval #528's own confirmation already
bounds. Worth keeping — it becomes load-bearing the moment a transport splits the
flow — but a reviewer should not count it twice.

Two more, both surfaced by review and now in AUTHORITY.md §5 rather than implicit:

- **`INTENT_UNAVAILABLE` is a coarse balance oracle at the level of control
  flow**, even though its payload carries nothing. An unfundable amount fails
  inside `propose_send`, which runs *before* `confirm_natively`, so a registered
  caller learns fundable-or-not from a silent refusal with no user interaction —
  and a descending binary search costs at most one visible dialog. Not
  exploitable today (two first-party callers, no transport); it becomes a real
  question the moment either changes.
- **"One-use" means one use per process.** A restart destroys the authorization
  *object*, but not the repeatability of the *request*: inside its ≤5-minute
  window the same bytes are admitted again by an empty ledger, the same dialog is
  shown, and two approvals are two payments. The barrier is a second human
  approval, not the ledger. The earlier wording ("a confirmed intent cannot be
  resurrected") overstated this and is corrected in both documents.

Also flagged, unchanged by this PR: the zcash plugin's seed lock is label-blind
(`WindowEvent::Focused(false)` ignores the window label). Nothing here touches
window events, and nothing here makes it worse.

Refs #905, #904.


skyl:
## Adversarial review — MERGE, after two cheap mutation gaps are closed

I could not find a way to reach `receive_intent` from a webview, execute a payment the user did not see, consume an authorization twice, cross-bind one request's approval to another, get an unattested caller treated as attested, or get a request-supplied string into an identity or authorization decision.

**Webview reachability — unreachable, more strongly than the PR claims.** `receive_intent` has *zero production callers anywhere in the repo*. Not a `#[tauri::command]`; no `.listen`, no `on_open_url`, no menu or window-event handler, no `.setup()` hook. `tauri_plugin_deep_link::init()` is registered at `lib.rs:20` but pre-exists this PR and nothing wires it here. So the Tauri-v2 "app-crate commands bypass capability ACL" hazard doesn't apply — there is no command and no caller. The `generate_handler!` parse test does what it claims, and its positive control genuinely works.

**Confirmation binding — every mechanical claim confirmed.** `consume(self, …)` is one-use by ownership with a constant-time compare and a re-checked deadline; the token is 32 CSPRNG bytes. Cross-request reuse is *structurally* impossible: `request_digest`, `review_digest` and `token` are locals in one stack frame and the approval never escapes it. `execute_send` cannot run without a consumed confirmation — it needs the proposal from `take_send_proposal`, which needs the token from `issue_send_confirmation`, which is only reached past the `confirm_natively` guard. `_send_operation` is held across the reviewed window and both `propose_send` and `discard_send_proposal` take that same lock, so no proposal can be swapped in. Every `Err` branch discards.

**Review correction:** `review_matches_request` does **not** compare fee — and that is correct, not a gap. The fee is advisory in the wire format; the displayed fee is always the wallet's own re-derivation, so a caller cannot make it wrong. My review brief asserted otherwise and was mistaken.

### Findings

**F1 (medium) — the native dialog verdict is undefended.** Drop the `!` in `if !confirm_natively(...)` at `intent.rs:303`/`:507`, or flip `unwrap_or(false)` → `unwrap_or(true)`, and **all 24 tests and all 26 mutation rows stay green**. The ordering test's step literal matches with or without the negation. That is a surviving mutation on the guard this design names as its *primary control*.

**F2 (medium) — the ambiguous-broadcast rule is undefended.** Inverting `!=` at `intent.rs:373-379` leaves the suite green; no test references `BroadcastStatus`. Both AUTHORITY.md §7 and the PR body state "never a txid" as a headline property.

**F3 (low)** — `INTENT_UNAVAILABLE` is a coarse **balance oracle**: an unfundable amount fails in `propose_send` *before* `confirm_natively`, so a hostile registered caller learns fundable/not-fundable with no user interaction. True of the payload, not the surface. Not exploitable today; belongs in the honest-limits list.

**F4 (low)** — "a *confirmed* one cannot be resurrected" (AUTHORITY.md §5) is misleading. True of the object, false of the effect: after a restart inside the ≤5-minute window the same bytes are admitted and the same dialog shown. The barrier is a second human approval, not the ledger.

**F5/F6 (nits)** — PROTOCOL.md §2's location table needs the new row; the unreachability test is name-based, not type-based.

Also undocumented anywhere: a request refused *after* admission burns its `request_id`, while parse/caller/window refusals do not — so an honest caller retrying after `INTENT_UNAVAILABLE` must mint a fresh id. Safe direction, but a real caller-facing contract detail.

**Verified with no finding:** the plugin refactor is a pure widening (`format_native_send_confirmation` reproduces its prior output byte for byte); the CI selector fix is exactly one insertion, 79→80 entries, zero deletions; CALLER-AUTHENTICATION §5 is fully intact and §6's "ship `execute-payment` first" staging is followed literally.

On the self-assessment: §5 is accurate and if anything *under*-sells the weakness. The sharper statement is that all three binding inputs are locals in a single stack frame, so `consume` is closer to a structural assertion than a runtime check — and the `request_digest` half becomes load-bearing the moment a transport splits the flow.

PR #535, opened 2026-08-23
feat(zuuli): prove first-party Rust WASM browser path

skyl:
## Summary

- add a dependency-free first-party Rust crate built with the canonical 1.97.1 toolchain for wasm32-unknown-unknown
- make Zuuli/Vite build, bundle, instantiate, and call the 116-byte module in a real browser
- bind generated bytes to exact Rust inputs, keep all generated output uncommitted, and verify production byte identity/reference
- add fail-closed CI/toolchain/action policy with adversarial tests for stale, skipped, decorative, narrowed, environment-injected, and post-checkout source-substitution controls
- document the build/versioning model, measured local and hosted costs, frontend recommendation, and real-crate requirements

## Proof

- GitHub Actions policy: 70 source-policy, 7 gate-verdict, 80 live-workflow mutations
- WASM boundary: 9 adversarial mutations
- toolchain verifier: 14 drift mutations
- actionlint and ShellCheck pass
- all four wallet Rust crates pass canonical rustfmt
- spike crate passes wasm-target clippy at -D warnings and its host test
- npm run build emits and re-verifies one byte-identical production WASM asset
- three artifact tamper tests pass
- Playwright calls the export in Chromium and observes 42

## Measurements and recommendation

Local clean-target build: 1.37 seconds wall; Rust build and verification: 1.02 seconds. Warm build: about 0.52 seconds. Release artifact: 116 bytes; deterministic gzip: 128 bytes.

Hosted replacement-head measurement: about 10.5 seconds of directly attributable sequential frontend overhead. That comprises a 10-second compiler/target install, 154 ms across three builds, 254 ms across artifact negative tests, and about 72 ms of final dist verification. The browser case passed inside the existing fully parallel suite; its individual hosted wall time is not claimed because the reporter exposes only aggregate suite time. The checked-in report links the exact run and explains this isolated-cost methodology.

Zuuli/Vite is the recommended consumer. CRA/react-scripts 5 should wait for migration. A real API should pin matching wasm-bindgen library/CLI versions, use serde-wasm-bindgen only where data exchange needs it, keep a browser-compatible core separate from native shells, and decide randomness/threads explicitly.

Playwright proves Vite/browser instantiation. Packaged Tauri WebView CSP now authorizes WASM compilation, but packaged/device runtime remains a distinct test caveat and is not claimed by this browser proof.

Closes #383

PR #920, opened 2026-09-04
feat(free2z): port Live, AI and Search out of ZUULI (#904 phase 2)

skyl:
Phase 2 of #904. Ports the three surfaces #912 deliberately deferred — `features/search`, `features/ai`, `features/live` — into `wallet/free2z/`.

> **Stacked on #912.** Branched from `feat/port-content-surfaces-free2z`, so until #912 squash-merges this diff shows #912's files too. Once #912 lands I will `git rebase --onto origin/main 5ffa877` and confirm the file/line counts shrink to this PR's own 47 files before asking for merge.

## The load-bearing property is intact

Nothing here needed privilege, and nothing here took any.

* `src-tauri/src/lib.rs` still registers **no `invoke_handler` at all**.
* `src-tauri/Cargo.toml` still links neither `tauri-plugin-zcash` nor `tauri-plugin-f2zmsg`.
* Both capability files are **byte-for-byte unchanged**; `surface-capability-authority.mjs` reports *"free2z grants no `zcash:*` or `f2zmsg:*` entry at all."*

The only `src-tauri/` change in this PR is three origins added to the CSP string.

## `live` — the CSP review

#816/#818 are the prior failure: ZUULI's packaged CSP blocked the bundled SDK's first request and "Join Free" died with `ERR0001`. `tests/realtimekit-csp.pw.ts` (ported and extended from ZUULI's) loads the **real packaged policy** and the **real bundled SDK**, so every claim below is executable.

### Admitted — three origins, each justified

| Origin | Justification |
| --- | --- |
| `https://api.realtime.cloudflare.com` | `RealtimeKitClient.init()`'s first call, `/v2/internals/participant-details`. The **negative-control test** removes it and reproduces `ERR0001` verbatim. |
| `wss://socket-edge.realtime.cloudflare.com` | The meeting transport. `SocketService.getSocketEdgeDomain()` composes `socket-edge.${baseURI}` via `fn({servicePrefix, baseURI}) => \`${s}.${t}\``. Without it no media session is established. |
| `https://rtk-assets.realtime.cloudflare.com` | `fetchEmojis()` in `@cloudflare/realtimekit-ui/dist/collection/utils/assets.js` fetches `/assets/emojis-data.json` for the reactions picker inside `<RtkMeeting>` (which `Stage.tsx` renders). It has **no `try`/`catch`**, so a block is an unhandled rejection, not a graceful degrade. |

### Refused — narrower than ZUULI, and asserted rather than merely omitted

`REFUSED_SOURCES` in the test carries the evidence for each, and a test fails if one creeps back in:

* **`location.realtime.cloudflare.com`, `location-legacy.realtime.cloudflare.com`** — a callstats **IP-address** lookup. Both fetches sit in a `try`/`catch` whose catch only logs `callstats::ipDetails:: failed to fetch ip using location service`. A content app should not beacon the reader's IP to a third party to produce a call statistic.
* **`da-collector.realtime.cloudflare.com`** — device-analytics events (`/api/v1/message`), guarded by a `try`/`catch` and a 3s `AbortController`.
* **`api-silos.realtime.cloudflare.com`** — OTel log shipping (`/otel/logs`), best-effort and caught.
* **`r2.cloudflarestorage.com`** — `grep -rl cloudflarestorage node_modules` returns **zero files**. Not referenced anywhere in the installed dependency tree.
* **`*.dyte.io` / `wss://*.dyte.io`** — **dead**. RealtimeKit 1.5.1 *throws* `"Dyte Base URIs are no longer supported. Use RealtimeKit Base URIs."` when the base URI contains `dyte.io`.
* **`*.zec.rocks`** — a lightwalletd origin. This app holds no seed (#904).

**The exclusions are measured, not assumed.** The initialization test asserts `violations: []` and reaches the `0004`/`Unauthorized` server boundary under the exact packaged policy — i.e. init completes with **zero CSP violations** without any of the four telemetry origins.

### `frame-src` stays `'none'` — RealtimeKit does not need frames

You asked me to make the argument rather than assume it, so: **the argument isn't needed, because the premise is false.** RealtimeKit renders the meeting as web components in the document; it frames nothing. ZUULI already ships the same SDK behind `frame-src 'none'`, and the join failure in #813 is unrelated to framing. `Live needs no frames, no eval and no wasm` asserts this. Had it needed frames I would have argued it is acceptable here — no privileged command exists for a hostile subframe to reach, which is the entire point of the split — but I am not spending that allowance on a need that does not exist.

Also unchanged, each with evidence:

* **`script-src 'self'`** — no `'wasm-unsafe-eval'`. ZUULI's came from #535 (first-party Rust WASM); `grep -rl WebAssembly` over the three `@cloudflare` packages returns nothing.
* **no `worker-src`** — the SDK's only `new Worker("data:application/javascript;base64,…")` lives in `EncryptionManager`, a **separate entry point this app never imports** (`index.es.js` contains zero `new Worker(`). It falls back to `default-src 'self'`.
* **`media-src` unchanged** — video attaches via `srcObject` (a `MediaStream`, which CSP does not govern); all three `createObjectURL` call sites in the SDK produce `` sources, already covered by the existing `img-src … blob:`.

## A real bug the port carried across

Three `<Link to="/wallet/fund">` buttons came over from ZUULI. free2z mounts `/fund` and **deliberately mounts nothing under `/wallet`**, so all three rendered NotFound — at exactly the moment a viewer is told they need more 2Zs to keep watching (`ai/index.tsx`, and both PPV gates in `live/Room.tsx`).

Nothing caught it. Typecheck can't see through a string; the unit suites never render those branches; and `surface-separation.pw.ts` asserting *"an unmounted wallet path renders NotFound"* is the **correct** behaviour and stayed green the whole time.

So `src/lib/route-targets.test.ts` now asserts every absolute in-app link against the mounted route table. Reintroducing one of the three fails it precisely:

```
FAIL  src/lib/route-targets.test.ts > only links to route segments this surface mounts
AssertionError: these paths render NotFound on this surface; ZUULI's route table is not this one
+ [ "../features/ai/index.tsx: /wallet/fund" ]
```

## Boundary guard is load-bearing — proof

Injecting a cross-project wallet import into `features/live/viewer-safety.ts`:

```
$ node wallet/zuuli/scripts/project-boundary.mjs
/…/wallet/free2z/src/features/live/viewer-safety.ts:1:27: relative import escapes wallet/free2z: ../../../../zuuli/src/store/wallet
free2z/vite.config.ts: constrained production Vite graph build failed; a plugin or module may have read outside wallet/free2z and wallet/shared: Could not resolve "../../../../zuuli/src/store/wallet"
EXIT=1
```

The `@/store/wallet` form fails too, and its failure is itself the point — free2z has no wallet store to resolve:

```
free2z/vite.config.ts: constrained production Vite graph build failed: [vite:load-fallback] Could not load
  /…/wallet/free2z/src/store/wallet (imported by src/features/live/viewer-safety.ts): ENOENT
EXIT=1
```

Reverted (`viewer-safety.ts` is byte-identical to ZUULI's), and clean:

```
Wallet project boundaries verified across 5 discovered projects, 538 source files, 2603 parsed
module references, … and 3 constrained production Vite builds.
```

That third build is now free2z's **with the RealtimeKit graph inside it**.

## Verification

| Check | Result |
| --- | --- |
| `npm run typecheck` / `typecheck:tests` | clean |
| `npx vitest run` | **785 passed**, 5 skipped, 61 files |
| `npx playwright test` | **46 passed** (5 new CSP specs) |
| `npm run test:ui-copy` | 4 passed |
| `npm run build` | ok — RealtimeKit code-splits into lazy `rtk-*` chunks |
| `npm audit --audit-level=high` | passes (new `uuid` advisory is moderate, under the floor) |
| `project-boundary.mjs` | 5 projects, 3 Vite builds |
| `project-boundary.node-test.mjs` | 92/92 |
| `surface-capability-authority.mjs` | free2z: zero `zcash:*`/`f2zmsg:*` |
| `check-workflow-gates.mjs` | passes |
| `check-github-actions-pins.mjs` | passes |

Ported tests are genuinely collected, not stranded: 9 `features/live/*` files and `features/search/pagination.test.ts` appear by name in the vitest run, and `tests/realtimekit-csp.pw.ts` matches `playwright.config.ts`'s `**/*.pw.ts`.

## Honest notes

* **Corrected — I previously wrote that "`features/ai` ships no tests, because ZUULI has none". That was wrong.** ZUULI has no *unit* tests for AI, but it has substantial **e2e** coverage that this port did not carry over: **992 lines across seven specs** — `ai-context-switch` (121), `media-preflight` (105), `participant-counts` (91), `private-live` (95), `search` (146), `search-autocomplete` (277), `search-pagination` (157) — plus `/ai` exercised inside `paid-actions.pw.ts` (394), `viewport.pw.ts` (379) and `touch-targets.pw.ts` (326). free2z's `tests/` gained exactly one file, the CSP proof. #912 *did* port its relevant specs, so this breaks that pattern. **AI coverage in free2z is zero**, and that is a choice this port made, not an absence in ZUULI. The porting work is filed separately.
* **All three surfaces inherit the `@tauri-apps/plugin-http` gap.** Search, AI and Live all reach the backend through `lib/api/free2z.ts` → `lib/api/http.ts`, which lazily imports `@tauri-apps/plugin-http`; that plugin is in `package.json` but is not registered in `src-tauri/`, so native HTTP fails in a packaged build. Untouched here on purpose — it needs a scope decision, since `surface-capability-authority.mjs`'s `REVIEWED_NAMESPACES` excludes `http`.
* **#813 is still an open P0 and this PR does not close it.** #818's CSP fix was necessary but not sufficient; #816's own acceptance criteria for observed WebSocket hostnames and signed device builds remain unchecked. This PR gives free2z a *narrower, better-evidenced* CSP than ZUULI's, not a fixed join.
* **Mobile media permissions are not wired.** ZUULI's `AndroidManifest.xml` declares `CAMERA`, `RECORD_AUDIO` and `MODIFY_AUDIO_SETTINGS`; free2z has no `src-tauri/gen/` yet, so `GoLiveDialog`/`MediaPreflight` will need those when free2z is initialized for mobile. Deliberately not added — creating a mobile project is not this PR.
* **Nav placement is a judgement call.** ZUULI reaches Search from a TopBar search box; free2z's TopBar has none, so rather than redesign a #912 file I added Search/Live/AI as `explore` nav entries. Easy to change.
* **`navLiveShort` was dropped.** free2z's nav model has no per-mobile label, and `i18n/source-policy.test.ts` correctly failed the orphan key — a nice catch by an existing guard.
* **Unrelated hygiene bug spotted:** `wallet/zuuli/scripts/project-boundary.node-test.mjs` leaks a fixture to repo-root `dist/.output-copied.ts`, which is **not** gitignored. I deleted it rather than committing it. Worth a follow-up issue.

Nothing is deleted from ZUULI, per #912's pattern — teardown is #904 phase 4.

---

**Second commit** fixes copy the port carried across verbatim: the camera-preview fallback told the reader to *"use a supported browser or the current **ZUULI** app"* — a different app, on the very surface being split away from it — and a neighbouring comment in `Room.tsx` described this app's own chrome as ZUULI's.

**Where the actual video goes.** A reviewer may reasonably ask why so few origins suffice for a livestream. The media itself never touches `connect-src`: RealtimeKit carries it over `RTCPeerConnection` to an SFU, negotiated through ICE/TURN (the SDK ships `turn:turn.dyte.in:443` among its ICE servers), and CSP does not govern ICE/TURN/STUN traffic at all. Only the **signaling** is CSP-visible — the REST call to `api.realtime.cloudflare.com` and the WebSocket to `socket-edge.realtime.cloudflare.com`. That is exactly the pair admitted above.

---

## Post-review corrections (third commit)

Adversarial review found three things, all fixed here.

**Two more `/wallet/...` targets the first guard missed.** Both were live, both were #912's
files, and neither was a `to=`:

* `lib/auth/login-destination.ts` — `safeLoginDestination` **returned** `/wallet/fund`,
  `/wallet/fund/activity` and `/wallet/fund/send` as sanctioned post-sign-in landings, and mapped
  `/buy → /wallet/fund`. Those go straight to `navigate()` from `ClassicLoginForm.tsx:44`,
  `SocialButtons.tsx:158` and `App.tsx:80`, so the sanitizer sanctioned three paths that render
  NotFound. Now the union is `/`, `/ai`, `/fund`, `/articles/*`, `/creator/*`, `/live/*` — exactly
  what `App` mounts. `/buy` still normalizes to `/fund` so in-flight logins survive; `/buy/send`
  has no counterpart on a surface with no Send, so it fails closed to `/`.
* `lib/api/free2z.ts` — `buyCheckout` sent `currentPath: "/wallet/fund"`, which is what the backend
  builds the Stripe return URL from. Latent today (no caller yet), live the moment funding ports —
  a paying customer returned to NotFound.

**The guard was too narrow, in four separate ways.** All four are now closed, and each was proven
to fail before the fix:

| Blind spot | Proof |
| --- | --- |
| bare allowlist (`safeLoginDestination`) | `"/buy" -> /wallet/fund`, `"/wallet/fund" -> /wallet/fund`, +3 more |
| `currentPath:` | `./api/free2z.ts: /wallet/fund` |
| object-literal `to:` (the whole `NAVIGATION` table was unscanned) | `../components/layout/navigation.ts: /wallet/fund` |
| first-segment-only matching (`/fund` is exact, has no children) | `../components/layout/navigation.ts: /fund/activity` |
| `href=` | `../features/live/Discovery.tsx: /wallet/fund` |

The guard now (a) scans `to=`, `href=`, `navigate(`, `to:` and `currentPath:`, (b) matches against
**real route patterns** — `:param` takes one segment, a trailing `/*` takes children, everything
else is exact — (c) asserts behaviourally that every value `safeLoginDestination` can *return* is
mounted, and (d) sweeps all executable source for any literal beginning with a ZUULI-only segment
(`/wallet`, `/messages`, `/kyc`, `/profile`, `/about`). Comments are stripped first, so the places
that name those routes precisely to disclaim them — `routes.ts`'s header, `lib/bridge/creator-tip.ts`
— are not flagged.

`login-destination.test.ts` and `paid-intent.test.ts` pinned the old behaviour and were updated;
`login-destination.test.ts` gained an explicit `refuses the unmounted destination` case for all
five bad paths. Suite is **791 passing** (up from 785), 46 Playwright, boundary and capability
checks unchanged.

### Two CSP notes from the review, both worth stating

* **Only one of the three admitted origins is exercised by a test.** `api.realtime.cloudflare.com`
  has a positive *and* a negative control. `rtk-assets` and `wss://socket-edge` rest on code-reading
  (`realtimekit-ui/dist/collection/utils/assets.js:1` and `index.es.js`'s
  `getSocketEdgeDomain`). The reviewer independently re-derived both and they hold. Over-admitting
  an origin the SDK demonstrably references is the low-risk direction — but the asymmetry is real.
* **My `media-src` reasoning was right in conclusion, incomplete in derivation.** Video does attach
  via `srcObject`, but the SDK also makes real `https:` media loads: a speaker-test clip from
  `rtk-assets.realtime.cloudflare.com/ui-kit/speaker-test.mp3` (already admitted), and notification
  sounds from **`rtk-uploads.realtime.cloudflare.com`** (`notification_join.mp3`,
  `notification_message.mp3`) — an SDK origin that appears in neither my admitted nor my refused
  list. Nothing breaks, because these are `media-src`, not `connect-src`, and the pre-existing
  `media-src 'self' https:` admits any HTTPS origin. Worth knowing that Live's sounds ride on that
  broad pre-existing allowance rather than on anything this PR reviewed.

### One more thing found while fixing

`PaidIntent` still carries a `{ kind: "send" }` member — ZUULI's ZEC Send flow. It has **no caller
in free2z** and never will; this surface has no Send. Left in place rather than removed, since
deleting a union member is outside this PR's scope, but it is dead weight on a surface defined by
not having a wallet.


skyl:
## Adversarial review — MERGE

Scope isolated to #920's own commits (`5ffa877..34e2503` = 47 files, +8846/−15); everything from #912 set aside and attributed separately.

**Privilege: clean, verified independently.** `git diff origin/main..34e2503 -- wallet/free2z/src-tauri/capabilities/` is **empty** — both files byte-for-byte unchanged across the whole stack. Parsed as JSON (not grepped): each grants exactly `core:default`, `deep-link:default`, `opener:default`. The only `src-tauri/` change in the entire stack is **one line** — the CSP string at `tauri.conf.json:23`. `lib.rs:11-17` registers no `invoke_handler` at all.

**The CSP is the strongest part.** The added directive is a strict *subset* of ZUULI's — free2z refuses 10 origins ZUULI carries, and drops `'wasm-unsafe-eval'`. Refusals are **asserted, not merely omitted**, and doubly: `realtimekit-csp.pw.ts:169-174` asserts exact ordered equality of `connect-src`, so any origin creeping in fails; `:176-183` re-asserts each named refusal with its evidence.

Every refusal justification was re-derived against the installed SDK rather than taken on trust:
- `index.es.js:28050-28051` — `if (o.baseURI?.includes("dyte.io")) throw new R("Dyte Base URIs are no longer supported…","0001")`. **Dead-origin claim confirmed verbatim.**
- `grep -rl cloudflarestorage` → **zero files**.
- `index.es.js:1329-1342` — both callstats fetches sit in `try`/`catch` that only log.
- `index.es.js:881`, `:5344`, `:5194-5198` — `da-collector` and OTel `api-silos` likewise.

**The negative control is real and was executed**: stripping only `apiOrigin` yields `requestCount === 0`, a `securitypolicyviolation` on `connect-src`, and the verbatim `[ERR0001] … Unable to connect to the server.` — the #813/#816 symptom, reproduced under the real packaged policy against the real bundled SDK. CI logged `46 passed`.

**Boundary guard confirmed load-bearing** by reading `project-boundary.mjs:1054-1057` — `inside()` is purely lexical, so it fires regardless of resolution; `:1241` throws, `:1256-1260` sets exit 1. free2z's Vite graph *with RealtimeKit in it* built inside the required gate.

### Findings

**F1 — a false claim in the body.** "`features/ai` ships no tests, because ZUULI has none" is true only at the *unit* level. ZUULI has `tests/ai-context-switch.pw.ts` plus `/ai` coverage in `paid-actions.pw.ts`, `viewport.pw.ts` and `touch-targets.pw.ts`. **992 lines across seven specs were not ported**, breaking the pattern #912 set. free2z's AI coverage is **zero** across 1,754 lines. Filed as **#923**; body being corrected.

**F2 — the route bug class is not closed.** The three `<Link to="/wallet/fund">` fixes are real and the test would catch a reintroduction. But two live targets survive in the tree it scans: `lib/auth/login-destination.ts:64-70` *returns* `/wallet/fund`, `/wallet/fund/activity`, `/wallet/fund/send` as sanctioned post-sign-in destinations (fed to `navigate()` at three call sites), and `lib/api/free2z.ts:2050` sends `currentPath: "/wallet/fund"` as the Stripe return path. Both are #912's files, not regressions here.

**F3 — the test regex is materially under-scoped.** It misses `href=` entirely, object-literal `to:` without `=` (so the **whole `NAVIGATION` table is unscanned**), bare string constants, and matches **first segment only** — `/fund` is mounted exact with no splat, so `to="/fund/activity"` passes and renders NotFound.

F2 and F3 are being fixed on this branch, with proof-of-failure before the fix.

### Minor, non-blocking
- Only one of the three admitted origins is test-exercised; `rtk-assets` and `wss://socket-edge` rest on code-reading (both re-derived and holding). Over-admitting is the low-risk direction.
- The `media-src` reasoning is right in conclusion but incomplete: three real `https:` media loads exist (`rtk-speaker-selector.js:112`, three sounds from `rtk-uploads.realtime.cloudflare.com`), all covered by the pre-existing `media-src 'self' https:`.

### Honest limits — all three confirmed accurate
#813 remains **OPEN P0** and this does not close it; the `plugin-http` gap (#918) is real; mobile media permissions are unwired (no `src-tauri/gen/`).

**Governance note:** `surfaces / frontend (free2z)` is in `check-workflow-gates.mjs` `UNGATED_WORKFLOWS`, so `realtimekit-csp.pw.ts` **runs but cannot block a merge**. Green on head SHA `34e2503`, so this PR's evidence is real — but the CSP invariant is advisory going forward until #915.

Issue #918, opened 2026-09-04
free2z: wire the native layer — http plugin, OAuth transport, and its own deep-link scheme


skyl:
Follow-up to #912 (#904 phase 1). #912 ported the article/creator content surface into `wallet/free2z/` and the **security property it was reviewed for is intact**: zero `zcash:*` and zero `f2zmsg:*` capability entries, neither privileged plugin linked, and `wallet/free2z/src-tauri/src/lib.rs` registers no `invoke_handler` at all. Nothing here weakens that.

What it also did was port three frontend code paths that branch on `isTauri() && !import.meta.env.DEV` **without** writing the Rust side they call into. Under those conditions the app is dead, and no gate can see it: vitest, Playwright and `npm run build` all run where `isTauri()` is `false` or `DEV` is `true` — precisely the branch that is *not* taken — and `cargo check --locked --all-targets` compiles a backend that is never asked whether the frontend expects commands from it. The JS→plugin contract is unchecked by construction.

#912 states this in `wallet/free2z/README.md` ("Not yet wired natively") and sets `bundle.active: false` in `wallet/free2z/src-tauri/tauri.conf.json`, so `tauri build` cannot emit an installer for a binary with a dead network layer. This issue is the work that earns the flag back.

## 1. HTTP is unwired

`wallet/free2z/package.json` depends on `@tauri-apps/plugin-http`, but `wallet/free2z/src-tauri/` was never touched to match:

* `wallet/free2z/src-tauri/Cargo.toml` — no `tauri-plugin-http`
* `wallet/free2z/src-tauri/src/lib.rs` — no `.plugin(tauri_plugin_http::init())`
* `wallet/free2z/src-tauri/capabilities/default.json`, `capabilities/mobile.json` — no `http:*` permission

Two call sites take the native path in a packaged build:

* `wallet/free2z/src/lib/api/http.ts:129-132` — `const useNativeHttp = isTauri() && !isDev;` then `await import("@tauri-apps/plugin-http")`. This is the transport for **every** API request.
* `wallet/free2z/src/components/common/RemoteMedia.tsx:50-51` — same branch for first-party image downloads, which additionally rely on the plugin's `maxRedirections: 0` transport option to validate each hop.

So in a packaged build every network call invokes `plugin:http|fetch`, which does not exist in that binary.

### The constraint that makes this a scope decision, not a one-liner

`wallet/zuuli/scripts/surface-capability-authority.mjs:56-61`:

```js
[...]

PR #924, opened 2026-09-04
feat(free2z): the caller side of the intent bridge — a ZEC tip that asks (#790)

skyl:
Closes the **caller half** of #790 (ZEC tipping), on the protocol #911 landed and against the authority #914 implements.

> **Stacked.** Branched from `feat/port-live-ai-search-free2z` (#920, itself on #912). The diff below therefore contains their commits until they merge. **Before merging this**, rebase off them and verify by the *shrink*:
> ```
> git fetch origin && git rebase --onto origin/main 34e2503 feat/free2z-intent-client
> ```
> If their files still show large diffs after that, stop — a bad rebase silently reverts merged work.
>
> **`wallet/shared/src/intent/error.ts` needs no hand-merge.** It is already byte-identical to `origin/main` (`git diff origin/main -- wallet/shared/src/intent/error.ts` is empty), so the hunk is present in the new base and either side of a conflict yields the same bytes. Verify afterwards that `Unavailable: 12,` appears exactly once and the `NAMES` row is not duplicated.

---

## What was actually missing, and why nothing shipped before

`wallet/free2z/src/lib/bridge/creator-tip.ts` only *validated* a tip target. It did not build an intent, and the file said why at length: `encodeExecutePaymentPayload` refuses `amountZatoshis <= 0`, and **no tip dialog in the product collected a ZEC amount**. That is the product gap #790 names, so this PR closes it first and the protocol work follows from it.

## 1. The tip dialog collects a ZEC amount

The creator dialog's ZEC tab now has four presets (`0.001 / 0.01 / 0.05 / 0.1`, carried as the exact decimal *strings* the parser reads — never floats) and an amount input, styled off the 2Z tab it sits beside.

`validateZec` in `lib/format.ts` wraps the existing `parseZecToZatoshis`, which was already exact-integer: it matches `^(\d+)(?:\.(\d{1,8}))?$`, pads the fraction to 8 places and builds a `BigInt` from the digits. **No JavaScript float is ever multiplied by 1e8.** Zero, negative, non-numeric, grouped, exponential and nine-decimal inputs are all refused outright rather than rounded; the amount is additionally bounded at 1,000 ZEC so a fat finger cannot reach ZUULI's confirmation.

Copy fits, and nothing truncates. The only shortened string anywhere in the flow is a txid, through `truncateAddress` — middle ellipsis, tail-weighted, no CSS clip on top.

## 2. A real `execute-payment` request

Built through `createIntentSession` / `encodeExecutePaymentPayload` from `@free2z/wallet-shared` — the single implementation. This PR declares none of the five names `project-boundary.mjs` reserves, mints no `free2z/intent/v1/` label, and writes no encoder, version gate or response matcher.

Matched to what `wallet/zuuli/src-tauri/src/intent.rs` actually validates, not to a guess:

| Field | Value | Why |
|---|---|---|
| `caller` | `cash.free2z.free2z` | the exact key in ZUULI's `REGISTERED_CALLERS` |
| `intent` | `ExecutePayment` | the only family #914's surface implements |
| window | 120 s | under the 300 s ceiling; #904's "nothing here is a continuous grant" |
| `purpose` | `Tip <username> on free2z` | the username, not the display name — it is the identity a payer can check against the page they came from |
| `memo` | empty | on-chain data the payer authors; ZUULI's own review is the surface that can show them what they are attaching |
| `fee_zatoshis` | 10 000 | advisory. ZIP-317's conventional minimum, so ZUULI's "the requesting app expected a fee of…" line appears when the payment really is costlier than a simple send, instead of on every tip |

Every previous validation is intact and still runs **first**: `recordCreatorTipIntent` is the argument check it always was. Strictly additive.

## 3. One transport seam, and it fails closed

`wallet/free2z/src/lib/bridge/intent-transport.ts`: one interface, one method (bytes in, bytes out), one shipped implementation.

`failClosedIntentTransport.exchange` **has no branch that returns bytes.** It rejects with `IntentTransportUnavailableError` (`code: "INTENT_TRANSPORT_UNAVAILABLE"`, reason naming #461). `installedIntentTransport` is the drop-in point — one binding, and nothing else in the app changes.

The UI says only what this app can prove. Five states, decided in one exhaustive map (`features/creator/tip-copy.ts`) with a `never` binding, so a sixth outcome is a compile error rather than a silent fall-through:

| Outcome | Can we prove no funds moved? | What it says |
|---|---|---|
| `sent` | — | the only branch that renders a txid |
| `no-transport` | **yes** | *"ZEC tips are signed in ZUULI"* — prepared a request, nothing was sent |
| `unsendable` | **yes** | *"free2z could not send this request"* — ZUULI was never asked |
| `refused` + `INTENT_NOT_CONFIRMED` | **yes** | *"ZUULI did not authorize this tip"* — your ZEC is untouched |
| `transport-failed`, `refused` + anything else | **no** | *"free2z did not get an answer"* — open ZUULI and check for a pending transaction |

The axis is *do we know what happened*, not *did it work* — see the review section below for why that mattered more than it looked.

**Interaction with #918** (free2z's native layer is deliberately unwired): none, deliberately. The seam does not `invoke()`, does not touch `@tauri-apps/plugin-http`, and does not need the bundle to be `active`. Whether the native layer is wired changes nothing about this path today — which is the correct coupling, because the transport this seam is waiting for is an OS link handler, not a Tauri command. If it ever wanted one, that would be the boundary violation, not the implementation.

## 4. The response is judged as if hostile

`IntentSession` is created **per exchange, with capacity one**. That is the correlation, not thrift: `accept` returns the family and payload but not *which* question was answered, so a shared map with two tips in flight could let the second's answer be read as the first's. One outstanding question cannot make that mistake, and the entry is consumed either way, so a replay finds nothing.

A txid is returned only when the response decoded (version gate, re-encode equality), correlated to this request, named this family, arrived inside the window, carried status 0, **and** held exactly 32 bytes (`decodeExecutePaymentResult`, added to the shared package rather than to a client).

### What this can and cannot guarantee — stated plainly

`CALLER-AUTHENTICATION.md` §5: **there is no signature over responses.**

- **Proved:** the responder saw the request. `request_id` is 32 CSPRNG bytes in exactly one outbound message, so an app that never saw it cannot forge an answer.
- **Not proved:** that the responder is ZUULI. An app that *received* the request holds the identifier and could answer. Correlation is not authentication.
- **Not proved:** that a `sent` outcome means the network took the transaction. That is ZUULI's `payment_outcome`, which returns a txid only for `BroadcastStatus::Accepted`. This side trusts that; it cannot verify it.

Response authenticity is a property of the transport, and the transport is #461. Which is why there is no transport.

That limit is stated **in a code comment at the seam** (`intent-transport.ts`), not only here — including the per-platform bar for whoever wires the real transport in: a Universal Link on iOS and not a custom scheme, `startActivityForResult` on Android with §3.1's unmeasured composition question flagged, and never a response payload in a URL query component (§4.1). A PR body is not read by that person; the code is.

## 5. The boundary is unchanged

`src-tauri/` is untouched. No `invoke_handler`, no `tauri-plugin-zcash`, no `tauri-plugin-f2zmsg`, no `zcash:*`, no `f2zmsg:*`. `surface-capability-authority.mjs` and `project-boundary.mjs` both confirm it below.

---

## Verification

Every command run to completion on this branch, on the real tree.

| Check | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run typecheck:tests` | clean |
| `npx vitest run` | **883 passed**, 5 skipped, 63 files |
| `npx playwright test` | **47 passed** |
| `node --test scripts/ui-copy-truncation.node-test.mjs` | 4 passed |
| `node wallet/zuuli/scripts/project-boundary.mjs` | 5 projects, 540 files, 2611 references, 5 guards, 3 constrained Vite builds |
| `node wallet/zuuli/scripts/surface-capability-authority.mjs` | "free2z grants no `zcash:*` or `f2zmsg:*` entry at all" |
| `wallet/zuuli` `npx vitest run src/lib/intent-bridge.test.ts` | 24 passed (the shared package is still byte-compatible) |

Playwright is a separate suite and it was run separately: `tests/surface-separation.pw.ts` gains an amount-entry path and an amount-validation spec, both asserting the rendered copy.

## Mutation evidence

Baseline: 190 green across the bridge, formatting and tip-copy suites. Every guard below is pinned alone. Every mutation was applied, run, pasted, and restored; the tree ends green. Recorded in `docs/intent-bridge/CONFORMANCE.md`.

**ZEC→zatoshi factor** (`padEnd(8, "0")` → `padEnd(7, "0")`):

```
× ZEC input becomes an exact number of zatoshis > converts 1 ZEC
  → expected { zatoshis: 10000000, error: null } to deeply equal { zatoshis: 100000000, error: null }
× converts 0.001 ZEC   → expected {  zatoshis: 10000 } … { zatoshis: 100000 }
× converts 0.01 ZEC    → expected {  zatoshis: 100000 } … { zatoshis: 1000000 }
× converts 0.1 ZEC     → expected {  zatoshis: 1000000 } … { zatoshis: 10000000 }
× converts 1.5 ZEC     → expected { zatoshis: 15000000 } … { zatoshis: 150000000 }
× converts 1000 ZEC    → expected { zatoshis: 10000000000 } … { zatoshis: 100000000000 }
```

**Response correlation** (`pending.findIndex((entry) => bytesEqual(…))` → `findIndex(() => true)`):

```
× rejects a response addressed to a different request
  → expected { kind: 'sent', …(1) } to deeply equal { kind: 'refused', error: 11 }
× rejects an identifier that is merely close
  → expected { kind: 'sent', …(1) } to deeply equal { kind: 'refused', error: 11 }
× answers one question once: a replayed response finds nothing pending
  → expected { kind: 'sent', …(1) } to deeply equal { kind: 'refused', error: 11 }
```

**`amountZatoshis <= 0`** — this started as a **negative** result, and closing it changed the code rather than the tests.

The first version had the check twice (`encodeExecutePaymentPayload` and `requestCreatorTipPayment`), and neither copy could be mutated alone — only removing both turned a test red. That is a defect, not a curiosity: either copy could have been deleted in a later refactor with the suite fully green. Measuring first showed the split was not where it looked:

| Value | Caller's `<= 0` | Encoder's `<= 0n` | Distinguishable? |
|---|---|---|---|
| `0` | refuses | refuses, at the same point, with the same code | **no** |
| `-1` | refuses | refuses — and `-1 < 0`, so the mutation never changed this value at all | **no** |
| `0.5`, `NaN`, `Infinity` | `isSafeInteger` refuses | **throws `RangeError`** out of `outcome()` | yes |
| `2^53` | `isSafeInteger` refuses | **accepts it and encodes it** | yes |

So the duplicate was deleted, not papered over: positivity belongs solely to the encoder (`PROTOCOL.md` §3.4), and the caller keeps *representability*, which nothing else enforces. That last row changed the design — `BigInt(2**53)` converts cleanly, so without `Number.isSafeInteger` a nonsense amount reaches the wire and is **sent**. The check that looked like a duplicate was, in the cases that matter, the only thing standing there.

Each guard is now pinned **alone**.

Encoder positivity (`<= 0n` → `< 0n`), red both directly and through the caller:

```
=== mutation: amount-positive-encoder -> exit 1 ===
× the encoder itself refuses 0n zatoshis
  → expected { ok: true, …(1) } to deeply equal { ok: false, error: 4 }
× refuses to build a request for 0 zatoshis
  → expected { kind: 'sent', …(1) } to deeply equal { kind: 'unsendable', error: 4 }
Tests  2 failed | 136 passed (138)
```

Caller representability (`Number.isSafeInteger` guard deleted):

```
=== mutation: caller-safe-integer -> exit 1 ===
× refuses 0.5 zatoshis as a refusal, never as a thrown RangeError
  → promise rejected "RangeError: The number 0.5 cannot be conv…" instead of resolving
× refuses -0.5 zatoshis …  → promise rejected "RangeError: The number -0.5 …"
× refuses NaN zatoshis …   → promise rejected "RangeError: The number NaN …"
× refuses Infinity zatoshis … → promise rejected "RangeError: The number Infinity …"
× refuses 9007199254740992 zatoshis as a refusal, never as a thrown RangeError
  → expected { kind: 'sent', …(1) } to deeply equal { kind: 'unsendable', error: 4 }
Tests  5 failed | 133 passed (138)
```

**Two existing creator-tip validations, to show they were not weakened:**

Trim-equality (`value === value.trim() &&` deleted):
```
FAIL … > refuses an untrimmed username
FAIL … > refuses an untrimmed label
AssertionError: expected true to be false // Object.is equality
```

Control-character rejection (the C0/DEL regex deleted from `isBoundedText`):
```
FAIL … > refuses a control character in the label
FAIL … > refuses a DEL in the username
AssertionError: expected true to be false // Object.is equality
```

Recipient-whitespace rejection (`!/\s/u.test(value.recipient)` deleted):
```
FAIL … > refuses an address split by a space
```

**The txid length check** — this has a permanent, named test (`never reports an empty payload as a payment`), not just a mutation run by hand, because its failure output is the whole argument for the guard:

```
=== mutation: txid-length -> exit 1 ===
× rejects a fulfilment carrying 1 txid bytes
  → expected { kind: 'sent', txid: '55' } to deeply equal { kind: 'refused', error: 1 }
× never reports an empty payload as a payment
  → expected 'sent' not to be 'sent' // Object.is equality
Tests  6 failed | 132 passed (138)
```

A *correlated* fulfilment carrying no transaction. The UI renders "ZUULI sent your ZEC tip" and a payer has no way to tell it did not happen.

**AFTER RESTORE: GREEN** — 138 passed, and `git diff --stat` confirms no mutation leaked into `wire.ts`, `session.ts` or `format.ts`.

---

## What remains blocked on #461

Everything that would make this *usable*. The wire format, the request, the guards and the failure copy are all correct regardless of how the bytes travel — and none of them travel. Per `PROTOCOL.md` §7, no intent carrying authority may be dispatched over a deep link until verified App Links / Universal Links exist, and `CALLER-AUTHENTICATION.md` §3.1 notes the Android answer may well be `startActivityForResult` rather than App Links, which nobody has measured on a device yet.

So today this ships a tip flow that collects an amount, builds a correct request, and honestly tells the reader that nothing was sent.

---

## Review round (F1 and two nits)

**F1 — four failure kinds, three toast branches.** `unsendable` and `transport-failed` fell through to *"the wallet did not complete this payment"*, which for `unsendable` is **false**: nothing left free2z and ZUULI was never asked. Reachable from untrusted profile data, not just developer error — a `recipient`/`username` carrying `U+200B` or `U+202E` passes free2z's C0/DEL check and is refused by `VisibleText`.

Fixing it surfaced something larger, and **two pre-existing defects**:

1. **`INTENT_UNAVAILABLE` does not mean "nothing happened".** `intent.rs`'s `payment_outcome` returns it for every `BroadcastStatus` but `Accepted` — *including `Unknown`*, where "the transaction exists locally and the wallet retains the exact bytes for `retry_pending_send`, but nothing establishes that the network took them". Copy saying "your ZEC is untouched" there is wrong in the one direction that matters.
2. **This branch's client could not see that status** — but that is a property of the stack, not of `main`. ~~A #911 gap that also affects e2e2z's client.~~ **Retracted: that claim was wrong**, and the correction is verified rather than taken on trust:

   ```
   git show origin/main:wallet/shared/src/intent/error.ts | grep Unavailable
     53:  Unavailable: 12,
     72:  [IntentErrorCode.Unavailable, "INTENT_UNAVAILABLE"],

   git merge-base origin/main HEAD              -> bd39b14 (#909), predates #914
   git show bd39b14:…/error.ts | grep -c Unavail -> 0
   ```

   #914 added status 12 to **both** halves when it introduced it, so the Rust and TypeScript definitions have never disagreed on `main`. This stack is based on #920 → #912, whose base is older than #914, which is the only reason it was missing here. The rebase brings main's definition in.

   What *is* real and worth the test: `IntentSession.accept` maps an unknown status through `intentErrorFromStatus(status) ?? Malformed`, so a client that does not know a status the wallet still emits reports it as "the answer was garbage" rather than "the wallet could not finish, go look" — and for a status that can mean *a transaction may exist*, that is the difference between reassuring a payer and sending them to their wallet.

Four new mutations, each red alone:

```
f1-original-bug         × never tells the payer the wallet declined when it was never asked
                          → expected 'creator.zecTip.declined.title' not to be itself
certainty-overclaim     × is certain only where certainty is earned
                          → expected [ 'no-transport', …(3) ] to deeply equal [ …(2) ]
unavailable-as-declined × treats an ambiguous broadcast as unknown, not as a refusal
                          → expected true to be false
error-status-12         × carries INTENT_UNAVAILABLE through as itself, not as a decode failure
                          → expected { kind: 'refused', error: 1 } to deeply equal { error: undefined }
```

That last one reproduces exactly the reading a client gets for a status it does not know: `error: 1` is `Malformed`. The test pins **main's** definition rather than a local one, so it also fails if a future change drops a status the wallet still emits.

`tip-copy.test.ts` asserts the honesty property against the **real shipped `en`/`es`/`fr` catalogs** through a real i18next instance, not against key names — a reassuring sentence added to the wrong message by a later translation is exactly what a key-level assertion sails past. It carries its own negative control so a typo in every regex cannot leave it green.

**Nit 1 — the capacity-one rationale was imprecise.** The reviewer is right: `accept` matches on `bytesEqual(entry.requestId, response.requestId)`, so cross-acceptance is impossible in a shared session too. What capacity-one buys is **attribution** — `accept` returns `{intent, payload}` without saying which question it answered, so a shared map would leave this function holding a txid it could not tie to a creator or an amount. Corrected in the code and in `PROTOCOL.md` §6.1.

**Nit 2 — the two clocks and the decimal separator.** `Date.now()` at accept vs the injected `now` at issuance now carries a comment saying why (the window must be judged at *arrival*; injecting only issuance is what makes the expiry path drivable). The ZEC error copy now names the decimal separator, since `inputMode="decimal"` on a comma-locale keyboard emits `,`. Playwright caught the stale assertion on that string — the separate suite doing exactly its job. Full i18n of both tabs' inline validation copy remains a shared follow-up with the adjacent 2Z tab.

**Noted, no change:** `installedIntentTransport` is an `export const`, so ES module bindings make it unwritable by importers — nothing to pin, and no scanner rule needed.


skyl:
## Adversarial review — MERGE

Scope isolated to #924's own two commits (`34e2503..ebf1a2a` = 17 files, +1537/−91); #912's and #920's content set aside.

**The ZEC→zatoshi conversion is exact, and I could not break it.** The parser is `/^(\d+)(?:\.(\d{1,8}))?$/` → `padEnd(8,"0")` → `BigInt` — no float ever multiplied by 1e8. Ran the full attack list against the real parser: `1e-8`/`1E8`, `+1`/`-1`, leading/trailing whitespace, `1,5`, empty, `.`, arabic-indic and fullwidth digits, and a trailing newline (JS `$` without `/m` refuses, unlike Python) — **all refused**. 9 decimal places refused rather than rounded, so truncation-vs-rounding is moot. `0.1`/`0.2`/`0.3` → exact, no `0.1+0.2`-class error possible. Long input bounded twice.

**The deleted duplicate `<= 0` was a decoy, not a guard — confirmed.** `encodeExecutePaymentPayload` refuses at `wire.ts:310`, the *first* statement inside `outcome()`, so `creatorTipRequest` returns `null` → `{kind:"unsendable"}` **before `session.issue` and before the transport** — same error, same shape. `-0 → BigInt(-0) → 0n` refused. No input class exists where the caller copy would refuse and the encoder would not. Both surviving halves are independently mutation-pinned, with `expect(transport.sent).toEqual([])` proving nothing dispatched.

**`Number.isSafeInteger` is genuinely the sole guard for `2^53`** — verified empirically: `isSafeInteger(2**53) === false`, `BigInt(2**53)` converts cleanly, and `ByteWriter.u64` only bounds `[0, 2^64-1]`, so without it the value encodes and reaches the wire.

**Fails closed, and tighter than #926.** `failClosedIntentTransport.exchange` is one statement: `return Promise.reject(...)`. And `installedIntentTransport` is an `export const`, not a mutable registry — ES module bindings are read-only to importers, so the setter hole #926 had to close **does not exist here**.

**Contract matches the merged authority as it actually is**, not as guessed: `CREATOR_TIP_CALLER` equals `intent.rs`'s `REGISTERED_CALLERS` key exactly; `purpose` respects `MAX_TEXT_BYTES = 255` and fails *closed* rather than truncating; and the advisory `10_000n` fee is right precisely because `review_matches_request` (`intent.rs:528-548`) does **not** compare fee — declaring `0n` would have fired ZUULI's divergence line on every tip.

**Nothing pre-existing was relaxed:** `git diff -- creator-tip.ts | grep '^-'` filtered of comments returns **zero** executable lines. All bounds, trim-equality, control-char rejection, `Object.freeze` and the 32-entry cap are byte-identical, and `recordCreatorTipIntent` still runs first.

**`{ kind: 'sent' }` is reachable only through a fixed 32-byte read**, with the empty-payload regression carrying a permanent named test (`never reports an empty payload as a payment`) plus a 1/16/31/33/64-byte table — not a hand-run mutation.

### Findings

**F1 (low, accuracy — being fixed).** `features/creator/index.tsx:940-951`: four `CreatorTipFailure` kinds, three toasts. `unsendable` falls through to *"ZUULI did not send this tip"* — but nothing left free2z and ZUULI was never asked. Reachable from untrusted remote profile data (a `U+200B`/`U+202E` in a username passes free2z's C0/DEL check but is refused by `VisibleText`; a 150-**code-point** username can exceed 255 **bytes**). Fails closed, never a false success — but it says a false thing.

**F2 (nit).** The `{ transport }` DI option on `requestCreatorTipPayment` has no mechanical check, contra the "structural" framing in the header comment. Strictly narrower than a writable registry; only the one production call site exists.

**F3/F4 (nits).** Mixed `Date.now()`/injected `now` inside one exchange (correct in production, worth a comment); ZEC error copy hardcoded English while the app ships `es`/`fr`, and `inputMode="decimal"` on a comma keyboard emits `,` with no message saying to use a period.

**F5 (process).** `BEHIND`; #912 and #920 must land first and the rebase-by-shrink check must actually be performed.

**Verdict: MERGE** after the stack lands, the shrink check passes, and the gate goes green. F1 is the only finding worth addressing and has no bearing on whether money can move.

PR #926, opened 2026-09-04
feat(e2e2z): build the issue-device-credential intent, still fail closed (#905)

skyl:
Closes the caller half of `#905`'s `issue-device-credential` intent in
`cash.free2z.e2e2z`. e2e2z's enrollment stops being a bare refusal and becomes a
**real, validated request built through the single intent-bridge client** — and
still fails closed, because there is no authenticated transport.

> **STACKED ON #913.** Branched from `origin/feat/port-messaging-e2e2z`, which is
> reviewed and green but not merged. #913 will squash-merge, so its commits will
> not be ancestors of `main` and this PR currently re-proposes its files. **Before
> merging this**, run:
>
> ```
> git fetch origin && git rebase --onto origin/main f715f60 feat/e2e2z-intent-client
> ```
>
> and verify by the **shrink** — the diff must drop to only the paths listed under
> "What changed" below (21 files, ~300 net lines). If #913's files still show
> large diffs, **STOP**; a bad rebase can silently revert merged work.

## The architecture this rests on

`docs/e2ee/ARCHITECTURE.md` §4.2 splits messaging key material in two:

- **Account keys** (`CeremonySigningKey`, `DirectoryAuthKey`, `BackupWrapKey`) —
  seed-derived, so restoring the mnemonic restores the identity. ZUULI holds the
  seed; e2e2z never will.
- **Device keys** (`DeviceSignatureKey`, `DeviceInitKey`, `QueueKey_{q}`) — OS
  CSPRNG, **never** seed-derived, **never** exported.

The consequence that makes the split work: **ongoing messaging never needs the
seed — only enrollment does.** So enrollment is one request to the wallet
authority, and everything else stays local.

## What changed

**Device keys come from Rust, not the renderer.**
`wallet/e2e2z/src-tauri/src/device.rs` adds `e2e2z_device_credential_keys`, an
**app-crate** command (§2.2: no `plugin:` prefix, no capability entry, not part
of `CLIENT-CONTRACT.md` §3's plugin surface). It calls
`tauri_plugin_f2zmsg::engine::Engine::prepare_device` — the only place in the
tree that samples a device key set, from `rand::rng()` — and returns the
**public** halves only. There is no `crypto.subtle` anywhere in
`src/lib/enrollment/`, and there must never be: a renderer keypair would be a
second cryptographic implementation holding a private key in a
garbage-collected JS heap, and it would still have to be smuggled back into the
engine before a credential over it meant anything.

**The request is built through `wallet/shared/src/intent`, not hand-rolled.**
`src/lib/enrollment/issueDeviceCredential.ts` supplies policy only — caller,
purpose, window — and hands every byte decision to `createIntentSession`,
`encodeIssueDeviceCredentialPayload` and `encodeIntentRequest`. It declares none
of the five names `project-boundary.mjs` reserves. One new decoder,
`decodeIssueDeviceCredentialResult`, lands **in the shared package** beside the
family encoder it mirrors, rather than as a local parser.

**The transport is ONE seam and it is shut.** `src/lib/enrollment/transport.ts`
declares an `IntentTransport` interface and the only implementation there is,
which rejects with a typed `IntentTransportUnavailableError`
(`reason: "intent-transport-not-built"`, `blockedOn: 461`). Two independent
guards: the client checks `available` **before** sampling a device key set —
`prepare_device` replaces the pending set and discards the previous secrets, so
sampling for an undeliverable request throws key material away for nothing — and
`dispatch` rejects **anyway**, unconditionally, without reading that flag. A
guard defeated by editing one boolean is not a guard, and there is a test for
exactly that. When #461 lands the work is `setIntentTransport(...)`; nothing
else on the path changes.

**`enroll` can only reject.** Even a fulfilled response yields a
`DeviceCredential`, and installing one is `install_identity`'s job in a process
this app registers no command for. The refusal is re-wrapped as #913's
`EnrollmentUnavailableError` with the cause attached, so every consumer branch
and every existing test is unchanged. Nothing synthesizes an `EnrollmentStatus`.

**User-visible copy is untouched.** The screen still says enrollment happens in
the wallet app and that the handover is not built yet.
`CLIENT-CONTRACT.md` §2.4's "Superseded" note is updated so it stays accurate
about what `enroll` now does, and every clause of it is still true.

## What the response validation proves — and what it does not

Implemented and tested even though nothing can deliver a response yet, against
**hand-assembled** bytes rather than an encoder that agrees with the decoder by
construction. Refused: an answer to a different request; an identifier differing
in one byte; a correlated answer echoing a different family; an unsupported
version; a truncated envelope; trailing bytes; an empty response; a truncated
family payload; bytes after the credential; a zero-length credential; a refusal
status (never mistaken for success); a refusal carrying a payload; an unknown
status; a replay; an answer after the window closed; a transport that resolves
with a non-response.

**`CALLER-AUTHENTICATION.md` §5 states plainly that there is no signature over
responses.** So all of the above establishes exactly one thing: *whoever
answered had seen the request*, because `request_id` is 32 CSPRNG bytes that
appeared in exactly one outbound message. **None of it establishes that the
responder was ZUULI.** An app that *received* the request holds the identifier
and can answer with a `DeviceCredential` of its own choosing, and this client
would accept it. Correlation is not authentication. §5 also records that there
is no iOS caller attestation and no per-caller enrollment keys. Closing that is
a transport that authenticates the response destination — #461 — which is why
the transport is shut rather than merely undocumented.

## Verification

| Check | Result |
|---|---|
| `wallet/e2e2z`: `npm run typecheck`, `typecheck:tests` | clean |
| `wallet/e2e2z`: `npx vitest run` | 207 passed (14 files) |
| `wallet/e2e2z`: `npx playwright test` | 4 passed |
| `wallet/e2e2z`: `node --test scripts/seed-authority-boundary.node-test.mjs` | 4 passed |
| `wallet/e2e2z`: `npm run build` | clean |
| `wallet/zuuli`: `npx vitest run` | 1394 passed, 5 skipped (95 files) |
| `wallet/zuuli`: typecheck + `messaging-contract.node-test.mjs` | clean, 9 passed |
| `wallet/zuuallet`, `wallet/free2z`: typecheck | clean |
| `node wallet/zuuli/scripts/project-boundary.mjs` | verified, 5 intent-bridge guards |
| `node wallet/zuuli/scripts/surface-capability-authority.mjs` | **e2e2z: zero `zcash:*`** |
| `cargo fmt --check`, `cargo clippy --locked --all-targets -D warnings` | clean |
| `cargo test --locked` (e2e2z) | 4 passed |

## Mutation matrix — the guards are load-bearing

Recorded in `docs/intent-bridge/CONFORMANCE.md`. **10 mutations, 10 failures, 0
survivors**; every one applied, watched to fail with a named assertion, and
restored. The tree ends green.

Fabricating a successful enrollment (after the transport check):

```
× never resolves to an EnrollmentStatus even on a fulfilled response
  → expected true to be false
× sends the wallet authority the request #905 specifies
  → promise resolved "{ enrolled: true, …(5) }" instead of rejecting
+   "enrolled": true,
+   "handle": "alice",
```

Fabricating it *before* the intent path — #913's negative control, intact:

```
× the enrollment gap > refuses every enrollment call with a typed refusal
  → promise resolved "{ enrolled: true, …(5) }" instead of rejecting
× the enrollment gap > never resolves to an EnrollmentStatus, however shaped
  → expected true to be false
```

Dropping correlation in `session.accept` (match any outstanding question):

```
× refuses an answer to a different request                → expected a refusal, resolved with 51,51,…
× refuses an answer whose identifier differs in one byte  → expected a refusal, resolved with 51,51,…
× refuses a replay of an answer it already accepted       → expected a refusal, resolved with 51,51,…
```

Seed-derived paths into e2e2z:

```
not ok 1 - e2e2z holds no route to the wallet seed
  + 'src-tauri/Cargo.toml: e2e2z must not link tauri-plugin-zcash; ongoing messaging never needs the seed'
  + 'src/lib/enrollment/deviceKeys.ts: e2e2z must not address the wallet plugin (plugin:zcash|)'
  + 'src/lib/enrollment/deviceKeys.ts: e2e2z must not name the seed command get_seed_phrase'
```

and the same crate mutation, caught repo-wide:

```
wallet/e2e2z/src-tauri/Cargo.toml: must not link tauri-plugin-zcash; a capability
can only reach a command the process registers
```

Also mutated to red: the shipping transport resolving instead of rejecting; the
transport's refusal gated on `available`; the family-result decoder's
trailing-byte and empty-credential guards.

**Two mutations survived, and CONFORMANCE.md records both rather than hiding
them.** (1) Fabricating a status *after* the transport check leaves
`enrollment-gap.test.ts` green — because with no transport `enroll` refuses
earlier on the same path; that is why the second row of the matrix exists. (2)
Weakening the family decoder's framing leaves `refuses every truncation` green —
`ByteReader.take` bounds-checks before every read, the same documented
redundancy the existing TypeScript matrix already reports.

## Hard constraints held

- e2e2z links **no** `tauri-plugin-zcash` (manifest test + `lib.rs` test +
  repo-wide surface script + new app-local scanner).
- e2e2z grants **zero** `zcash:*` — `surface-capability-authority.mjs` confirms.
- e2e2z reaches **no** seed. I did not at any point need one: §4.2's split is
  exactly why the design works, and the request carries public keys only.
- The plugin's §3 command population is unchanged, so
  `messaging-contract.node-test.mjs`'s three-way comparison is untouched.

## Still blocked on #461

No intent carrying authority can be dispatched. `sign-challenge` and
`execute-payment` have no caller here at all. There is no iOS caller
attestation, no per-caller enrollment key, and no signature over responses. The
credential round trip cannot complete end to end even in principle: this app
registers no command that installs one, and `IssueDeviceCredentialResultV1`
carries no `identity_pk` or `BackupWrapKey`, which `install_identity` needs.
That is follow-up work for whoever lands the transport, and it is named here
rather than discovered then.

### DIFF
--- a/docs/e2ee/CLIENT-CONTRACT.md
+++ b/docs/e2ee/CLIENT-CONTRACT.md
@@ -294,6 +294,22 @@ what travels in the clear.
 > intent, which does not ship before
 > [#461](https://github.com/free2z/zuu/issues/461): a custom-scheme deep link is
 > not an authenticated channel.
+>
+> **`enroll` is now that intent, and it still refuses.** Since #904's caller
+> half landed, `enrollment.enroll` runs
+> `wallet/e2e2z/src/lib/enrollment/issueDeviceCredential.ts`: it reads this
+> device's **public** keys from the app-crate command
+> `e2e2z_device_credential_keys` (which calls
+> `tauri_plugin_f2zmsg::engine::Engine::prepare_device` in process — OS CSPRNG,
+> never seed-derived, private halves never leaving that process), builds the
+> request through `@free2z/wallet-shared`, and hands it to the single transport
+> seam, which rejects. The refusal is re-wrapped as
+> `EnrollmentUnavailableError` with the cause attached, so the paragraph above
+> stays true in every clause: `enroll` cannot resolve, cannot produce an
+> `EnrollmentStatus`, and cannot install a credential — this app registers no
+> command that could. `e2e2z_device_credential_keys` is an app-crate command
+> (§2.2: no `plugin:` prefix, no capability entry) and is deliberately **not**
+> part of §3's plugin surface: it grants nothing and returns no secret.
 
 Messaging mounts as a new top-level route. In `src/lib/routes.ts`:
 
--- a/docs/intent-bridge/CONFORMANCE.md
+++ b/docs/intent-bridge/CONFORMANCE.md
@@ -123,6 +123,67 @@ what keeps a *future* reader safe if someone adds a decode path that forgets
 `finish()` — and claiming an independent test for each would have been the
 overstatement this document exists to avoid.
 
+## The e2e2z caller — `wallet/e2e2z/src/lib/enrollment`
+
+The first application to build an intent. `#904` splits the messaging surface
+away from the wallet, so e2e2z holds device keys and never anything
+seed-derived, and the one operation it cannot perform alone is enrollment
+(`ARCHITECTURE.md` §4.2). It therefore builds an `issue-device-credential`
+request — really, through the shared implementation, over its own OS-CSPRNG
+device public keys — and fails at the transport, because there is not one.
+
+Baseline: 42 tests green across `transport.test.ts`, `deviceKeys.test.ts` and
+`issueDeviceCredential.test.ts`, plus 3 in
+`wallet/e2e2z/src/lib/messaging/enroll-intent.test.ts`, plus the 4 in
+`wallet/e2e2z/scripts/seed-authority-boundary.node-test.mjs`. Reproduce with
+`cd wallet/e2e2z && npx vitest run && node --test scripts/seed-authority-boundary.node-test.mjs`.
+
+| Guard, as mutated | Test that must fail | Result |
+|---|---|---|
+| `enroll` refuses instead of returning a status, **after** a fulfilled response | `enroll builds a real intent and still cannot enroll > never resolves to an EnrollmentStatus even on a fulfilled response` | FAILS |
+| `enroll` refuses instead of returning a status, **before** the intent path | `the enrollment gap > never resolves to an EnrollmentStatus, however shaped` | FAILS |
+| the shipping transport rejects rather than resolving | `the intent transport > never resolves to bytes a caller could mistake for a response` | FAILS |
+| the shipping transport's refusal is not gated on `available` | `the intent transport > does not gate its refusal on the availability flag` | FAILS |
+| session correlation, seen from the caller | `refuses an answer to a different request`, `refuses an answer whose identifier differs in one byte`, `refuses a replay of an answer it already accepted` | FAILS |
+| `IssueDeviceCredentialResultV1` trailing-byte refusal **and** re-encode equality, together | `the issue-device-credential family result > refuses trailing bytes` | FAILS |
+| `IssueDeviceCredentialResultV1` non-empty credential | `the issue-device-credential family result > refuses a zero-length credential` | FAILS |
+| seed authority: the wallet plugin as a dependency | `e2e2z holds no route to the wallet seed` | FAILS |
+| seed authority: a `plugin:zcash\|` invoke in the renderer | `e2e2z holds no route to the wallet seed` | FAILS |
+| seed authority: `get_seed_phrase` named in executable code | `e2e2z holds no route to the wallet seed` | FAILS |
+
+**10 mutations, 10 failures, 0 survivors.** Every one was applied, watched to
+fail with a named assertion, and restored; the tree ends green.
+
+### What survived, and what that means
+
+Two observations are recorded rather than smoothed over, because each says
+something true about where a guard actually lives:
+
+| Mutation | Result | Why |
+|---|---|---|
+| fabricating a status **after** the transport check | `enrollment-gap.test.ts` **still green** | with no transport, `enroll` refuses before it reaches the fabricated line. #913's negative control is intact — it is simply testing an earlier point on the same path, which is why the *second* row above exists |
+| weakening `decodeIssueDeviceCredentialResult`'s framing | `refuses every truncation` **still green** | `ByteReader.take` bounds-checks before every read, so a truncation refuses without `finish()`. The same redundancy the TypeScript table above documents, in the same format, for the same reason |
+
+The seed-authority scan additionally carries its own coverage anchors — a
+fabricated violation of each of its three routes, an assertion that prose about
+the seed is *not* a violation while executable code is, and an assertion that
+the reader saw a manifest, a capability file and more than twenty sources. A
+boundary scanner that has silently stopped finding files reports success
+forever (`#553`).
+
+### What none of this proves
+
+`CALLER-AUTHENTICATION.md` §5: **there is no signature over responses.** Every
+case in `a response is judged as if the responder were hostile` establishes one
+thing — that whoever answered had seen the request, because `request_id` is 32
+CSPRNG bytes that appeared in exactly one outbound message. Not one of them
+establishes that the responder was ZUULI. An app that *received* the request
+holds the identifier and can answer with a `DeviceCredential` of its own
+choosing, and this client would accept it. Only a transport that authenticates
+the response destination closes that, which is
+[#461](https://github.com/free2z/zuu/issues/461), which is why the transport is
+shut.
+
 ## The boundary scanner
 
 `wallet/zuuli/scripts/project-boundary.node-test.mjs` gains seven cases, and
--- a/docs/intent-bridge/CONFORMANCE.md
+++ b/docs/intent-bridge/CONFORMANCE.md
@@ -133,10 +133,11 @@ request — really, through the shared implementation, over its own OS-CSPRNG
 device public keys — and fails at the transport, because there is not one.
 
 Baseline: 42 tests green across `transport.test.ts`, `deviceKeys.test.ts` and
-`issueDeviceCredential.test.ts`, plus 3 in
-`wallet/e2e2z/src/lib/messaging/enroll-intent.test.ts`, plus the 4 in
-`wallet/e2e2z/scripts/seed-authority-boundary.node-test.mjs`. Reproduce with
-`cd wallet/e2e2z && npx vitest run && node --test scripts/seed-authority-boundary.node-test.mjs`.
+`issueDeviceCredential.test.ts`, plus 5 across
+`wallet/e2e2z/src/lib/messaging/enroll-intent.test.ts` and
+`enroll-chunk-failure.test.ts`, plus the 6 in
+`wallet/e2e2z/scripts/authority-boundary.node-test.mjs`. Reproduce with
+`cd wallet/e2e2z && npx vitest run && node --test scripts/authority-boundary.node-test.mjs`.
 
 | Guard, as mutated | Test that must fail | Result |
 |---|---|---|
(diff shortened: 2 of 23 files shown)


skyl:
## Adversarial review — MERGE (after the rebase the body specifies)

Scope isolated first: #926's own work is `f715f60..abd7e40` = **21 files, +2,114/−9**, matching the stated list. `git merge-tree origin/main abd7e40` is **clean**.

**The central question — can anything seed-derived reach e2e2z? No. Confirmed by reading, and I could not construct a path.**
- `engine.rs:475-478` — `prepare_device` calls `prepare_device_material(&mut rand::rng())`; `:3383-3403` fills signing key, queue seed and KEM bytes via `rng.fill_bytes`. **OS CSPRNG, no seed input, no derivation path.**
- `DevicePublicKeys` has exactly two fields, both public; secrets stay in `inner.pending_device`. The `the_response_carries_only_public_halves` test asserts the serialized field set is exactly `["deviceKemPk","devicePk"]`, so a later private field goes red.
- `Cargo.lock` — **zero** matches for `zcash*`, `zip32`, `bip0039`, `orchard`, `sapling`, `secp256k1`, `mnemonic`, `tauri-plugin-zcash`.
- Capabilities **parsed as JSON**: 46 permissions each — `core:default`, `deep-link:default`, `opener:default`, 43 `f2zmsg:allow-*`. Zero `zcash:*`, no blanket `f2zmsg:default`, no `allow-enroll*`.

**Fail-closed: both guards confirmed genuinely independent.** Guard 1 (`issueDeviceCredential.ts:215`) refuses *before* `readKeys()`, pinned by a spy assertion. Guard 2 (`transport.ts:142`) is a bare `Promise.reject` that never reads `available`, pinned by a test that spreads `{...unavailableIntentTransport, available: true}` and still fails. Delete guard 1 and dispatch is still impossible — it only wastes a key sample first. #913's `enrollment-gap.test.ts` negative control is **byte-identical** and unweakened.

**Both admitted survivors checked hard, both benign — not the #914 class.** (a) is in fact *pinned* by `enroll-intent.test.ts`, which installs an available transport returning a fulfilled response so control reaches the very line `enrollment-gap.test.ts` cannot. (b) For a one-field `opaque24`, `finish()` and re-encode-equality are **mutually total** — delete either and the other catches trailing bytes — so neither deletion loses anything, and CONFORMANCE honestly records they were mutated together.

**App-crate command ACL:** `e2e2z_device_credential_keys` *is* webview-invokable with no capability, which is correct per CLIENT-CONTRACT §2.2 and the same mechanism `f2zmsg_enroll` and `oauth_loopback_start` already ship on. Repeated invocation yields public keys only. The `pending_device` clobber is latent, not live — e2e2z registers no `install_identity` path, and when one exists `install_identity` refuses if the credential does not bind the prepared key, so a mid-flight swap fails closed rather than sealing a mismatch.

**Guards survive:** `messaging-contract.node-test.mjs` genuinely untouched (the new command appears in none of its three inputs, and the CLIENT-CONTRACT addition is a blockquote the `^\|` table parser cannot see). `project-boundary.mjs`'s five reserved names are declared by none of this. `seed-authority-boundary.node-test.mjs` **is load-bearing** — a `plugin:zcash|get_seed_phrase` invoke produces two failures, and the file proves this itself with a fabricated violation.

### Findings — none blocking

1. **[Follow-up, not this PR — and it corrects #928]** `IdentityInstall.wrap_key` is documented at `engine.rs:206-217` as **the seed-derived `BackupWrapKey`**. So widening `IssueDeviceCredentialResultV1` to carry it would ship seed-derived material into e2e2z — the exact thing §4.2 forbids. The round trip cannot be completed without redesigning where sealing happens. #928 has been corrected; this PR correctly does not carry it.
2. **[Nit]** `bridge.ts:682` — the dynamic `import()` sits *outside* the `try`, so a chunk-load failure escapes as an untyped error rather than `EnrollmentUnavailableError`. Still fails closed, but breaks the typed contract.
3. **[Suggestion]** `transport.ts:160` `setIntentTransport` is exported from shipping code with nothing pinning that no production module calls it — the symmetric case of this PR's own "a guard defeated by editing one boolean is not a guard".
4. **[Informational, pre-existing]** `device_kem_pk` is 1,216 random bytes **with no private half** (`engine.rs:3391-3393`), a documented placeholder. #926 is the first code that would transmit it, so once a transport lands ZUULI would sign a credential attesting a key nobody holds. `device.rs:71-79` carries the caveat faithfully rather than inventing a key — the right call.

Items 2 and 3 are being addressed. Neither is a reason to hold.

skyl:
## Follow-up: both review findings addressed (`f8404e1`)

Neither changed what the app can do — both paths were already fail-closed — but each was true by accident rather than pinned. Thanks for both; the second in particular is my own argument pointed back at me.

**1. `bridge.ts` — the lazy `import()` moves inside the `try`.**

Correct as reported. A chunk that fails to load escaped as an untyped error, and while it could not produce an `EnrollmentStatus`, the screen branches on `isEnrollmentUnavailable` and would have shown "something broke" instead of the standing "enrollment happens in the wallet app" state — the one true thing this app can say. Every way `enroll` can fail now wears the typed refusal, with the real cause still attached.

New test in its own file, `src/lib/messaging/enroll-chunk-failure.test.ts`: it uses `vi.resetModules()` + a throwing `vi.doMock` factory, and a reset registry orphans module instances other tests in a file have already captured — `enroll-intent.test.ts` registers a transport into exactly such an instance, so putting it there turned that test red for an unrelated reason. The file comment records why.

Mutation (import moved back above the `try`):

```
× a chunk that never loads > still refuses with the typed enrollment refusal
  → expected Error: [vitest] There was an error when m… to be an instance of EnrollmentUnavailableError
```

Restored; 2 passed.

**2. `setIntentTransport` is now pinned to test files.**

Accepted without reservation: a guard defeated by one `setIntentTransport(...)` call from any renderer module has exactly the shape of the guard defeated by one boolean that `transport.ts` argues against. The registry is not a guard if anything in the renderer may write to it.

The scanner grew the rule, and grew a name to match: `scripts/seed-authority-boundary.node-test.mjs` → **`scripts/authority-boundary.node-test.mjs`**, because it now judges *dispatch* authority alongside *seed* authority and the old name understated it. `resetIntentTransport` is deliberately unrestricted — it restores the fail-closed default and can only ever narrow what the app can do.

Two coverage anchors came with it, since a rule about a name nobody declares is a rule about nothing:

- `setIntentTransport` is permitted from `*.test.ts` / `*.pw.ts` and from its declaring module, refused elsewhere, and prose that merely names it is not an installation.
- If `transport.ts` ever stops exporting `setIntentTransport`, the scan **fails** rather than silently protecting nothing (#553).

Mutation (a `setIntentTransport` call added to `src/features/messages/index.tsx`):

```
not ok 1 - e2e2z holds neither seed authority nor dispatch authority
  + 'src/features/messages/index.tsx: only a test may call setIntentTransport;
     installing a transport is dispatch authority, and #461 has not landed'
```

Restored; 6 passed.

**CONFORMANCE.md is now 12 mutations, 12 failures, 0 survivors**, and records why these last two rows exist.

Re-verified after the change: e2e2z typecheck ×2 clean, `npm test` — vitest **209 passed (15 files)**, node tests **8 passed**, playwright **4 passed** — `npm run build` clean; `project-boundary.mjs` verified (5 intent-bridge guards); `surface-capability-authority.mjs` — e2e2z zero `zcash:*`; `messaging-contract.node-test.mjs` 9/9.

---

## Correction to this PR's "Still blocked on #461" section

The description above says the round trip could be completed by widening `IssueDeviceCredentialResultV1` to carry `identity_pk` and `BackupWrapKey`. **That is wrong and I am striking it.** `IdentityInstall.wrap_key` is documented at `engine.rs:206-217` as the seed-derived `BackupWrapKey` (§4.2), so widening the wire format that way would ship a seed-derived key into e2e2z — breaching the exact split this PR upholds. The real question is *where sealing happens* (e2e2z sealing under a device-local key, or ZUULI sealing and returning only sealed material), which is a design decision, not a format change. See #928, which has been corrected.

This PR does not carry that, and nothing here needs to change.

Also on #928: `device_kem_pk` is 1,216 random bytes with no private half, so once a transport lands ZUULI would sign a credential attesting a key nobody holds. `device.rs:71-79` carries that caveat from `DevicePublicKeys` faithfully rather than inventing a key.

**Not rebasing yet** — #913 has not merged. On your signal: `git rebase --onto origin/main f715f60 feat/e2e2z-intent-client`, verified by the shrink.

skyl:
## Rebased onto `ed672401` — the shrink is confirmed

`git rebase --onto origin/main f715f607 feat/e2e2z-intent-client` — **clean, no hand-resolution on any file.** New head `19791d5f`.

| | before | after |
|---|---|---|
| changed files | 89 | **22** |
| additions | 5,820 | **2,338** |
| deletions | 399 | **10** |

22 rather than the 23 I predicted: the `seed-authority-boundary.node-test.mjs` → `authority-boundary.node-test.mjs` rename collapses to a single add against `main`, where the old filename never existed. My earlier count double-counted it.

**#913's files are gone from the diff.** Explicitly checked for the ones you named — `wallet/e2e2z/src/features/messages/**`, the i18n catalogs, `components/ui`, `Transcript`, `FirstContact`, `mock.ts`, `types.ts`, `events.ts`: **none present.** `bridge.ts` does appear, at 59 changed lines, and its diff is exclusively my `enroll` rewrite and the `cause` field — nothing of #913's is re-proposed.

**On the #914 hazard:** I never touched `wallet/shared/src/intent/error.ts` or `rs/crates/f2z-intent/src/error.rs`, so nothing was hand-merged. Confirmed on the rebased tree: `Unavailable: 12` appears exactly once (`error.ts:53`) and its `NAMES` row exactly once (`:72`). My only shared change is one added line in `index.ts` re-exporting `decodeIssueDeviceCredentialResult`, which merged cleanly alongside #914's exports.

## Re-verified against the new main

Full reinstall in both `wallet/e2e2z` and `wallet/zuuli` first, since `main` moved.

| Check | Result |
|---|---|
| e2e2z `npm run typecheck` / `typecheck:tests` | clean |
| e2e2z `npm test` — vitest | **209 passed (15 files)** |
| e2e2z `npm test` — node tests incl. `authority-boundary` | **8 passed** |
| e2e2z `npm test` — **playwright** | **4 passed** |
| e2e2z `npm run build` | clean |
| zuuli `npx vitest run` | **1394 passed, 5 skipped (95 files)** |
| zuuli `src/lib/intent-bridge.test.ts` | 28 passed |
| `messaging-contract.node-test.mjs` | 9 passed |
| `project-boundary.mjs` | verified, 5 intent-bridge guards |
| `surface-capability-authority.mjs` | **e2e2z zero `zcash:*`** |
| `cargo fmt --check`, `cargo clippy --locked --all-targets -D warnings` | clean |
| `cargo test --locked` (e2e2z) | 4 passed |

`gh pr view 926` now reports `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — awaiting checks, not `DIRTY` and not `BEHIND`.

Mutation campaign not re-run: the rebase forced no hand-resolution in `issueDeviceCredential.ts`, `transport.ts`, `bridge.ts` or the wire code, per your point 3.

## Noted: #930

Thanks — `IntentSession.accept`'s `intentErrorFromStatus(status) ?? Malformed` collapsing an unknown status into "the answer was garbage" rather than "the wallet is newer than I am" does reach my client, since `requestDeviceCredential` surfaces whatever `accept` returns. My `refuses an unknown refusal status rather than guessing` case asserts the *current* behaviour (status 4242 → `Malformed`), so if #930 changes it to a distinguishable code that test will need updating in the same change — flagging it so whoever takes #930 knows to expect it rather than being surprised by a red e2e2z suite.

Issue #915, opened 2026-09-04
ci: free2z and e2e2z test suites are advisory — not required by branch protection


skyl:
Found during adversarial review of #912.

Branch protection on `main` requires exactly two contexts:

```
["gate", "rs / gate"]
```

`zuuli.yml`'s `gate` depends on `frontend`, which runs `project-boundary.mjs` and
`surface-capability-authority.mjs` — so the **security** guards for the new apps
*are* required. That part is fine and is the load-bearing half.

But the actual test suites for the two new apps run in `wallet-surfaces.yml` as
`surfaces_frontend`, which is **not** a required context. As of #912 that is
677 vitest + 41 Playwright + ui-copy tests for `wallet/free2z`, and the
equivalent for `wallet/e2e2z` once #913 lands. **They can go red without
blocking a merge.**

This is pre-existing from #906, not introduced by #912 — but it matters more now
that the two apps hold real ported surface rather than placeholder screens, and
it undercuts the evidentiary weight of the test tables our PR bodies present.

### Acceptance criteria

- [ ] Decide whether `surfaces_frontend` becomes a required context, or whether
      `gate` should depend on it (prefer whichever keeps ONE required gate rather
      than growing the required-contexts list — see #846 on gate cost).
- [ ] Verify the exact context name as GitHub reports it before adding it;
      a wrong name silently blocks every PR forever.
- [ ] Confirm the change does not deadlock PRs that touch neither new app —
      the check must still report (or be skipped in a way branch protection
      accepts) when its paths are untouched. This is the classic
      required-but-skipped deadlock; get it right or the repo stops merging.
- [ ] Prove it: a PR with a deliberately failing free2z test must be unmergeable.

### Related

- #912 (where this was found), #906 (scaffold that introduced the workflow),
  #910 (path-filter cost), #846 (gate cost), #904 (architecture epic).

PR #934, opened 2026-09-05
fix(e2e2z): hold the moved chat components to the RTL/bidi source policy

### LINKED ISSUE #917: e2e2z: port RTL/bidi source policy enforcement for the moved chat components
Disclosed by #913 as a known follow-up; filing so it is tracked.

`wallet/zuuli/scripts/rtl-source-policy.mjs:85` `CHAT_OWNED_RESIDUALS` is now `{}`
— which is correctly **stricter** for ZUULI, since the chat components left. But
the components themselves (`Transcript.tsx`, `FirstContact.tsx`,
`BrowserGuarantee.tsx`) are now held to **nothing** in `wallet/e2e2z/`.

Impact is low today: e2e2z ships en/es/fr only, none RTL. It becomes real the
moment an RTL locale is added, and chat is exactly where bidi handling matters —
user-authored content rendered next to UI chrome is the classic bidi-override
spoofing surface.

### Acceptance criteria

- [ ] An RTL/bidi source policy runs against `wallet/e2e2z/` components,
      equivalent to what `rtl-source-policy.mjs` enforced before the move.
- [ ] Proven load-bearing: the check FAILS on a deliberately introduced
      violation, then passes after revert.
- [ ] Decide whether the checker is shared across surfaces or duplicated —
      prefer shared, since #912 already opened a duplication drift window.

Related: #913, #904.

skyl:
Closes #917.

`wallet/zuuli/scripts/rtl-source-policy.mjs` enforced RTL/bidi handling on the chat components until #913 moved them to `wallet/e2e2z/`. Emptying ZUULI's `CHAT_OWNED_RESIDUALS` was correct and made ZUULI **stricter** — but it left `Transcript.tsx`, `FirstContact.tsx` and `BrowserGuarantee.tsx` held to **nothing**. Chat is precisely where bidi handling counts: a message body is user-authored text rendered next to identity chrome, which is the classic bidi-override spoofing surface.

## Shared checker, not a duplicated one

`rtl-source-policy.mjs` now walks a `SURFACES` list, in the same shape `surface-capability-authority.mjs` already uses for the same two apps:

```js
export const SURFACES = Object.freeze([
  Object.freeze({ directory: "zuuli", reviewedResiduals: CHAT_OWNED_RESIDUALS, requiredDirectionalTransforms: ZUULI_REQUIRED_DIRECTIONAL_TRANSFORMS }),
  Object.freeze({ directory: "e2e2z", reviewedResiduals: E2E2Z_OWNED_RESIDUALS, requiredDirectionalTransforms: Object.freeze({}) }),
]);
```

Each surface carries only what is genuinely its own — its reviewed residual inventory and its reviewed directional-transform sites, keyed by paths relative to that surface's project root. The ~1100 lines that actually judge physical Tailwind utilities, physical CSS declarations and shorthands, inline styles, directional Lucide icons, numeric bidi isolation and the document-direction bootstrap stay single-copy. #912 already opened one duplication drift window with the Markdown/RemoteMedia copies; a second copy of an AST policy this size would be a worse one, because the copies would diverge silently in exactly the direction that weakens them.

`assertRtlSourcePolicy(sources, surface = ZUULI_SURFACE)` keeps its original one-argument call shape, so all 17 existing tests are unchanged.

## e2e2z gets no exemption — the residuals are fixed at the source

`E2E2Z_OWNED_RESIDUALS` is `{}`, like ZUULI's. The moved screens did not carry their old counted exemption across; the four residuals the policy found are fixed instead:

- **`Transcript.tsx`** — the `Send` composer icon mirrors (`rtl:-scale-x-100`); the delivery receipt is `text-end`, not `text-right`.
- **`index.tsx`** — the conversation list is `text-start`, not `text-left`; the relay and witness counters are `bidi-number numeral`. "3 of 5" reads as "5 of 3" under an RTL base direction without isolation.
- **`index.css`** — the `.bidi-number { direction: ltr; unicode-bidi: isolate; }` rule those counters name.

## The bootstrap that makes any of it reachable

e2e2z had **no document-direction install**, so every `rtl:` variant in the tree was dead code no locale could ever exercise. `src/i18n/index.ts` already owns `<html lang>`, so ZUULI's kernel is restated as `src/lib/document-direction.ts` — a copy for the same documented reason `src/i18n/locale.ts` is one (`project-boundary.mjs` forbids one wallet app importing another) — and installed before the first render. Adding an RTL catalog is now a catalog change and nothing else.

Verified in the built CSS, not just in source: `[dir=rtl] … {--tw-scale-x: -1}`, `text-start{text-align:start}`, `text-end{text-align:end}`, `.bidi-number{direction:ltr;unicode-bidi:isolate}`.

## Where it runs — inside the required gate

`scripts/check-workflow-gates.mjs` records `wallet-surfaces.yml` as **ungated** (#915), so a check that only ran there could not block a merge. This needs no new wiring: `zuuli.yml`'s `frontend` job already runs `node scripts/rtl-source-policy.mjs` and `node --test scripts/rtl-source-policy.node-test.mjs`, and its path filter already includes `wallet/e2e2z/*`. Covering e2e2z from this script therefore puts it behind the required `gate`, the same way `project-boundary.mjs` and `surface-capability-authority.mjs` are. `wallet/zuuli`'s `npm run test` picks it up too. No workflow file is touched.

## Proven load-bearing

Failures now name **file and line**. Residuals report `file (token@line)`; a non-mirroring directional icon reports `file:line`.

Violation introduced (this is byte-for-byte the code on `main` today, before this PR's fixes):

```
$ node scripts/rtl-source-policy.mjs
RTL source policy: zuuli OK
Error: src/features/messages/Transcript.tsx:224 Send (Send) must mirror with literal rtl:-scale-x-100
exit=1
```

With only the `text-right` residual left in place:

```
$ node scripts/rtl-source-policy.mjs
RTL source policy: zuuli OK
Error: e2e2z: RTL residual paths changed
expected:
actual: src/features/messages/Transcript.tsx (text-right@70)
exit=1
```

Reverted byte-identically (`sha256 766c9d1e…9843` before mutation, identical after revert):

```
$ node scripts/rtl-source-policy.mjs
RTL source policy: zuuli OK
RTL source policy: e2e2z OK
exit=0
```

## Verification

- `node --test scripts/rtl-source-policy.node-test.mjs` — **25 pass, 0 fail** (the 17 existing tests unchanged, 8 e2e2z tests added). `CHAT_OWNED_RESIDUALS` is still `{}` and still enforced as empty.
- All three moved components carry a residual probe, so an exemption cannot be smuggled back in as a directory rule a single carrier file would not reveal.
- `npm run test:rtl` in `wallet/zuuli` — vitest + `tests/bidi-identifiers.pw.ts` (2 passed).
- `wallet/e2e2z`: `npm run typecheck`, `npm run typecheck:tests`, `npx vitest run` (**221 passed**, incl. 12 new `document-direction.test.ts`), `npx playwright test` (**4 passed**), `node --test scripts/ui-copy-truncation.node-test.mjs scripts/authority-boundary.node-test.mjs` (**8 passed**), `npm run build`.
- `node scripts/project-boundary.mjs` — clean across 5 projects / 561 source files.
- `node scripts/surface-capability-authority.mjs` — clean; e2e2z still grants **zero** `zcash:*`.

## What is still not covered, honestly

- **`wallet/free2z` and `wallet/zuuallet` are not in `SURFACES`.** free2z is a separate component tree needing its own residual review; that is its own change, not a line added here. This is documented in the `SURFACES` doc comment rather than left silent.
- **No runtime RTL browser test for e2e2z.** ZUULI has `tests/bidi-identifiers.pw.ts`; e2e2z's coverage is source-level plus the unit-tested direction kernel. Worth adding when e2e2z actually ships an RTL catalog.
- **e2e2z ships `en`/`es`/`fr` only**, none RTL. The policy and the bootstrap are in place ahead of that, which was the point of #917: cheap now, awkward later.

### DIFF
--- a/wallet/e2e2z/src/features/messages/Transcript.tsx
+++ b/wallet/e2e2z/src/features/messages/Transcript.tsx
@@ -67,7 +67,7 @@ function MessageRow({ message }: { message: Message }) {
           {message.body.text}
         </p>
         {outbound && (
-          <p className="text-right text-xs text-muted-foreground">
+          <p className="text-end text-xs text-muted-foreground">
             {DELIVERY_COPY[message.delivery.state]}
           </p>
         )}
@@ -221,7 +221,7 @@ export function Transcript({ conversation }: { conversation: Conversation }) {
           aria-label="Send message"
           className="min-tap"
         >
-          <Send className="size-4" aria-hidden />
+          <Send className="rtl:-scale-x-100 size-4" aria-hidden />
           Send
         </Button>
       </form>
--- a/wallet/e2e2z/src/features/messages/index.tsx
+++ b/wallet/e2e2z/src/features/messages/index.tsx
@@ -80,7 +80,7 @@ function EngineSummary({ status }: { status: EngineStatus }) {
 
       <div className="rounded-xl border border-border bg-card p-4">
         <dt className="eyebrow text-muted-foreground">Relays</dt>
-        <dd className="numeral mt-1 text-foreground">
+        <dd className="bidi-number numeral mt-1 text-foreground">
           {status.relaysConnected} of {status.relaysConfigured}
         </dd>
         <dd className="mt-1 text-sm text-muted-foreground">
@@ -90,7 +90,7 @@ function EngineSummary({ status }: { status: EngineStatus }) {
 
       <div className="rounded-xl border border-border bg-card p-4">
         <dt className="eyebrow text-muted-foreground">Witnesses</dt>
-        <dd className="numeral mt-1 text-foreground">
+        <dd className="bidi-number numeral mt-1 text-foreground">
           {status.independentWitnesses}
         </dd>
         <dd className="mt-1 text-sm text-muted-foreground">
@@ -373,8 +373,8 @@ export default function MessagesFeature() {
                 }
                 className={
                   conversation.conversationId === selectedId
-                    ? "w-full rounded-xl border border-primary/50 bg-card p-3 text-left"
-                    : "w-full rounded-xl border border-border bg-card p-3 text-left transition-colors hover:border-primary/50"
+                    ? "w-full rounded-xl border border-primary/50 bg-card p-3 text-start"
+                    : "w-full rounded-xl border border-border bg-card p-3 text-start transition-colors hover:border-primary/50"
                 }
               >
                 <span className="mono-id block text-foreground">
--- a/wallet/e2e2z/src/index.css
+++ b/wallet/e2e2z/src/index.css
@@ -177,6 +177,19 @@
   }
 
   /*
+   * A number is written left-to-right in every locale, including RTL ones. Left
+   * to the paragraph's own direction, "3 of 5" reorders to "5 of 3" under an
+   * RTL base direction, and a run of digits adjacent to RTL text drags its
+   * neighbours around with it. Isolating each numeric run keeps the reading
+   * that was measured. `rtl-source-policy.mjs` requires `bidi-number` wherever
+   * `numeral` or `tabular-nums` appears, so this rule is what that class means.
+   */
+  .bidi-number {
+    direction: ltr;
+    unicode-bidi: isolate;
+  }
+
+  /*
    * Opaque identifiers — handles, fingerprints, error codes. Same face as
    * amounts, with the slashed zero on, because here a 0 really can be misread
    * as an O and the value is transcribed by hand.
--- a/wallet/e2e2z/src/lib/document-direction.test.ts
+++ b/wallet/e2e2z/src/lib/document-direction.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from "vitest";
+import {
+  directionForLocale,
+  installDocumentDirection,
+  syncDocumentDirection,
+  type DirectionElement,
+  type DirectionObserver,
+} from "./document-direction";
+
+describe("document locale direction", () => {
+  it.each(["ar", "ar-EG", "fa-IR", "he-IL", "ur-PK"])(
+    "maps %s to RTL",
+    (locale) => expect(directionForLocale(locale)).toBe("rtl"),
+  );
+
+  it.each(["en", "es-419", "fr-FR", "", "not_a_locale"])(
+    "fails %j safely to LTR",
+    (locale) => expect(directionForLocale(locale)).toBe("ltr"),
+  );
+
+  it("writes the direction derived from the current language", () => {
+    const root = { lang: "fa-AF", dir: "ltr" };
+    expect(syncDocumentDirection(root)).toBe("rtl");
+    expect(root).toEqual({ lang: "fa-AF", dir: "rtl" });
+  });
+
+  it("observes only lang and re-syncs every locale change", () => {
+    const root: DirectionElement = { lang: "en", dir: "rtl" };
+    let callback: (() => void) | undefined;
+    let observed:
+      | {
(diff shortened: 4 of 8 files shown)


skyl:
Rebased onto `3ba3ffa8` and fixed.

**Cause, precisely:** #943 already removed `src/features/profile/LinkedAccounts.tsx` from `ZUULI_REQUIRED_DIRECTIONAL_TRANSFORMS` on main, and the rebase carried that in cleanly. The stale reference was the hand-written key list in *my* new test `the shared checker keeps each surface's contract to itself`, which still expected three entries. Corrected to the surviving two.

**Sweep result:** the only stale path in either file. Every remaining path in `SURFACES` (`src/components/ui/switch.tsx`, `src/features/auth/ZcashLoginFlow.tsx`, plus `index.html` / `src/main.tsx` / `src/index.css` for both surfaces) exists. Nothing in this PR referenced `features/{articles,creator,live,ai,search,profile,kyc}` or the deleted content tree — main's own rebase of the pre-existing tests had already moved those mutation carriers to `home/Hero.tsx`, `home/VaultActions.tsx`, `wallet/funding/{ActivityTab,SendTab}.tsx` and `ui/dropdown-menu.tsx`. ZUULI's entry does not name `LinkedAccounts.tsx` in any form, so it makes no claim on free2z's copy; free2z stays out of `SURFACES` per #940.

**Added the guard that should have caught it.** Correcting a hand-written list does not stop the next deletion, so there is now a sweep asserting that every path in every surface's reviewed inventory — directional transforms *and* residuals, both surfaces — names a file the collected tree still contains. An inventory entry pointing at a deleted file is not stricter enforcement; it is a dead entry that reads like coverage.

Proven load-bearing by reinstating the exact stale entry:

```
not ok 25 - every reviewed inventory path names a file that still exists
  error: zuuli reviews src/features/profile/LinkedAccounts.tsx, which no longer exists
```

It names the cause rather than a list diff. Reverted byte-identically (`sha256 b35fa1b9…d9fa` before and after).

**Load-bearing re-proved on e2e2z after the rebase** — `bidi-number` dropped from both counters and one `text-start` flipped:

```
RTL source policy: zuuli OK
Error: e2e2z: RTL residual paths changed
expected:
actual: src/features/messages/index.tsx (numeral@83, numeral@93, text-left@376)   → exit 1
```

Reverted byte-identically (`sha256 7edd70b5…803d`) → `zuuli OK` / `e2e2z OK`, exit 0.

**Re-verified on the rebase**, with all four projects reinstalled from their current lockfiles:

- `node --test scripts/rtl-source-policy.node-test.mjs` — **26 pass, 0 fail** (main's 17 unchanged, 9 added). `CHAT_OWNED_RESIDUALS` and `E2E2Z_OWNED_RESIDUALS` both still `Object.freeze({})`.
- `node scripts/rtl-source-policy.mjs` — `zuuli OK` / `e2e2z OK`.
- `npm run test:rtl` in zuuli — green, incl. `bidi-identifiers.pw.ts` (2 passed) against the post-#943 tree.
- e2e2z: typecheck ×2, vitest **221 passed**, playwright **4 passed**, node tests **8 passed**.
- `project-boundary.mjs` — clean (5 projects, 472 source files; was 561 before #943).
- `surface-capability-authority.mjs` — clean; e2e2z still **zero** `zcash:*`.

Nothing else in this PR touches ZUULI paths — the four e2e2z residual fixes, the `document-direction` install and the built-CSS verification are unchanged.

Issue #929, opened 2026-09-04
intent-bridge: response authenticity — correlation is not authentication, and two clients now depend on it


skyl:
Raised independently by both caller-side implementations (#924 free2z, #926 e2e2z).
`docs/intent-bridge/CALLER-AUTHENTICATION.md` §5 already records that **there is
no signature over responses**. This issue exists because that limitation now has
two concrete dependents and becomes load-bearing the moment #461 lands.

### The precise statement

Both clients correlate a response to their outstanding request. Correlation
proves the responder **saw** the request. It does **not** prove the responder
**is** ZUULI.

Concretely: any app that received the request holds the 32-byte request
identifier, and can answer with a payload of its own choosing — a fabricated
`DeviceCredential` for e2e2z, or a fabricated txid for free2z. Both clients would
accept it as well-formed. #924 additionally notes a `sent` outcome cannot be
verified against the network; that fact is ZUULI's `BroadcastStatus::Accepted`,
which the caller has no independent way to check.

Today this is unreachable — there is no transport, so nothing can deliver a
response at all. **It becomes live on the same day #461 does.**

### The design tension, from §5

Adding a response signature needs a ZUULI response key, which is a **second
identity for the wallet alongside the seed hierarchy** — and §5 argues that
minting one to paper over a transport gap is how a system ends up with two
identities that disagree. Its position is that the transport is the right layer.
That may well be correct; the point of this issue is that the decision must be
made **explicitly, before the transport ships**, rather than inherited by default.

### What must be decided

- [ ] Does response authenticity come from the **transport** (e.g. Android
      `startActivityForResult`, where the OS attests the responding component),
      or from a **signature**, or from both per-platform?
- [ ] If transport-derived: what is the iOS story? §5 states plainly there is no
      iOS caller attestation of any kind, and §6 concludes that on iOS the native
      confirmation is the only control. Does that asymmetry hold for *responses*
      too, and is that acceptable for `issue-device-credential`?
- [ ] Whatever is chosen, both clients must **verify**, not assume — with a test
      proving a response from a non-authority is rejected.

Until then, neither client should present a response as authoritative, and
neither currently does. Both fail closed.

Related: #924, #926, #461, #905, #904, #367.

Jhonattan2121:
Looked into this closely. All three checkboxes are already answered in what's on `main`, so I think this can close.

[...]

Issue #928, opened 2026-09-04
intent-bridge: IssueDeviceCredentialResultV1 cannot complete the enrollment round trip


skyl:
Found while building the e2e2z caller in #926, and stated there rather than left
to be discovered after the transport lands.

Even if #461 shipped a verified transport tomorrow, **the
`issue-device-credential` round trip could not complete**:

1. `IssueDeviceCredentialResultV1` carries **no `identity_pk` and no
   `BackupWrapKey`** — both of which `install_identity` requires.
2. **e2e2z registers no install command**, so there is no process step that could
   consume a returned credential even if it were complete.

So `enroll` in e2e2z can only ever reject today. That is correct fail-closed
behaviour given no transport exists, but the gap is in the **protocol result
type**, not just the transport — and it will not be noticed by anyone testing the
transport alone, because the failure appears one step later.

### Why this is worth fixing before the transport, not after

The transport is the thing everyone is waiting on (#461). When it lands there
will be pressure to declare enrollment working. This gap guarantees it will not
be, and diagnosing it at that moment — with a freshly-landed transport as the
prime suspect — is strictly worse than fixing the result type now while the
protocol is still unshipped and cheap to change.

### Acceptance criteria

- [ ] `IssueDeviceCredentialResultV1` carries everything `install_identity` needs
      — at minimum `identity_pk` and `BackupWrapKey` — with the wire format
      updated in **both** halves (`rs/crates/f2z-intent`, `wallet/shared/src/intent`)
      and canonical decode/re-encode equality preserved (WIRE.md §3.3).
- [ ] A decision, written down, on **where the install step lives** in e2e2z and
      what capability it needs — bearing in mind e2e2z must keep granting zero
      `zcash:*` and holding nothing seed-derived (ARCHITECTURE.md §4.2).
- [ ] Conformance vectors updated, and a test that would fail if a field
      `install_identity` needs is dropped from the result.
- [ ] ZUULI's authority side (`wallet/zuuli/src-tauri/src/intent.rs`) currently
      refuses `issue-device-credential` with `INTENT_UNKNOWN_INTENT`; note what
      implementing it will require, but do not treat that as in scope here.

Related: #926, #905, #461, #904, #305.

skyl:
## Correction — my acceptance criteria above are WRONG and must not be followed

Adversarial review of #926 found what I missed when filing this. Flagging it
prominently because the fix I proposed would breach the architecture.

[...]

PR #935, opened 2026-09-05
docs(e2ee): ADR 0016 — where enrollment sealing happens (#928)

skyl:
Closes the design question in #928 — the one your correction comment reframed
from "widen the wire format" into "decide where sealing happens". **`docs/` only;
no code changes.**

New: `docs/e2ee/decisions/0016-enrollment-sealing-boundary.md`, following the
existing ADR convention. Cross-linked from `docs/e2ee/README.md` (index row +
the `0001`–`0015` range), `docs/intent-bridge/PROTOCOL.md` §3.3 (where the
result struct is defined, so an implementer finds it), and `docs/status.md`
(the #928 "what is blocked" row).

## What it decides

**A. Where sealing happens — the app that generated the secrets seals them,
under a per-device `DeviceWrapKey` it samples from the OS CSPRNG and holds.**
`IdentityInstall.wrap_key` is removed — for ZUULI as well as e2e2z. One sealing
scheme, not two.

The argument that decides it is a fact #928 does not mention: the seal's
*contents* are `{ device_signing_key, queue_seed }`, both sampled from
`rand::rng()` (`engine.rs:3383-3402`) and by construction **not recoverable
from the seed** — so an account-scoped key sealing them gains nothing for
restore and costs blast radius (one key, every device, life of the mnemonic).
Meanwhile the job §4.2 actually assigns `BackupWrapKey` — wrapping local
history — is unimplemented (`store.rs:40-50`, which says so). The key is doing
a job it was not specified for and not doing the one it was.

The other two options are argued and rejected on the record:

- **ZUULI seals and returns sealed material** is *impossible*, not merely
  undesirable: ZUULI cannot seal what it does not have, and the device secrets
  live in e2e2z's plugin process (`engine.rs:489`). Getting them to ZUULI means
  exporting device private keys over the bridge — forbidden by §4.2 and by
  `PROTOCOL.md` §1's stated invariant. It trades one breach for a worse one.
- **ZUULI holds the wrap key and unseals per launch** fails three ways
  independently: runtime coupling #904 split the apps to remove; a continuous
  grant wearing a five-minute window (`PROTOCOL.md` §3.4); and it still delivers
  the key at the end.
- A **hybrid** (per-device seal + an account-level recovery envelope only in
  ZUULI) is *deferred, not rejected* — it is the right answer if OS secret-store
  availability turns out worse than desktop `keyring` suggests.

The ADR carries the attacker table you asked for (e2e2z alone / ZUULI alone /
store copy / store + secret store), and the restore analysis. **Restore is not
broken**, because restore was never a property of the seal: §4.2 says restoring
the mnemonic restores the *identity*, and `engine.rs:221-223` says in as many
words that it "does not resurrect a device". One regression is named and scoped:
mnemonic + store but no secret-store item is no longer recoverable, in ZUULI
only.

**B. `IssueDeviceCredentialResultV1` needs no new fields.** `identity_pk` and
`handle` are already *inside* the signed `DeviceCredential`
(`f2z-kt-core/src/entry.rs:73`). A second copy beside it would be **unsigned**,
and under #929 unsigned means chosen by whoever answered — and the unsigned one
would win, because it is the one `install_identity` stores today
(`engine.rs:537-538`). So the rule is the inverse of the widening:
`install_identity` reads both out of the credential and stops taking them as
arguments. `IdentityInstall` becomes `{ credential, submitted_at }` — three
fields removed, none added, no wire change, §3.6's vector untouched.

That also closes a live gap found while writing this: **nothing today checks
`credential.handle == install.handle` or `credential.identity_pk ==
install.identity_pk`.** `MlsEngine::new` → `validate_for_leaf` checks the
signature, the window and the `device_pk` binding, but not those two
(`engine.rs:501-556`). Harmless while one process does both halves; not harmless
the moment the credential arrives over a bridge and the handle from a renderer.

**C. `device_kem_pk` is left open — explicitly, with the consequence spelled
out.** It is not resolvable inside a document about sealing: it is a KT wire
question with an upstream OpenMLS dimension. What the ADR *does* add is why the
likely fix is to **remove or redefine** rather than fill: the field has no
consumer, `WIRE.md` §12.6 never reads it, ADR 0015's rejected alternative 4
already explains why the obvious consumer would be worse than none, and the init
key is arguably already authenticated transitively (credential binds
`identity_pk → device_pk`; RFC 9420 has the `KeyPackage` signed under that same
key; the `KeyPackage` carries the init key). §6.1 states the consequence
bluntly: **when #461 lands and the round trip completes, that means the
transport works, not that enrollment does.**

**D. The install step is an e2e2z app-crate command**,
`e2e2z_install_device_credential`, beside `e2e2z_device_credential_keys`.
**No capability entry, no `zcash:*` grant** — `authority-boundary.node-test.mjs`
passes unchanged. One argument: the credential. That short argument list is
itself the proof-obligation, because the credential comes back *through the
renderer*, so anything in that signature has been in the webview's
garbage-collected JS heap — which is precisely why a `wrap_key` parameter could
never have been right, independently of §4.2.

## What it does not assume

§7 handles #929 head-on. The exchange is **public material in both directions**,
so there is no secret for a hostile responder to steal — but on a *first*
enrollment the caller has never seen the account's real `identity_pk`, so it
cannot distinguish a genuine credential from a fabricated one. There is no key
to check against. That is trust-on-first-response and no shape of this result
type fixes it. What contains it is the KT directory, not the bridge: a
fabricated credential's `device_pk` is absent from the real `DirectoryEntry`, so
the failure mode is "messaging silently does not work", not "the attacker
becomes you".

## What I am not certain about

§8 lists five, with what would change the recommendation. The two that matter
most:

- **Mobile has no secret store at all.** `keychain.rs:332-352` returns
  `UnavailableStore` for every non-desktop target, and `keyring` lives in the
  crate e2e2z must not link. §3.5's "refuse to enroll rather than enroll into a
  state you cannot reopen" is my inferred policy, not one I read anywhere — and
  it matters, because auto-re-enrollment would mint a directory entry per launch
  and §4.2 requires every one to be surfaced as a wiretap.
- **I did not verify against OpenMLS that no path accepts a `KeyPackage` whose
  init key falls outside the leaf-signature coverage.** If §6's claim 2 is
  wrong, the recommendation to remove `device_kem_pk` is wrong and the answer is
  a real long-term KEM key.

Every claim about current behaviour cites `file:line`. Where intent is inferred
rather than read, the ADR says so.

## One deliberate scope departure: `scripts/check-markdown-links.mjs`

This PR was `docs/` only until #933's Markdown link checker landed in the
required gate and flagged the ADR's own table-of-contents link
(`#6-device_kem_pk-is-not-resolved-here`) as unresolvable, suggesting
`#6-devicekempk-is-not-resolved-here` instead.

**The link was right and the checker was wrong**, so the fix is in the checker
and the ADR is unchanged. Taking the suggestion would have written an anchor
that passes CI and is **broken on GitHub** — the rot #933 exists to prevent,
pointed the other way.

Verified against GitHub's own renderer rather than reasoned from CommonMark,
because the behaviour that matters is GitHub's:

```
gh api -X POST /markdown

## 6. `device_kem_pk` is not resolved here  ->  #6-device_kem_pk-is-not-resolved-here
## snake_case heading                       ->  #snake_case-heading
## _leading emphasis_ here                  ->  #leading-emphasis-here
```

GitHub keeps an underscore that did not delimit emphasis and drops one that did.

The defect is two functions contradicting each other, with the wrong one
winning: `slugOf`'s character class deliberately keeps `_` and its comment says
so, but `headingText` ran `replace(/\*\*|__|\*|_|~~/g, "")` first and destroyed
every underscore before the slugger saw one. Both rules #933 calls out as easy
to get wrong — no trim, no space collapsing — were implemented carefully; this
was a strip that never asked whether the underscore was emphasis. Now: code
spans are protected through the emphasis strip and restored after it, and the
strip is per-occurrence, keeping runs flanked by letters or digits (CommonMark
forbids intraword `_` emphasis).

Four self-test cases added, each rendered by GitHub before being pinned, and
mutation-verified **in both directions**:

| Mutation | Test that fails |
|---|---|
| revert to the unconditional strip | `snake_case heading` is `snakecase-heading` |
| never strip an underscore | `_leading emphasis_ here` is `_leading-emphasis_-here` |

`--self-test` 41 → 45 cases; the full scan is green across all 113 tracked
documents (917 links, 491 anchors). The self-test had **no underscore case at
all** before this, which is why the defect shipped — the same shape as
`AUTHORITY.md` §8's surviving mutations: the guard existed, nothing exercised
it.

Kept here rather than split, by the reviewer's call: splitting would either
block the ADR on another PR or require someone to write the knowingly-broken
anchor to get green, and a document whose subject is *not shipping things that
only look correct* cannot be the place that happens.

Related: #926, #905, #904, #929, #461, ADR 0002, ADR 0006, ADR 0015.


skyl:
Converted to draft to hold the merge watcher off while two review amendments land. Marking ready again once they're pushed.

**Review verdict: MERGE, conditional on two additive amendments.** All ~45 `file:line` citations verified accurate; §4.2 preservation, restore semantics, the TOFU containment argument and the `device_kem_pk` handling all independently confirmed. The reviewer resolved this PR's own §8.3 uncertainty (exactly 5 `IdentityInstall` construction sites — the caution was over-cautious, not wrong).

**Amendment 1 — §4/§5 discard the one check an unauthenticated channel permits.** The rule "take handle and identity_pk from the credential" is right for values the *responder* supplies, but generalizes silently to one the **caller already knows**: `requestDeviceCredential(handle)` holds the handle the user typed. Since `validate_at` verifies the signature under the credential's *own embedded* `identity_pk`, a hostile responder can mint a self-consistent credential for **any** handle — refused if compared against what this session requested, installed under the rule as written. §4.1 identifies the divergence as the danger and then closes it by deleting one side of the comparison rather than performing it.

**Amendment 2 — §3/§3.3's containment claim rests on an unstated, platform-false assumption.** Per-device isolation depends on the OS secret store separating apps. True on macOS/Windows; **false on Linux** Secret Service (per-user collection, no per-app isolation) — where a compromised e2e2z could read ZUULI's item and unseal its device secrets, partially reconnecting the two apps the split exists to separate. Also: §3/§5 never specify the secret-store service/account namespace, and the key lives in a plugin linked into **both** apps, so a plugin-level constant collides.

Plus three one-liners for §10: nothing lifts `locked` after this change; no migration path for already-sealed stores; and the claim that `messaging-contract.node-test.mjs` pins §6.1's prose is false — that test pins §3's command set, §5.1's events, relay error codes and Rust body digests, none of which cover it.

Neither amendment reverses a decision. The central call — sealing belongs to the app that generated the secrets, under a key it holds — is right, and proved from the tree rather than asserted.

skyl:
Both amendments pushed in `b4382580`, plus the four corrections. No decision reversed; all additive.

**Amendment 1 — the check the rule was deleting.** You are right, and the failure I had was in the scope of the argument rather than the argument itself. "A second copy beside the credential is unsigned, so the responder chooses it" is true of values the *responder* supplies; I generalized it to a value the *caller already holds*, and `requestDeviceCredential(handle)` holds exactly that at exactly the right moment.

§4 now carries it as an explicit MUST immediately after the rule it qualifies, with the concrete reason: `validate_at` verifies under the credential's **own embedded** `identity_pk` (`f2z-msg-mls/src/credential.rs:157-162`), so a hostile responder mints its own ISK and signs a well-formed credential for any handle, and every `validate_for_leaf` check passes. It is also the only authenticity check TOFU permits — so I added the line that it is *"the difference between '#929 is an open question' and '#929 is an open question and we also skipped the one thing we could have done about it'"*, because the failure mode here is a reader deleting the check on the strength of §4's headline.

§5's signature now takes `expectedHandle` (renderer-side comparison explicitly allowed; performing it nowhere explicitly not), and I rewrote the proof-obligation paragraph, which was the actual defect: it conflated a **secret** with **an argument**. The test is now stated as *what authority does this carry, and where did it come from* — `wrap_key` fails it, `expectedHandle` passes it and is required.

**Amendment 2 — the unstated platform assumption.** Stated in §3.3, scoped per platform: accurate as written on macOS (keychain ACLs by code signature) and Windows; on Linux the `e2e2z alone` row's containment is conditional on Secret Service isolation this repo does not have. I kept your framing that it is adjacent to a pre-existing question rather than a new hole — the mnemonic is already in that collection (`keychain.rs:702`) — and noted that it does not exist today at all, because ZUULI's wrap key is never persisted (`messaging.rs:219-243`).

Namespacing is now a §3 decision bullet, not left implicit: the plugin takes the service name from the host app rather than declaring one, following `SERVICE = "cash.free2z.zuuli.seed.v1"` (`keychain.rs:17`), with the note that on the platform of the stated assumption this is the difference between two device stores and one.

**The three §10 items**, all real:

- **`locked` had no exit.** §3 now specifies a seed-free retry that re-asks the secret store, reachable from e2e2z — which is the app that cannot fall back to enrollment. §10 says the duty must be *replaced*, not merely removed.
- **Migration.** Now a §3 bullet, with the point that "nothing has shipped" is load-bearing rather than convenient and that the window closes on the first enrollment — the same asymmetry §6 already invokes for `device_kem_pk`, which I had indeed applied in only one place.
- **The false pin claim is corrected in place rather than deleted**, naming what `messaging-contract.node-test.mjs` actually pins and stating that no test catches the §6.1 rewording. "There is a safety net" and "there is no safety net" are opposite instructions to whoever implements this.

**Also taken:** §3.5's refusal is now reconciled with Fact 6 — I answer that the refusal is not really protecting the seal, it is preventing re-enrollment-per-launch minting directory entries, which is durable and append-only and independent of what the seal is worth — and §9 gains **option F, degrade rather than refuse**, conditional on the degradation being visible in the UI and engine status rather than silent. §8's uncertainty 3 is marked resolved with your five sites rather than deleted, since an over-cautious doubt is worth as much to the next reader as a real one. And it is 42, not 43.

On **#936**: agreed it needs the same correction — the binding gap is real, but the fix is compare-against-what-we-requested, not only take-from-the-credential. Happy for you to update it; say the word if you would rather I did.

skyl:
Pushed `7d967f90`. **The link was right and the checker was wrong** — so the fix is in the checker, not the ADR, which is unchanged.

Taking the suggested `#6-devicekempk-is-not-resolved-here` would have produced an anchor that passes CI and is **broken on GitHub**. That is the rot #933 exists to prevent, pointed the other way, so I verified before editing rather than after.

Verified at the real conditions, against GitHub's own renderer rather than reasoned from CommonMark:

```
gh api -X POST /markdown   # mode: markdown

## 6. `device_kem_pk` is not resolved here
  -> id="user-content-6-device_kem_pk-is-not-resolved-here"
## snake_case heading
  -> id="user-content-snake_case-heading"
## _leading emphasis_ here
  -> id="user-content-leading-emphasis-here"
```

GitHub keeps an underscore that did not delimit emphasis and drops one that did.

**The bug is two functions contradicting each other, and the wrong one winning.** `slugOf` already knows underscores survive — its character class deliberately includes `_` and its comment says so — but `headingText` runs `replace(/\*\*|__|\*|_|~~/g, "")` first and destroys every underscore before the slugger sees one. Both of the flagged rules (no trim, no space collapsing) were implemented carefully; this one was a strip that never asked whether the underscore was emphasis.

Two independent reasons an underscore survives, both now handled:

- **Inside a code span** there is no emphasis processing at all. Code spans are now protected through the emphasis strip and restored after, instead of being inlined ahead of it. This alone covers ```device_kem_pk```.
- **Between two word characters** CommonMark forbids intraword `_` emphasis, so `snake_case` is literal. The strip is now per-occurrence and keeps runs flanked by letters or digits.

Four self-test cases added, each rendered by GitHub before being pinned, and mutation-verified **in both directions** so neither over- nor under-stripping survives:

| Mutation | Test that fails |
|---|---|
| revert to the unconditional strip | `snake_case heading` is `snakecase-heading` |
| never strip an underscore | `_leading emphasis_ here` is `_leading-emphasis_-here` |

`--self-test` 41 → 45 cases; full scan green across all 113 tracked documents (917 links, 491 anchors). The ADR file is byte-identical to before.

Two notes:

- The self-test had **no underscore case at all** before this, which is why it shipped. That is the same shape as `AUTHORITY.md` §8's surviving mutations — the guard existed, nothing exercised it.
- It rides on this PR because the ADR cannot go green without it, and the alternative was writing a knowingly-broken anchor into a document whose subject is not shipping things that only look correct. **Happy to split it into its own PR if #933's author would rather own the change** — say the word and I will move it.

You were right that the checker caught something real, incidentally: it is the first thing it has flagged, it named file, line and the exact anchor, and that precision is what made it a ten-minute investigation instead of a guess.

Issue #922, opened 2026-09-04
hygiene: project-boundary.node-test.mjs leaks a fixture to un-gitignored dist/.output-copied.ts


skyl:
Noticed in passing during #920.

`wallet/zuuli/scripts/project-boundary.node-test.mjs` leaves a fixture behind at
repo-root `dist/.output-copied.ts`. That path is **not** gitignored, so running
the test suite dirties the working tree.

Low severity, but it matters more here than in most repos: our doctrine is that
the primary checkout is never dirty, and a stray untracked file from a *test run*
is exactly the kind of noise that makes "is this checkout clean?" stop being a
reliable signal. It also risks being committed accidentally by an agent doing a
broad `git add`.

#920 deleted the file rather than committing it.

### Acceptance criteria

- [ ] The test cleans up its own fixture, or writes it somewhere already ignored
      (a temp dir is preferable to extending `.gitignore` — ignoring it hides the
      leak rather than fixing it).
- [ ] `git status --porcelain` is empty after a full local test run.

Related: #920.

skyl:
Reconciled against current main: PR #943 already shipped this fix in b78fb970. The output-copy fixture now resolves config.build.outDir against config.root, so it stays inside the temporary fixture directory removed by withFixture in its finally block. The constrained-build rejection remains covered. The merged PR explicitly records this issue as completed; closing the remaining stale tracker.

Issue #916, opened 2026-09-04
zuuli: drop blanket f2zmsg:default now that ZUULI has no messaging frontend


skyl:
Found during adversarial review of #913.

After #913, `wallet/zuuli/src/` contains **zero** messaging code — the surface moved
to `wallet/e2e2z/`. But `wallet/zuuli/src-tauri/capabilities/default.json` still
grants blanket `f2zmsg:default` to ZUULI's desktop webview.

Net effect: the **seed-holding process** retains webview-reachable authority over
43 messaging commands it can no longer use. That is exactly backwards from the
direction of #904 — the wallet authority should be shrinking its reachable
surface, not carrying dead grants.

This matters more than it looks because of #367: the whole reason for the split is
that a confused/hostile frame resolving as the trusted main window can invoke
whatever the capability set allows. Every unused grant on the seed-holding app is
free attack surface.

### Not a one-liner

`wallet/zuuli/src-tauri/src/lib.rs:124-127` actively **requires** `f2zmsg:default`
to be present, so deleting the grant alone will fail. The plugin registration and
that assertion have to be reconciled together, and the three app-crate enroll
commands (`f2zmsg_enrollment_status`, `f2zmsg_enroll`, `f2zmsg_unenroll`) need a
decision too — note they are registered via `generate_handler!` and so are
**not** subject to capability ACL at all.

Related: the review established that ZUULI still registers `f2zmsg_enroll` in Rust
but has no frontend caller anywhere after #913, so no app can claim a handle until
#905/#461 land. Whatever is decided here should be consistent with that.

### Acceptance criteria

- [ ] ZUULI grants no `f2zmsg:*` permission it cannot use, or the grant is
      justified in writing next to the capability file.
- [ ] `lib.rs:124-127`'s assertion updated to match reality rather than pinning
      the stale grant.
- [ ] Decide and document the ACL story for the three app-crate enroll commands,
      given `generate_handler!` bypasses the capability system entirely.
- [ ] `surface-capability-authority.mjs` still passes and still proves the
      property, with a negative control.

Related: #913, #904, #367, #905.

Issue #753, opened 2026-08-26
A messaging store that will not open takes the whole ZUULI wallet down at launch


skyl:
Found while linking the plugin into ZUULI (#750). It is a consequence of that
link, not a defect in the plugin as it was merged — until #750 the plugin's
`setup` never ran in a shipping app, so this failure mode was theoretical.

`tauri_plugin_f2zmsg::init()`'s setup hook is fallible end to end:

```rust
.setup(|app, _api| {
    let data_dir = app.path().app_data_dir()?;
    std::fs::create_dir_all(&data_dir)?;
    let backend = f2z_msg_store::SqliteBackend::open(&data_dir.join(STORE_FILE))
        .map_err(|error| std::io::Error::other(error.to_string()))?;
    …
    let engine = Arc::new(engine::Engine::new(backend, sink, platform)…?);
    app.manage(state::F2zMsg::new(app.clone(), Arc::clone(&engine)));
    Ok(())
})
```

An `Err` from a plugin's setup makes `tauri::Builder::build()` fail, and
`wallet/zuuli/src-tauri/src/lib.rs` ends in
`.expect("error while running tauri application")`. **So a messaging store that
cannot be opened takes the whole wallet down at launch.**

`SqliteBackend::open` is deliberately strict — it verifies every pragma at open
and *refuses* otherwise, because `CLIENT-CONTRACT.md` §11.2 says a client that
cannot promise durability must not ACK, and a silent fallback to a
non-durable store would stop deleting relay copies without telling anyone. That
strictness is right. What is wrong is its blast radius: WAL is unavailable on
some filesystems, the data directory can be full or read-only, and a corrupt
`f2zmsg.sqlite` from a killed write is a real state a device reaches. In every
one of those the correct product behaviour is *messaging is unavailable and the
wallet works*, and today it is *ZUULI does not start*.

### Why it is not a one-line fix

Failing soft cannot just skip `app.manage(..)`: `F2zMsgExt::f2zmsg` is
`self.state::<F2zMsg<R>>().inner()`, so every one of the 43 commands would panic
instead of refusing. The state has to exist and answer.

`EngineState::Faulted` already exists and is exactly the right answer, so the
shape is probably:

* `F2zMsg` holds the outcome rather than an `Engine` — either the engine, or the
  `ErrorCode` that stopped it from being built.
* `get_engine_status` reports `state: "faulted"` with `lastError`, and every
  other command refuses with that code (`storage-full` / `durability-unavailable`
[...]

PR #943, opened 2026-09-05
feat(zuuli): harden ZUULI into a wallet vault (#904 phase 4)

### LINKED ISSUE #921: zuuli CSP: remove dead dyte.io origin — RealtimeKit 1.5.1 rejects Dyte base URIs
Found while evidencing free2z's RealtimeKit CSP in #920.

`wallet/zuuli/src-tauri/tauri.conf.json` still allows `dyte.io` in its CSP. That
origin is **dead**: RealtimeKit 1.5.1 *throws* `"Dyte Base URIs are no longer
supported"` when given a dyte.io base URI. It is not a fallback, not a legacy
path — the SDK actively refuses it.

The entry is presumably residue from #816/#818, when the CSP was widened to get
RealtimeKit working after the packaged policy blocked its first request.

### Why this is worth doing rather than leaving

This is the app that holds the **master seed**. Under #367, a remote subframe on
Android resolves as the trusted main window, so the CSP is one of the few real
containment layers ZUULI has. Every origin in it that nothing uses is free
surface on the wallet authority, in the same category as #916's dead blanket
`f2zmsg:default` grant. The direction of #904 is that ZUULI's reachable surface
shrinks; this is a cheap step in that direction.

### Evidence to reuse

#920 did the static analysis of the bundled SDK and encoded per-origin
justification in `wallet/free2z/tests/realtimekit-csp.pw.ts`. Its findings for
the whole RealtimeKit origin set:

- **Needed (3):** `api.realtime.cloudflare.com` (init's first call — a negative
  control reproduces `ERR0001` exactly, the #813 symptom),
  `wss://socket-edge.realtime.cloudflare.com` (meeting transport),
  `rtk-assets.realtime.cloudflare.com` (`fetchEmojis()`, which has no try/catch).
- **Not needed (7):** `location` / `location-legacy` (a callstats **IP-address**
  lookup), `da-collector` / `api-silos` (analytics and OTel logs) — all four sit
  in a catch that only logs; `r2.cloudflarestorage.com` appears in **zero** files
  in the installed tree; and `*.dyte.io` per above.

Note the video path is WebRTC/ICE, which CSP does not govern at all — only
signaling is CSP-visible, which is exactly the admitted pair.

### Acceptance criteria

- [ ] `dyte.io` removed from ZUULI's CSP, and the rest of ZUULI's RealtimeKit
      origin list reconciled against #920's measured set.
- [ ] Live/Join Free still works on ZUULI — verified against a **real active
      room** on signed device builds, not a mock. This is the #816/#818 failure
      mode: the bug only appears under the packaged CSP.
- [ ] A negative control proves the test would catch a missing required origin
      (reproduce `ERR0001`), so the check is load-bearing rather than decorative.

Does **not** close #813, which needs an actual successful join on signed builds.

Related: #920, #813, #816, #818, #367, #916, #904.

skyl:
## Phase 4 of #904: harden ZUULI into a wallet vault

Phases 1–3 **copied** the content surfaces into `wallet/free2z` and
`wallet/e2e2z`. Nothing was deleted from ZUULI except `features/messages`.
This is the deletion, and it is the part that actually reduces risk.

The reason, restated so a reviewer does not have to go find it: under **#367**
Wry injects Tauri's IPC bridge into *every* frame on Android and its IPC
reports the **top-level URL** as the origin, so a remote subframe resolves as
the trusted main window and can invoke whatever the capability set allows.
ZUULI holds the master seed and the spending keys. The split exists so that the
app holding the seed renders **no** content we do not control.

`118 files deleted, 30,828 lines removed.`

Closes #921, #916 and #922. Part of #904.

---

## 1. Deleted from `src/features/`

`articles`, `creator`, `live`, `ai`, `search`, `profile`, `kyc`.
Kept: `wallet`, `auth`, `home`, `about`.

`features/home` was itself a discovery dashboard — `LiveRail`, `ArticlesGrid`,
`AiCta`, `CreatorsRow`, every one rendering remote copy and remote images
inside the WebView that holds the seed. It is replaced by a vault overview
whose every destination is inside this app's own wallet.

## 2. Deleted the content-rendering tree

`Markdown.tsx` (+2 test files), `markdown.css`, `Mermaid.tsx`,
`mermaid.worker.ts`, `mermaid-policy.ts` (+test), `RemoteMedia.tsx`,
`ImagePrivacySetting.tsx` (+test), `lib/markdown/**`, `lib/media/**`, plus
`lib/cover.ts`, `lib/search-route.ts`, `lib/discovery-autocomplete.ts` and
`lib/utils/bio.ts`.

Each was verified unreachable first, by walking the import graph from
`src/main.tsx` rather than by grep.

Thirteen dependencies then had no importer: `react-markdown`, `mermaid`,
`dompurify`, four `remark-*`, three `rehype-*`, `unist-util-visit`, and the
three `@cloudflare/realtimekit*` packages. **The lockfile goes from 709 to 351
packages.** `linkedom` stays (the vitest DOM parser); `qrcode.react` stays
(Receive).

The GHSA-2v8p-3f2j-5mp7 guard (`check-mermaid-security.mjs` +
`mermaid-security.node-test.mjs`) is **not** deleted. Its subject moved to
`wallet/free2z`, which still renders creator-controlled Markdown and has no
node-test gate of its own, so the guard moved with it and gained a companion
pinning that ZUULI depends on none of `mermaid`, `dompurify` or
`react-markdown`. See the `changes` comment below — the pinned `test` script is
what caught me deleting it.

## 3. The routes, and the #920 bug class

Deleting a route that something still links to produces a NotFound at runtime
that typecheck cannot see. Swept:

- `lib/routes.ts` / `App.tsx` mount only `/`, `/login`, `/about`, `/wallet/*`
  and the legacy `/buy` redirect.
- `navigation.ts` lost Live, Articles, AI, Profile and Revenue share; the
  desktop `explore`/`tools` groups collapsed to one `overview` group; the
  mobile bar's hard-coded `grid-cols-5` became a count derived from the
  placement contract, so it does not leave two dead columns.
- `TopBar` lost the search chrome (its route is gone) and the Profile and
  Revenue share menu items.
- **`lib/auth/login-destination.ts`** no longer allowlists `/ai`,
  `/articles/*`, `/creator/*` or `/live/*`. This one matters: a `returnTo`
  survives the deletion in `localStorage`, so a *successful* sign-in would have
  landed the user on a NotFound.
- **`lib/auth/paid-intent.ts`** keeps only the `send` kind. A retired kind is
  now destroyed on read rather than restored to a route that does not exist —
  it was a private draft that nothing could ever hand back.
- i18n: `MESSAGE_KEYS` goes from 67 keys to 32, and the `live`/`search`/`articles` namespaces are removed from
  `messages.ts` and all **three** locale catalogs (`en`/`es`/`fr` — the brief
  said six; there are three), kept key-identical. `errorNotFoundBack` no longer
  says "Back to Discover".

`navigation.test.ts` gains a check that every navigation target is a segment
`App.tsx` still mounts, with a negative control over the removed segments, so
this defect class fails a test rather than a page load.

## 4. CSP (#921)

**Before**

```
default-src 'self'; img-src 'self' data: blob: https:; media-src 'self' https:;
frame-src 'none'; connect-src 'self' https://free2z.cash https://*.free2z.cash
https://free2z.com https://*.free2z.com https://*.zec.rocks https://*.dyte.io
wss://*.dyte.io https://api.realtime.cloudflare.com
https://api-silos.realtime.cloudflare.com https://da-collector.realtime.cloudflare.com
https://location.realtime.cloudflare.com https://location-legacy.realtime.cloudflare.com
https://r2.cloudflarestorage.com https://rtk-assets.realtime.cloudflare.com
wss://socket-edge.realtime.cloudflare.com https://*.stripe.com
https://checkout.stripe.com; style-src 'self' 'unsafe-inline';
script-src 'self' 'wasm-unsafe-eval'
```

**After**

```
default-src 'self'; base-uri 'self'; object-src 'none'; worker-src 'none';
img-src 'self' data:; media-src 'none'; frame-src 'none';
connect-src 'self' https://free2z.cash https://*.free2z.cash;
style-src 'self' 'unsafe-inline'; script-src 'self' 'wasm-unsafe-eval'
```

19 sources removed across three directives; three directives added. Every
removal is justified per-directive in
`scripts/csp-policy.mjs`:

| removed | why it is dead |
|---|---|
| `img-src https:` | admitted any host on the internet |
| `img-src blob:` | RemoteMedia's validated local bytes |
| `media-src https:` | Markdown `<video>/<audio>` and the live stage |
| `*.dyte.io`, `wss://*.dyte.io` | **#921**: RealtimeKit 1.5.1 *throws* `"Dyte Base URIs are no longer supported"`. Already dead before this PR. |
| 7 × `realtime.cloudflare.com`, `r2.cloudflarestorage.com` | RealtimeKit; the SDK is not a dependency any more |
| `*.zec.rocks` | lightwalletd, reached from **Rust** by `tauri-plugin-zcash`. Never a WebView connection; CSP never governed it. |
| `*.stripe.com`, `checkout.stripe.com` | Checkout is opened in the OS browser by `tauri-plugin-opener`, or navigated to on web. Neither is a `connect-src` fetch, and no code path in `src/` fetches a stripe.com URL. |
| `free2z.com`, `*.free2z.com` | no consumer; not in the `http:default` allowlist either |

`'wasm-unsafe-eval'` **stays** — checked, not assumed: `src/lib/wasm-spike.ts`
instantiates the first-party Rust/WASM module (#535) and
`scripts/wasm-boundary.mjs` pins that exact string in two places, including a
self-test mutation.

## 5. Capabilities (#916)

`capabilities/default.json` granted blanket `f2zmsg:default`;
`capabilities/mobile.json` enumerated 42 named `f2zmsg:` permissions. Both
removed: **43 messaging commands are no longer webview-reachable** on the
process that holds the seed.

`src-tauri/src/lib.rs:124-127`'s assertion is **inverted, not deleted**. It
required those grants so losing them would be a red test; it now requires that
no shipping capability grants any `f2zmsg:` permission, with a positive control
on the Zcash grants so it cannot pass against an empty document.
`mobile-webview-authority.mjs` does the same for mobile, and its test pushes
all 42 reviewed identifiers one at a time and requires each to be refused.

**The plugin stays registered, deliberately.** `src-tauri/src/messaging.rs`'s
enrollment trio needs its engine and durable store, and those three cannot move
to `wallet/e2e2z`: enrollment is the one messaging operation that needs the
wallet seed, and the seed never crosses IPC.

**The ACL story #916 asked for, written at the call site:** there is none.
`generate_handler!` routes app-crate commands unconditionally, so
`f2zmsg_enroll` / `f2zmsg_enrollment_status` / `f2zmsg_unenroll` are reachable
from the WebView and, under #367, from a frame that resolves as it. Residual
risk is bounded but real — a hostile caller could submit an attacker-chosen
handle, or unenroll behind a confirmation string. Closing it is **#905's** job:
the intent authority is already `.manage()`d for exactly this reason and these
three move behind it when it ships. Adding a capability entry would not gate
them, only look like it did.

## 6. #922 — the leaked fixture

`project-boundary.node-test.mjs`'s Vite-transform fixture wrote to
`config.build.outDir` verbatim. In a resolved Vite 6 config that is still the
raw default string `"dist"` (Vite resolves it against the root only inside
`getResolvedOutDirs()`), so it landed at `wallet/zuuli/dist/.output-copied.ts`
in the real working tree. Resolved against `config.root` instead, so the copy
lands in the temp directory the test already deletes. `.gitignore` deliberately
not extended — that hides the leak rather than fixing it.

---

### The reduced CSP blocks what the old one allowed

`tests/csp-policy.pw.ts` runs each retired capability twice in a real browser —
under the **previous** policy (the positive control) and under the shipped one:

```
✓ the shipped CSP is the one this file measures
✓ the previous policy loaded a remote image; the shipped one blocks it
✓ the previous policy reached RealtimeKit; the shipped one blocks it
✓ both policies refuse an arbitrary host, so the probe is not trivially true
```

Under the previous policy a `https://tracker.example/pixel.png` image loads
(`succeeded: true`, zero violations) and a fetch to
`https://api.realtime.cloudflare.com/v2/internals/participant-details` resolves.
Under the shipped policy each raises `securitypolicyviolation` (`img-src`,
`connect-src`) and fails. The third test fetches a host **neither** policy
admits, so a passing result cannot come from the probe simply being broken.

What is measured is the renderer's verdict, not whether a packet left the
machine: Playwright's `page.route` interception is consulted *before* the
renderer's CSP check, so a request event fires for a blocked load too.
Asserting on it would be a measurement taken where the difference cannot
appear, and the file says so.

`scripts/csp-policy.mjs --self-test` mutates the shipped policy twelve ways —
each widening, each re-added dead origin — and requires every one to be
rejected:

```
CSP self-test: 12 mutations rejected.
Packaged CSP matches the reviewed directive contract.
```

`tests/mobile-webview-authority.pw.ts` additionally sweeps every mounted route
with `__TAURI_INTERNALS__` defined (where #367's defect lives) asserting zero
`iframe`/`frame`/`object`/`embed` and zero provider requests, with a positive
control that each route actually rendered — plus a negative control proving
`frame-src 'none'` refuses an injected iframe.

### The reduced capability set refuses what the old one permitted

This one proved itself: removing the grants turned **three Rust tests red**,
because `tauri::test`'s mock harness resolves the same `capabilities.json` the
packaged app does. The ACL is a real runtime boundary here, not a JSON file we
merely read. Before the change `plugin:f2zmsg|send_message` reached its body
and answered argument validation. After it:

```
f2zmsg.send_message not allowed. Permissions associated with this command:
f2zmsg:allow-send-message, f2zmsg:default
```

That is the requested proof, at the IPC layer.
`shipping_router_registers_the_messaging_surface_and_enrollment` is rewritten
as `the_capability_set_refuses_plugin_commands_but_not_the_enrollment_trio`,
which asserts both halves and the difference:

- a `plugin:f2zmsg|…` command is refused *before* its body runs, **and the
  refusal must name the permissions that would have been needed** — without
  that second assertion the test would also pass for a command that does not
  exist, a different failure with the same shape;
- `f2zmsg_enroll` / `f2zmsg_unenroll` still answer argument validation, proving
  they were routed. That is `generate_handler!` bypassing the ACL entirely —
  #916's caveat, now executable rather than only a comment.

The two #753 census tests drive the plugin one layer *below* the ACL, so they
grant themselves a **test-only** capability via Tauri's dynamic ACL
(`Manager::add_capability`), with identifiers that say so
(`f2zmsg-census-test-only`, `f2zmsg-engine-test-only`) and a comment pointing at
why the shipping app has none. Their property is unchanged.

Statically, in addition:

```
$ node wallet/zuuli/scripts/mobile-webview-authority.mjs
Mobile main has 30 named Zcash permissions and no messaging permission;
native frames are disabled and the WebView admits no remote image or connect origin.
```

`mobile-webview-authority.node-test.mjs` pushes each of the 42 reviewed
`f2zmsg:` identifiers onto the capability and requires every one to be refused,
and asserts the refused population is the real 42 rather than an empty list.
The Rust `shipping_capabilities_grant_no_messaging_permission` reads the
shipping artifact and requires zero `f2zmsg:` entries in both capabilities,
with a positive control on the Zcash grants so it cannot pass against an empty
document.

One more guard went red for the right reason:
`intent::tests::the_intent_authority_is_not_reachable_from_the_webview` parses
the `generate_handler!` list and requires it to contain no "intent" — and the
ACL comment I first wrote *inside* that list contained the word. The prose
moved above the call.

`node scripts/check-tauri-plugin-permissions.mjs` no longer lists ZUULI's
capability files under `f2zmsg` (it fails a registered file that grants
nothing) and reports `f2zmsg: 43 commands, 86 permission identifiers, 2
capability grants` — the two are now `wallet/e2e2z`'s.

### Everything else

```
npm run typecheck                    ✓
npm run typecheck:tests              ✓
npx vitest run                       72 files, 1072 passed, 5 skipped
npx playwright test                  99 passed
npm run build                        ✓ (WASM artifact verified in dist)
cargo build --locked                 ✓
cargo test --locked                  48 passed
cargo fmt --check                    ✓
cargo clippy --all-targets -D warnings  ✓
```

Checkers, each run, `--self-test` where it exists:

```
csp-policy.mjs --self-test               12 mutations rejected
project-boundary.mjs                     5 projects, 470 files, 2031 refs, 3 Vite builds
project-boundary.node-test.mjs           92/92
surface-capability-authority.mjs         ✓
mobile-webview-authority.mjs (+test)     12/12
check-tauri-plugin-permissions.mjs       ✓
messaging-contract.node-test.mjs         9/9
check-zuuli-android-gate.mjs --self-test 34 mutations
rtl-source-policy.mjs (+test)            17/17
ui-copy-truncation.node-test.mjs         4/4
seed-capture-boundary.mjs (+test)        104/104
safe-area-contract.node-test.mjs         6/6
wasm-boundary.mjs                        ✓
```

`git status --porcelain --untracked-files=all` is empty after a full local test
run — #922's acceptance criterion.

---

## What I did NOT do, and why

**`STATUS.md` is untouched.** It still has per-surface rows for Articles, Live,
AI, Creator profile and the More-sheet destinations, which are now wrong. Its
canonical hash is pinned in `scripts/status-freshness.mjs`
(`STATUS_DOCUMENT_SHA256`), and re-deriving it is the release-evidence ritual in
`docs/releasing.md` — an audit SHA, a first-parent ancestry check, and a
reviewed prose diff — not a drive-by edit inside a deletion PR. **This is a
required follow-up before the next release**, and the freshness gate will
demand it anyway once these paths change.

**The store listing and screenshots are untouched.**
`store/locales/*/apple.json` and `play.json` describe "articles and
livestreams"; `store/manifest.json` and all 20 shipped screenshots are of
`/articles`, article search, the article reader and a creator profile; and
`scripts/store-screenshot-contract.mjs`'s `CAPTURE_SHOTS` hard-codes those four
routes. Editing `CAPTURE_SHOTS` changes `contractDigest` and makes
`store/capture-record.json` stale on every run, and re-capturing needs a build
that can reach four routes that no longer exist. That is a screenshot-capture
change with its own review, not part of this one. CI stays green today because
`enforceCurrentSource` is only true on the publish path. **Required follow-up.**

**Android/iOS still request camera and microphone.**
`media-permission-manifests.node-test.mjs` pins `CAMERA`, `RECORD_AUDIO`,
`MODIFY_AUDIO_SETTINGS` and usage strings that promise livestreaming — grants
that existed only for `features/live`. Removing them means editing the
committed generated Android manifest and both Info.plists, which the repo warns
against doing casually, and it should be verified on a device. **This is a real
remaining over-grant on the seed-holding app and deserves its own issue.**

**`src/lib/api/free2z.ts` keeps its `ai`, `articles`, `live`, `discover`
sections.** They are unreferenced now, but the file is the shared typed contract
and `auth-session-boundary.node-test.mjs` anchors on its `// ─── Profile`
section marker. Dead TypeScript in a module nothing imports is tree-shaken and
is not attack surface; pruning it is a contract change, not a deletion. Noted in
`CLAUDE.md`.

**`useZcashAssociate.ts` and `SocialButtons`' `associate` mode are kept**
although their only caller (the profile "Linked identities" card) has gone. They
are auth-domain, and an account-settings surface in this app will want them
rather than reinventing the gating. Both comments were updated to say so instead
of pointing at a deleted file.

**`zod` is still a dependency with no importer in this app.** Out of the stated
scope ("the markdown/mermaid/rehype/remark stack"); flagging it rather than
removing it in the same change.

### One real loss of test coverage, stated plainly

`tests/creator-zec-tip.pw.ts` goes from 8 tests to 2.
`createCreatorTipRouteState` keeps its issued snapshot in module memory and
hands out only a lookup nonce — deliberately, so a route state cannot
authenticate itself — and the only issuer was the creator tip dialog, now in
`wallet/free2z`. Nothing in this app can mint one, so the six tests that entered
through that dialog **cannot be driven from a browser at all**; the route now
always fails closed.

Kept: the fail-closed case, plus a **new** one — a forged, well-formed route
state, the shape a compromised renderer produces under #367, must still be
refused. The validation contract itself remains fully unit-covered in
`src/lib/wallet/creator-tip.test.ts` (untouched). The route and
`lib/wallet/creator-tip.ts` are **kept** as the landing point #905's
authenticated issuer will drive.

### A defect this PR found rather than introduced

Widening `tests/touch-targets.pw.ts` to the vault's route set put `/about` into
the tap-size audit for the first time — it was only in the *viewport* matrix
before, which measures overflow, not tap size. The About feedback composer's
"Include sanitized diagnostics" switch renders at **44×24**, eleven pixels under
the floor. Probed against `origin/main`: same 44×24 box; `touch-targets` was
green there only because `/about` was not in its `ROUTES`. Fixed with `min-tap`
plus `justify-self-start`.

---

## Playwright: what was deleted and why

35 specs → 20. **Deleted (16 + 2 helpers)** — each exists only to cover a
deleted surface: `ai-context-switch`, `article-drafts`, `article-tags`,
`creator-pagination`, `creator-responsive`, `media-preflight`,
`participant-counts`, `private-live`, `search`, `search-autocomplete`,
`search-pagination` (the AI/article/creator/live/search routes);
`mermaid-security` (re-pointed at free2z, not deleted), `remote-media-consent`, `first-party-images` (the Markdown
renderer and RemoteMedia); `ui-copy` (its two subjects were the Search heading
and the article search field); `realtimekit-csp` (asserted an exact
`connect-src` by loading `RealtimeKitClient` from a package this app no longer
has — replaced by the stronger `csp-policy.pw.ts`); and
`helpers/creator-catalog-audit.ts` + `helpers/mock-capture.ts`.

**No spec covering a surviving surface was weakened.** Where a spec's *subject*
moved, the subject moved with it: the mirrored-icon RTL mutations to
`features/home/parts.tsx` and the funding Activity tab, the long-text tap
target to Wallet Send's recipient field, scroll restoration to Wallet and
`/login`, the paid-intent destruction tests to Wallet Send & Tip. Two claims got
*stronger*: `mobile-webview-authority` went from "one article externalizes its
embeds" to "no route creates a frame", and `TopBar.test.tsx` from "search is
visible" to "no search chrome exists anywhere in the shell".

`src/lib/wallet/mock.ts`'s three-transaction ledger is padded to 24 so
`/wallet/history` is a route that actually scrolls — scroll restoration needed a
long route once the Articles feed left, and a three-row wallet was never a
realistic fixture.

---

One flake to be aware of: `auth-lifecycle.pw.ts:103` failed once inside the
chained `npm test` run (27s) and passes in isolation (4.6s) and in a standalone
full-suite run. It is timing under machine load, not a regression from this
change.


skyl:
## `changes` fixed — and it caught something real

`scripts/check-github-actions-pins.mjs` pins ZUULI's `test` script **verbatim**
and uses it as the base for its mutation suite, so removing an entry without
updating the pin makes the checker refuse to run at all. That refusal is right:
a mutation base it cannot trust is worse than none.

**The delta is exactly one removal and one addition**, and chasing the removal
found a coverage loss I was about to ship:

| | |
|---|---|
| removed | `scripts/mermaid-security.node-test.mjs` |
| added | `scripts/csp-policy.node-test.mjs` |

### The removal was wrong, and the pin is why I noticed

I deleted the Mermaid guard because its subject was gone from this app. Its
subject is not gone — it **moved**. `wallet/free2z` depends on `mermaid`
11.17.0 and `dompurify` 3.4.14, ships `Mermaid.tsx` and `mermaid.worker.ts`,
and renders creator-controlled Markdown — and it has no node-test gate of its
own (`npm test` there is `vitest && ui-copy-truncation && playwright`).
Deleting this file would have left the repository with **no** guard on
GHSA-2v8p-3f2j-5mp7's patched floors, for the one app that still parses
untrusted diagrams.

So the guard follows its subject rather than dying with ZUULI's copy of the
renderer. `check-mermaid-security.mjs` and `mermaid-security.node-test.mjs` now
read `wallet/free2z`'s manifest, lock and two components. They stay in
`wallet/zuuli/scripts` for the reason `messaging-contract.node-test.mjs`
already states in its own header — `wallet/zuuli`'s `npm test` is what the
protected `zuuli / frontend` job runs — and they only *read* free2z, modifying
nothing there.

The manifest test gains a companion pinning the other half of the move: **ZUULI
must depend on none of `mermaid`, `dompurify` or `react-markdown`.** If the
renderer ever returns to the app that holds the seed, someone decides that
deliberately (#904, #367) instead of it arriving with a dependency bump.

`npm test` also now runs `check-mermaid-security.mjs` and
`csp-policy.mjs --self-test` alongside the other live checkers, so the version
floors and the 12 CSP mutations are enforced in CI and not only locally.

### The reverse hazard, checked rather than assumed

All 25 `.mjs` entries in the pinned script were verified to exist on disk. The
five flagged as likely to cover surviving surfaces all remain in the chain, and
three of them were edited *earlier in this PR* precisely so they keep enforcing
against surviving files instead of being dropped:

- `surface-capability-authority` — its subject is free2z/e2e2z, untouched here.
- `seed-capture-boundary` — lost only its `features/profile/LinkedAccounts.tsx`
  read; every other file it reads survives.
- `mobile-webview-authority` — lost the `RemoteMedia.tsx` redirect assertion,
  gained exact-match `img-src`/`connect-src` rules and the #916 refusal.
- `ui-copy-truncation` — kept all its prose rules; only the coverage anchor
  moved off `features/search/index.tsx`.
- `send-review-boundary` — untouched; all 14 files it reads survive.

### Proof the pin is still load-bearing

```
$ # drop scripts/seed-capture-boundary.node-test.mjs from the test script
$ node scripts/check-github-actions-pins.mjs; echo $?
GitHub Actions fail-closed policy failed:
1
$ # restore it
$ node scripts/check-github-actions-pins.mjs; echo $?
GitHub Actions policy passed: … (212 external reference(s), 20 file(s))
0
```

### Companion gates, self-test and live

```
check-github-actions-pins.mjs --self-test   74 source-policy, 11 gate-verdict,
                                            148 current-workflow, 3 project-discovery,
                                            18 frontend-build, 4 classic-frontend-audit,
                                            128 Rust-root ownership cases
check-workflow-gates.mjs --self-test        39 cases
check-workflow-gates.mjs                    2 gates, 69 jobs, 19 workflow files
check-zuuli-android-gate.mjs --self-test    34 mutations
check-zuuli-android-gate.mjs                armv7-linux-androideabi type-check
check-tauri-plugin-permissions.mjs          zcash 39 / f2zmsg 43 commands
check-markdown-links.mjs                    920 links, 491 anchors, 113 documents
mermaid-security.node-test.mjs              6/6 (was 5; +the ZUULI-has-no-renderer pin)
```

### Follow-ups now filed rather than living only in this PR's prose

- **#945** — ZUULI still requests **camera and microphone** on Android and iOS,
  with usage strings promising livestreaming, for a capability it no longer
  has. Same category as #921's dead origin and #916's dead grant, except the
  grant comes from the OS, so neither the CSP nor the capability work here
  contains it. `media-permission-manifests.node-test.mjs` is what currently
  *keeps* it there; it touches the committed generated native projects and
  wants signed-device evidence.
- **#946** — `STATUS.md` and the store listing (copy, questionnaire notes, all
  20 screenshots) still describe articles, livestreaming, AI and creator
  profiles. `status-freshness.mjs` pins the STATUS hash and
  `store-screenshot-contract.mjs` pins the four capture routes.

### Rebase note

Rebased onto `main` after #932 landed. The new
`scripts/check-markdown-links.mjs` passes, and `wallet/zuuli/CLAUDE.md`'s
architecture reference is now a real relative link rather than a bare
repo-root path. I have **not** rebased onto #941/#942 — neither has merged. If
#942 lands first I will re-run `surface-capability-authority.mjs` against its
per-surface `REVIEWED_NAMESPACES`; this PR removes ZUULI's `f2zmsg:` grants and
adds no namespace anywhere, so I expect no conflict beyond the file itself.

skyl:
`zuuli / frontend` fixed.

The failure was mine and mechanical: I rewrote `wallet/zuuli/package.json` with Python's `json.dumps`, whose default `ensure_ascii=True` escaped the em dash in `description` as a `\u2014` escape. `store-screenshot-contract.mjs` requires canonical pretty JSON matching Node's `JSON.stringify` — and it is one of the capture-record digest inputs, so a byte difference is a real failure rather than a style nit. Round-tripped through Node.

Checked the diff versus `origin/main` is only what this PR intends, since a reformat can silently reorder things: the `test` script, `test:mermaid-security` losing its deleted Playwright spec, `test:realtimekit-csp` becoming `test:csp`, `test:ui-copy` losing `tests/ui-copy.pw.ts`, and the thirteen removed dependencies. No key reordered, nothing else restyled.

Re-ran the `zuuli / frontend` program locally before pushing:

```
npm run test:store-listing                    60/60
npm run store:validate                        exit 0
npm run icons:check                           brand assets current
node scripts/project-boundary.mjs             5 projects, 470 files, 3 constrained Vite builds
node scripts/surface-capability-authority.mjs passed
node scripts/rtl-source-policy.mjs            OK
node scripts/check-github-actions-pins.mjs    passed
npm run typecheck && npm run typecheck:tests  clean
npx playwright test                           99 passed
```

Agreed on `failedProviders: ["play"]` — that is the store-audit job reporting the Play listing does not exist yet (#948), not this PR.

On #944: this PR touches only `capabilities/default.json`, `capabilities/mobile.json`, `src/lib.rs` and `tauri.conf.json` under `src-tauri/`. It does **not** touch `src/messaging.rs`, so #944's `IdentityInstall` edit should not conflict.

Issue #238, opened 2026-08-06
Release: immutable cross-platform train and TestFlight/Play delivery


skyl:
## Goal

Turn one reviewed ZUULI commit and one version into reproducible, signed iOS/iPadOS, Android, macOS, and Linux artifacts, then deliver the mobile builds to TestFlight and Play internal testing. Windows remains a compile canary for v1.

## Invariants

- one SemVer value across package.json, Cargo.toml, tauri.conf.json, iOS, and Android
- one immutable source SHA and provenance manifest for every artifact
- release is tag/manual only; ordinary PR CI never receives signing secrets
- exact Rust/Node/Java/Xcode/Android toolchains and locked dependencies
- unsigned packaging smoke tests on PRs; signed/notarized uploads only from protected release environment
- artifact hashes, SBOM, checksums, and retention
- TestFlight + Play internal delivery is idempotent and cannot overwrite a different SHA at the same version/build number
- macOS universal bundle/DMG and Linux AppImage/deb/rpm (as supported by current Tauri) move on the same train
- rollback/stop-the-line controls and documented founder runbook
- Claude CLI adversarial review evidence plus all automated gates green before founder admin merge

## Acceptance

- [ ] version consistency verifier and release metadata schema
- [ ] iOS archive/export lane with App Store Connect upload dry-run/test mode
- [ ] Android signed AAB lane with Play internal upload dry-run/test mode
- [ ] macOS universal signed/notarized lane and Linux packaging lane
- [ ] immutable provenance/checksum/SBOM artifacts
- [ ] GitHub release orchestration and promotion runbook
- [ ] first TestFlight build installed and exercised on a physical device
- [ ] first Play internal build installed and exercised on API 29 and current Android

Tracked by #227.

skyl:
Release-train implementation is ready in #270 at `9cf5b8deba0cc7f7f40ab639e2b01d7bda848ad8`.

Automated evidence is fully green:

- required `gate` (frontend, shared plugin, full ZUULI Rust backend)
- release identity and immutable manifest contract
- Android unsigned AAB packaging on a clean Ubuntu runner: 13m13s
- iOS unsigned simulator packaging on Xcode 26.6: 9m11s
- universal macOS app/DMG packaging: 22m58s
- Linux AppImage/deb/rpm packaging: 26m31s
- source SBOM, checksums/provenance, and retained artifacts in every platform lane

[...]

PR #955, opened 2026-09-06
feat(zuuli): finish the vault cleanup #943 deferred (#945, #946, #916)

### LINKED ISSUE #916: zuuli: drop blanket f2zmsg:default now that ZUULI has no messaging frontend

### LINKED ISSUE #945: zuuli: drop the camera and microphone grants that only Live needed
Found while doing #904 phase 4 (#943), and deliberately left out of that PR.

`wallet/zuuli` still requests **camera and microphone** on Android and iOS, and
ships usage strings promising livestreaming — for a capability the app no longer
has. #904 phase 4 deleted `features/live` and the RealtimeKit dependency, and
narrowed the CSP so the WebView cannot reach a streaming origin at all.

### What is still granted

`wallet/zuuli/src-tauri/gen/android/app/src/main/AndroidManifest.xml`

```
android.permission.CAMERA
android.permission.RECORD_AUDIO
android.permission.MODIFY_AUDIO_SETTINGS
```

`Info.ios.plist` / `Info.macos.plist` / `gen/apple/zuuli_iOS/Info.plist`

> "ZUULI uses the camera when you broadcast or join a live video stream."
> "ZUULI uses the microphone when you broadcast or join a live stream."

### Why it matters

This is the process that holds the master seed. #367 is the whole reason for the
split: a remote subframe on Android resolves as the trusted main window. An OS
permission the app cannot use is exactly the same category as #921's dead
`dyte.io` origin and #916's dead `f2zmsg:default` grant — free surface on the
wallet authority — except this one is granted by the *operating system* rather
than by our own ACL, so our CSP and capability work does not contain it.

It is also a store-review liability: the app asks for camera and microphone and
tells the user it is for livestreaming, which it cannot do.

### Why #943 did not do it

`wallet/zuuli/scripts/media-permission-manifests.node-test.mjs` pins the exact
set and the exact copy:

```js
const CAMERA_COPY = "ZUULI uses the camera when you broadcast or join a live video stream.";
const MICROPHONE_COPY = "ZUULI uses the microphone when you broadcast or join a live stream.";
const EXPECTED_ANDROID_PERMISSIONS = [
  "android.permission.INTERNET",
  "android.permission.CAMERA",
  "android.permission.RECORD_AUDIO",
  "android.permission.MODIFY_AUDIO_SETTINGS",
];
```

so the guard is what *keeps* them there — stripping the manifest without editing
the test is a red build, and editing both is a deliberate change, not a
side-effect of a deletion. It touches the committed generated native projects
under `src-tauri/gen/{android,apple}`, which `wallet/zuuli/CLAUDE.md` warns
against regenerating casually, and the result should be observed on a signed
device rather than asserted from a manifest diff.

### Acceptance criteria

- [ ] `CAMERA`, `RECORD_AUDIO` and `MODIFY_AUDIO_SETTINGS` are gone from the
      Android manifest, and the camera/microphone usage strings are gone from
      `Info.ios.plist`, `Info.macos.plist` and the generated iOS plist —
      including `Info.ios.plist` as the regeneration source of truth.
- [ ] `media-permission-manifests.node-test.mjs` still passes the *property*
      (an exact, non-duplicated permission set with reviewed copy) against the
      reduced set, rather than being deleted.
- [ ] Verified on a signed device build that ZUULI installs and runs with no
      camera or microphone prompt, and that wallet create/restore/send still
      work — the manifest diff alone is not evidence.
- [ ] `scripts/macos-keychain-entitlements.mjs` (which also covers media-capture
      authority) still passes and still proves its property.

Related: #904, #367, #921, #916, #943.

### LINKED ISSUE #946: zuuli: STATUS.md and the store listing still describe the deleted content surfaces
Found while doing #904 phase 4 (#943), and deliberately left out of that PR
because both halves are reviewed release artifacts with their own rituals.

After #943, `wallet/zuuli` ships no articles, creator profile, livestreaming, AI,
search, profile editing or KYC. Two release surfaces still say it does.

### 1. `wallet/zuuli/STATUS.md`

It carries per-surface rows for things that no longer exist, e.g.

- L166 — names *"Articles, Messages, Profile, Revenue share, and About"* as the
  More-sheet destinations. After #943 the More sheet holds **Log in** and
  **About** only.
- L168 — *"Public Articles, creator listing, search, Live discovery, AI models,
  and pricing"*.
- L174 — *"Livestream room/media"*.
- L175 — *"Articles read/write/comments/tips … markdown/media render inside the
  privileged webview"*. Nothing renders markdown inside the privileged webview
  any more; that is the whole point of the phase.
- L176 — *"Creator public profile and self-edit"*.
- L178 — *"2Z send/tip/membership, and creator ZEC tips"* — the ZEC tip's
  *issuer* moved to `wallet/free2z`; the wallet-side landing route remains and
  now always fails closed until #905 lands an authenticated issuer.

This is a document whose entire purpose is to be true about what is proven, so
staleness here is worse than staleness anywhere else.

**Why #943 did not do it:** `scripts/status-freshness.mjs` pins the canonicalized
document hash (`STATUS_DOCUMENT_SHA256`), requires named sections to match
reviewed prose byte-for-byte, and — at release time — requires the recorded audit
SHA to be a first-parent ancestor with no release-impacting path changed since.
Re-deriving it is the ritual in `docs/releasing.md`, not a drive-by edit inside a
deletion PR. The gate will demand it before the next release regardless, because
#943 changes `wallet/zuuli/**`.

### 2. The store listing and its screenshots

`wallet/zuuli/store/locales/en-US/apple.json`

> `"subtitle": "Zcash wallet. Creator tools."`
> `"keywords": "zcash,wallet,creators,livestream,articles,AI,payments"`
> `"description": "… Explore articles and livestreams, use AI tools, support creators …"`

`play.json` says the same. `store/manifest.json` declares *"User-generated
articles, profiles, comments, and livestreams require store questionnaire
review"* and *"AI-generated content and live interaction require age-rating
questionnaire review"* — both now false, and both are questionnaire answers.

All **20** shipped screenshots (5 device sets × 4) are of the deleted surfaces:
`01-articles-fresh`, `02-semantic-search`, `03-article-reader`,
`04-creator-profile`.

**Why #943 did not do it:** `scripts/store-screenshot-contract.mjs`'s
`CAPTURE_SHOTS` hard-codes those four routes, and editing it changes
`contractDigest`, which makes `store/capture-record.json` stale on **every** run.
Re-capturing needs a build that can reach four routes that no longer exist, so
new shots have to be chosen first (Wallet overview / Send / Receive / History are
the obvious four). CI stays green today only because `enforceCurrentSource` is
true on the publish path alone.

### Acceptance criteria

- [ ] `STATUS.md` describes the surfaces ZUULI actually ships, with its rows for
      deleted surfaces removed rather than reworded into something unfalsifiable,
      and `STATUS_DOCUMENT_SHA256` recomputed in the same commit.
- [ ] `node --test wallet/zuuli/scripts/status-freshness.node-test.mjs` passes,
      and the re-derivation marker is refreshed against the merge SHA.
- [ ] Store copy in both locales describes a Zcash wallet, and
      `store/manifest.json`'s questionnaire notes no longer claim user-generated
      articles, livestreams or AI content.
- [ ] `CAPTURE_SHOTS` names four routes this app mounts, screenshots are
      re-captured for all five device sets, and `npm run store:validate` passes
      on the **publish** path (`enforceCurrentSource`), not only the PR path.

Related: #904, #943, #387.

skyl:
Finishes the ZUULI vault cleanup that #904 phase 4 ([#943](https://github.com/free2z/zuu/pull/943)) deliberately deferred.

Closes #945. Closes #946. Closes #916.

Follow-ups filed: #956 (store screenshots), #958 (`USE_FINGERPRINT`).

---

## #945 — the camera and microphone grants are gone

ZUULI holds the master seed. #943 deleted `features/live` and the RealtimeKit dependency and narrowed the CSP so the WebView cannot reach a streaming origin — but the app was still asking the OS for capture hardware it cannot use. That is free surface under #367, granted by the operating system rather than by our own ACL, so neither the CSP work nor the capability work contains it.

**Verified unused before removing anything**, not inferred from the feature name. Walked `src/`, `tests/` and `src-tauri/src/` for `getUserMedia`, `mediaDevices`, `videoinput`/`audioinput`, any QR/barcode scanner, and every `camera`/`microphone`/`RECORD_AUDIO` identifier. The **only** hit anywhere in `src/` is the BIP-39 word `"camera"` in `src/lib/bip39-english.ts`. Nothing in this app reads a camera or a microphone.

Removed:

| Where | What |
|---|---|
| `gen/android/.../AndroidManifest.xml` | `CAMERA`, `RECORD_AUDIO`, `MODIFY_AUDIO_SETTINGS` — `android.permission.INTERNET` is now the only permission this file declares (the merged manifest adds two reviewed biometric entries; see below) |
| same | the `camera`/`camera.autofocus`/`camera.any`/`microphone` `uses-feature` overrides |
| `Info.ios.plist`, `Info.macos.plist`, `gen/apple/zuuli_iOS/Info.plist`, `gen/apple/project.yml` | `NSCameraUsageDescription`, `NSMicrophoneUsageDescription` |
| `Entitlements.plist` | `com.apple.security.device.camera`, `com.apple.security.device.audio-input` |

**Kept:** `android.hardware.touchscreen` and `faketouch` at `required="false"`. Those are ChromeOS reach, not capture. The capture `uses-feature` entries went with the permissions because Play derives its implied hardware filter from the *permissions* — with `CAMERA` gone there is nothing left for the override to undo, and keeping one half without the other is the exact mistake the new test asserts against.

`Info.macos.plist` is kept as a pinned empty dictionary rather than deleted: it is the file anyone would edit to add a usage string back, and `macos-keychain-entitlements.mjs` matches it byte-for-byte, so a reintroduction is a red build rather than a quiet new TCC prompt on the wallet.

### The checkers keep their property and gain a negative control

None of them was deleted; each one's assertion is **inverted**, because the thing that must be red is now a *reintroduced* grant rather than a missing one.

- `media-permission-manifests.node-test.mjs` — still asserts an exact, non-duplicated Android permission set (now `[INTERNET]`) and keeps its three parser self-tests verbatim, because a permission audit that cannot see a permission would pass this file no matter what shipped. The Apple half now covers `project.yml` as well, since XcodeGen would otherwise regenerate the key back.
- `android-device-catalog.node-test.mjs` — asserts the capture features and permissions are absent, with a positive control on the touch features that remain.
- `macos-keychain-entitlements.mjs` / its node-test — the entitlements and the macOS Info.plist are still pinned byte-for-byte; what they are pinned to now excludes capture, and the two `rejects missing … authority` tests became `rejects reintroduced … authority`, which name the offending entitlement in the failure.

Proven by mutation, not by assertion:

```
control 1: re-add android.permission.CAMERA         → 4 failed
control 2: re-add NSCameraUsageDescription (iOS)    → 1 failed
control 3: re-add com.apple.security.device.camera  → checker exit=1,
    "macOS entitlements must not claim com.apple.security.device.camera"
restored                                            → 6 passed, checker green
```

### Proving it on a built artifact — and what that proof found

The source manifest is not the thing that ships, so the assertion is on the **merged** manifest, in both `zuuli-packaging.yml` (on the AAB this PR builds — so it runs here) and `zuuli-release.yml` (through the `bundletool dump manifest` that job already runs). This is the shape #941 established for e2e2z.

**The guard failed on its first real run, and it was right to.** It printed:

```
android.permission.DUMP, android.permission.INTERNET,
android.permission.USE_BIOMETRIC, android.permission.USE_FINGERPRINT
```

Investigated rather than allowlisted-to-green. Three separate answers:

| Name | Traced to | Verdict |
|---|---|---|
| `DUMP` | `androidx.profileinstaller`'s AAR manifest | **Not a permission this app requests — see below.** |
| `USE_BIOMETRIC` | `tauri-plugin-zcash/android/src/main/AndroidManifest.xml`, and `androidx.biometric:1.1.0` | **Load-bearing — verified in source, not assumed.** `ZcashPlugin.kt:295` calls `prompt.authenticate(info, BiometricPrompt.CryptoObject(cipher))` over the seed cipher under `BIOMETRIC_STRONG or DEVICE_CREDENTIAL`. Removing it breaks wallet unlock. |
| `USE_FINGERPRINT` | `androidx.biometric:1.1.0`'s manifest only | **Genuinely vestigial.** It exists for the library's pre-API-28 `FingerprintManagerCompat` path; both the app module and the plugin module set `minSdk = 29`, so it is unreachable. Filed as **#958**. |
| `cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION` (second run) | injected by `androidx.core` at merge time | **Legitimate.** Targeting API 33+ makes androidx.core declare this so dynamically registered `BroadcastReceiver`s are not exported. It is *this app's own* permission — note the package prefix — and signature-level. e2e2z's assertion accounts for the same class (#941). |

### What `DUMP` turned out to be, and how it went away

Worth recording, because it is the most interesting thing the guard found and because it did **not** go away by being fixed in the app.

`android.permission.DUMP` was never a permission ZUULI requests. It appears in the merged manifest exactly once, as an attribute:

```xml
<receiver android:name="androidx.profileinstaller.ProfileInstallReceiver"
          android:exported="true"
          android:permission="android.permission.DUMP" >
```

`androidx.profileinstaller` is pulled in transitively, and that `android:permission` is the permission a **caller** must hold to send this receiver an intent — it is what restricts profile installation to shell/adb. It is a *restriction*, not a grant, and `DUMP` is `signature|privileged`, so a normal app could not be granted it anyway.

Traced by extracting the `AndroidManifest.xml` from every AAR in the Gradle module cache and grepping each for the three names — `profileinstaller-{1.3.1,1.4.0}.aar` for `DUMP`, `biometric-1.1.0.aar` for `USE_BIOMETRIC`/`USE_FINGERPRINT` — then reading profileinstaller's manifest to see which element the string sat on.

**So nothing in the app changed. The check did.** My first version grepped the raw string anywhere in the manifest and could not tell a receiver guard from a request. Both workflows now fold newlines, split on `<`, and read `android:name` from `uses-permission` elements only. `DUMP` is therefore deliberately **not** on the allowlist, and two assertions pin the distinction: promoting it to a real `<uses-permission>` fails the build, and the `ProfileInstallReceiver` guard must remain in place — that receiver being shell-only is a property worth keeping, so the check now defends it rather than merely tolerating it.

So both halves of the check changed:

- **It reads elements, not strings** (above).
- **The list is ZUULI's, not e2e2z's.** `INTERNET`-only is the right boundary for a messaging client with no wallet; a wallet unlocks a seed. The assertion is an exact reviewed allowlist with one line of justification per entry, recorded in the workflows, in the source manifest comment, and in STATUS.md — and every entry is package-qualified, so a differently prefixed injected permission still fails.
- **The injected self-permission's benign-ness is checked, not asserted.** Its `<permission>` element must declare `protectionLevel="signature"`; downgrading it to `normal` is a red build.

Extraction self-test, run locally against a manifest shaped like the real merged one:

```
exact allowlist match                          PASS
signature-level self-permission                PASS
  downgraded to normal                         PASS (caught)
  com.evil.app.* prefix instead of ours        PASS (caught)
CAMERA added                                   PASS (caught)
DUMP as a receiver guard                       PASS (not read as a request)
DUMP promoted to a uses-permission             PASS (caught)
```

**`USE_FINGERPRINT` is allowlisted, not removed, and that is a deliberate call.** The fix is one line of `tools:node="remove"` — but it lands on the seed-unlock path and this PR has no signed device to check it on. A `SecurityException` out of `authenticateCipher` locks a user out of their own seed, which is far worse than one dead `normal`-protection permission that produces no prompt and no store questionnaire. #958 carries the removal with a device check as an acceptance criterion. The exact-set assertion means it cannot silently spread in the meantime.

The AAB-bytes half is kept for the three dead capture names, where grepping the protobuf member for *absence* is sound, plus a positive control so it cannot pass against an unreadable file.

**Honest limits.** I could not build the AAB locally: `gen/android` needs `tauri.settings.gradle` and `tauri.build.gradle.kts`, which the Tauri CLI generates and which are gitignored, so `./gradlew` cannot even evaluate settings from the committed tree. The merged-manifest assertion is CI evidence — which is how the four-name finding surfaced at all. #945's *"verified on a signed device that ZUULI installs with no camera or microphone prompt"* criterion is **not** met by this PR; that stays open on #238 and is recorded as unmet in the new STATUS row.

### Cross-scope workflow edits (minimal, and forced)

`.github/workflows/` is another agent's area right now. These edits were unavoidable, because the workflows **asserted the copy I removed** and would have gone red:

- 5 `plutil` blocks (1 in packaging, 4 in release) inverted from *"the packaged Info.plist carries this exact livestreaming copy"* to *"it declares no such key"*, with `CFBundleIdentifier` as the positive control.
- 2 signed-entitlements blocks (`macos-sign`, `macos-finalize`) that required `device.camera`/`.audio-input` to be `true` now require them to be `null`, and the exact key-set assertion drops both.
- the 2 new merged-manifest assertions above.

`apple-credential-boundary.mjs`'s opaque job digests for `android-sign-upload` and `macos-sign` are updated deliberately, having reviewed exactly what changed in each job, and its marker list is inverted the same way (with `rejectText` entries so neither job can reassert the old copy).

---

### STATUS.md, genuinely re-derived

Re-anchored from the build-20 anchor `39ce2e6e` to **`e129284c`**, whose own push-triggered [wallet/zuuli](https://github.com/free2z/zuu/actions/runs/33971858264), [rs](https://github.com/free2z/zuu/actions/runs/33971858271), [wallet/surfaces](https://github.com/free2z/zuu/actions/runs/33971858260) and [packaging smoke](https://github.com/free2z/zuu/actions/runs/33971858278) runs all concluded successfully on this exact tree. The 39 first-parent commits since the previous anchor are enumerated, organised around the three-app split, with the point stated plainly: **no release has been cut since `0.1.0+20`**, so none of it is in any signed artifact.

Six matrix rows are **removed, not reworded into something unfalsifiable** — public articles/creator/search/Live/AI/pricing, AI conversations and billing, livestream room/media, articles read/write/comments/tips, creator public profile, and KYC. Five are corrected (runtime transport and its CSP, app shell, the creator-ZEC-tip landing, E2EE messaging, distribution). One is **added**: a *Vault boundary* row for the property that is now the product — no remote content, no capture authority, no `f2zmsg:` grant — with its own evidence and its own explicit "not device-observed" limit.

The anonymous production probes were **re-run fresh** on 2026-09-06 against the endpoints ZUULI actually reaches, and the endpoints it no longer calls are deliberately dropped with a note saying why. `STATUS_DOCUMENT_SHA256` is recomputed in the same commit, and `status-freshness.mjs`'s pinned `macos-packaging` boundary cell is updated in step with it (118/118 tests pass).

### Store listing

Both locales now describe a self-custody Zcash wallet. `manifest.json`'s questionnaire notes no longer claim user-generated articles, livestreams or AI content — they are questionnaire *answers*, so a false one is a review problem. Categories move from `SOCIAL_NETWORKING`/`SOCIAL` to `FINANCE` (+ `UTILITIES` secondary), since the app is a wallet. `package.json`'s description said "AI, livestreaming, and articles" and now doesn't. `npm run store:validate` passes.

### Screenshots: reported and filed, not faked

**All 20 shipped screenshots (5 device sets × `01-articles-fresh`, `02-semantic-search`, `03-article-reader`, `04-creator-profile`) depict routes this app no longer mounts.** They are not re-captured here, because `CAPTURE_SHOTS` hard-codes those four routes, editing it changes `contractDigest` and makes `capture-record.json` stale for every set, and capture is pinned to a `linux/amd64` Playwright container that cannot be reproduced from this machine. Recorded in the STATUS distribution row and filed as its own issue.

---

## #916 — verified already closed by #943, deliberately not changed again

The issue describes `capabilities/default.json` still granting blanket `f2zmsg:default` and `lib.rs:124-127` requiring it. **Both were already fixed by #943** (`b78fb970`), which the issue predates. Confirmed rather than assumed:

- neither capability file carries any `f2zmsg:*` entry, and each says why in its `description`; the only two `f2zmsg` capability grants in the repo belong to `wallet/e2e2z`;
- `lib.rs`'s assertion is **inverted** — `shipping_capabilities_grant_no_messaging_permission` reads the shipping `capabilities.json` and requires that *no* capability grants any `f2zmsg:` permission, with a positive control on the Zcash grants so it cannot pass vacuously;
- the ACL story for the three app-crate enroll commands is decided and written down at `lib.rs:95-114`: they are exempt from the capability system because `generate_handler!` does not consult it, they are kept because deleting them deletes the only path that can issue a `DeviceCredential`, the residual risk is stated, and closing it is #905's job. *"Do not add a capability entry for them; that would not gate them, it would only look like it did."*
- the plugin stays registered, per #943's reasoning and ADR 0016.

**Proof that the narrowing bites** (the evidence #916's fourth criterion asks for). Putting `f2zmsg:default` back on the desktop capability and rebuilding:

```
test tests::shipping_capabilities_grant_no_messaging_permission ... FAILED
test tests::the_capability_set_refuses_plugin_commands_but_not_the_enrollment_trio ... FAILED

  the default capability must grant no messaging permission it cannot use,
  found ["f2zmsg:default"]

  the shipping capability set must refuse plugin:f2zmsg|send_message,
  got invalid args `args` for command `send_message`
```

That second line is the whole point: **with** the grant, `plugin:f2zmsg|send_message` resolves all the way to argument validation; **without** it, the ACL refuses it before its body runs and the refusal names the permissions that would have been needed. Green again once reverted.

---

## Verification

| Check | Result |
|---|---|
| `npm run typecheck`, `typecheck:tests` | pass |
| `npx vitest run` | 1072 passed, 5 skipped (72 files) |
| `npx playwright test` | 99 passed |
| `npm test` (vitest + 23 node-test suites + 6 checkers + playwright) | pass |
| `cargo build --locked` | pass |
| `cargo test --locked` | 48 passed |
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets --locked -- -D warnings` | pass |
| `surface-capability-authority.mjs` + node-test | pass (28) |
| `check-tauri-plugin-permissions.mjs` | pass — zuuli holds 0 of the 2 `f2zmsg` capability grants |
| `messaging-contract.node-test.mjs` | pass (9) |
| `status-freshness.node-test.mjs` | pass (118) |
| `media-permission-manifests` + `android-device-catalog` | pass (6), and red under 3 mutations |
| `macos-keychain-entitlements.mjs` + node-test | pass (12), and red under mutation |
| `release-identity.mjs`, `apple-credential-boundary.mjs` + node-tests | pass (197 + 15) |
| store contract / state audit / screenshot contract node-tests | pass (83) |
| `npm run store:validate` | pass |
| merged Android manifest permission set | **CI only — could not build the AAB locally. The first CI run caught four names; see above** |
| merged-manifest extraction logic | self-tested locally: exact-set match, `CAMERA` rejected, DUMP-as-request caught, DUMP-as-guard correctly ignored |
| signed-device install with no capture prompt | **not done — stays open on #238** |


skyl:
CI evidence at `e97f35d5` — all three runs green.

**The merged-manifest guard now passes on the real artifact.** `Android / unsigned AAB` ([job](https://github.com/free2z/zuu/actions/runs/34052034064/job/101537291244)) built the AAB and the new step read `base/manifest/AndroidManifest.xml` out of it, so the reduced Android permission set is asserted against what actually ships, not against the source file — the #941 standard. It took three iterations to get right, and each one was a real finding rather than a flake:

1. **`DUMP`** — never a permission ZUULI requests. It is the `android:permission` guard on `androidx.profileinstaller`'s `ProfileInstallReceiver`, restricting that receiver to shell/adb. The check was grepping strings; it now reads `uses-permission` elements, and separately asserts that guard stays in place.
2. **`USE_BIOMETRIC` / `USE_FINGERPRINT`** — the `INTERNET`-only rule was e2e2z's. `USE_BIOMETRIC` is load-bearing (`ZcashPlugin.kt:295` authenticates a `BiometricPrompt.CryptoObject` over the seed cipher); `USE_FINGERPRINT` is genuinely vestigial at `minSdk 29`, allowlisted with the reasoning recorded rather than removed blind off the unlock path — #958.
3. **`cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION`** — androidx.core's injected self-permission. Allowlisted package-qualified, and its `protectionLevel="signature"` is now checked rather than assumed.

**The macOS half is proven on a packaged app too.** `Desktop / macOS universal app` ([job](https://github.com/free2z/zuu/actions/runs/34052034064/job/101537291272)) passes with the `plutil` assertions inverted: the packaged `Info.plist` must declare *no* `NSCameraUsageDescription` or `NSMicrophoneUsageDescription`, with `CFBundleIdentifier` as the positive control.

| Run | Result |
|---|---|
| [`wallet/zuuli`](https://github.com/free2z/zuu/actions/runs/34052034071) | success, including the required `gate` |
| [`ZUULI / packaging smoke`](https://github.com/free2z/zuu/actions/runs/34052034064) | success — all five jobs |
| [`rs`](https://github.com/free2z/zuu/actions/runs/34052034060) | success |

24 checks pass, 0 fail. Rebased onto #954.

Still **not** done and recorded as such: no signed-device install has been observed, so #945's device criterion stays open on #238; the 20 store screenshots still show deleted routes (#956).

Issue #969, opened 2026-09-06
e2e2z-release: merging a build bump starts a real dual-store release; it fails closed today only because the Play listing is missing


skyl:
`.github/workflows/e2e2z-release.yml` triggers on push to `main` filtered to
`wallet/e2e2z/release.json`, and that path hardcodes both the target and the
dry-run flag:

```yaml
# :152
RELEASE_TARGET: ${{ github.event_name == 'push' && 'mobile' || inputs.target }}
# :202
dry_run=false
```

So **merging any PR that bumps the e2e2z build number starts a real,
non-dry-run, both-stores release**, with no dispatch and no confirmation.

### Today this fails closed — which is the only reason it is not already a problem

`prepare` calls `store-identity.mjs --require=apple --require=android` for the
`mobile` target, and `google.playConsoleListing` is still `"missing"`, so the run
dies in the first job. Nothing is built and nothing is uploaded. #968 will
produce exactly one red run on `main` for this reason.

### Why it should be fixed before Play is set up

The failure is protecting us by accident. The moment the Play listing is
recorded — which is a data change to `store-identity.json`, not a workflow
change, and not obviously release-affecting to whoever makes it — this same
trigger becomes a **fully automatic dual-store upload on merge**. The safety
property disappears silently, from an edit that looks unrelated.

Consuming a build identity is not reversible: Apple and Play both reject reuse
of a version+build, so an unintended run burns the identity even when the upload
is later abandoned.

### Also: `mobile` is the wrong default for e2e2z right now

We are deliberately TestFlight-only for e2e2z and free2z; Play is deferred. The
push path forcing `mobile` means the *only* automatic path targets a store we
have decided not to ship to yet. `wallet/free2z`'s new workflow (#963) defaults
its push path to `ios` for this reason.

### Options (owner picks; each is small)

1. **Make the push path dry-run** and require a manual dispatch to actually
   release. Strongest, and matches how the first TestFlight upload is being done
   by hand anyway.
2. **Make the push path `ios`** while Play is deferred, so it matches free2z.
3. **Drop the push trigger** and make release always an explicit dispatch.
4. Keep it, but add a guard that refuses a non-dry-run push release unless an
   explicit opt-in marker is present in the commit.

Compare `zuuli-release.yml`, which has a working, exercised release path — its
choice here should inform the answer rather than being assumed correct.

### Acceptance

- [ ] Merging a build bump cannot start a non-dry-run upload without an explicit human step
[...]

PR #977, opened 2026-09-07
feat(intent-bridge): verified App Links / Universal Links for all three apps (#461)

### LINKED ISSUE #461: zuuli: replace custom-scheme checkout return with verified Universal Links / App Links

skyl:
Closes the client half of #461. The blocking input — the three Play **App
Signing** SHA-256 fingerprints — existed for the first time today.

## The domain, established rather than guessed

**`free2z.com`.** `free2z.cash` is the canonical product host and permanently
redirects; neither Apple nor Android follows a redirect for these documents, so
the deployment repository serves both from exact-match `location =` blocks on
the `free2z.com` server block, ahead of that redirect.
`https://free2z.com/.well-known/apple-app-site-association` is **live today**
with exactly the three App IDs and the path scheme used here — the document
rendered by this PR is byte-identical to it (verified by `diff`).

Migrating to `free2z.cash` is recommended elsewhere and is deliberately **not
one change**: an association is domain-scoped and both platforms cache, so the
host must serve and warm both documents *before* any build claims it. Claiming
`.cash` in this PR would have shipped exactly that inversion.

## The path scheme

```
/oauth/callback     -> cash.free2z.zuuli   (pre-existing, live, unchanged)
/bridge/zuuli/*     -> cash.free2z.zuuli   (intent REQUESTS into the wallet)
/bridge/free2z/*    -> cash.free2z.free2z  (intent RESPONSES back to free2z)
/bridge/e2e2z/*     -> cash.free2z.e2e2z   (intent RESPONSES back to e2e2z)
```

The second segment names the **receiving** app: a URL's only job is to reach the
right process, and `CALLER-AUTHENTICATION.md` §5 is explicit that the bridge's
own `caller` field is not authenticated by the message. No two apps may match
one URL, or the OS picks arbitrarily — proved pairwise disjoint by the guard.

## What landed

- **All three `tauri.conf.json`** gain one `https` app-link route on
  `free2z.com`, scoped to that app's own `/bridge/<app>/` prefix. Existing
  custom schemes are untouched and keep carrying OAuth and checkout returns.
- **Generated trees regenerated, not hand-edited.** The deep-link plugin's own
  build script produced the `android:autoVerify="true"` intent filters and the
  `com.apple.developer.associated-domains` entitlements, driven by the same env
  the Tauri CLI sets; each app's existing normalizer then ran, so the release
  path's `git diff --exit-code` fixed point still holds.
- **Two normalizers had to learn that an app link is not a URL type.** The
  plugin filters `!is_app_link()` before writing `CFBundleURLTypes`, so counting
  all routes would expect one more duplicate dictionary than a build writes —
  free2z's derived count and ZUULI's exact pinned route array both needed it.
- **`wallet/zuuli/scripts/app-link-association.mjs`** (runs inside the protected
  `gate` via ZUULI's `npm test`) holds four surfaces to one reviewed record:
  each app's config, its generated manifest, its generated entitlements, and the
  two association documents.
- **Each app's `release-identity.mjs`** re-asserts `autoVerify` + host + path
  prefix against the manifest it is about to ship. Proved load-bearing: deleting
  ` android:autoVerify="true"` turns all three red with
  `Android manifest App Link autoVerify filters: expected 1, got 0`.
- **The three fingerprints** land in `store-identity.json` for free2z and e2e2z
  (`google.appSigningCertificateSha256` only — `playConsoleListing` is left at
  `"missing"` on purpose, see #969) and in the reviewed record for all three.
  ZUULI's **upload** certificate is named as forbidden, because it verifies in a
  sideloaded test build, fails for every real install, and Android caches that
  failure.

## Tests

23 cases in `app-link-association.node-test.mjs`, each applying one corruption
to the **real committed bytes** and asserting a named failure — mutations live
in the test rather than a patch/restore loop, so none can be left applied.
Tabulated in `CONFORMANCE.md`. Includes the two #461 asks: a fingerprint equal
to the known upload key is refused in both the record and the rendered document,
and every shipped package / bundle ID must appear in both documents.

## What this deliberately does NOT do

- **No transport.** `PROTOCOL.md` §7's refusal stands verbatim;
  `wallet/e2e2z/src/lib/enrollment/transport.ts` is untouched and still rejects
  twice. That is the next PR.
- **No association file in a web-served directory.**
  `ts/react/free2z/public/.well-known/` is the wrong home and would be actively
  harmful: a `.json` there is served on `free2z.cash` — a host no app claims,
  with no reviewed generator or `503` gate behind it — while
  `apple-app-site-association` has no extension and would not be served at all.
  Half an unreviewed association on the wrong host is the exact
  negative-caching failure the serving side's `503` contract exists to prevent.
  The same directory already lost `security.txt` to nginx for this reason.

## Still needed, outside this repo

1. **Add the three `<package> <SHA-256>` pairs to the serving repository's
   reviewed fingerprint list, regenerate, commit both files, deploy.** Until
   then `https://free2z.com/.well-known/assetlinks.json` returns `503` and
   **Android App Links do not verify for any of the three apps** — the
   all-or-nothing property is deliberate. The exact document to produce is
   `docs/intent-bridge/association/assetlinks.json`.
2. **Verify on signed device builds, both platforms.** `adb shell pm
   get-app-links <pkg>` reporting `verified`, and an iOS device opening a
   `/bridge/…` URL in the app. A simulator or emulator check is not evidence.
3. Note for iOS: all three provisioning profiles already carry
   `associated-domains: *`, so no profile reissue is needed.

## Verification run locally

`app-link-association` guard + its 23 tests; ZUULI vitest (1072); free2z vitest
(924) + node tests; e2e2z vitest (221) + node tests; all three
`release-identity.mjs`; all three Android manifest normalizers; all three iOS
project normalizers (+ self-tests where they have one); `check-markdown-links`;
`check-public-repo-boundary`; `surface-capability-authority`; ZUULI and free2z
typechecks. Two pre-existing failures are environmental and unrelated:
`wasm-boundary` needs a built WASM artifact and `artifact-sbom`'s DMG case needs
a built DMG. Playwright was not run — this change touches no UI copy, routing or
component structure.

Closes #461's repo-side work. Does not close #461 itself: it stays open on the
server-side Android deployment and the on-device verification above.


skyl:
## Fixed: `Rust / ZUULI backend`

Confirmed from the job log rather than inferred — two assertions, both
`left: 3, right: 2`:

```
oauth::tests::generated_config_names_only_canonical_mobile_redirects   (oauth.rs:1074)
oauth::tests::generated_android_manifest_registers_exact_mobile_redirect (oauth.rs:1098)
```

**Caused by this branch, not pre-existing.** Both encoded "ZUULI registers
exactly two mobile deep-link routes, so the manifest has two browsable
intent-filters". Adding the verified App Link makes it three of each. The counts
were correct; what they counted changed.

This is the same stale expectation the two iOS normalizers held, approached from
the other side. There it was *"every route contributes a `CFBundleURLTypes`
entry"* — an app link contributes none, because the plugin filters
`!is_app_link()` before writing URL types. Here it is *"every route is one of the
two custom-scheme routes"*.

### The fix sharpens the pin rather than loosening the count

- The config test now pins the third route exactly — `https`, `free2z.com`,
  `/bridge/zuuli/`, `appLink: true` — **and** asserts that exactly one route
  claims a domain. Both directions matter and they are opposite: a custom-scheme
  route that quietly became an app link would claim a domain, and an app link
  that quietly became `appLink: false` would register an unverified web link any
  app may also claim, which is what `CALLER-AUTHENTICATION.md` §4 refuses to
  carry authority over.
- New `generated_android_manifest_registers_the_verified_app_link`: asserts
  `autoVerify`, the https scheme, the host, ZUULI's own path prefix, and the
  **absence** of the other two apps' prefixes.
- New `generated_ios_entitlements_claim_the_association_host`: asserts the claim
  is in the entitlements and **not** in `Info.plist` — so a regeneration that put
  it in the wrong file is caught rather than passing quietly.

Both new tests proved load-bearing by mutation: deleting
` android:autoVerify="true"` fails the first; emptying the entitlements to
`<dict/>` fails the second. Restored, and full suites green afterwards — zuuli
backend 50, free2z 3+3, e2e2z 4.

The OAuth callback contract is untouched: the mobile scheme, host and path are
unchanged, and `MOBILE_REDIRECT_URI` still terminates in
`cash.free2z.zuuli://oauth/callback`.

### Rebased onto `origin/main` (`540021d4`, #962)

Re-verified after the rebase: the association guard and its 23 tests, all three
`release-identity.mjs` (free2z now at build 2 from #972), both iOS normalizer
self-tests, `check-github-actions-pins`, `check-markdown-links`, and the full
ZUULI backend suite. Working tree clean.

PR #993, opened 2026-09-08
chore(zuuli): remove orphaned content helpers

### LINKED ISSUE #982: zuuli: article-tags, private-live and participant-count are orphaned after the content-API removal
After #979 removed the unreachable content-API surface from `wallet/zuuli/src/lib/api/`, three modules outside that directory are left with no importers:

- `wallet/zuuli/src/lib/article-tags.ts`
- `wallet/zuuli/src/lib/private-live.ts`
- `wallet/zuuli/src/lib/participant-count.ts`

plus their `.test.ts` files. Their only callers were the `articles` / `live` export objects deleted in #979.

Left alone there deliberately — they sit outside that PR’s stated scope, and quietly widening a removal PR is how a reviewable change becomes an unreviewable one. Recording them so they are not simply forgotten.

Same method as #979 applies: walk the import closure from `main.tsx`/`App.tsx` and the four surviving features (`about`, `auth`, `home`, `wallet`), confirm nothing reaches them, and let `npm run typecheck` be the proof. Check the tests too — a module whose only importer is its own test is dead, but a test that covers behaviour still reachable elsewhere is not.

ZUULI is a wallet vault now (`#904` phase 4); article, live and participant-count helpers have no place in it.

skyl:
ZUULI no longer mounts article or livestream surfaces, but retained three unused helpers and two tests that exercised only those helpers. Remove `article-tags`, `private-live`, and `participant-count` from the wallet vault.

Closes #982.

Validation:

- TypeScript's module resolver walked 118 local modules from `main.tsx`, `App.tsx`, and the four surviving feature entrypoints; none reaches any deleted helper. Repository-wide import and exported-symbol searches confirm the only ZUULI importers are the two deleted tests. Free2Z's separate copies and live callers remain intact.
- `npm run typecheck` and `npm run typecheck:tests`: pass.
- Existing Vitest suite: 61 files, 980 tests pass, 5 skipped.
- Public repository boundary guard and diff whitespace check: pass.

This deletion changes no mounted component or user-visible flow. No new tests or package/script changes.


skyl (commented):
Independent review of cdb7b6260931193e8a98fcd03e6315bf5c067cf3: no blocking findings. The diff remains exactly five unreachable helper/test deletions, with the compiler-resolved entrypoint closure confirming no production imports and the Free2Z copies preserved. Both typechecks and 980 unit tests passed locally. Current-head gate and rs / gate are green, and hosted frontend/native selection matches the new policy. Separate packaging smoke is still running and will continue to be monitored.

PR #996, opened 2026-09-09
fix(intent): preserve uncertainty for unfamiliar response statuses

### LINKED ISSUE #930: intent-bridge: PROTOCOL.md must state that INTENT_UNAVAILABLE does not mean "nothing happened"
Found while writing free2z's failure copy in #924. Filing because it is
**caller-facing protocol semantics**, not one client's wording, and every future
caller will hit it.

### The fact

`wallet/zuuli/src-tauri/src/intent.rs`'s `payment_outcome` returns
`INTENT_UNAVAILABLE` for **every** `BroadcastStatus` except `Accepted` — and that
includes **`Unknown`**, the case where the transaction exists locally but nothing
establishes that the network took it.

So `INTENT_UNAVAILABLE` spans two very different situations:

| situation | did funds move? |
|---|---|
| the wallet could not start | **no** |
| broadcast returned `Unknown` | **unknown — possibly yes** |

### Why it matters

A caller that reads `INTENT_UNAVAILABLE` as "nothing happened" will tell a user
their funds are untouched when a transaction may have broadcast. That is wrong in
the single direction that actually harms someone, and it is the natural reading
of the name.

#924 hit exactly this: its first-cut copy said *"your ZEC is untouched"* on that
status. Fixed there by reorganising the copy around **"do we know what
happened?"** rather than **"did it work?"** — with an exhaustive map and a `never`
binding so a new outcome is a compile error instead of a silent fall-through.
That axis is the right one and should be the documented guidance, not a local
discovery each caller repeats.

Note this is deliberate and correct on the wallet side: `intent.rs` reports
`UNAVAILABLE` rather than a txid precisely because an ambiguous broadcast must
never be presented as a completed payment. AUTHORITY.md §7 already pins that with
a mutation. The gap is that **the caller-side consequence is written down
nowhere**.

### Acceptance criteria

- [ ] `docs/intent-bridge/PROTOCOL.md` states normatively that
      `INTENT_UNAVAILABLE` carries **no** claim about whether the action took
      effect, and that a caller must not present it as "nothing happened".
- [ ] It names the `BroadcastStatus::Unknown` case explicitly as the reason.
- [ ] Guidance on the honest partition for callers: which outcomes permit
      certainty (`no-transport`, request never encoded, `INTENT_NOT_CONFIRMED` —
      which `intent.rs` returns *before* `execute_send`) and which do not
      (`UNAVAILABLE`, transport failure after dispatch).
- [ ] `CONFORMANCE.md` gains a row so a caller claiming certainty on an
      uncertain status is a conformance failure, not a style opinion.

Related: #924, #914, #929, #905, #904.

skyl:
A well-formed intent response carrying an unfamiliar status was reported as `INTENT_MALFORMED`, hiding the difference between a decoding failure and an outcome this client cannot interpret. The shared session now preserves the original uint16 in a client-only `unknown-status` outcome, after canonical decoding, correlation, family and expiry checks; the outstanding question is consumed exactly once.

Free2Z carries the outcome to its existing uncertain-payment copy without a txid or automatic retry. E2E2Z exposes a distinct typed error without returning a credential or claiming enrollment. The protocol and conformance table now require uncertainty for `INTENT_UNAVAILABLE` (including `BroadcastStatus::Unknown`), unfamiliar statuses and failures after dispatch, and distinguish the cases where no dispatch occurred. Wire statuses, native payment authority and transport availability remain unchanged.

Closes #930, including its forward-compatibility and consumer-test follow-ups.

Validation:

- ZUULI shared intent suite: 33 tests and full test typecheck passed. Statuses 13, 4242, 60000 and 65535 preserve uncertainty and reject replay; malformed payloads, wrong identifiers/families and expired responses still fail their original checks.
- Free2Z full `verify`: typechecks, 929 unit tests (5 skipped), 13 source checks and 93 Playwright tests passed. Existing known-`INTENT_UNAVAILABLE` coverage remains intact; shipped en/es/fr copy checks include the new outcome.
- E2E2Z full `verify`: typechecks, 273 unit tests, 13 source checks and 9 Playwright tests passed.
- Restoring the old unknown→Malformed fallback in each isolated installed shared package made 4 ZUULI, 1 Free2Z and 1 E2E2Z regressions fail; exact sources restored afterward.
- Shared-project boundary audit, Markdown links, public-boundary guard and `git diff --check` passed.

### DIFF
--- a/docs/intent-bridge/CONFORMANCE.md
+++ b/docs/intent-bridge/CONFORMANCE.md
@@ -253,6 +253,7 @@ it work*, and the honest axis is **do we know what happened**:
 | `unsendable` | yes — nothing was encoded, the wallet was never asked |
 | `refused` + `INTENT_NOT_CONFIRMED` | yes — ZUULI returns it before `execute_send` |
 | `transport-failed` | **no** — the request may have arrived, the answer did not |
+| `unknown-status` | **no** — the response is well-formed but its meaning is unfamiliar |
 | `refused` + anything else | **no** |
 
 The sharpest case is `INTENT_UNAVAILABLE`: `intent.rs`'s `payment_outcome`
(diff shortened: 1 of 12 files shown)


skyl:
Independent review of d47929796635469dea04e60a94472892ae1c5e3e found no blocking issue. Canonical decoding, correlation, family, expiry, and one-use consumption precede unfamiliar-status handling; neither consumer can turn the outcome into a transaction ID, credential, or automatic payment retry. The protocol now explicitly distinguishes uncertainty from no effect, and the existing INTENT_UNAVAILABLE regression remains. Merge awaits current-head required CI.

skyl:
The current Free2Z CI failure is the existing @xmldom/xmldom dependency audit (high-severity advisory), not an intent test failure. A separate minimal dependency repair is being prepared on main. Remaining runs on this head were intentionally cancelled; this branch will be refreshed and fully gated after that repair merges.

PR #997, opened 2026-09-09
test(auth): anchor session guards to declarations instead of comments

### LINKED ISSUE #981: auth-session-boundary test slices source by literal section-header text, so unrelated edits break it
`wallet/zuuli/scripts/auth-session-boundary.node-test.mjs` asserts that `auth.completeSocialOAuth` never touches `setToken`/`localStorage` and that OAuth completion goes through `withOAuthSession`. Good properties.

It establishes them by slicing `free2z.ts` between **literal comment text** — the boundary marker is `"};\n\n// ─── Profile"`.

That means the test depends on a *comment header for an unrelated section* remaining in the file at a particular place. In #979 the dead `profile` export object was removed — `auth` was not touched at all — and the test failed (`not ok 236/238`) purely because the text anchor disappeared. The code under test was identical.

The failure is also misleading: it reports an auth/session boundary violation, which reads as a security regression, when nothing about auth changed. That is the same shape as #964, where a permission assertion fails on a correct manifest and reports a boundary violation.

### Why it matters beyond the one break

The property is real and worth keeping. The *anchoring* is the problem: it couples an auth security assertion to the presence and ordering of unrelated sections, so any future reorganisation of `free2z.ts` produces a spurious security-looking failure. Whoever hits it is invited to "fix" it by restoring a comment.

### Suggested

Anchor on the thing being tested rather than its neighbours — slice from the `completeSocialOAuth` declaration to the end of its own function body (brace matching or a real parser), or extract the method to its own module so no slicing is needed. Either removes the dependency on unrelated file content.

Repointed in #979 to `"// ─── Tuzi"` so the suite passes; assertions and sliced text are unchanged. That is a repair, not a fix — the next reorganisation breaks it again.

skyl:
Renaming or removing an unrelated section header made the auth-session security suite fail even when the guarded methods were unchanged. The suite now uses the existing TypeScript parser to locate each top-level auth method or transport function and inspect its complete declaration, including default parameters.

The token-publication and OAuth session-binding assertions remain intact. Missing, ambiguous and syntactically invalid declarations fail closed. The unrelated recovery-flow assertions retain their existing scope.

Closes #981.

Validation: all eight auth-session boundary tests passed, including neighboring-header removal/rename, nested braces and marker strings, reordered/decoy methods, token-write mutations and removed OAuth session binding. Reinstating the old comment slicer made the header-removal regression fail with a missing-marker error; the exact source was restored afterward. `git diff --check` and the public-boundary guard passed. No application source changed.

Current-main refresh: normally merged `b19bb88d` after #1000. The authored guard file is byte-identical to the reviewed original head; all eight boundary tests, existing mutation controls, public boundary, and diff checks pass on the combined tree.

### DIFF
--- a/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
+++ b/wallet/zuuli/scripts/auth-session-boundary.node-test.mjs
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict";
 import { readFileSync } from "node:fs";
 import test from "node:test";
+import ts from "typescript";
 
 function between(source, start, end) {
   const from = source.indexOf(start);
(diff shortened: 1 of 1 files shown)


skyl:
Independent review of 5d7c8578c17faa37bc47a498ee5d2eccecb47143 found no blocking issue. Declaration discovery targets the top-level auth object and method itself, includes default parameters, and fails closed on missing or ambiguous declarations. Existing token-publication and OAuth-session assertions remain; the reorganization regression fails against the old comment slicer. Merge awaits current-base required CI.

skyl:
Paused current builds while the shared Free2Z dependency baseline is repaired: the same base fails the high-severity @xmldom/xmldom audit in #996. Runs were intentionally cancelled to avoid rebuilding a known-blocked base. This branch will receive a normal main merge and fresh required CI after the dependency repair lands.

PR #1004, opened 2026-09-09
test: name Android gate wrong-digest controls accurately

### LINKED ISSUE #970: check-zuuli-android-gate: the stale-digest self-test control is two bumps old and no longer tests 'stale'
`scripts/check-zuuli-android-gate.mjs` pins a digest of the reviewed change-detector step, and its self-test includes a "digest is stale" control fixture pinning `cad27317…`.

That constant has not moved while the real digest has: `cad27317…` → `464b7a52…` (#957, adding `e2e2z-release.yml`) → `81ed7dc8…` (#963, adding `free2z-release.yml`). The fixture is now **two bumps old**.

It still works as a *wrong-digest* control — any non-matching value fails — but it no longer tests what it is named for. The interesting stale case is the **immediately-previous** digest: someone edits the selector, updates the comment, and re-pins to the value they had before rather than the one they just produced. A control that tracks the previous digest would catch that; one pinned to an arbitrary old value only catches obviously-wrong input.

Not urgent and not a correctness bug — the 34 mutation controls and the structural fail-closed assertions are what actually hold the gate, and those all fire. This is about keeping the control honest to its own name.

Suggested: derive the stale fixture from the previous digest as part of the bump, or rename the control to `wrong-digest` so it stops implying a property it does not test.

Noticed while re-deriving the digest in #963, and deliberately left alone there — changing a check's own controls does not belong in an unrelated feature PR.

skyl:
The fixed historical digest fixture does not track the immediately previous selector digest, so calling it “stale” overstates what it checks. Name the two fixtures as nonzero and all-zero wrong-digest controls and explain that both are fixed values. Digest constants and policy behavior are unchanged.

Validation: all 35 existing self-test mutations pass, the live Android gate policy passes, and the public boundary check passes. No new tests are needed for these diagnostic labels.

Closes #970.

### DIFF
--- a/scripts/check-zuuli-android-gate.mjs
+++ b/scripts/check-zuuli-android-gate.mjs
@@ -422,12 +422,13 @@ function runSelfTest(workflow, toolchainEnv) {
   if (baseline.length > 0) {
     throw new Error(`cannot self-test an invalid baseline:\n${baseline.join("\n")}`);
   }
+  // Fixed wrong-digest controls; neither tracks the immediately previous selector.
   for (const [name, digest] of [
     [
-      "the reviewed change-detector digest is stale",
+      "the reviewed change-detector digest is wrong (nonzero)",
       "cad2731734c97401e94ae2b25a722141f09488a107ddde6f28618a5454786c02",
     ],
-    ["the reviewed change-detector digest is wrong", "0".repeat(64)],
+    ["the reviewed change-detector digest is wrong (all zeroes)", "0".repeat(64)],
   ]) {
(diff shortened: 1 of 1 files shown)


skyl:
Independent review of c996ee3538cd6d650ef8fdfb88bdd028ecd9c026: the diff changes only two self-test diagnostic labels and their explanatory comment. Digest inputs and all assertions are unchanged; the labels now accurately describe fixed wrong-digest controls. No blocking findings. Merge still requires the current-main QA gates.

Issue #937, opened 2026-09-05
e2e2z mobile: no OS secret store — DeviceWrapKey has nowhere to live on iOS or Android


skyl:
Found while writing ADR 0016 (#935), and it blocks e2e2z being genuinely usable
on the platforms we are trying to ship to.

ADR 0016 recommends sealing device secrets under a per-device `DeviceWrapKey`
held in the **OS secret store**. On desktop that works. On mobile it currently
has nowhere to live:

- `keychain.rs:332-352` returns `UnavailableStore` off desktop.
- The `keyring` crate lives in the crate e2e2z **must not link**
  (`tauri-plugin-zcash`), so e2e2z cannot simply reuse it without breaching the
  boundary that `authority-boundary.node-test.mjs` enforces.

So on iOS and Android there is no answer yet to *where the wrap key is stored*.

### Why this matters for shipping

`cash.free2z.e2e2z` is being prepared for TestFlight and Play internal testing.
A build that cannot store a device wrap key cannot complete enrollment on the
platform it ships to — and enrollment is already blocked on #461 and #928, so
this would surface only after those clear, which is the worst time to discover it.

### The open question ADR 0016 flags explicitly

Its rule — *refuse to enroll rather than enroll into a state you cannot reopen* —
is **inferred from the tree's fail-closed doctrine, not read anywhere**. It needs
deciding, because the alternative is worse than it looks: auto-re-enrolling on
each launch would mint a **new directory entry per launch**, and §4.2 requires a
new device entry to be surfaced to the user as a possible wiretap. Silently
minting them would train users to ignore exactly the signal that protects them.

### Acceptance criteria

- [ ] Decide where `DeviceWrapKey` lives on iOS (Keychain, with the right
      accessibility class — consider `kSecAttrAccessibleAfterFirstUnlock` vs
      `WhenUnlocked` and what each means for background queue processing) and on
      Android (Keystore / EncryptedSharedPreferences), **without** e2e2z linking
      `tauri-plugin-zcash`.
- [ ] Decide and document the behaviour when the store is unavailable — refuse,
      or degrade — with the wiretap-signal consequence stated.
- [ ] A test that e2e2z refuses to enroll rather than enrolling into an
      unreopenable state, proven load-bearing.
- [ ] Confirm `authority-boundary.node-test.mjs` and
      `surface-capability-authority.mjs` still pass — e2e2z must keep zero
      `zcash:*` and link no wallet plugin.

Related: #935, #928, #904, #461.

PR #1009, opened 2026-09-10
feat(e2ee): seal under a per-device wrap key and let e2e2z install a credential (#928)

### LINKED ISSUE #928: intent-bridge: IssueDeviceCredentialResultV1 cannot complete the enrollment round trip

Jhonattan2121:
Closes #928. Implements [ADR 0016](docs/e2ee/decisions/0016-enrollment-sealing-boundary.md) §3, §4 and §5.

**The wire format does not change.** No field is added to `IssueDeviceCredentialResultV1`, so `PROTOCOL.md` §3.6's 130-byte vector, the `f2z-intent` conformance vectors and `wallet/zuuli/src/lib/intent-bridge.test.ts` are untouched. This follows the correction on the issue rather than its original acceptance criteria, which would have shipped a seed-derived key into e2e2z.

## Sealing moves off the seed (§3)

`install_identity` samples a `DeviceWrapKey` from the OS CSPRNG, writes it to the custody #937 built, and seals under it. `IdentityInstall` loses `wrap_key`; `Engine::unlock` takes no key and re-asks the store.

Two ordering decisions, both written down at the call site:

- **The custody write precedes the seal and the commit.** The inverse order fails unrecoverably: secrets committed under a key the store then refused to keep are an enrolled device that can never open its own seal, whose only exit is `unenroll`. This way round leaves a stray key that opens nothing.
- **The key is sampled at install, not at `prepare_device`** where §3's sketch puts it. A round trip to the wallet authority sits between the two and can be refused, abandoned, or answered with a credential this device must reject; a key written at prepare time outlives all three under a name the next enrollment overwrites. `prepare_device` still probes custody, so §3.5's fail-closed gate is unchanged.

## The install step (§5)

e2e2z gains two app-crate commands, neither needing a capability entry nor a `zcash:*` grant:

- `e2e2z_install_device_credential` — the step whose absence meant the round trip could not complete even with a transport. Two arguments, both public: the credential's bytes and the handle **this session requested**. `deny_unknown_fields` refuses a third.
- `e2e2z_retry_device_unlock` — the seed-free exit from `locked` that §3 requires an app with no enrollment path to have. It replaces `f2zmsg_enroll` re-deriving the wrap key from the mnemonic, which only ZUULI can do.

`bridge.ts`'s `enroll` installs what comes back and returns the engine's own status. It still rejects in every shipping build, for one reason: there is no transport (#461).

ZUULI stops passing `account.backup_wrap`, and its already-enrolled branch no longer reads the seed at all.

## What this does not mean

`device_kem_pk` stays open (ADR 0016 §6). When #461 lands, a completed round trip proves the **transport** works, not that enrollment does. `status.md` records that so nobody infers otherwise. ZUULI's authority side still answers `issue-device-credential` with `INTENT_UNKNOWN_INTENT` — the issue puts that out of scope.

## Docs

`CLIENT-CONTRACT.md` §6.1 is reworded in both places ADR 0016 §10 flags as covered by no test, and its state diagram with them. `THREAT-MODEL.md`'s at-rest claim is narrowed to what the seal actually covers — the SQLite database beside it is not encrypted, and Fact 6 says so.

## Verification

- `tauri-plugin-f2zmsg`: `cargo test` 90 + 2 + 14 + 5, plus the `relay-harness` suite; `clippy --all-targets --features relay-harness` and `fmt` clean
- `e2e2z`: 7 Rust tests, 281 vitest, 9 Playwright, both typechecks, `clippy` and `fmt` clean
- ZUULI: `cargo check --all-targets`, `clippy --all-targets`, `fmt` clean
- `messaging-contract.node-test.mjs` 9/9, `project-boundary.mjs`, `surface-capability-authority.mjs`, `authority-boundary.node-test.mjs` (unchanged, which is the mechanical statement that the boundary held), `check-markdown-links.mjs`, `check-tauri-plugin-permissions.mjs`

New tests, rather than adjusted ones:

- a store that passes the probe and then refuses the real key installs nothing — pins the ordering above
- a custody that lost the key reports `durability-unavailable`, not `engine-locked`; the distinction is what tells the UI whether the §3.5 retry is worth offering
- the install arguments carry a credential and a handle and nothing else, asserted in both Rust and TypeScript
- `enroll` installs the credential that came back over the bridge, for the handle it asked for, and returns the engine's status field for field

`enroll-intent.test.ts` was rewritten rather than patched: its premise was that this app has no command that could install a credential, which is no longer true. The property it now pins is the one that was always the point — the only `EnrollmentStatus` this app can return is one the engine produced.

### DIFF
--- a/wallet/e2e2z/src-tauri/src/device.rs
+++ b/wallet/e2e2z/src-tauri/src/device.rs
@@ -53,8 +53,11 @@
 //! established that a transport exists: preparing a device for a request that
 //! cannot be sent throws away key material for nothing.
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use tauri::{AppHandle, Runtime};
+use tauri_plugin_f2zmsg::engine::IdentityInstall;
+use tauri_plugin_f2zmsg::error::Error;
+use tauri_plugin_f2zmsg::models::{EngineStatus, EnrollmentStatus};
 use tauri_plugin_f2zmsg::{F2zMsgExt as _, Result};
 
 /// The public halves of this device's key set, hex-encoded.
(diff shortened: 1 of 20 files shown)


wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs
```diff
@@ -686,7 +681,20 @@ impl<B: StorageBackend> Engine<B> {
...
+        // prepare time is an orphan under a name the next enrollment
+        // overwrites. `prepare_device` still probes custody, so the fail-closed
+        // gate is unchanged.
+        //
+        // The custody write precedes the seal and the commit because the
+        // inverse failure is unrecoverable: secrets committed under a key the
+        // store then refused to keep leave an enrolled device permanently
+        // `locked`. This way round leaves a stray key that opens nothing.
+        let wrap_key = device_wrap_key();
+        self.wrap_key_custody.put_wrap_key(&wrap_key)?;
```
skyl, line 714: [P1] Refuse replacement of an installed identity before overwriting custody. Neither prepare_device nor install_identity rejects an already-enrolled engine, and the new app command exposes install directly. A second valid install writes a fresh key into the single custody slot here before the identity/seal transaction commits. If that transaction fails or the process exits in between, the previous seal remains persisted while its only key has been overwritten, so the existing device becomes permanently locked. Check for an existing identity under the engine mutex before changing custody, and cover repeated install plus preservation of the original seal/key.

wallet/e2e2z/src-tauri/src/device.rs
```diff
@@ -109,6 +112,113 @@ pub async fn e2e2z_device_credential_keys<R: Runtime>(
...
+pub async fn e2e2z_retry_device_unlock<R: Runtime>(app: AppHandle<R>) -> Result<EngineStatus> {
+    app.f2zmsg().engine_handle()?.unlock().await
+}
+
+/// Decode lowercase hex. Local for the reason [`hex`] is.
+fn unhex(value: &str) -> Result<Vec<u8>> {
+    let bytes = value.as_bytes();
+    if !bytes.len().is_multiple_of(2) {
+        return Err(Error::internal("the credential is not whole bytes of hex"));
+    }
```
skyl, line 191: [P2] Make unlock retry preserve an already-unlocked engine. At device.rs:183, the new public retry command calls Engine::unlock unconditionally, but unlock rebuilds MLS and sets state to Stopped without closing the live relay connections or clearing the running caches. Calling retry while Running/Degraded therefore stops the engine state and makes the next start reconnect live relays. ZUULI explicitly avoids that call unless Locked. Put the idempotency check under the engine mutex so repeated/concurrent retries preserve the active state, with a lifecycle regression test.

skyl (changes requested):
The per-device sealing boundary and unchanged wire format match ADR 0016, and all 34 checks are green. Independent review found two lifecycle regressions that need to be fixed before merge. I am preparing a focused follow-up with regression coverage for both.

wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs
skyl: Fixed in 48cf3e4e698f2fc200415456bccf43d70a2a0c27: both prepare and install reject an existing identity under the engine mutex before custody mutation. The regression injects a database write failure for a late second install, checks the original wrap key is unchanged, and reopens the original seal. It fails on the original PR implementation and passes with the guard.

wallet/e2e2z/src-tauri/src/device.rs
skyl: Fixed in 48cf3e4e698f2fc200415456bccf43d70a2a0c27: unlock checks identity under the mutex and returns the existing status when MLS is already loaded. A regression starts the engine, retries unlock twice, and verifies the active state is preserved; it fails on the original PR implementation. Unenrolled unlock also now returns NotEnrolled before querying custody. Full pinned plugin tests and clippy pass; CI is rerunning.

skyl (approved):
Reviewed the final tree independently. Both lifecycle blockers are fixed: installed identities cannot have custody overwritten by a repeated install, and unlock retries preserve an already-unlocked engine. All three new regression tests fail on the original implementation and pass with the fix. Local verification passed 118 plugin tests, clippy, fmt, both frontend typechecks, 281 unit tests and 9 browser tests; all 34 CI checks, including both required gates and native platform/packaging jobs, are green on this exact commit. The wire format and seed/device authority boundary remain intact. Cross-app transport, authority handlers, directory setup and KEM binding remain separate unfinished work.

PR #1019, opened 2026-09-14
feat(bridge): issue device credentials, and carry them over the App Link

Jhonattan2121:
Closes the two halves that stood between `issue-device-credential` and a
working round trip: ZUULI never implemented the family, and neither app
listened on the verified App Links #977 declared.

## The authority side

`intent.rs` answered `issue-device-credential` with `INTENT_UNKNOWN_INTENT`.
It now mirrors `execute_payment`'s shape:

- The native confirmation is shown **before** the seed is touched, the same
  ordering `execute_payment` uses for `ensure_active_seed_loaded`.
- Only on approval: `messaging::account_keys()` — reused rather than
  duplicated — then `identity.issue_device_credential()`.
- The credential's lifetime is ZUULI's own policy, the same constants
  `f2zmsg_enroll` already issues under. The caller's requested window is
  shown and not enforced, the treatment `execute_payment` gives an expected
  fee.
- The confirmation carries the same "Identity: NOT CONFIRMED" line for an
  unattested caller, so this introduces no trust class that payments had not
  already accepted.

The wire format is unchanged: `IssueDeviceCredentialResultV1` still carries
only `credential`, matching what #1009 built against.

## The transport

Four links in the chain were missing, and all four are here: e2e2z
dispatches, ZUULI receives, ZUULI answers, e2e2z resolves.

**ZUULI receives in Rust.** OAuth and checkout returns are received in
TypeScript because neither carries authority. This one does, so the bytes
never enter the WebView — `receive_intent` stays unreachable from the
renderer and its test still says so.

**Payloads travel in the fragment, never the query.** `CALLER-AUTHENTICATION.md`
§4.1 makes it a rule for exactly this work: a verified link degrades to the
web when the association is stale or the app is absent, and a query component
then reaches server logs, history and `Referer`.

**The reply address comes from the registry**, keyed by the caller the gate
admitted, so a request cannot name its own return address. `IntentOutcome`
carries it out of `receive_intent` rather than letting the transport re-parse
a request the gate already read.

**e2e2z dispatches through an app-crate command.** `@tauri-apps/plugin-opener`
is not a dependency there and adding it would mean granting `opener:*` in a
capability file. The renderer hands over hex and never owns the URL.

Hex rather than base64url: `hex` is already a dependency of both crates, and a
new dependency in the app holding the seed costs more than 2x encoding does.

## What this deliberately does not do

- **No Android `startActivityForResult`.** That is the only transport that
  attests a caller (§6), it is a native plugin change, and it is not here.
  Every request arrives `CallerTrust::Claimed` and the dialog says so.
- **No signature over responses** (#929). This inherits that gap rather than
  widening it: the response half rests on the domain association, which is
  what §4 says it rests on.
- **No `sign-challenge`.** Still refused, for the reason the module documents.
- **No cold-start recovery.** An answer that arrives after e2e2z was killed
  finds no pending dispatch and is ignored; enrolling again is cheaper than
  holding an authority round trip across a process boundary.
- **No device round trip.** This has never run between two signed apps on a
  phone. Everything below is local.

## The dispatch-authority rule

`wallet/e2e2z/scripts/authority-boundary.node-test.mjs` held
`setIntentTransport` to test files, and its header named the exit: the rule is
amended when #461's transport arrives. This PR is that transport, so `7d9925c`
amends it to admit exactly one production caller,
`src/lib/enrollment/appLinkTransport.ts`, by exact path. A second production
writer still fails, including a sibling in the same directory or the same
basename elsewhere.

`5028641` updates the prose that still called the seam shut: the header of
`transport.ts` and the e2e2z section of `CONFORMANCE.md`.

## Verification

- e2e2z, the full `npm test` chain: vitest 292, `node --test` 15 (including
  `authority-boundary.node-test.mjs`), Playwright 9; `typecheck`,
  `typecheck:tests` and `build` clean
- ZUULI: `cargo test` 58 passed, `clippy --all-targets -D warnings` and `fmt`
  clean
- Gate scripts: `project-boundary.mjs` (including its single-implementation
  intent-bridge guard count), `surface-capability-authority.mjs`,
  `check-tauri-plugin-permissions.mjs`, `messaging-contract.node-test.mjs`,
  `check-github-actions-pins.mjs`, `check-markdown-links.mjs`
- CI on `5028641`: 29 checks passed, 5 skipped, none failed

Draft because the round trip has not yet run between the two apps on one
device. Issuing and installing the credential needs neither a relay nor the
witness threshold in #770; those gate the directory merge and the messaging
that come after it.


skyl:
Pushed a commit amending the `zuuli / surfaces (e2e2z)` guard, which was the one real failure here.

**Why it was failing.** `wallet/e2e2z/scripts/authority-boundary.node-test.mjs` held `setIntentTransport` to test files only, and this branch adds `src/lib/enrollment/appLinkTransport.ts`, which installs a transport from production code.

**Why that licenses an amendment rather than a workaround.** The rule states its own exit condition in its header: *"`setIntentTransport` may be referenced only from test files, **until #461's transport arrives and this rule is deliberately amended along with it**."* #461 is closed — verified App Links and Universal Links landed as #977 — and this PR is that transport. So the rule is amended, on the terms it set for itself.

**What the amendment keeps.** The invariant was never "no transport exists"; it was that the registry is not a guard if *anything* in the renderer may write to it. So production may now reference the installer from exactly **one** module, `src/lib/enrollment/appLinkTransport.ts`, named by **exact path** — not `src/lib/enrollment/*`, and not a filename pattern. A second production writer still fails. Widening the set means editing a named constant in the guard, which is a diff a reviewer sees, rather than adding an import nobody notices.

Three anchors were added so the amendment cannot be read as the rule's removal:

- a judge case where a sibling transport in the same directory as the sanctioned one fails (the obvious "it's right next to the allowed one" reading);
- a judge case where the same basename under a different directory fails (the exemption is the path, not the name);
- a live anchor: if the exempted module ever stops existing or stops calling the installer, the scan fails instead of leaving a dangling exemption at a path a future file would silently inherit. That is #553's lesson applied to the amendment rather than to the rule.

The live test is renamed to what is now true — `e2e2z holds no seed authority, and no unreviewed dispatch authority` — and the prose stating the old rule is updated with it: the scan header, the `setIntentTransport` docblock in `transport.ts`, `docs/intent-bridge/CONFORMANCE.md` (mutation row, coverage-anchor paragraph, test count) and `wallet/e2e2z/README.md`.

**Verification, both directions.** Before the change the scan failed with exactly CI's message. After, all 8 of its cases pass. With a second production writer planted at `src/lib/enrollment/plantedTransport.ts`, the live scan goes red again naming that file; with `appLinkTransport.ts` moved away, the new exemption anchor goes red. Full `npm test` for e2e2z is green (292 vitest, 15 node, 9 Playwright), as are `typecheck`, `typecheck:tests` and `npm run build`.

**Two things left for you, deliberately untouched.** `transport.ts`'s module header still opens "The one seam where intent bytes would leave this process — and it is shut", and `CONFORMANCE.md`'s e2e2z section still says the request "fails at the transport, because there is not one". Both are narrative about this PR's design, not statements of the guard rule, and both are yours to reword rather than mine to guess at. Left as a draft and unmerged, per your note.

skyl:
Reviewed on the owner's behalf: adversarial review came back SHIP. Seed ordering, the `&'static` reply address, fragment-only payloads, and the pending-dispatch exits were all verified in the code. CI is green on c778de47 (`gate` and `rs / gate` both pass). Merging so enrollment and chat can build on it. Non-blocking follow-ups (stale prose in `bridge.rs`/`intent.rs`, a missing handle check against ZUULI's own enrollment, no inbound length cap, desktop `available: true`) are being picked up separately. Thanks @Jhonattan2121.

Issue #1022, opened 2026-09-17
Epic: Start encrypted chat — free2z → 2Z → e2e2z on TestFlight


skyl:
# Epic: "Start encrypted chat" from free2z → e2e2z, on TestFlight / internal test

**Goal.** Find someone on free2z → tap **Start encrypted chat** → pay 1 2Z → if they don't have e2e2z yet, they get a notification telling them to install it → both people chat end-to-end encrypted in `cash.free2z.e2e2z`.

This issue holds the **shared contracts** every workstream builds against. Change a contract here first, then in code.

## Owner decisions (2026-09-17)

1. **Directory: disposable internal log.** free2z runs the KT log *and* the only witness. The existing "this log is its own only witness" warning stays visible. The log is **explicitly disposable** and gets wiped before any public launch. That also postpones the `device_kem_pk` wire decision (ADR 0016 §6) to before the first *public* user.
2. **Whole flow at once**, not staged.
3. **Proven username → handle mapping now**, not a guessed one.
4. **Existing free2z infra** runs the relay, log and witness.

## Working assumptions (the owner may override these)

- **The 1 2Z fee is credited to the recipient**, on the existing donation ledger. It is not burned.
- **Handles.** An eligible account (lowercased username matches `^[a-z0-9_]{1,30}$`) gets that handle reserved automatically. Where usernames differ only by case, the **earliest-created** account keeps the lowercase handle, and later accounts must claim another. An ineligible account (~10%) claims an eligible handle once. Uniqueness is case-insensitive and enforced in the database.
- **A chat request never auto-sends a message.** The deep link only opens e2e2z's first-contact screen with the handle filled in.

## Contract A — the chat-request API (private backend; the free2z app calls it)

```
GET  /api/e2ee/chat-requests/price/            → 200 {"cost": 1}
POST /api/e2ee/chat-requests/{username}/
     Authorization: Token <knox>   Idempotency-Key: <uuid-v4>   body: {}
  200 → {"balance": int, "charged": int, "replayed": bool,
         "request_id": str,
         "recipient": {"username": str, "handle": str|null,
                       "handle_status": "bound"|"unclaimed"}}
  402 → insufficient 2Z (same body shape as donate-idempotent: carries the authoritative balance)
  404 → unknown user   409 → cannot message yourself   429 → rate limited
```

- The server sets the price; the client never sends an amount. The client refuses any `charged` that differs from the `cost` it showed the user, the same rule as `normalizeDonationResult`.
[...]

PR #1023, opened 2026-09-17
fix(release): check the compiled Android self-permission level bundletool prints (#1016)

### LINKED ISSUE #1016: Android release: accept signature-level compiled manifest value from bundletool
Build 22's Android release compiled its universal AAB successfully, then the credential-free inspection in the protected signer rejected the correct signature-level self-permission because it compared bundletool's numeric output to a source-XML string.

Evidence: [protected release run 34540973883, Android signer job 103089142101](https://github.com/free2z/zuu/actions/runs/34540973883/job/103089142101), `Inspect attested Android artifact without credentials`, failed at `2026-09-10T23:36:22Z`. The credential-materialization/sign/upload step never ran. This is independent of the iOS provisioning-profile repair in #1014/#1015; keep TestFlight recovery separate.

Reproduced from retained unsigned artifact `10177909601`, named `unsigned-zuuli-android-0.1.0+22-05b5f39f2ac5a5ec7e7d048382d0a296608dd29e`. Archive members were audited before extraction, all checksums verified, and the embedded verifier matched the pinned bundletool 1.18.3 SHA-256 `a099cfa1543f55593bc2ed16a70a7c67fe54b1747bb7301f37fdfd6d91028e29`.

Actual source-bound artifact:

- Source: `05b5f39f2ac5a5ec7e7d048382d0a296608dd29e`.
- AAB SHA-256: `52418ea41c8c4f5eabbdc3763bee8fa4781d99139b21d6184e4e625520109eb3`, 60,173,700 bytes.
- Manifest package/version/code: `cash.free2z.zuuli` / `0.1.0` / `22`.
- Exact declared permission set: `android.permission.INTERNET`, `android.permission.USE_BIOMETRIC`, `android.permission.USE_FINGERPRINT`, `cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION`.
- Exact native ABI set: `arm64-v8a,armeabi-v7a,x86,x86_64`.
- Dumped self-permission:

```xml
<permission android:name="cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION" android:protectionLevel="0x00000002" />
```

The local installed Android API 36 definition independently confirms `android.content.pm.PermissionInfo.PROTECTION_SIGNATURE = 2`. The compiled permission is therefore correct; the failure is the verifier's assumption about rendering.

Reproduction with the downloaded artifact's verified AAB and bundletool:

```bash
java -jar bundletool-all-1.18.3.jar dump manifest \
  --bundle=ZUULI-android-unsigned.aab > manifest-dump.xml
set -o pipefail
tr '\n' ' ' < manifest-dump.xml | tr '<' '\n' |
  grep -a '^permission ' |
  grep -aF 'android:name="cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION"' |
  grep -qF 'android:protectionLevel="signature"'
# Exit 1, reproducing the protected workflow's assertion.
```

The assertion is in `.github/workflows/zuuli-release.yml` under `Inspect attested Android artifact without credentials`. Raw merged-manifest XML can retain `signature`; bundletool's decoded compiled manifest renders `0x00000002`. A test that only feeds source XML cannot establish this protected-artifact check works.

Acceptance:

- Validate the compiled manifest's actual protection-level value while retaining the exact application-scoped self-permission and exact declared permission/ABI allowlists.
- Exercise the checker with a real bundletool dump (or a faithful retained compiled-output fixture). Preserve negative controls for missing/wrong self-permission and `normal`, `dangerous`, or additional protection flags; do not replace the check with mere permission existence.
- Print a useful mismatch diagnostic instead of a bare `test`/`grep` exit.
- Review and update the protected credential-job integrity seal and relevant tests when changing its verification program.
- A future authorized Android-only recovery/new identity must pass artifact inspection before signing and Play upload. Do not re-upload any already-delivered iOS build.

No repository changes, signing credentials, Apple profiles, or store state were modified during this diagnosis. Related: #1010, #945, #958.

skyl:
Workstream 2 of #1022.

## What was wrong

`android-sign-upload` → *Inspect attested Android artifact without credentials* ran `grep -qF 'android:protectionLevel="signature"'` on `bundletool dump manifest` output. Bundletool decodes the compiled manifest and prints `android:protectionLevel="0x00000002"`, so the grep failed on the correct AABs for 0.1.0+22 and +23. `bash -e` then stopped the job with no message, and signing and the Play upload never ran.

## What changed

- **New `wallet/zuuli/scripts/android-self-permission.sh <app-id> <manifest>`.** It follows the `aab-payload-digest.sh` pattern: the credential job can't check out source, so it installs a verbatim heredoc copy in a new step, *Install the shared Android self-permission checker*. The check requires:
  - exactly one `<permission>` element, named `<app-id>.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION`;
  - exactly one `android:protectionLevel`, either `signature` (source XML) or hex (compiled);
  - a value of exactly **2** (`PermissionInfo.PROTECTION_SIGNATURE`).
- **Flag bits are rejected on purpose, not masked off.** Each flag lets more apps hold the permission: `privileged` 0x10, `development` 0x20, `appop` 0x40, `knownSigner` 0x8000000. androidx.core injects plain `signature`, so an extra flag means the injector changed and someone should review it.
  - `normal`, `dangerous`, `signatureOrSystem` and a missing attribute also fail. Android treats a missing attribute as `normal`.
  - Each failure prints the expected and actual values, the base level and flags decoded, and the offending element.
- **The protected inspection writes the dump to `$RUNNER_TEMP` and runs the installed checker.** It deletes the dump with the other temp files. The declared-permission and ABI `test`s in that step now print expected/actual instead of failing silently.
- **`zuuli-packaging.yml` runs the same script on AGP's merged manifest.** It replaces its inline `signature` grep, so the source and compiled forms are checked by one rule.
- **`free2z-release.yml` / `e2e2z-release.yml` don't have this assertion.** They check only the `uses-permission` list, via xpath, so they aren't broken and are unchanged here. Their comments say the self-permission is signature-level, but nothing checks it. That could be a follow-up.
- **Docs:** `wallet/zuuli/docs/releasing.md` has a short paragraph on the check.

## Seals changed (full job / document reviewed)

| Seal | File | Old → new | Why |
|---|---|---|---|
| `android-sign-upload` credential job digest | `wallet/zuuli/scripts/apple-credential-boundary.mjs` | `4e80aa8c…745c4` → `e95217ff…fc2c` | New install step, the checker call in place of the grep, diagnostics on the permission/ABI checks, one extra temp file in the `rm`. No new secrets, actions, `scripts/` references, network or build commands. The boundary checker's semantic rules still pass. |
| `releasing.md` reviewed-document digest | `wallet/zuuli/scripts/status-freshness.mjs` | `352cf01b…b100` → `eeaed2fe…81f8` | The new documentation paragraph above. |

`apple-credential-boundary.node-test.mjs`: the "verifier tooling retained into secret scope" mutation now targets the new `rm` line.

**How the new test runs.** `wallet/zuuli/package.json` is left byte-identical to main. It is an input to the free2z/e2e2z store-capture contract digest, so editing it would invalidate recorded capture evidence. Instead, `android-self-permission.node-test.mjs` runs in its own step in the required `zuuli / frontend` job, *Verify Android self-permission checker*. `scripts/check-github-actions-pins.mjs` requires that step exactly, and its self-test now covers two new mutation cases: the step deleted, and the step allowed to fail.

## Verification

- **Real bundletool output.** Retained artifacts 10177909601 (+22) and 10180979793 (+23) had expired. Instead I downloaded bundletool 1.18.3, checked it against the pinned SHA-256 `a099cfa1…8e29`, and built real AABs:
  1. `aapt2 link --proto-format` against android-36 on a manifest shaped like ZUULI's (the same uses-permissions, the injected self-permission and the DUMP-guarded ProfileInstallReceiver);
  2. `bundletool build-bundle`;
  3. `bundletool dump manifest`.
- **Positive dump matches the issue.** The `signature` dump shows `<permission android:name="cash.free2z.zuuli.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION" android:protectionLevel="0x00000002"/>`, the same as the build 22 dump quoted in #1016. The old grep exits 1 on it, which reproduces the failure.
- **Those dumps are now the fixtures in `scripts/fixtures/android-self-permission/`.** Each case and its rendered value:

  | Case | Rendered value |
  |---|---|
  | signature | `0x00000002` |
  | normal | `0x00000000` |
  | dangerous | `0x00000001` |
  | signatureOrSystem | `0x00000003` |
  | signature\|privileged | `0x00000012` |
  | signature\|development | `0x00000022` |
  | signature\|knownSigner | `0x08000002` |
  | missing level | attribute absent |
  | missing self-permission | no `<permission>` |
  | duplicate self-permission | two elements |
  | extra `<permission>` | two elements |

  There are also AGP-style source fixtures for `signature` and `signature|privileged`.
- **`android-self-permission.node-test.mjs` (9 tests)** covers:
  - both renderings pass;
  - every negative control fails, with the diagnostic it should print;
  - a correct manifest checked under a different app id fails;
  - the heredoc in `zuuli-release.yml` is byte-identical to the script, is installed before the inspection, and runs before the credential step;
  - the inspection calls the installed checker and has no inline `signature` grep;
  - the packaging gate calls the script.
- **End to end against the real AABs:**
  1. Took the new install step and the manifest part of the inspect step out of the workflow YAML.
  2. Ran them with the pinned bundletool.
  3. Results: the signature AAB passes; the normal, privileged, knownSigner, absent and extra-permission AABs fail with diagnostics.
- **Ran the script on Linux** (`ubuntu:24.04` docker, GNU coreutils) against every fixture. The results match macOS.
- **Local checks:**
  - `npm run test:store-listing` passes (85/85), and `check-github-actions-pins.mjs --self-test` passes (150 current-workflow cases).
  - These pass: `node scripts/apple-credential-boundary.mjs`, `apple-credential-boundary.node-test.mjs` (203/203), `status-freshness.node-test.mjs` (118/118), `aab-payload-digest.node-test.mjs`, `scripts/check-github-actions-pins.mjs`, `scripts/check-workflow-gates.mjs`, `scripts/check-markdown-links.mjs`, `scripts/check-public-repo-boundary.sh`, `actionlint` on both workflows, and `shellcheck` on the new script.
  - I ran the full `wallet/zuuli` node-test list: 642 passed, 3 failed. The 3 failures are all in `wasm-boundary.node-test.mjs`, which needs a WASM file that isn't generated in a fresh local worktree (ENOENT). This change doesn't touch it.

## Not verified

- I couldn't use the actual build 22/23 AABs because the artifacts expired. The fixtures are real bundletool 1.18.3 output from equivalent aapt2-compiled manifests.
- The protected `android-sign-upload` job can only run in a real protected release. Android recovery for +22/+23 still needs a separately authorized run, and #1016 says not to re-upload any iOS build. `release.json` changed nothing here.
- Because this touches the release-impacting surface, the next promotion needs the usual `STATUS.md` re-derivation.

Closes #1016


skyl:

- **Where the new code runs.** The install step and the rewritten check both run before "Materialize, sign, verify… credentials". Inside the credential step, the only change is one extra temp file in the cleanup `rm`. No new secrets, actions, network calls or build commands, and the credential-boundary checker still passes. The seal moved `4e80aa8c…` → `e95217ff…` for exactly that reason.
- **The check itself.** It is exact: protection level must equal `PROTECTION_SIGNATURE` (2), and extra flag bits fail on purpose (privileged, development, appop, knownSigner). The failure was reproduced with real bundletool 1.18.3 output, and the fixtures cover every negative case.
- **Store-capture fix.** The follow-up commit keeps `wallet/zuuli/package.json` byte-identical to main, so the recorded store-capture contract is still valid. The new test runs as its own step in the required `zuuli / frontend` job, pinned by the actions-pins checker.

PR #1024, opened 2026-09-17
feat(free2z): Start encrypted chat on the creator page (#1022, workstream 4)

skyl:
Workstream 4 of #1022: a paid **Start encrypted chat** button on the free2z creator page, built against Contracts A and B.

## What the reader sees

`Tip` · `Subscribe` · **`Start encrypted chat`** → a confirmation sheet:

> **Start an encrypted chat with @zooko**
> 1 2Z, paid to @zooko. Messages are end-to-end encrypted in the e2e2z app.
> Chat request **1 2Z** · Your balance · *e2e2z opens with @zooko filled in. Nothing is sent until you write it.*
> `Pay 1 2Z and start chat`

The primary button follows `paidActionGate`: *Checking account*, *Sign in to start chat* (saves a new `start-encrypted-chat` paid intent, and signing in reopens the sheet without charging), *Not enough 2Z — buy more*, or *Pay*.

| Outcome | Title | Actions | May say "nothing was charged"? |
|---|---|---|---|
| 200, `bound` handle | Chat ready with @x | **Open e2e2z**, Get e2e2z | n/a |
| 200, `unclaimed` or unusable handle | @x has been notified (they need to claim a messaging handle in e2e2z first, so there is no link yet) | Done, Get e2e2z. No link is built. | n/a |
| 402 | Not enough 2Z (shows the authoritative balance from the body and updates the session) | Buy 2Zs | yes |
| 404 | @x can't be found | Done | yes |
| 409 `price_changed` | Back to the confirmation with a "The price changed" notice (focused) and the **new** price; the next Pay is a new attempt with a new key | Pay {new cost} | yes |
| 409 (any other) | This is your own profile (the button is also hidden on your own page) | Done | no |
| 422 `sender_handle_unavailable` | Claim your messaging handle in e2e2z first | **Get e2e2z**, Done | yes |
| 429 | Too many chat requests | Try again (**new** key) | yes |
| 401/403 | Sign in again | Sign in | yes |
| other 4xx (including a 405 or a 422 without that code) | free2z refused this request | Done, then the balance is refreshed | no |
| charged ≠ shown cost, or the recipient differs from the creator on screen | free2z's answer didn't match | Done, then the balance is refreshed | no |
| network loss or 5xx | free2z did not get an answer ("…whether 1 2Z was charged. Trying again is safe…") | Try again (**same** key) | no |

- **Idempotency:** a uuid-v4 key is created on the first Pay. It is kept, along with the cost shown, while the outcome is unknown. Closing and reopening the sheet shows the uncertain state again rather than starting a new key. A definite answer ends the attempt.
- **Message again:** after a success, the button becomes *Message again*. It opens e2e2z directly, or reopens the "notified" status if the recipient has no handle yet. It never charges again. This memory lives only in process memory, keyed by account and creator, and nothing is written to storage.
- **Contract B:** `e2e2zChatLink` accepts only `^[a-z0-9_]{1,30}$`, and the handle goes in the fragment only. `openE2e2z` refuses any URL it did not build itself (for example the intent reply route, a query string, another host, or an uppercase handle). In the app it uses `openUrl` (`opener:default`, already granted). In a browser it uses `window.open(…, noopener)`.
- **Honest-copy guard:** `chat-request-copy.ts` is an exhaustive `switch` with a `never` check, in the same style as `tip-copy.ts`. `chat-request-copy.test.ts` renders the shipped **en/es/fr** catalogs and fails if a message that can't prove it claims nothing was charged. A negative control per locale checks that the matcher can actually match.
- **Scope:** changes are limited to `wallet/free2z/`. No new capability, plugin or `invoke_handler`. `surface-capability-authority.mjs` passes.
- **Accessibility:** accessible names contain the visible label (`Start encrypted chat with @x`). Focus moves to the next action whenever the sheet changes state and returns to the opener when it closes. Escape and outside clicks are ignored while a charge is in flight. Footer buttons wrap instead of clipping, and a 320px check asserts that no dialog text overflows. Motion follows the global `prefers-reduced-motion` rule.

## Contract choices, updated to the backend's real behavior (please confirm)

The chat-request backend now exists, and it differs from the original Contract A text in ways this client now matches:

1. **Money arrives as decimal strings** (`"balance": "9.000"`, `"charged": "1.000"`). The existing parsers already accepted this. Tests now pin it, including refusing strings like `"1.500"`, `"1e0"` and `"1,000"`.
2. **The POST body is `{"expected_cost": <shown cost>}`.** A **409 with `code: "price_changed"`** is a certain refusal: it ends the attempt, the price is fetched again, and the payer re-confirms at the new amount. **Any other 409** is still read as "cannot message yourself". Because donate-idempotent uses 409 for key mismatches, that copy does not say "nothing was charged".
3. **A 422 with `code: "sender_handle_unavailable"`** means the payer has no bound handle. It gets its own outcome: nothing was charged, "Claim your messaging handle in e2e2z first", and Get e2e2z. Refusal codes are read defensively: a top-level `code` string, or the DRF-style `detail.code`. Anything else (a string body, an array, a non-string code) counts as no code, so a malformed 422 is a generic refusal that doesn't claim nothing was charged.
4. **A creator named `price`** (compared case-insensitively) has no button, because that path is the price endpoint. A 405 from anywhere else is still a generic refusal.
5. **"Get e2e2z" URL.** Contract B defines only the `#peer=` link. *Get e2e2z* opens `https://free2z.com/bridge/e2e2z/chat/` with **no fragment**. That path is inside the App Link prefix, so a device that already has e2e2z opens the app with no `peer`, which e2e2z must refuse harmlessly (workstream 5). If the backend prefers a separate install URL, it is one constant: `E2E2Z_INSTALL_URL`.
6. **A `bound` status with a handle that fails the pattern** is shown as unclaimed (honest copy, no link) rather than an error. An unknown `handle_status`, or a `recipient.username` that differs case-insensitively from the requested user, is refused as a contract mismatch.
7. **`request_id`** must be a non-empty string of at most 128 characters. The price must be a positive whole number of 2Z.
8. **Retry keys.** A 429 retry uses a new key, on the assumption that throttling happens before the view records the key. A 402, a `price_changed` or a `sender_handle_unavailable` refusal also ends the attempt. Only an unknown outcome (network loss, a 5xx) reuses the key.

The sender's record-only event needs no client work.

## Verification (local, `wallet/free2z`)

- `npm run typecheck` and `npm run typecheck:tests`
- `npm test`: vitest (73 files, 1092 tests), plus the ui-copy truncation and release-path node tests, plus all 104 Playwright specs. `creator-responsive.pw.ts` now expects the third action target. Its touch-height reading is snapped to Chromium's 1/64px layout grid, because the third button intermittently read 43.99997px from float noise; a real 43.984px target still fails.
- New Playwright spec `tests/encrypted-chat.pw.ts` covers: the sign-in detour keeping the intent without charging; success opening the exact Contract B URL, then *Message again* opening it with no second charge; an unclaimed recipient (no `#peer=` link is ever opened); a 402 with the authoritative 0 balance, which then gates to *buy more*; a lost response that stays uncertain across close and reopen, where *Try again* replays and the balance is debited once; a 429 starting a new attempt; a 422 `sender_handle_unavailable` (Get e2e2z opens the install URL, nothing is charged); a `price_changed` 409 that shows the new 2 2Z price, after which paying debits 2 exactly once; a creator named `price` and your own profile both having no button; the 320px no-clipping check.
- New unit and contract tests: `chat-request.test.ts` (normalizers and classification), `chat-request-http.test.ts` (routes, empty body, key header, refusing a mismatched charge, no request sent with a malformed key, the 402 balance), `e2e2z-chat.test.ts`, `chat-request-copy.test.ts`, `chat-request-session.test.ts`, and a paid-intent round trip.
- `npm run build`, `node wallet/zuuli/scripts/surface-capability-authority.mjs`, `scripts/check-public-repo-boundary.sh`, and `rtl-source-policy` (node test and verdict) all pass.
- **Not run locally:** `wallet/zuuli/scripts/project-boundary.mjs`. It needs `wallet/zuuli`'s own `npm ci`, and the machine was nearly out of disk. CI's `zuuli` job runs it. The new files import only `@/…`, `react`, `lucide-react`, `sonner` and `@tauri-apps/plugin-opener`, all of which free2z already depends on.

Screenshots were captured at a 390px width with Playwright (confirmation, chat ready, notified, uncertain, insufficient, and the *Message again* action row) and reviewed locally. They aren't attached because `gh` can't upload images to a PR body. The Playwright spec asserts the same rendered copy.

The backend endpoint (workstream 6) is being built in parallel. This PR is tested only against the HTTP-layer mock and the in-process `VITE_MOCK` ledger. The mock now returns the real decimal-string shapes and enforces `expected_cost`. It takes a fault queue from `sessionStorage["zuuli.mock.chat-request"]`, the same way `zuuli.mock.creator-pages` works.


skyl:

- **No new capability.** No `invoke_handler`, plugin, capability or Tauri config change. The only reach into another app is `openE2e2z`, which accepts exactly the install URL or a `#peer=` link it can rebuild from a pattern-valid handle. It can't be used as a general URL opener.
- **Contract A.** The server sets the price. A `charged` that differs from the shown cost, or a recipient other than the creator on screen, is refused. The idempotency key is reused only when the outcome is unknown. Only 402/404/429/401/403 may say "nothing was charged", and an exhaustiveness test enforces that across en/es/fr.
- **Contract B.** The handle goes in the fragment only, validated before the URL is built.
- **Store capture.** No store-capture contract inputs are touched.

**Update (`8fe117c4`), aligned with the real backend:**
- Decimal-string amounts are pinned by tests.
- 422 `sender_handle_unavailable` has its own outcome ("Claim your messaging handle in e2e2z first", nothing charged).
- The POST sends `expected_cost`; a 409 `price_changed` ends the attempt and re-confirms at the new price with a new key.
- The button is hidden for a creator literally named `price`.

PR #1025, opened 2026-09-17
feat(e2e2z): open chats from links and enroll through ZUULI (#1022)

skyl:
Workstream 5 of #1022: the e2e2z side of "Start encrypted chat". Contract B is implemented as written.

## What changed

**Contract B chat link** (`src/lib/chat/chatLink.ts`, installed from `main.tsx`)
- Handles `https://free2z.com/bridge/e2e2z/chat/#peer=<handle>` on cold start (`getCurrent`, deduplicated across webview reloads) and warm start (`onOpenUrl`).
- Parsing is strict: https only, host `free2z.com`, exact path, no query, no credentials or port, a single `peer` key, and `^[a-z0-9_]{1,30}$` with no decoding or case folding. Any other URL is ignored. A malformed link on this route shows "That chat link didn't work" and is never acted on.
- The link only fills in first contact (the field is filled, focused, and says "Nothing has been sent"). It **never auto-sends**.
- If the device isn't enrolled, or its handle isn't active yet, the peer is kept as pending (local storage, 24 h, dismissible) and filled in once first contact renders.
- The intent reply route now matches `/bridge/e2e2z/` **exactly** (`REPLY_PATH`), so the two routes cannot collide. Tests pin both directions.

**"Enroll with ZUULI"** (`features/messages/Enrollment.tsx`, `lib/enrollment/outcome.ts`)
- Flow: explain, then a username field. `check_handle_eligibility` produces the candidate, labeled as a *request*: the free2z directory decides the binding (Contract C).
- "Enroll with ZUULI" dispatches `issue-device-credential` through the #1019 path. It then waits, with Cancel and a live countdown to the request's own 2-minute deadline, and installs the result.
- Each failure has its own copy: cancelled, expired, declined (`NOT_CONFIRMED`), ZUULI not ready (`UNAVAILABLE`), caller not registered, link failed or ZUULI not installed, **handle mismatch**, durability unavailable, unknown status, and defect (with the detail shown).
- **Handle mismatch (ADR 0016 §4): confirmed that #1019 already does this.** `bridge.enroll` passes the *requested* handle to `e2e2z_install_device_credential`. `Engine::install_identity` compares it with the credential's signed handle and refuses with `handle-ineligible` before it consumes the pending secrets. The UI maps that refusal, from the install step only, to "ZUULI signed a different handle".
- New seed-free app-crate command **`e2e2z_enrollment_status`**, which calls `Engine::enrollment_status`. Without it the screen could not tell an enrolled device from an unenrolled one. It is not an `f2zmsg_*` name, so `lib.rs`'s guard still holds. `WIRE_COMMANDS` and bridge parity are unchanged. `enroll` still goes through the transport, and `unenroll` still refuses.
- Desktop and browser builds keep the standing "Enrollment happens in the wallet app" state. The copy was updated; the title is unchanged.

**Engine state**
- "Messaging service not configured in this build" appears when `relaysConfigured === 0`.
- The relay card shows connected, not connected, or none configured.
- The existing witness warning now also shows before enrollment.
- A locked device gets "Messages are locked on this device" with a seed-free **Unlock** button (`e2e2z_retry_device_unlock`, which previously had no caller). The locked copy no longer mentions a seed.
- "Submitted, not yet active" shows the engine's own `blocked` code.
- No free-form add-relay screen.

**#1019 review follow-ups**
- (a) `bridge.rs` rationale corrected: `opener:default` includes `allow-default-urls` (I checked the plugin's `default.toml`), so the command is a validation point, not a boundary.
- (b) `installAppLinkIntentTransport` installs only on iOS, iPadOS and Android, detected from the user agent. Desktop stays on the fail-closed transport (`available: false`) instead of opening a browser and waiting two minutes.
- Link-open failures are now typed (`AuthorityLinkError`), and the transport gained `cancel()`.

## Verified locally
- `npm run typecheck` and `npm run typecheck:tests` pass.
- vitest: 25 files, 373 tests pass. New: `chatLink.test.ts`, `outcome.test.ts`, `Enrollment.test.tsx`, `index.chat-link.test.tsx`, and FirstContact prefill cases. `enrollment-gap.test.ts` was deliberately updated: `getEnrollmentStatus` now reads through `e2e2z_enrollment_status`, never an `f2zmsg_*` command, and never resolves with anything but a parsed engine answer.
- Node tests: `ui-copy-truncation`, **`authority-boundary`** and `release-path` all pass (15/15). zuuli's `messaging-contract` and `surface-capability-authority` also pass.
- Playwright, 26 tests passing across the mocked and unmocked servers:
  - New `tests/chat-link.pw.ts`: cold and warm start, malformed links, other routes and intent replies ignored, pending-peer resume through a real enrollment, survives a reload, and dismiss.
  - New `tests/enrollment-flow.pw.ts`: the happy path with the real request and response envelope and install arguments, plus declined, wallet not ready, handle mismatch, durability, link failed, cancel (a late answer is ignored), the 2-minute expiry with `page.clock`, service not configured, a 360px width with no overflow, and a mocked enrollment.
  - `tests/enrollment-gap.pw.ts` deliberately updated: the desktop user agent still gets the gap. Its stub now answers `e2e2z_enrollment_status`. It asserts no `f2zmsg_*` command, no key sample and no dispatch.
- `npm run build` passes.
- App crate: `cargo fmt --check` passes, `cargo clippy --locked --all-targets -D warnings` passes, and `cargo test --locked` passes (11 tests).
- `scripts/check-public-repo-boundary.sh` passes.

## Dependencies and follow-ups
- **Workstream 3 (engine):** first contact appears only once `mergedAtEpoch` is non-null. Until workstream 3 sets it from the log receipt, an enrolled device stays on "Submitted, not yet active" (the engine currently reports `blocked: directory-unreachable`), and a pending chat link waits. The "not configured" notice goes away once workstream 3 bundles a default relay. No engine code was touched here.
- **Store capture (release):** `wallet/zuuli/scripts/surface-store-capture.mjs`'s e2e2z native stub does not answer the two new startup calls, `e2e2z_enrollment_status` and `plugin:deep-link|get_current`, so the `unenrolled-native-v1` capture would now render the failure screen. Updating that stub changes the recorded `contractDigest`, so it has to land together with a re-capture. I left it untouched.
- **Workstream 8:** when ZUULI isn't installed, the dispatched App Link opens `free2z.com/bridge/zuuli/` in a browser. That page needs an "Install ZUULI" landing, like Contract B's. The e2e2z copy tells the user what to do, but there is no store URL to link to yet.

Refs #1022


skyl:

- **The new command.** `e2e2z_enrollment_status` only reads this device's own stored enrollment. It is seed-free, needs no capability, is named `e2e2z_*` rather than `f2zmsg_*` (which the crate test forbids here), and is recorded in `docs/architecture.md`.
- **Links.** The chat link is parsed strictly (https, free2z.com, exact path, fragment only, pattern-valid `peer`). The ZUULI reply route now requires the exact path `/bridge/e2e2z/`, so the two routes can't be confused, and nothing is ever sent automatically.
- **Enrollment.** The returned-handle check (ADR 0016 §4) is enforced by the install command and the engine.
- **#1019 follow-ups.** The stale `bridge.rs` rationale is corrected, and the transport reports `available: false` on desktop.
- **Known dependency.** First contact appears only once `merged_at_epoch` is set (workstream 3 of #1022). The e2e2z store capture needs a re-capture at release time; no contract inputs changed here.

PR #1026, opened 2026-09-17
feat(e2ee): activate messaging on a disposable internal directory (#1022)

skyl:
Workstream 3 of #1022. Records [ADR 0017](docs/e2ee/decisions/0017-internal-directory-activation.md) and implements it end to end in source. The shipped config still holds placeholders, so **no build from this source connects anywhere yet.**

## What changes

- **ADR 0017** records the disposable internal directory posture: free2z runs the log and its only witness, `t = 1`, and that witness is not independent. The ADR states exactly what this posture does and does not guarantee, how config reaches devices, the enrollment-to-submission sequence, Contract C (byte format ratified from `f2z-authority`), what `mergedAtEpoch` means, and what must change before public launch.
  - `KT.md` §12, `docs/status.md`, `docs/e2ee/README.md` and `rs/README.md` are updated to match.
- **Config**: `wallet/plugins/tauri-plugin-f2zmsg/internal-directory.conf` is compiled in and **fails closed**.
  - With every value `PLACEHOLDER`, the build stays on `NoDirectory` with no relay.
  - A partly filled file is malformed, and a unit test fails the build on it.
  - `log_id` is checked against the log key, and `vrf_public_key` is pinned.
  - There is no environment override, and the file cannot mark a witness independent.
- **Wiring** (`lib.rs`, `directory.rs`, `engine.rs`):
  - The engine uses a lazily connecting `BundledDirectory` plus a default relay, added through `add_relay` on `start_engine`.
  - `get_witness_set_state` and `list_witnesses` report the bundled policy. `set_witness_set` is refused while that policy applies.
  - `start_engine` refreshes the directory's root on a blocking thread.
  - **The last verified tree head now persists across restarts** via `KtClient::open`. A log reset under the same key is refused as a rollback; a new `log_id` sets the old checkpoint aside.
- **Submission**:
  - `f2z-msg-identity::AccountKeys::sign_directory_submission` signs the entry with the `DirectoryAuthKey` and the assertion binding with the identity key.
  - `f2z_kt_client::submit` posts the entry and verifies the receipt against the pinned log key and against the submitted handle, version and entry hash.
  - ZUULI's `f2zmsg_enroll` (new optional `knoxToken` argument) then does the rest:
    1. starts the engine;
    2. fetches the assertion (`POST /api/kt/handle-assertion/`);
    3. prechecks it with `f2z-authority` against the bundled authority key;
    4. submits and keeps the receipt;
    5. sets `mergedAtEpoch` **only from a verified lookup** of its own handle.
  - `blocked` now reports the real reason.
- **Log** (`f2z-kt`): new adversarial tests refuse an assertion from an unconfigured authority, and one forged under the configured `authority_id`.
- **Contract C vectors**: `rs/crates/f2z-authority/tests/fixtures/handle-assertion-v1.vectors` is pinned by `tests/contract_c_vectors.rs`. The layout, hash derivations and Ed25519 signature were reproduced independently in Python.
- **e2e2z**: the witness warning stays up until two independent witnesses cosign. It has separate copy for "threshold met by free2z's own witness".
- **CI and guards**: the ZUULI change selector now names `rs/crates/f2z-authority` (with a new digest), because ZUULI links it directly. The label register, dependency register and `messaging-contract` `unenroll` digest are updated, each with a note.

## Verified locally

- **Plugin** (`--features relay-harness`): all tests pass, including both two-process relay tests; `clippy -D warnings` is clean.
- **rs**:
  - `f2z-msg-identity`, `f2z-authority`, `f2z-codec`, `f2z-kt`, `f2z-kt-core` and `f2z-kt-client`: all suites pass.
  - `scripts/check-rust-clippy.sh --root rs` is clean.
  - The wasm32 build of the client-linked crates succeeds.
- **ZUULI**: `cargo test` passes (68) and `clippy --all-targets -D warnings` is clean.
- **e2e2z**:
  - Rust: `clippy` is clean and `cargo test` passes.
  - TypeScript: `vitest` (377), `tsc` for build and tests, and Playwright (26) all pass.
- **Formatting**: `check-rust-fmt.sh` passes for `wallet/` and `rs/`.
- **Guard scripts**: `project-boundary`, `surface-capability-authority`, `messaging-contract`, the store-listing tests, `check-zuuli-android-gate` (with self-test), markdown links, hash labels (with self-test), dependency register (with self-test), workflow gates, action pins and plugin permissions all pass.
- **Not run**: `cargo-deny` (not installed here), and Linux CI (macOS only).
- **Mutations** (ADR 0017 §7), each caught and restored green:
  - authority rules 12 and 13;
  - `KtClient::open` failing open, or ignoring the checkpoint;
  - ZUULI's precheck accepting everything;
  - the e2e2z warning keyed on the threshold alone.

## Not in this PR

- **e2e2z device publication.** It needs the requester's contact endpoint in the credential request. ADR 0017 §4.1 specifies `IssueDeviceCredentialRequestV2` and argues it against ADR 0016 §4. The follow-up must also let e2e2z open its contact queue before install.
- **Real config values.** The deployment's values are ready except `handle_authority_pk`, which has not been minted. Because a partly filled file is malformed, the values should land as one follow-up once the key exists and the services answer.
- **A ZUULI UI trigger** for `f2zmsg_enroll`.
- **Persisting per-handle pins and alarms.** Only the tree-head checkpoint persists today.
- **Staging builds.** The file holds one environment, so dev builds cannot point at staging yet.
- **Device testing.** Nothing here has run against a deployed log or on a device.


skyl:
Adversarial security review for the owner (#1022): **SHIP**. The checked-in `internal-directory.conf` holds only placeholders, so every build from this PR stays on `NoDirectory` and connects nowhere.

**Verified sound**
- **Tree heads and witnesses:** STH checks are pinned to the log key and `log_id`. Witness cosignatures are checked against the configured keys and threshold; below the threshold, lookups fail closed. The VRF key is pinned.
- **Rollback:** rollback and monotonicity rules hold against a resumed checkpoint. There is no fallback to trusting the relay.
- **Seed and token:** the seed stays in ZUULI. The Knox token is redacted in `Debug`, redirects are disabled, and the token is sent only to the bundled https URL and never stored.
- **Precheck:** it covers handle, identity key, `log_id`, the authority key and the validity window.
- **`mergedAtEpoch`:** set only from a verified lookup that lists this device.
- **Log side:** assertion enforcement still holds.
- **Boundaries and warnings:** no new plugin commands or permissions. Witness independence is hard-coded to false, so the warning conditions match ADR 0017.

**Must land before the real values go into `internal-directory.conf` (tracked in #1022)**
1. **Vouching check.** Require the log's signed authority policy to be "vouched" and to list exactly the bundled `handle_authority_pk`. Today `refresh_authority_policy()` is discarded, so a log deployed without its authority would accept first-come registrations.
2. **Checkpoint handling.** A checkpoint whose `log_id` bytes were flipped, or that was deleted, is silently re-trusted, because `WrongLog` is checked before the signature. Correct the ADR 0017 §3 wording, and move the checkpoint into integrity-protected storage with fsync and a recovery path.

**Follow-ups**
- Config values: printable ASCII only, reject `#` inside URLs, and set a minimum `reset_cooldown_seconds`.
- Parse the config at compile time.
- A user-intent step before signing (#905).
- An end-to-end test of `f2zmsg_enroll` against a real log.
- Correct the `status.md` §2.4 wording.

PR #1027, opened 2026-09-17
fix(e2ee): require the log's vouching, and seal the directory checkpoint (#1022)

skyl:
The security fixes the adversarial review of #1026 said must land **before** the real values go into `wallet/plugins/tauri-plugin-f2zmsg/internal-directory.conf`. Workstream 3c of #1022.

Every build from this branch still carries only `PLACEHOLDER`s, so it stays on `NoDirectory` and connects nowhere.

## 1. Vouching is required, not merely reported (review item 1)

`KtClient::refresh_authority_policy()`'s result was discarded by `KtDirectory::connect`, so a log deployed **without** its handle authority would have accepted first-come registrations while this client went on resolving them — which is the whole value of Contract C.

- New `KtClient::require_authority_policy(&[PublicKey])`: fetch §4.6's `SignedAuthorityPolicy`, verify it under the accepted log key, and require it to **vouch** and to list **exactly** the expected authorities (no key missing, none extra; duplicates and an empty requirement are a `Configuration` error). The comparison is on public keys, and `authority_id = H("free2z/kt/v1/authority-id", pk)` is a function of the key, so equal key sets are equal id sets.
- `DirectoryConfig::required_authority` carries the bundled `handle_authority_pk`. A bundled `KtDirectory` checks the policy on connect and on **every root refresh**, pins the first policy it verified (with `published_at_ms` zeroed) and **latches unusable** on any later change. Lookups, `submit` and `start_engine`'s refresh then fail closed.
- A policy that cannot be fetched at all is a refusal, not a pass.
- New closed-union code **`directory-unvouched`**, surfaced in `EngineStatus.lastError` (start_engine now records the directory refresh error) and in e2e2z — a destructive callout on the messages page and first-contact copy.

## 2. Checkpoint handling (review item 2)

`SignedTreeHead::verify` checks `log_id` **before** the signature, so `KtError::WrongLog` said nothing about who signed the bytes. Flipping one `log_id` byte in `f2zmsg-kt-checkpoint.bin`, or deleting the file, made the device trust the log's next head on first use.

- `KtClient::open(.., retired_log_pks)` now: resumes this log's head; refuses a head that verifies under the **current** key while naming a foreign id (`WrongLog` — the log misbehaving); sets a head aside **only** when its id is `H(label, pk)` for a `pk` the build **names as retired** *and* the signature verifies under that `pk`; otherwise refuses with the new `ClientError::UnrecognisedCheckpoint` (what a tampered id looks like) or `BadSignature`. Nothing bootstraps over an unexplained checkpoint.
- `retired_log_public_key` is a new, repeatable config key. ADR 0017 §8.3's wipe procedure now says to list the old genesis key; a build that forgets leaves devices refusing until they re-enroll, which is the conservative direction.
- **The loose file is gone.** The head is stored in the engine's own SQLite store, sealed with ChaCha20-Poly1305 under a key derived from this device's unsealed queue seed (`engine::SealedCheckpoints`). Durability is SQLite's `synchronous = FULL` + WAL, one transaction per save, rather than a hand-rolled `fsync`. The first save is mandatory; later saves are best-effort (saving less is safe and only loses progress).
- Refused, with the new **`directory-state-invalid`**, and never silently re-trusted: a record that does not decode or does not authenticate under this device's key (edited, truncated, transplanted), and a **missing** record on a device that has already relied on the directory (it holds a receipt, a merged epoch, or a conversation).
- **The safe recovery is explicit and documented**: there is no "reset directory state" button — a button that clears the evidence of what this device saw is a button an attacker wants pressed. The recovery is `f2zmsg_unenroll`, which already takes a typed confirmation and now deletes the checkpoint in the same transaction as the identity and its secrets; `shutdown` and `unenroll` detach the store. The UI says exactly that.
- What this does **not** claim: a party who can rewrite `f2zmsg.sqlite` arbitrarily can delete the record and the evidence beside it. That party can equally rewrite stored peer identities, and the store's missing at-rest encryption is `f2z-msg-store`'s recorded gap. ADR 0017 §3 states the limit.
- ADR 0017 §3 is rewritten to match the implemented behaviour exactly.

## 3. Config parser strictness (follow-up)

- A `#` is a comment **only** at the start of a line. Inside a value it is an error — the old parser cut every line at its first `#`, silently turning a URL with a fragment into a different URL.
- Values must be printable ASCII with no spaces: a zero-width space, a no-break space or a Cyrillic look-alike in a host cannot pass review. Comment lines may carry any UTF-8.
- URLs must name a plain host: no userinfo (`https://free2z.cash@elsewhere`), no backslash, non-empty host of letters/digits/`-`/`.`/`:port`.
- `reset_cooldown_seconds` has a floor of **604800** (seven days) — ADR 0014's proposed value and the log's own default, justified in `syntax.rs`: the client enforces the value on every platform reset it sees, the window protects only a user who opens the app inside it, and a smaller number in a reviewed file is far more likely a test fixture's `60` than a decision. Lowering it is a code change.
- **Well-formedness is now a compile-time guarantee.** The grammar moved to `src/internal_directory/syntax.rs` (std only, no crate deps) and `build.rs` includes it through `#[path]` and parses the checked-in file, so a malformed file fails the build of every crate that links the plugin — not only a unit test. The `log_id` derivation is checked too, via a tiny `blake2` build-dependency (already in every lockfile; the unit test `the_build_time_log_id_is_the_protocol_log_id` holds it to `f2z_kt_core::labels::log_id`).

## 4. Docs honesty (follow-up)

`docs/status.md` §2.4 claimed `f2zmsg_enroll` was "tested against a real log and witness". The real-log test drives `AccountKeys` + `f2z-kt-client` directly. The page now separates what is proved against a real log from what is not, and says why an end-to-end test through `publish_steps` cannot live where the code does: `f2z-kt` and `f2z-witness` are AGPL-3.0, so a dev-dependency from ZUULI or from the plugin would put them in a shipping lockfile (`rs/README.md`'s boundary), while `f2z-kt-client` — which may dev-depend on them — cannot depend on Tauri. The honest closure is the deployed-log observation §2.4 already lists as missing.

## Tests, and the mutation proof for each guard

Each guard was broken, the suite run, and the guard restored (`docs/intent-bridge/CONFORMANCE.md`'s method). ADR 0017 §7 carries the table; in full:

| Guard mutated | What failed |
|---|---|
| `KtClient::open`: another log's checkpoint bootstraps with no retired list | `only_a_named_retired_generation_is_set_aside_and_anything_else_wrong_is_refused` |
| `KtClient::open`: the foreign-id-under-the-current-key check removed | same test, on the `WrongLog` assertion |
| `KtClient::open`: the checkpoint ignored entirely | same test + `a_log_wiped_under_the_same_key_is_refused_after_a_restart` |
| `require_authority_policy`: `exact = true` | `the_bundled_client_requires_exactly_the_bundled_authority` |
| a `#` put inside a URL in `internal-directory.conf` | **the build**: `internal-directory.conf is malformed (ADR 0017 §3): line 90: \`#\` is only allowed at the start of a comment line` |
| parser: the cooldown floor removed | `the_reset_cooldown_has_a_floor` |
| parser: the printable-ASCII rule removed | `values_are_printable_ascii_only` |
| parser: a retired key may equal the current log key | `retired_generations_are_real_distinct_and_not_the_current_log` |
| `directory_config()`: `required_authority: None` | `the_bundled_authority_becomes_the_required_authority` |
| `BundledDirectory`: connects without a checkpoint store | both `directory::tests` |
| `SealedCheckpoints::load`: a record that does not open treated as absent | `an_edited_or_transplanted_record_is_refused_rather_than_ignored` |
| `SealedCheckpoints::load`: the missing-but-relied-on rule removed | `a_missing_record_is_refused_once_the_device_has_relied_on_the_directory` |
| `clear_identity`: the checkpoint left behind | `unenrolling_clears_the_checkpoint_and_detaches_the_store` |
| e2e2z: the unvouched banner never rendered | `names an unvouched log and what it means` |

New tests: `seed_submission.rs` gains `the_bundled_client_requires_exactly_the_bundled_authority` (vouched / unvouched / another key / an extra key / a policy signed by an impostor / never served / an empty or duplicated requirement) and rewrites the checkpoint test to cover the flipped id, the unlisted log, the forged retired head, the foreign id signed by the current key, and a truncated file; the plugin gains 8 `internal_directory` tests, 2 `directory` tests and 4 `engine::directory_state_tests`; e2e2z gains 3.

**Green locally**: `cargo test -p f2z-kt-client` (all 6 suites), the plugin's full suite **including the two-process relay tests** (`--features relay-harness`, real `f2z-relay` binary), `cargo clippy -D warnings` on the `rs` workspace, the plugin (`--all-targets --all-features`) and **ZUULI** (`--all-targets`, on the pinned 1.97.1), `cargo fmt --check` on both trees, `cargo-deny` over `wallet/` (advisories, bans, licences, sources — 8 crates ok), e2e2z `typecheck` + `typecheck:tests` + vitest (380) + node guards (15) + **Playwright (26)**, `wallet/zuuli` `messaging-contract` and `surface-capability-authority` node tests, and the repo guards `check-hash-domain-labels`, `check-markdown-links`, `check-public-repo-boundary`, `check-akd-doc-evidence`, `check-kt-sth-repeat-agreement`, `check-tauri-plugin-permissions`.

One reviewed digest moved: `messaging-contract.node-test.mjs`'s pinned bodies for `unenroll` and `shutdown`, because both now detach the checkpoint store. The comment beside each says why; the ordering those digests exist to pin is unchanged and is also checked structurally.

Out of scope, deliberately: the user-intent step before signing (#905) and `IssueDeviceCredentialRequestV2` (ADR 0017 §4.1), which belong to other workstreams.


skyl:
Review fixes pushed as `69fbec18`. F1, F2, F3 and F5 are addressed; F4 and F6 are documented as the review allowed, with the reasoning below.

## F1 (blocker) — only a statement about vouching is latched

`recheck_policy` latched every policy-fetch failure, so a 429 from the log's own `/authority` endpoint — or a captive portal answering 200 with a non-canonical body — left the directory refusing everything for the life of the process, under a code §8 says to retry, with no UI keyed on it.

Latched now: `AuthorityPolicyMismatch`, `Protocol(BadSignature)` and `Protocol(WrongLog)`. **Why those two protocol errors belong:** neither is a bad moment. A document that does not verify under the log key this client has *already accepted* is the log using a key it is not entitled to, or a party on the path substituting the document; a policy naming another `log_id` is a replay of a different log's document. Both are statements, and asking again cannot change them. Everything else — unreachable, refused (including `ERR_RATE_LIMITED`), an answer that does not decode, a catch-up checkpoint — is returned to the caller and forgotten, so the next call asks the log again.

The first attempt at this only tested the `latching()` predicate, and a mutation that deleted the non-latching arm at the call site **survived**. So the decision and its application were extracted into `policy_verdict` (what happened → `Unchanged` / `Pin` / `Report` / `Latch`) and `apply_verdict` (what that does to the pin, the latch and `threshold_met`), both of which a test drives with no log, no transport and no connection. What is left untested is a two-line function that fetches and calls them.

**Mutation proof (F1):**

| Mutation | Test that failed |
|---|---|
| `policy_verdict`: every `Err` becomes `Latch` (the pre-review behaviour) | `a_policy_that_could_not_be_checked_is_reported_and_forgotten` |
| `apply_verdict`: a `Report` falls through to the latch path | same test |
| `apply_verdict`: `Pin` stops pinning | same test |
| `latching()`: `Protocol(_) | Refused(_)` added | `only_a_statement_about_vouching_is_latched` |

The test asserts each transient error maps to its own code (`directory-rate-limited`, `directory-unreachable`, `directory-protocol-violation`) as a **report**, that a report leaves `unusable` empty and `threshold_met` untouched, and that a latch sets both.

## F2 — `directoryBlocked`, written by directory state and nothing else

`EngineStatus` gains `directoryBlocked: ErrorCode | null`, read from the directory itself on every status call (`Directory::directory_blocked`), so nothing on the engine can overwrite it. `BundledDirectory` reports its connection's latch, and also remembers a **connect** refusal that is one of the two persistent codes — a policy mismatch at connect leaves no `KtDirectory` to hold a latch, and that is exactly the case the UI has to explain. It is cleared by a connection that succeeds and by `attach_checkpoints` (new secrets, new question). Mirrored in `types.ts`, `mock.ts`, the Playwright host and `CLIENT-CONTRACT.md` §3.1, which states the difference from `lastError` in the shape a reader will need it.

`checked`/`invoke` in `bridge.ts` now take `z.ZodType<T, z.ZodTypeDef, unknown>`: the old `z.ZodType<T>` forced input and output equal, so `.default(null)` (there so an older native build and the capture fixture still parse) would have made `T` carry an optional field. No capture contract input was touched.

**Mutation proof (F2):** keying the banner back on `status.lastError` fails `stays up while the relay weather rewrites lastError`, plus both banner tests. On the Rust side, `a_blocked_directory_is_reported_apart_from_whatever_failed_last` sets `inner.last_error` the way `pump_inbound` does and asserts `directoryBlocked` is unmoved — and that a directory with nothing wrong reports `None` however badly the rest of the engine is doing.

## F3 — the pin compares sets, not order

`comparable_policy` sorts `authorities` and `asserted_versions` before comparing. Mutation: drop the sorts → `a_reordered_policy_is_the_same_policy` fails. The test also holds it to still being a comparison (a different authority, a different version set) and to ignoring `published_at_ms`.

## F5 — a locked database no longer gives destructive advice

`RecordStore::kt_checkpoint_bytes` returns the record as bytes with the store's own error code, and the decode happens in `SealedCheckpoints::load`. A `SQLITE_BUSY` or io error now propagates as itself; only a record that does not decode or does not authenticate is `directory-state-invalid` with the unenroll advice. Mutation: map the read failure to `checkpoint_invalid` again → `a_store_that_will_not_answer_is_not_a_damaged_record` fails (it asserts the code is not `directory-state-invalid` and that the message does not contain "unenroll").

## F4 — documented, not changed

ADR 0017 §3 now says the policy pin is **per connection**, that `attach_checkpoints` rebuilds the connection (a lock/unlock cycle, an unenroll, a restart) and re-asks, and why persisting the document was rejected for this posture: it would turn a legitimate `max_validity_ms` change on a disposable log into a support call, while the property it would add is already covered for the only field that decides anything — the authority key, which is compiled into the build and is the durable pin. §10's line is corrected to match.

## F6 — documented, with the two cases the review did not name

Left as accepted, and §3 now says the "has relied on the directory" signals (a receipt, a merged epoch, a conversation) are plaintext records beside the sealed one, so a party who can rewrite the store can delete both; that a **resolve-only** device and one holding **only a pending contact request** carry none of those signals either, so their missing checkpoint is indistinguishable from a first run; and that the fix is a monotonic counter inside the sealed record — the seal is authenticated, so a counter in it cannot be rolled back without the device key — to land with persisted pins and alarms, which need the same record.

## Verification

Plugin: 126 lib tests + the full integration suite **including the two-process relay tests** (`--features relay-harness`, real `f2z-relay` binary), `cargo clippy --all-targets -D warnings`, `cargo fmt --check`. `rs`: `cargo fmt --check`, `cargo clippy --workspace` (libs; the previous head's `--all-targets` run was clean and this round's only `rs` change is a doc comment). e2e2z: `typecheck`, `typecheck:tests`, vitest **382**, the three node guards, Playwright **26**. `wallet/zuuli` `messaging-contract` (9), and `check-markdown-links`, `check-public-repo-boundary`, `check-hash-domain-labels`. No dependency changed in this pass, so `cargo-deny`'s verdict from the previous head stands.

skyl:
Adversarial security review for the owner (#1022): **SHIP** at `69fbec18`.

The earlier pass blocked on one finding: `recheck_policy` latched the directory unusable on ANY policy-fetch failure, so a 429 or a captive portal permanently disabled messaging for the process, storing a code the contract marks retryable, with no UI explanation. Fixed: only `AuthorityPolicyMismatch`, `Protocol(BadSignature)` and `Protocol(WrongLog)` latch; everything transient is reported and forgotten. The author's first attempt at proving this **survived a mutation**, so the logic was split into `policy_verdict`/`apply_verdict` and re-proved — that is the standard this tree should hold.

Also fixed in the same pass: the destructive callout now keys on a dedicated `directoryBlocked` field that the 5-second inbox poll cannot clobber; the policy pin compares order-insensitively; and a locked database is no longer reported as a damaged record telling the user to unenroll.

**Verified sound in review (attempts to break it failed):** vouching cannot be made vacuous (an empty authority set with `vouched` is rejected by `validate()`; no policy, unsigned, retired-key-signed and replayed-from-another-log all refuse); exactness defeats subset, superset and duplicates; no read path bypasses the latch and nothing answers before the policy check; `KtClient::open` never treats a decode error as absent and a retired generation needs both the derived id and that key's signature; the seal is ChaCha20-Poly1305 under a per-device key, so a transplanted record does not open; the build-time parse is real for release builds, with `rerun-if-changed` on all three inputs; both new error codes are non-retryable, absent from the wire mapping and complete in Rust and TypeScript; the moved reviewed digest hides no behavior change (re-run: 9/9).

**Accepted and documented:** an attacker with write access to the app's database file can clear the plaintext "has relied on the directory" signals and get trust-on-first-use. The durable fix is a monotonic counter inside the sealed record, to land with persisted pins.

PR #1028, opened 2026-09-17
feat(e2ee): publish an e2e2z device through ZUULI (#1022, ADR 0017 §4.1)

skyl:
Closes the last step of the enrollment round trip in #1022: an e2e2z device that
enrolls through ZUULI's App Link (#1019) now ends up **published in the
directory**, so another person can find it, and it reaches `mergedAtEpoch` from
its **own** verified lookup rather than from anything ZUULI said.

## What was missing, and what carries it

A `DirectoryEntry` publishes the `contact_addr` the **relay** issued to the
device that created the queue. Only that device knows it, so ZUULI could not
build a publishable entry from `IssueDeviceCredentialRequestV1` — which is why
version 1 answered with a credential and published nothing, and still does.

`IssueDeviceCredentialRequestV2` carries the endpoint. It is a **family code
(4)**, not a version field inside the payload: version 1's payload carries no
tag, so the only selector that refuses instead of best-guessing is the one
`payload` is already opaque behind. Code 2 keeps meaning exactly what it meant,
its wire vector is frozen (builds in testers' hands send it), and neither
version decodes as the other — version 2 is trailing data to version 1's
decoder, version 1 is a truncation to version 2's.

## The path, end to end

- **e2e2z opens its queue before it asks.** `prepare_device_with_endpoint` adds
  the bundled relay through the ordinary `add_relay` path, creates the contact
  queue from the *pending* device's queue seed, and `install_identity` commits
  it with the identity. A second preparation discards it; a build with no relay
  refuses (`relay-unreachable`) rather than enrolling a device nobody could
  reach.
- **ZUULI establishes whose handle it is** — from the entry the log already
  publishes under this wallet's identity key, or from Contract C's
  `HandleAssertion` verified against the bundled authority key — and refuses if
  that is not the handle the app asked for.
- **The confirmation says PUBLISHED**, names the free2z **account** the wallet's
  session belongs to, the handle **the log publishes** when it publishes one
  (never the request's — a first entry has no such name yet and says so), the
  device's key fingerprint, the relay host, and that every new conversation will
  reach this device *first*, ahead of the devices already published. "Adding a
  device to an account is a wiretap" is on the screen.
- **Only then** does it read the signing keys, issue the credential, sign the
  entry, precheck it locally against `f2z-authority`'s rules, submit it and
  verify the log's receipt. The credential is answered only if the log admitted
  the entry, so a refusal leaves e2e2z with nothing to install rather than with
  a credential nobody can find.
- **e2e2z reaches active on its own evidence**: `e2e2z_enrollment_status` asks
  the log while the entry is unmerged, and the screen moves from "Submitted, not
  yet active" to "Handle active" only when a verified lookup shows this device.

## Two decisions argued in ADR 0017 §4.1

1. **Nothing about this wallet leaves the process, and the seed is not read,
   until somebody says yes.** An earlier revision of this PR derived the
   account's public identity key and fetched the assertion *before* the dialog,
   so the dialog could name an authority-signed handle. Review established what
   that cost: a session somebody else wrote into the slot would hand this
   wallet's `identity_pk` to their account with no user interaction, and an
   unattested caller could make ZUULI decrypt custody by sending a request. So
   the handle moved behind the approval and the **account** moved onto the
   screen. The trade: a first entry's dialog names the account rather than a
   signed handle — publication is still refused unless the authority signs the
   handle that was asked for.
2. **A write-only native session slot** (`session.rs`). An intent arrives from
   the operating system with no argument to carry a Knox token on, and a
   renderer-supplied session *per intent* would let the WebView name the account
   an inbound request is answered under. Nothing reads the slot back over IPC —
   but the command is not capability-gated, so what bounds a *foreign* session
   is the flow above: it costs a declined dialog. free2z's answer about the
   session is cached for a minute against the session's generation, which is the
   rate limit on the single pre-approval request.

`INTENT_HANDLE_UNAVAILABLE` (13) is added for the one refusal a caller can act
on and the wallet can state with certainty (no free2z session in ZUULI, no bound
handle, or a different handle). Everything else stays `INTENT_UNAVAILABLE` under
`PROTOCOL.md` §6.2. The response still carries only the credential.

## Evidence

- `rs/crates/f2z-kt-client/tests/intent_publication.rs` — against a **real** log,
  witness and relay: a real V2 request admitted by the real `IntentGate`, the
  wallet's submission, the device seeing its own key published at the address
  the relay issued, and a stranger who knew only the handle claiming a key
  package there and authenticating it against the proved entry. Plus the log's
  own refusals.
- `tests/endpoint_before_credential.rs` — the queue exists before any identity
  does, the install commits exactly that queue, a second preparation discards
  the first, and a `ws://` default relay is judged like any other.
- ZUULI: the handle read from the authority's *signature* rather than its JSON,
  an assertion that does not verify naming no handle, a handle published under
  another identity refused before anything is asked, a log refusal publishing
  nothing, the confirmation copy, and the publish order.
- Playwright: the request carries the endpoint and is family 4; status 13 reads
  as "ZUULI couldn't confirm @alice is yours"; a build with no relay stops
  before asking; and the merge transition.
- **25 mutations, 25 failures, 0 survivors**, each restored from a saved copy of
  the working file (never `git checkout`, which would discard the change under
  test) and re-run green — including the two guards review found broken: the
  session mirror's ordering (the implementation this branch first shipped now
  fails its own suite) and the post-submission certainty mapping. Tables in `CONFORMANCE.md`.
- `authority-boundary`, `project-boundary` (92/92), `surface-capability-authority`,
  the crate manifest tests, clippy `-D warnings`, fmt and cargo-deny are green on
  both trees.

## What review changed

Adversarial review of the first revision found two blockers and a design gap,
all fixed here rather than documented: a dead ordering guard in the session
mirror whose test asserted the bug (F1), `INTENT_HANDLE_UNAVAILABLE` reachable
*after* the log accepted an entry (F2, now decided by phase — everything from
the submission onward is `INTENT_UNAVAILABLE`), and the pre-dialog disclosure
above (F3). The confirmation also now says the new device answers **first**,
ahead of the existing ones, and shows its key fingerprint (F5). Two residuals
are recorded as accepted in ADR 0017 §4.1: the contact endpoint is the caller's
to choose, bound only by the hostname the user reads (relay pinning is the
follow-up), and the pre-dialog lookup makes ZUULI a handle-resolution oracle for
a registered caller.

## What this does not prove

No shipping build can run any of it: `internal-directory.conf` still holds
placeholders, so there is no relay to open a queue at and no log to publish to.
Nothing here has run against a deployed log or on a device. ZUULI's own device
is still unpublished because nothing in its UI calls `f2zmsg_enroll` — an e2e2z
device does not need it.

Docs: ADR 0017 §4.1 → implemented (with the session slot and the ordering
argued), `PROTOCOL.md` §3.3/§3.6/§6.2, `AUTHORITY.md`, `CONFORMANCE.md`,
`docs/status.md`, `docs/architecture.md`, `wallet/e2e2z/README.md`.


skyl:
Fixed F1, F2, F3 and F5; F4's claim corrected in the PR body; F6/F10 recorded as accepted with rationale; F8's rows added. Details, and the one question answered, below.

## F1 — the ordering guard was dead code, and the test blessed it

Correct on both counts. `applied` was written and never read, and the native write happened *inside* `invoke`, so the check could not undo anything. The test asserting `["null","slow"]` asserted the bug.

The publisher is now a **single-slot queue**: one `invoke` in flight at a time, and while one is in flight only the newest pending value is kept, so a value overtaken before dispatch is never sent and the wallet converges on the last value the renderer asked for. The tests assert **the slot's final value**:

| Mutation | Test that failed |
|---|---|
| the queue replaced by the generation counter this branch first shipped | `leaves the wallet holding the last value asked for, not the last to arrive` |
| an overtaken value sent anyway (`pending = pending ?? {token}`) | `drops a value that was overtaken before it was ever sent` |

Both `FAILS`, both restored green.

## F2 — certainty now comes from the phase

`refusal_after_submission(_code) -> Unavailable` ignores its argument, the submission moved out of `publish_device` into the handler (so the two sides of the boundary are two calls), and a source-asserted test keeps the certain mapping on its own side.

| Mutation | Test that failed |
|---|---|
| `refusal_after_submission` delegating to `publication_refusal` | `nothing_after_the_submission_claims_that_nothing_happened` |
| the submission's refusal mapped with `publication_refusal` | `the_certain_refusal_mapping_is_never_applied_after_the_submission` |

## F3 — nothing is disclosed before the approval (option (a), plus (b)'s account name)

I did not take (b) alone, because naming the account only helps if the user reads it. The order changed instead, so a poisoned slot is **inert as well as visible**:

- **Pre-dialog**, two answers that disclose nothing about this wallet: free2z names the account the session belongs to (`GET /api/auth/user/`, the app's own call, derived from the bundled authority's origin — carries the session's own token and nothing else), and the log answers the handle lookup unauthenticated.
- **The dialog names that account**, the handle the *log* publishes (or says a first entry has none yet, rather than rendering the app's string), the device fingerprint, the relay host, and the first-contact ordering.
- **After the approval**: the seed is read, the assertion is fetched (this is the request carrying `identity_pk`), and publication is refused unless the authority signs the handle that was asked for.

So a declined or ignored request leaks nothing about the victim — no identity key, no custody read, no assertion issued — and binding the victim's key to a foreign account now requires a human to approve a dialog naming that account. Ordering mutations (`the assertion fetched before the confirmation`, `the seed read before the confirmation`) both fail `the_publish_order_is_plan_then_prompt_then_seed_then_assertion_then_submit`.

`docs/architecture.md` §3.3 now says what the intent path *does* add to the invoke surface, that it is not capability-gated, that ZUULI's CSP (`frame-src 'none'`) means the #367 deputy needs an XSS in ZUULI's own origin rather than an embed, and what bounds the damage.

## F4 — the PR body was wrong, and the thing it was wrong about is gone

It said "the signing keys are read after the approval"; `account_identity` read the seed and derived the full `AccountKeys` before the dialog. With F3's reordering there is no pre-dialog custody read at all, so an unattested caller can no longer make ZUULI decrypt custody. What it *can* still do is cause one authenticated `GET` — and that is rate-limited: free2z's answer is cached for 60 s against the session's generation, so a caller looping intents gets one request a minute per session, not one per intent.

## F5 — the copy says what approving actually does

> From now on, every NEW conversation will reach this device first — ahead of your N existing device(s) — at `<host>`.

plus `The device being added: abababab…ababababab` (head-and-tail `device_pk`).

## F6 / F10 — accepted, with the reasoning written down (ADR 0017 §4.1)

- **The endpoint is attacker-chosen.** A caller past the registry and the confirmation can point first contact at a relay it controls: that relay learns *that* someone is contacting the handle and can withhold delivery; it cannot read anything, because the `Welcome` is encrypted to a key package authenticated against the proved entry (ADR 0015). Follow-up recorded: refuse an endpoint whose `relay_id` this wallet has not itself authenticated a connection to — with one relay in the deployment the cheap form is "the published relay must be the bundled one", which I did not add blind before the deployment exists.
- **The pre-dialog lookup is a handle-resolution oracle** for a registered caller. The log is a public directory and the same lookup is available to anyone who can reach it; what it adds is that *this wallet* performed it. Accepted for the internal deployment.

## Your question: does the authority record the binding at issuance?

**I cannot answer it from this repository, and the design no longer depends on it.** What the artifacts here do establish: `KT.md` and Contract C make the *issuer* responsible for never reusing an `(authority_id, nonce)` pair and for maintaining `account_epoch` as a durable counter, so issuance necessarily writes some server-side state. Whether the `identity_pk` is retained with it is workstream 7's decision and is not written down anywhere in `zuu`. With the reordering, nothing is issued unless a human approved, so a declined flow leaves no server-side state either way — but it should be written into Contract C, and if issuance *does* bind, the endpoint should say so. Recorded as an open question in ADR 0017 §4.1.

## Verification re-run on this head

ZUULI 94 lib tests; plugin 129 lib + the endpoint suite; `f2z-intent` 69; `f2z-msg-identity` 40; **the real log/witness/relay round trip (`intent_publication`, 2)**; ZUULI 1007 vitest + both typechecks; e2e2z 385 vitest + **29 Playwright** (the merge-transition spec now drives the re-read instead of racing one focus event); fmt and clippy `-D warnings` clean on the wallet tree. Mutation total is now **25, 25 failures, 0 survivors**.

skyl:
All four addressed, plus the optional wording. Pushed as `3c7a4aff`.

**1 — the wedge.** You are right that the trap is real and that my own test file supplies the callback that triggers it. `draining` is now cleared in a `finally`, and the reporter is called inside its own `try` — a reporting failure must not abandon the value pending *behind* it, which the `finally` alone would not have covered (the loop would exit and the pending value would sit unsent). New test drives a refusing wallet **and** a throwing reporter and asserts the next publication still reaches `invoke`. Both halves mutation-proved:

| Mutation | Test that failed |
|---|---|
| the `finally` that re-arms the queue | `keeps publishing after a reporter throws` |
| the guard around a throwing reporter | same |

**2 — `session.rs`'s module note.** Rewritten. It now says what widens (the ability to state the wallet's session is *somebody else's*), that an app-crate command is not capability-gated and ZUULI's CSP makes that an XSS rather than an embed, and what bounds it: nothing about this wallet is disclosed before the confirmation, the confirmation names the **account**, the authority's handle is established after the approval — so a poisoned slot costs a declined dialog. That was the one place still arguing the superseded design.

**3 — the unrunnable rows.** Both updated to `the_publish_order_is_plan_then_prompt_then_seed_then_assertion_then_submit`; `grep` for the old name across `docs/` and `wallet/` now returns nothing. I also added the two new renderer rows and corrected the totals (27 mutations).

**4 — the PR body bullet.** Replaced: it now says the confirmation names the free2z account, the handle only when the log publishes one (a first entry says it has none yet), the device fingerprint, the relay host, and the first-contact ordering.

**Optional — status 13's wording.** Taken, since it was two sentences. `PROTOCOL.md` §6.2 and the ADR's refusal table now say "every site that produces it is on the pre-submission side of the phase boundary", and name the one site that runs after a credential has been minted in memory, noting that issuance is pure and persists nothing. That is the honest form of the claim rather than the convenient one.

Verified on this head: 66 `src/lib/auth` vitest, `typecheck:tests`, `cargo fmt` clean across the wallet tree, and the ZUULI crate rebuilt from scratch. No behaviour changed outside the queue fix; the rest is prose.

skyl:
Adversarial review for the owner (#1022), two passes plus a delta pass: **SHIP** at `3c7a4aff`.

**The first pass blocked on three findings, all now fixed and re-verified:**
1. **A dead safety check.** The guard meant to stop a stale sign-in overtaking a sign-out never ran (its variable was written and never read, and the write happened inside `invoke`), and the test asserted the broken order. Now a single-slot queue, with tests asserting the slot's FINAL VALUE. The reviewer proved it by execution: 2000 randomized interleavings always end holding the last value asked for, while the reverted version fails 5 of 6 named scenarios.
2. **A false certainty.** `INTENT_HANDLE_UNAVAILABLE` promised "nothing was submitted" but was reachable after the log accepted the entry (a receipt with a malformed handle). Certainty now comes from the phase: every path from the submit call onward maps to `Unavailable`, verified by walking the error types, not the happy path.
3. **Pre-consent disclosure under a poisoned session slot.** A hostile frame writing someone else's token could, before any dialog, get the victim's identity key POSTed under the attacker's account. Now nothing wallet-identifying leaves the device before approval: the pre-dialog work is an unauthenticated log lookup plus `GET /api/auth/user/` carrying only the session's own token. The seed is read and the identity key sent only after approval. No TOCTOU (the same token is reused after the dialog) and the 60s cache is generation-bound and cleared on every write, so it cannot serve a stale account. A poisoned slot now costs a declined dialog.

**Also fixed:** the dialog states the first-contact ordering ("ahead of your N existing devices") and shows the device fingerprint — approving used to silently move first contact to the new device; and a wedge where a throwing reporter would stop the wallet from ever seeing the session again (the author found the `finally` alone was insufficient and guarded the reporter too).

**Verified sound:** handle provenance has no TOCTOU and the request's handle never reaches the dialog; V1 and V2 never decode as each other; the credential is answered only after a verified receipt; dialog copy is injection-resistant. 27 mutations, 0 survivors, and the rows that named a renamed test are corrected — a mutation table naming a missing test is worse than no row.

**Accepted, with rationale in ADR 0017 §4.1:** the contact endpoint is attacker-chosen and bound only to a hostname the user reads (relay-pinning follow-up recorded), and the pre-dialog log lookup makes ZUULI a handle-resolution oracle.

PR #1031, opened 2026-09-18
docs(zuuli): re-derive STATUS at the encrypted-chat candidate (#1022)

skyl:
Step 1 of the release runbook for workstream 10 of #1022. **Merge this first**; the ZUULI build-24 PR (#1034) records this PR's squash SHA in the `STATUS.md` marker and cannot merge before it.

Rebased onto `42ef4be2` (#1033, the filled `internal-directory.conf`), which the marker now names.

## Why this is required

`status-freshness.mjs` treats everything under `wallet/zuuli/**` and `wallet/plugins/**` as release-impacting, and **63 such paths changed** since the recorded marker `737af8e8` (2026-09-11). A build-24 release would be rejected by `Pin immutable source` exactly as build 21 was. Because the re-derivation is substantive, `STATUS_DOCUMENT_SHA256` is rotated to the new reviewed surface.

## What it corrects

- **Build 23 is on TestFlight.** [Its iOS job](https://github.com/free2z/zuu/actions/runs/34549813662/job/103112775184) recorded `uploaded`, `processed` and `availableToInternalTesters` for build `c74a1089-9e47-4f6a-8deb-92b4f49bdd51`, `VALID`/`IN_BETA_TESTING`, with the exact internal-group relationship read back. The old claim that every signed ZUULI artifact predates the vault split is false. [Its Android job](https://github.com/free2z/zuu/actions/runs/34549813662/job/103115292958) failed at `Inspect attested Android artifact without credentials`, so **Play internal still holds build 20** and the two stores sit at different builds.
- **The shipping directory posture changed under the old text's feet.** #1033 fills the bundled config, so these are the first ZUULI and e2e2z binaries configured to reach a production service at all — not `NoDirectory` any more. The document now says that, and says what it does not mean.
- **A blocker the release cannot fix, recorded plainly:** the deployed log's witness is **not cosigning**. #1033 observed zero cosignatures on 2026-09-17; I re-checked independently for this audit — `GET https://kt.free2z.cash/kt/v1/sth` returns a 314-byte bundle carrying the bundled `log_id` and VRF key whose trailing cosignature-vector length is **zero**. The bundled `threshold = 1` is unmet, so `resolve_handle` and `start_conversation` refuse with `witness-threshold-unmet`. **Two internal testers cannot exchange a message until `f2z-witness` posts cosignatures.** That is a deployment fix; cutting the builds is still right, because the config already points at the log and cosignatures appear on new tree heads without a new binary.
- **The ungated app-crate surface grew from three commands to four.** `free2z_session_sync` joins the enrollment trio, `f2zmsg_enroll` now carries a Knox token and attempts directory publication, and ZUULI is additionally an intent authority over a verified App Link (`bridge.rs`, non-IPC, dialog-gated, `CallerTrust::Claimed`). No capability file, CSP directive, entitlement or manifest declaration changed.
- **Fresh production probes**, re-run 2026-09-18 at 01:36:05 UTC: `/api/e2ee/authority/` answers 200 with prod's public key `332e237e…934f` and log id `05b5dc5a…c7e9`; `assetlinks.json` returns **200**, byte-identical to the reviewed declaration (`286e6580…64d5`), where the previous audit recorded 503.
- **A second, newer store-media problem**, with its fix already in flight: `wallet/e2e2z`'s `unenrolled-native-v1` capture can no longer be reproduced, because the app now makes startup calls the harness answers with `Unexpected native command`. It breaks a *re-capture*, not the gate — `validateSurfaceCaptureRecord` checks the committed record against the current **contract** digest, not current source — and no release path consults these records. #1035 fixes it with a real two-pass capture and is deliberately held until after this release.

## What it refuses to say

Nothing in the encrypted-chat flow has run on a phone. No signed build containing it has been installed. Every new claim is source, fixture, in-process-loopback or CI evidence, and the document names which, in each place: the two-instance relay test substitutes the directory with a shared file, and `directory_publish.rs`'s tests talk to an in-process HTTP server.

## Verification

```
node --test wallet/zuuli/scripts/status-freshness.node-test.mjs   # 118 pass
verifyReleaseEvidencePolicy(releasing.md, STATUS.md)              # OK
node scripts/check-github-actions-pins.mjs                        # OK (incl. crypto-probe digests)
node wallet/zuuli/scripts/apple-credential-boundary.mjs           # OK (credential-job seals)
npm --prefix wallet/zuuli run test:store-listing                  # 85 pass
```

The sealed `## Release-path execution disposition` prelude and its six-row evidence inventory are untouched, and `docs/releasing.md` is untouched, so `RELEASING_DOCUMENT_SHA256` is unchanged.

Ref #1022.


skyl:
Reviewed for the owner (#1022): **SHIP**.

`status-freshness` was the one stale release gate — **63 release-impacting paths** changed since the old marker, which is exactly what killed build 21. Re-derived at the encrypted-chat candidate.

It must stay honest about the two things that are true right now: a great deal is implemented and CI-verified, and **nothing has been exercised on a real device against the deployed servers**. It also records that the deployed log carries zero cosignatures, so at threshold 1 every handle lookup refuses — the client failing closed as designed, and a server-side gap rather than a client one.

Merged first so the ZUULI bump can point its marker at this commit's squash SHA.

PR #1033, opened 2026-09-18
feat(e2ee): point shipping builds at the deployed internal directory (#1022)

skyl:
Workstream 9's services are deployed, so `internal-directory.conf` is filled in.
Both ZUULI and e2e2z link `tauri-plugin-f2zmsg`, so both apps now construct a
real key-transparency client and a default relay instead of keeping
`NoDirectory`. Part of #1022.

| Key | Value |
|---|---|
| `log_url` | `https://kt.free2z.cash` |
| `relay_url` | `wss://relay.free2z.cash/relay/v1` |
| `handle_assertion_url` | `https://free2z.cash/api/kt/handle-assertion/` |
| `threshold` / witnesses | 1 / one free2z-operated witness, **never** independent |
| `reset_cooldown_seconds` | 604800 (the floor, and the log's own configured value) |
| `retired_log_public_key` | none — nothing has been wiped yet |

## Each value was checked against the live service, not taken on trust

Run through the **shipped path** — `KtDirectory::connect` with the
`DirectoryConfig` this file produces — against the real deployment:

- `GET /kt/v1/sth` verifies under `log_public_key` and carries `log_id`; the
  client bootstrapped at **epoch 40, tree_size 40**.
- `log_id` equals `H("free2z/kt/v1/log-id", log_public_key)`, derived
  independently of the file as well as by `build.rs`.
- The published descriptor (`/.well-known/free2z-kt/v1/log`) carries the same
  genesis key, the same `vrf_public_key`, the same `reset_authority_pk`, and
  `reset_cooldown_seconds = 604800`. Its operator name is
  `free2z DISPOSABLE-INTERNAL prod log - will be wiped before public launch`.
- The §4.6 authority policy is **vouched** and lists **exactly one** authority —
  the bundled `handle_authority_pk` — so #1027's `require_authority_policy`
  passes (`vouching = Claimed`, `max_validity_ms = 900000`).
- `GET https://free2z.cash/api/e2ee/authority/` publishes that same
  `public_key`, `authority_id` `a28e39a1…f144eb2`, and the same `log_id`.
- The relay answers a `free2z-relay.v1` upgrade with `101 Switching Protocols`
  (over HTTP/1.1; an h2 request cannot carry that upgrade).

## ⚠️ The deployed log is not being cosigned — first contact does not work yet

Every tree head the log serves carries **zero cosignatures**, at epochs 1, 5,
20, 35, 38, 39 and 40, against a configured *t* = 1. So a client bootstraps,
verifies the head, accepts the vouched policy — and then refuses:

```
resolve(skylar)      -> Err(WitnessThresholdUnmet)
resolve(skyl)        -> Err(WitnessThresholdUnmet)
resolve(nobodyatall) -> Err(WitnessThresholdUnmet)
```

That is `KT.md` §8.3 working as designed (*"an unverified key here is the
MITM"*), and it means **nobody can look up a handle against this deployment**,
so e2e2z cannot reach "Handle active", which comes only from its own verified
lookup. `/kt/v1/cosign` is enabled and recognises witnesses — it answers
`ERR_MALFORMED`, not `ERR_NOT_A_WITNESS` — so `f2z-witness` is simply not
posting cosignatures. **This is a server-side fix, not a client change.** Until
it lands, `witness_pk` is also the one bundled value with no live signature to
check it against.

Merging this is still right: the client fails closed correctly, and the
configuration is verified correct for everything the log does publish.

## The fail-closed proof moves rather than disappearing

The shipped file can no longer demonstrate it, so
`src/internal_directory/placeholder.conf` — a **test-only** fixture, held to the
shipped file's key set by `the_placeholder_fixture_declares_the_shipped_files_keys`
— carries it via `a_placeholder_file_fails_closed`. New alongside it:
`the_shipped_file_configures_the_internal_deployment` asserts every real value
and re-derives `log_id`, and
`the_shipped_file_gets_the_internal_directory_and_its_relay` asserts the engine
gets the relay, one witness, zero independent, the bootstrap disclaimer, and a
threshold that is conservatively unmet before first use — so a device with no
network fails closed rather than blocking a launch.

## Also here

ADR 0017 §4.1's open question — *does the handle authority record the
handle → `identity_pk` binding at issuance?* — is replaced with the answer from
the backend side: the assertion is never recorded (`ktauth` has no models), the
only possible write is `ensure_username_handle` reserving an account **its own**
handle, and that is a one-time, rate-limited (10/hr/account), per-account
transition that cannot be farmed. Phrased as the deployed authority's behaviour,
not a protocol guarantee. `docs/status.md` and `docs/architecture.md` are
re-derived: the e2e2z transport dispatches, the directory is configured, and
what remains unproven — nothing has run on a device, and no automated check
reaches the deployed services.

## Verification

- `cargo test --locked --all-targets --features relay-harness` and the default
  feature set: green, including both two-process relay tests against the real
  `f2z-relay` binary.
- ZUULI `npm run verify` (typecheck ×2, vitest, 23 guard scripts, Playwright)
  and e2e2z `npm run verify`: green.
- `check-rust-fmt.sh`, `check-rust-clippy.sh` (`-D warnings`),
  `check-rust-deny.sh`, `check-markdown-links.mjs`,
  `check-hash-domain-labels.mjs`, `check-kt-sth-repeat-agreement.mjs`,
  `check-tauri-plugin-permissions.mjs`, `check-public-repo-boundary.sh`: green.
- Two transient failures were chased to ground and are **not** this change:
  two e2e2z Playwright specs timed out in page setup under concurrent load and
  pass on rerun, and `zuuli/tests/about.pw.ts` fails on any dirty worktree
  because `build-identity.mjs` reports `sourceCommit: null` — it passes with the
  tree clean.


skyl:
Reviewed for the owner (#1022): **SHIP**.

Every bundled value was verified against the live deployment by running the shipped code path (`KtDirectory::connect` with the config this file produces), not by trusting the brief:
- log bootstrapped to epoch 40; STH signature verifies under the bundled log key; `log_id` **derived independently** via `f2z_kt_core::labels::log_id` and byte-identical.
- VRF key pinned at connect; reset-authority key from the log's own descriptor; `reset_cooldown_seconds` **read from the deployed descriptor** (604800) rather than assumed from the default.
- handle authority: the log's §4.6 policy lists exactly the bundled key, and `GET /api/e2ee/authority/` publishes the same key, authority id and log id.
- relay: 101 with `free2z-relay.v1`.
- **Vouching (the #1027 requirement) passes** on the deployed log: `vouching = Claimed`, one authority, the expected key — so `connect` does not refuse.

**The finding that matters more than the config:** the deployed log carries **zero cosignatures** at threshold 1 (checked across epochs 1→40; independently confirmed: no `f2z-witness` Deployment exists in either namespace, and the head is a bare 314 bytes). The client therefore fails closed on every lookup with `WitnessThresholdUnmet`, so e2e2z can never reach "Handle active". That is `KT.md` §8.3 working as designed, and it is a server-side gap, not a client one — the witness activation is now the last blocker for the epic. Recorded in `status.md` §2.4, ADR 0017 §8, the commit message and the PR body rather than left to be discovered on a device.

Merging because no value disagrees with a live service, the client fails closed correctly, and the release needs this commit.

PR #1034, opened 2026-09-18
chore(zuuli): prepare internal build 0.1.0+24 (#1022)

skyl:
Step 3 of the release runbook for workstream 10 of #1022, and the commit that makes the real store transaction. **Merge this last, and this is the one that ships.**

Rebased onto `16fd11de` (#1031), and the `STATUS.md` marker now names that exact squash SHA. The runbook's freshness rule is satisfied end to end — proven, not assumed:

```
$ node wallet/zuuli/scripts/status-freshness.mjs --source-sha=ccecc54cb4febd68a0bcdfaf638ff49324981326
{"sourceSha":"ccecc54c…","parentSha":"16fd11de…","auditSha":"16fd11de…","auditDate":"2026-09-18"}
```

That exit-0 is all four conditions the protected `Pin immutable source` step re-checks: the recorded marker is on the release parent's first-parent history, nothing release-impacting but `STATUS.md` changed between them, this commit touches only `release.json` plus `release:bump`-owned files plus `STATUS.md`, and every generated file reproduces mechanically from the parent.

## The change

1. `npm run release:bump -- --version=0.1.0 --build=24` from a clean worktree — `release.json` plus the four generated representations the script owns, written atomically, reviewed in the diff:
   ```
   $ npm run release:verify
   {"applicationId":"cash.free2z.zuuli","channel":"internal","version":"0.1.0","build":24,"identity":"0.1.0+24","tag":"zuuli-v0.1.0+24"}
   ```
2. One line of `STATUS.md`: the re-derivation marker, advanced to #1031's merge commit.

## What merging does

`ZUULI / protected release` fires on the push, pins the exact pushed SHA, and resolves `target=mobile`, **`dry_run=false`**. It builds, signs and uploads to **TestFlight and Play internal**. Android `versionCode` 24, iOS `CFBundleVersion` 24, marketing `0.1.0`, application `cash.free2z.zuuli`. Store build numbers are append-only: if an upload is ambiguous or a different package is needed, the answer is build 25, never re-cutting 24.

**Build 24 should be the first ZUULI to reach Play internal since build 20.** Builds 21, 22 and 23 all failed before a Play upload; build 23's Android job failed at `Inspect attested Android artifact without credentials`, which is #1016, fixed in source by #1023 and **not yet executed by any protected run**. This release is the first real-input test of that fix — read an Android failure here as evidence about the fix rather than as a new fault, and look at the decoded `base <name>, flags 0x…` diagnostic the new checker prints. iOS is the lower-risk lane: build 23 already signed, exported and reached TestFlight on the current profile pin.

## After it runs

Require the iOS evidence to show all three booleans — `uploaded`, `processed`, `availableToInternalTesters` — rather than treating a green workflow as delivery, and confirm the Play internal release is available to the tester list. Neither is device evidence.

For whoever runs the two-tester test afterwards: **the deployed log's witness is not cosigning.** The current signed tree head carries zero cosignatures (independently verified for #1031), so with the bundled `threshold = 1` every handle lookup refuses with `witness-threshold-unmet` and first contact cannot complete until `f2z-witness` starts posting. That is a server-side fix and needs no new build — these builds are still the right thing to cut.

Ref #1022.


skyl:
Reviewed for the owner (#1022): **SHIP** — and this one is a real release.

**What merging does:** `zuuli-release.yml` fires on the push, `target=mobile`, `dry_run=false`. Real upload to **TestFlight and Play internal**, version `0.1.0`, Android `versionCode` 24, iOS `CFBundleVersion` 24.

**Why now:** every server this build talks to is live and verified in production — relay (101 with the expected subprotocol), key-transparency log (signed head), handle authority (returning prod's key), and as of minutes ago the **witness cosigning**, with `check-kt-witnessed.py prod` verifying the signature under the key the app bundles. Until that last one, the client correctly refused every handle lookup, so a build shipped earlier could not have started a conversation.

**Build 24 is also the first real-input test of #1023's Android fix.** Builds 21-23 never reached Play: 21 died on a stale STATUS marker, and 22/23 died in the compiled-manifest protection-level check. Build 23's iOS leg *did* reach TestFlight.

**Release gates checked:** `status-freshness` re-derived at the candidate (63 release-impacting paths had changed, which is what killed 21) and its marker re-pointed at #1031's squash commit; `check-github-actions-pins` including the crypto-probe digests; the Android and three Apple credential-job seals; `Release identity`.

The earlier `gate` failure on this branch was `result: cancelled` from a superseded run, not a real failure; reruns confirmed green before merge.

**After merge I will verify delivery rather than trust a green workflow** — the runbook's three booleans (`uploaded`, `processed`, `availableToInternalTesters`) — and watch whether the Android leg finally clears.

PR #1032, opened 2026-09-18
chore(release): prepare free2z 0.1.0+4 and e2e2z 0.1.0+5 (#1022)

skyl:
Workstream 10 of #1022. The sibling halves of the encrypted-chat cut: free2z's paid **Start encrypted chat** and Contract B link (#1024), and e2e2z's chat-link route, "Enroll with ZUULI" transport and its own device publication (#1025, #1026, #1027, #1028).

**Draft until the `internal-directory.conf` commit lands.** e2e2z's backend links `tauri-plugin-f2zmsg`, so the bundled directory configuration is compiled into the e2e2z binary. Shipping build 5 before that file is filled would ship a build that is `Unconfigured` by construction and can never reach the relay. I will rebase onto that commit and mark this ready.

## The change

Mechanical identity files only — `release.json` plus the four generated Apple/Android representations per app, in exactly the shape #1011 used. Nothing else. Verified:

```
node wallet/e2e2z/scripts/release-identity.mjs
{"applicationId":"cash.free2z.e2e2z","version":"0.1.0","build":5,"identity":"0.1.0+5","tag":"e2e2z-v0.1.0+5"}
node wallet/free2z/scripts/release-identity.mjs
{"applicationId":"cash.free2z.free2z","version":"0.1.0","build":4,"identity":"0.1.0+4","tag":"free2z-v0.1.0+4"}
```

Android `versionCode` and iOS `CFBundleVersion` both become 4 (free2z) and 5 (e2e2z); marketing version stays `0.1.0`. Neither app is ZUULI-release-impacting, so this does not disturb `STATUS.md` freshness and can merge before or after the ZUULI build-24 PR.

## What merging actually does — neither app is delivered by the merge

- **`free2z / protected release`** fires and resolves `target=ios`, **`dry_run=true`**. The workflow says why in full: merging a bump is a code-review event, not a decision to publish, and the upload carries free2z's `ITSAppUsesNonExemptEncryption` declaration, which is under owner/counsel review in #961/#967. It builds, signs and runs `altool --validate-app`, which creates no App Store Connect build record, and stops. **Delivering build 4 is an explicit `workflow_dispatch` with `target=ios`, `dry_run=false`.**
[...]

PR #1035, opened 2026-09-18
fix(store): answer e2e2z's real unenrolled startup contract and re-capture (#1022)

skyl:
## ⚠️ MUST NOT MERGE BEFORE #1034 (ZUULI build 24)

**This PR is release-impacting: it touches `wallet/zuuli/scripts/**`, which
`isReleaseImpactingPath` matches via the `wallet/zuuli/` prefix.** Landing it
first puts a release-impacting change between #1031's `STATUS.md` marker and the
build-24 bump's parent, and `Pin immutable source` will reject build 24. It is a
deliberate follow-up to the release train, not part of it. **Merge #1034 first.**

## The problem

`wallet/e2e2z`'s committed store-surface capture could no longer be reproduced.
The capture harness (`wallet/zuuli/scripts/surface-store-capture.mjs`) installs a
fake `window.__TAURI_INTERNALS__.invoke` that answers a fixed list of native
commands and refuses everything else. Since #1025 / #1026 / #1028 the e2e2z app
issues further commands at startup that the fake did not answer, so the capture
hung on the screen it was waiting for.

### Verbatim failure, before this change

Running the documented command on the then-current `main` (`ee06a8d5`):

```
$ cd wallet && node zuuli/scripts/surface-store-capture.mjs e2e2z --verify-reproducible
AssertionError [ERR_ASSERTION]: render inputs differ from source commit
+ actual - expected

+ 'wallet/e2e2z/release.json\n' +
+   'wallet/e2e2z/src-tauri/src/bridge.rs\n' +
+   'wallet/e2e2z/src-tauri/src/device.rs\n' +
+   'wallet/e2e2z/src-tauri/src/lib.rs\n' +
+   'wallet/e2e2z/src/features/messages/Enrollment.tsx\n' +
...
```

That is `assertSourceCommit` refusing the *stale* `sourceSha`
(`a057f7f42947f9600a38a70cf97c245ab610f940`) before the browser ever starts. With
`sourceSha` pointed at the commit actually being rendered, the run reaches the
browser and produces the real failure:

```
locator.waitFor: Timeout 30000ms exceeded.
Call log:
  - waiting for getByText('Enrollment happens in the wallet app', { exact: true }) to be visible

    at preparePage (/work/zuuli/scripts/surface-store-capture.mjs:121:83)
    at async capturePass (/work/zuuli/scripts/surface-store-capture.mjs:164:46)
    at async main (/work/zuuli/scripts/surface-store-capture.mjs:239:19)
Error: docker exited 1
```

## What the complete native-command set actually is

Found empirically, not guessed: the built `wallet/e2e2z/dist` was loaded in
Chromium behind a *recording* fake that logged every `invoke` and answered the
known commands, for both capture actions (`enrollment-unavailable` and
`local-diagnostics`). The complete set the app issues is:

| command | already answered? |
| --- | --- |
| `plugin:event\|listen` (`deep-link://new-url`, `f2zmsg://engine-state`) | yes |
| `plugin:event\|unlisten` | yes |
| `plugin:f2zmsg\|get_engine_status` | yes |
| `plugin:f2zmsg\|get_device_info` | yes |
| `plugin:deep-link\|get_current` | **no — added** |
| `e2e2z_enrollment_status` | **no — added** |

Nothing else is invoked, in either action, and the probe reported no unanswered
command once these two were answered.

### Where each new answer comes from in the Rust source

- **`e2e2z_enrollment_status`** → `wallet/e2e2z/src-tauri/src/device.rs`'s
  `e2e2z_enrollment_status`, which delegates to `Engine::enrollment_status`.
  `wallet/plugins/tauri-plugin-f2zmsg/src/engine.rs` takes the `identity: None`
  branch on a device with nothing installed: `enrolled: false`, `handle: None`,
  `eligibility: handle::not_signed_in()` (`src/handle.rs` →
  `{ eligible: false, candidate: None, reason: NotSignedIn }`), and
  `directory_entry_version` / `submitted_at` / `merged_at_epoch` / `blocked` all
  `None`. Serialized `camelCase` with `kebab-case` enum variants per
  `src/models.rs`, that is exactly:
  `{"enrolled":false,"handle":null,"eligibility":{"eligible":false,"candidate":null,"reason":"not-signed-in"},"directoryEntryVersion":null,"submittedAt":null,"mergedAtEpoch":null,"blocked":null}`.
  This is byte-for-byte the `NOT_ENROLLED_STATUS` that
  `wallet/e2e2z/tests/enrollment-gap.pw.ts` already pins for the same contract.
- **`plugin:deep-link|get_current`** → `tauri-plugin-deep-link` 2.4.9
  (`src/commands.rs`: `-> Result<Option<Vec<url::Url>>>`; the desktop
  `get_current` returns the plugin's `current`, which is `None` unless the
  process was launched from a link). The plugin is registered for every platform
  in `wallet/e2e2z/src-tauri/src/lib.rs` with `deep-link:default` in
  `capabilities/default.json`, and `main.tsx` calls `installChatLinkListener()`
  before the first render, so a packaged e2e2z really does issue this call and
  really does get `null`. Answered as `null`.

Additionally, `plugin:f2zmsg|get_engine_status` now also carries
`directoryBlocked: null` — `EngineStatus` grew that field
(`tauri-plugin-f2zmsg/src/models.rs`), the fake had not, and
`enrollment-gap.pw.ts` already includes it.

No invented enrolled state, no handle, no credential, no conversation, no
transport. The default branch is still
``throw new Error(`Unexpected native command: ${cmd}`)`` — there is no permissive
catch-all. `NATIVE_CALLS` is updated in step and the capture's own assertion now
requires `e2e2z_enrollment_status` to have been observed, alongside
`get_device_info` and `get_engine_status`.

## Rebased onto #1033 — and re-checked against a *configured* directory

Base is now **`42ef4be2`** ("point shipping builds at the deployed internal
directory", #1033), which moved `plugins/tauri-plugin-f2zmsg/src` — a member of
`renderInputs('e2e2z')` — so the earlier record documented a tree that no longer
exists on `main`. Both apps were re-captured from `42ef4be2`.

#1033 makes a shipping build **`Bundled::Configured`** rather than
`Unconfigured`, so the question is whether that changes the unenrolled startup
path. **It does not, and no native call was added.** Three independent checks:

1. #1033 changed only Rust, `internal-directory.conf` and docs — **no frontend
   TypeScript**. Rebuilding `wallet/e2e2z` on the new base produces byte-identical
   bundle hashes (`index-D5fQyM2S.js`, `index-C_bJ84m0.js`, …), and the
   re-captured e2e2z PNGs are byte-identical to the pre-rebase ones; only digests
   and `sourceSha` moved.
2. The recording probe was re-run against the rebuilt `dist`: the same six
   commands, in the same order, no unanswered command, no page error.
3. The two answers stay faithful. `Engine::enrollment_status`'s
   `directory_configured()` check lives only in the `Some(identity)` branch, so an
   unenrolled device never reaches it and `blocked` is unconditionally `None`.
   And `directoryBlocked` stays `null`: `KtDirectory::directory_blocked` reports
   only a live connection's block or a remembered connect refusal, and a
   never-started engine has neither (`state.rs` says so explicitly — "an engine
   that never opened its store has no directory state to be blocked by").

One note for the reviewer: the fake's pre-existing `relaysConfigured: 1` is now
*more* accurate than it was, since a configured build does add a relay. It is
unchanged by this PR either way.

## Why free2z is re-captured too

`contractInputs()` in `surface-store-capture-contract.mjs` includes
`zuuli/scripts/surface-store-capture.mjs`, so editing the harness changes the
`contractDigest` for **both** apps. `npm --prefix wallet/zuuli run
test:store-listing` fails for free2z as well until its record is regenerated.
Both apps were therefore re-captured from the same commit. free2z's pixels are
byte-identical; only its digests and source pointer move.

New digests, all produced by real capture runs:

| app | sourceDigest | contractDigest |
| --- | --- | --- |
| e2e2z | `1020b8500c469e66…` | `d1dce6b74503a24e…` |
| free2z | `6fc999e3a9a75195…` | `99859560de8720a8…` |

No digest was hand-edited anywhere. The only field changed by hand in either
`store/capture.json` is `sourceSha`. `publicationReady` stays `false`.

## Verification

All run from `wallet/`, under the pinned
`mcr.microsoft.com/playwright@sha256:c091b21d…` image on `linux/amd64`, on base
`42ef4be2`:

- `node zuuli/scripts/surface-store-capture.mjs e2e2z --write` — exit 0
- `node zuuli/scripts/surface-store-capture.mjs free2z --write` — exit 0
- `node zuuli/scripts/surface-store-capture.mjs e2e2z --verify-reproducible` — exit 0
- `node zuuli/scripts/surface-store-capture.mjs free2z --verify-reproducible` — exit 0
- `npm --prefix zuuli run test:store-listing` — **85 pass, 0 fail**
- `npm --prefix zuuli run store:validate` — exit 0, both apps `publicationReady: false`

Refs #1022.


skyl:
Heads-up from the release train (workstream 10 of #1022): the three release PRs are now open — #1031 (STATUS re-derive), #1032 (free2z 4 / e2e2z 5), #1034 (ZUULI 24) — and #1033 has merged as `42ef4be2`.

Two things for this PR:

1. **Merge order stands: this one goes after #1034.** It touches `wallet/zuuli/scripts/surface-store-capture.mjs`, which `isReleaseImpactingPath` matches, so landing it before #1034 puts a release-impacting change between #1031's marker and the bump's parent and `Pin immutable source` will reject build 24.
2. **#1033 moved this PR's render inputs.** `renderInputs('e2e2z')` includes `plugins/tauri-plugin-f2zmsg/src`, and #1033 changed `src/lib.rs` and `src/internal_directory.rs` (and added `src/internal_directory/placeholder.conf`). The committed `sourceSha` here is `ee06a8d5`, so it is already behind. Nothing fails — `validateSurfaceCaptureRecord` runs with `enforceCurrentSource: false`, so `test:store-listing` stays green — but the record documents a tree that is no longer current. Worth a rebase and another `--write` at the new base before this merges, so the capture that lands is pinned to something real.

Recorded in #1031's STATUS.md re-derivation as a known, non-blocking store-media gap with this PR named as the fix.

skyl:
**Status for whoever picks this up.** Left open deliberately. It must land **after** the release bumps (#1034 merged and delivered; #1032 next), because `wallet/zuuli/scripts/**` is release-impacting and landing this first would invalidate the `STATUS.md` marker that #1034's `Pin immutable source` step re-checks.

**This does not gate the release.** Store capture gates store *listing* only: `validateSurfaceCaptureRecord` runs with `enforceCurrentSource: false`, no workflow runs the capture, and no release path reads these records.

**What it is:** e2e2z's app now makes startup calls the capture harness's fake native layer did not answer (`e2e2z_enrollment_status`, `plugin:deep-link|get_current`), so the `unenrolled-native-v1` capture would not render. The harness was extended to answer the real not-enrolled contract (enumerated from the built app, default still throws) and **both apps were re-captured** in the pinned `linux/amd64` container, two passes each, `--verify-reproducible` exit 0. **No digest was hand-edited** — that would forge capture evidence.

**The trap to check right before merging:** this capture is pinned to the tree it was taken from, and `test:store-listing` will **not** catch it going stale (`enforceCurrentSource: false`). If anything touching `wallet/e2e2z/src`, `wallet/e2e2z/src-tauri/src`, `wallet/shared/src` or `wallet/plugins/tauri-plugin-f2zmsg/src` has landed since, re-run the capture with `--write` rather than trusting the green gate. Main has moved since this was captured, so **assume it needs a re-capture** and verify.

